PR #2, opened 2019-07-01
admin/release/new: always consider all mappings

crawford:
Previously, additional mappings provided as args weren't considered when
creating a new release image from an ImageStream, an exsting release
image, or a directory. The was particularly problematic when attempting
to add new images:

    oc adm release new \
      --from-release quay.io/openshift-release-dev/ocp-release:4.2 \
      --to-image quay.io/crawford/openshift-release \
      cluster-etcd-operator=quay.io/hexfusion/cluster-etcd-operator

Without this patch, the previous command would create a release image
but it wouldn't the cluster-etcd-operator manifests.

Instead of adding new mappings in the default case (when not using an
existing ImageStream, release image, or directory), add them
unconditionally. Clayton thinks that this defect may have been
introduced when the validation code was being reworked.

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -586,15 +586,6 @@ func (o *NewOptions) Run() error {
 			}
 		}
 		fmt.Fprintf(o.ErrOut, "info: Found %d operator manifest directories on disk\n", len(ordered))
-
-	default:
-		for _, m := range o.Mappings {
-			if exclude.Has(m.Source) {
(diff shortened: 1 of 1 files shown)


tnozicka:
/close

this is just a published code from staging, open your PR here: https://github.com/openshift/origin/tree/master/staging/src/github.com/openshift/oc

crawford:
Moved to https://github.com/openshift/origin/pull/23337.

PR #12, opened 2019-07-19
Fix images

sttts:


Makefile
```diff
@@ -19,4 +19,6 @@ GO_BUILD_PACKAGES :=$(strip \
 # $2 - Dockerfile path
 # $3 - context directory for image build
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-$(notdir $(GO_PACKAGE)),./images/Dockerfile,.)
+$(call build-image,origin-cli,./images/cli/Dockerfile,.)
```
tnozicka: to run `yum install` in UBI (rhel) based image you must use `-mount ~/projects/origin-repos/4.2/:/etc/yum.repos.d/` with imagebuilder - hence the generated make targets are wrong

tnozicka: /hold
this won't work

PR #16, opened 2019-07-19
Make sure packages are installed

tnozicka:
/cc @deads2k @sttts


tnozicka:
ci is down but images step completed successfully, need it for follow up pr

PR #18, opened 2019-07-19
quick and dirty way to get a kubectl version

deads2k:
Much better option than https://github.com/openshift/origin/pull/23438

It's still not pretty, but it's slightly better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,14 @@
 all: build
 .PHONY: all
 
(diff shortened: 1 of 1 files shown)


deads2k:
I think we want to pull this information from glide.yaml instead of hardcoding, but right now we really just need to unbork CI.

smarterclayton:
Please make sure have a bug opened to track reverting this

twiest:
Bug filed: https://bugzilla.redhat.com/show_bug.cgi?id=1731530

PR #22, opened 2019-07-23
Make 'oc adm prune auth' can prune ServiceAccount

barleyer:
As title,since this **[BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1724596)**,I think this PR can make some sense.

This can prune scc,clusterrolebinding and rolebinding of sa; you can use  `oc adm prune auth sa " " -n " "` to prune specific sa.

### DIFF
--- a/pkg/cli/admin/prune/auth/serviceaccount.go
+++ b/pkg/cli/admin/prune/auth/serviceaccount.go
@@ -1,0 +1,55 @@
+package auth
+
+import (
+	"fmt"
+	"io"
+
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	authv1client "github.com/openshift/client-go/authorization/clientset/versioned/typed/authorization/v1"
+	securityv1client "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
+)
+
+func reapForServiceAccount(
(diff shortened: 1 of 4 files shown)


barleyer:
@deads2k @enj could you please help to take a look?

pkg/cli/admin/prune/auth/serviceaccount.go
```diff
@@ -0,0 +1,55 @@
...
+		if len(retainedUsers) != len(scc.Users) {
+			updatedSCC := scc
+			updatedSCC.Users = retainedUsers
+			if _, err := securityClient.Update(&updatedSCC); err != nil && !kerrors.IsNotFound(err) {
+				errors = append(errors, err)
+			} else {
+				fmt.Fprintf(out, "securitycontextconstraints.security.openshift.io/"+updatedSCC.Name+" updated\n")
+			}
+		}
+	}
```
soltysh: What about groups?

soltysh: I mean any group you can add scc to through `oc adm policy add-scc-to-group` invocation.

barleyer: We already that feature,see below:
```
# oc adm prune auth -h
Removes references to the specified roles, clusterroles, users, and groups.
Other types are ignored
```

barleyer: And I just add serviceaccout to the list.

soltysh: I mean you're missing the code for removing your particular SCC from groups, compare your code with https://github.com/openshift/oc/blob/master/pkg/cli/admin/prune/auth/user.go#L55

PR #24, opened 2019-07-23
Add tnozicka

tnozicka:
/cc @mfojtik


sttts:
Button merging, as this is only an OWNERS update.

PR #27, opened 2019-07-24
add cli tools

deads2k:
Adds cli tools from origin that were use to make completions and verify conventions.


Makefile
```diff
@@ -35,3 +36,11 @@ image-ose-deployer: image-ose-cli
 $(call build-image,ose-recycler,./images/recycler/Dockerfile.rhel,.)
 image-ose-recycler: image-ose-cli
 
+update:
+	hack/update-generated-completions.sh
+.PHONY: update
+
+verify:
```
tnozicka: i think make will complain on overriding a target. make a specific target and then verify depend on it. same for update

hack/verify-cli-conventions.sh
```diff
@@ -0,0 +1,12 @@
...
+
+function cleanup() {
+    return_code=$?
+    exit "${return_code}"
+}
+trap "cleanup" EXIT
+
+go build github.com/openshift/oc/tools/clicheck/
+
+${OS_ROOT}/clicheck
```
tnozicka: I guess this all is really just
```make
verify-cli-conventions:
    go run ./tools/clicheck
.PHONY: verify-cli-conventions
```

tools/clicheck/OWNERS
```diff
@@ -0,0 +1,4 @@
+reviewers:
```
tnozicka: already in top level

hack/update-generated-completions.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
tnozicka: given we don't have the custom gopath now, can we just reduce it to committing output of `oc completion ${shell}` to a folder and drop `.files_generated`. seems like that's what it will do in the end https://github.com/openshift/origin/tree/master/contrib/completions

Makefile
```diff
@@ -35,3 +36,20 @@ image-ose-deployer: image-ose-cli
 $(call build-image,ose-recycler,./images/recycler/Dockerfile.rhel,.)
 image-ose-recycler: image-ose-cli
 
+update: update-generated-completions
+.PHONY: update
+
+verify: build verify-cli-conventions verify-generated-completions
```
tnozicka: verify doesn't depend on build

deads2k: > verify doesn't depend on build

it requires `oc`

tnozicka: not the `verify` target, the `verify-*` might

tnozicka:
/lgtm
(rest can be done as followups)

PR #29, opened 2019-07-25
Build an openshift-clients and openshift-clients-redistributable RPM

eparis:
The former is just oc + man pages for the current arch.
The later is oc for linux, windows, and macos

This is wrong/incomplete. But it's a start. Problems I can identify:
1. for non-x86_64 arches we don't really need to do a cross build.
2. It isn't properly handling the fact that the tags are per platform and per arch.
3. It doesn't make 'kubectl' only oc (not symlink or copy)
4. The horrible hack/build-cross.sh doesn't respect ENV vars
4. probably other stuff....


hack/lib/build/version.sh
```diff
@@ -0,0 +1,121 @@
...
+		KUBE_GIT_MAJOR=${BASH_REMATCH[1]}
+		KUBE_GIT_MINOR="${BASH_REMATCH[2]}+"
+	fi
+}
+readonly -f os::build::version::kubernetes_vars
+
+# Saves the environment flags to $1
+function os::build::version::save_vars() {
+	cat <<EOF
+OS_GIT_COMMIT='${OS_GIT_COMMIT-}'
```
tnozicka: these are not the same variables as ART sets - they have named it SOURCE_GIT_COMMIT now - I suppose this breaks versioning our binaries - maybe origin was special

hack/lib/constants.sh
```diff
@@ -0,0 +1,400 @@
...
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/oc/pkg/version.buildDate" "${buildDate}"))
+
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/openshift-apiserver/pkg/version.majorFromGit" "${OS_GIT_MAJOR}"))
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/openshift-apiserver/pkg/version.minorFromGit" "${OS_GIT_MINOR}"))
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/openshift-apiserver/pkg/version.versionFromGit" "${OS_GIT_VERSION}"))
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/openshift-apiserver/pkg/version.commitFromGit" "${OS_GIT_COMMIT}"))
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/openshift-apiserver/pkg/version.gitTreeState" "${OS_GIT_TREE_STATE}"))
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/openshift-apiserver/pkg/version.buildDate" "${buildDate}"))
+
+  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/github.com/openshift/openshift-controller-manager/pkg/version.majorFromGit" "${OS_GIT_MAJOR}"))
```
tnozicka: sets flags it shouldn't

hack/lib/constants.sh
```diff
@@ -0,0 +1,400 @@
...
+# Returns:
+#  None
+function os::util::list_go_src_dirs() {
+    go list -e ./... | grep -Ev "/(third_party|vendor|staging|clientset_generated)/" | LC_ALL=C sort -u
+}
+readonly -f os::util::list_go_src_dirs
+
+# os::util::list_go_deps outputs the list of dependencies for the project.
+function os::util::list_go_deps() {
+  go list -f '{{.ImportPath}}{{.Imports}}' ./test/... ./pkg/... ./cmd/... ./vendor/k8s.io/... | tr '[]' '  ' |
```
tnozicka: this seems out of date as well

hack/lib/constants.sh
```diff
@@ -0,0 +1,400 @@
...
+              -o -path '*assets/node_modules' \
+              -o -path '*test/*'              \
+              -o -path '*pkg/proxy'           \
+              -o -path '*k8s.io/kubernetes/cluster/gce*' \
+        \) -prune                             \
+    \) -name '*_test.go' | xargs -n1 dirname | sort -u | xargs -n1 printf "${OS_GO_PACKAGE}/%s\n"
+
+    local kubernetes_path="vendor/k8s.io/kubernetes"
+
+    if [[ -n "${TEST_KUBE-}" ]]; then
```
tnozicka: probably not relevant either

hack/lib/constants.sh
```diff
@@ -0,0 +1,400 @@
...
+
+# Removes the .syso file used to add compile-time VERSIONINFO metadata to the
+# Windows binary.
+function os::build::clean_windows_versioninfo() {
+  rm ${OS_ROOT}/vendor/github.com/openshift/oc/cmd/oc/oc.syso
+}
+readonly -f os::build::clean_windows_versioninfo
+
+# OS_ALL_IMAGES is the list of images built by os::build::images.
+readonly OS_ALL_IMAGES=(
```
tnozicka: not relevant

hack/lib/log/system.sh
```diff
@@ -0,0 +1,243 @@
...
+# os::log::system::internal::prune_datafile removes the given columns from a datafile created by 'sadf -d'
+#
+# Globals:
+#  None
+# Arguments:
+#  - 1: datafile
+#  - 2: comma-delimited columns to remove, with trailing comma
+# Returns:
+#  None
+function os::log::system::internal::prune_datafile() {
```
tnozicka: not sure what is this needed for

hack/lib/util/ensure.sh
```diff
@@ -0,0 +1,116 @@
...
+# tests that the user can list iptables rules, trying with
+# `sudo` if it fails without.
+#
+# Globals:
+#  None
+# Arguments:
+#  None
+# Returns:
+#  None
+function os::util::ensure::iptables_privileges_exist() {
```
tnozicka: not sure oc needs this

hack/lib/util/environment.sh
```diff
@@ -0,0 +1,296 @@
...
+# os::util::environment::use_sudo updates $USE_SUDO to be 'true', so that later scripts choosing between
+# execution using 'sudo' and execution without it chose to use 'sudo'
+#
+# Globals:
+#  None
+# Arguments:
+#  None
+# Returns:
+#  - export USE_SUDO
+function os::util::environment::use_sudo() {
```
tnozicka: probably not a case for oc

tnozicka:
I was hoping the split is an opportunity to take what we need and have issues in just one release w.r.t. ART. Doing this will risk another set of release bugs in the next release.

eparis:
This is, literally, just cp stuff from origin hack/* then creating the smallest oc.spec and hack/build-cross.sh I could. I agree that 98%+ of this is garbage and should get removed.

Our problem is that you (the entire team "you", not "you" singular) have broken our ability to build and deliver the product. This type of work should have been done before the repo split. Not after. But here we are, so we just need to get things working again as quickly as possible. I don't care what path the team takes to build the clients rpms. But that path needs to be complete in O(days) not O(weeks).

If this is not a useful starting point for the team, no problem, close it. But if this is the fastest way to begin to fix the problems this team created for others then I would hope we would prioritize fixing what we broke and then worry about 'doing it right'.

Then again this isn't even 'right and would take more PRs to finish cleaning the mess we're in.

sttts:
/hold

I am against mergin this as is. This creates a mess of hack/ files which we were trying to avoid with the shared build system.

Having said that, if we don't have the resource to build RPMs cleanly through the library-go build system, rename the hack dir here to something like `legacy-rpm-build-hack` and could live with it. Then at least we avoid creating a kitchen sink for other scripts in the future because people think hack/ is back.

tnozicka:
the CI failed with 
```
Pulling image docker-registry.default.svc:5000/ci-op-licwphqt/pipeline@sha256:d9984a98ad7f7940f897a00c6f240743c7518d89b4fa6dc623d3266be257d3f6 ...
--> FROM docker-registry.default.svc:5000/ci-op-licwphqt/pipeline@sha256:d9984a98ad7f7940f897a00c6f240743c7518d89b4fa6dc623d3266be257d3f6 as 0
--> RUN ["/bin/bash","-c","set -o errexit; umask 0002; make build-rpms; ln -s $( pwd )/_output/local/releases/rpms/ /srv/repo"]
hack/build-rpms.sh
[INFO] Building release RPMs for /go/src/github.com/openshift/oc/oc.spec ...
error: Architecture is not included: x86_64
[ERROR] hack/build-rpms.sh:55: `rpmbuild -b${srpm} "${OS_RPM_SPECFILE}" --define "skip_dist 1" --define "make_redistributable ${make_redistributable}" --define "version ${OS_RPM_VERSION}" --define "release ${OS_RPM_RELEASE}" --define "commit ${OS_GIT_COMMIT}" --define "os_git_vars ${OS_RPM_GIT_VARS}" --define 'dist .el7' --define "_topdir ${rpm_tmp_dir}"` exited with status 1.
[ERROR] hack/build-rpms.sh exited with code 1 after 00h 00m 33s
make: *** [build-rpms] Error 1
error: build error: running '/bin/bash -c set -o errexit; umask 0002; make build-rpms; ln -s $( pwd )/_output/local/releases/rpms/ /srv/repo' failed with exit code 2
```

tnozicka:
I can see the same error `error: Architecture is not included: x86_64` on non-hack based PR ttps://github.com/openshift/oc/pull/34 (via https://github.com/openshift/library-go/pull/492) but I have based the spec file on yours. It builds fine locally, maybe this is an issue with base image or being built in a container

deads2k:
we went with https://github.com/openshift/oc/pull/34 instead

/close

PR #40, opened 2019-08-02
Bug 1671685: remove-(cluster-)role-from-*: Properly report binding kind if not found

stlaz:


pkg/cli/admin/policy/modify_roles.go
```diff
@@ -644,9 +644,13 @@ func (o *RoleModificationOptions) RemoveRole() error {
...
 		}
-	}
-	if len(roleBindings) == 0 {
-		return fmt.Errorf("unable to locate RoleBinding %s for %s %q", o.RoleBindingName, o.RoleKind, o.RoleName)
+		if len(roleBindings) == 0 {
+			bindingType := "ClusterRoleBinding"
+			if len(o.RoleBindingNamespace) > 0 {
+				bindingType = "RoleBinding"
+			}
+			return fmt.Errorf("unable to locate %s for %s %q", bindingType, o.RoleKind, o.RoleName)
```
tnozicka: why did you drop `o.RoleBindingName` from the message?

stlaz: Not only that, I moved the whole thing into the else branch of the previous condition.

`getRoleBindingAbstraction()` returns error even in `NotFound()` cases and therefore you'd never get to the original message with `o.RoleBindingName` set, which caused extra spaces in the message (as seen in the BZ)

stlaz: `getRoleBindingAbstraction()` is the function called in the `if` branch of this `if-else` condition

tnozicka: should the message be `unable to locate any %s ...`?

stlaz: that sounds a bit be better, true, let me get that in

PR #41, opened 2019-08-02
Bug 1707679: Don't install CRD SCC types into global scheme

sttts:
SCCs are custom resources. Having them in the global oc scheme leads to strange effects like broken `oc edit` and `oc patch`.

### DIFF
--- a/cmd/oc/oc.go
+++ b/cmd/oc/oc.go
@@ -10,6 +10,9 @@ import (
 
 	"github.com/spf13/pflag"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	utilflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/logs"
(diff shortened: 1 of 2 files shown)


sttts:
/hold

How can we test this?

/assign @tnozicka @sallyom

sallyom:
@tnozicka I've tested this with `Cluster version  4.2.0-0.ci-2019-08-05-103049`
both working as expected:
`oc edit scc restricted`
`oc patch scc restricted -p '{"allowedUnsafeSysctls":["kernel.msg*"]}' --type=merge`

sallyom:
/lgtm

thanks @sttts and I've learned something new, too.

tnozicka:
/approve
@sallyom thx, for testing it

cmd/oc/oc.go
```diff
@@ -100,3 +104,18 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func installNonCRDSecurity(scheme *apimachineryruntime.Scheme) error {
```
soltysh: Although this is somewhat valid approach, I think we should rather split the installation at the API level, iow. inside openshift/api's security group we should have two separate calls for installing all resources (including CRDs) and just native ones. IIRC security is the only group that holds both and having the distinction at API level will prevent diverge in installing those when new resources will be added. /cc @deads2k @sttts

sttts: The right solution would be to make kubectl/oc stop using the scheme membership as an indication of the patch type.

soltysh: Yeah, that'd be nice too.

PR #43, opened 2019-08-05
Bug 1737127: idle: fix null dereference when pod's controller can't be found

dcbw:
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1737127


dcbw:
@squeed @openshift/networking

dcbw:
@deads2k does this one look OK?

PR #46, opened 2019-08-08
Add oc target

tnozicka:
/cc @soltysh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,11 @@ RPM_EXTRAFLAGS := \
 
 IMAGE_REGISTRY :=registry.svc.ci.openshift.org
 
+oc: GO_BUILD_PACKAGES:=./cmd/oc
+oc: build
+
+.PHONY: oc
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -35,6 +35,11 @@ RPM_EXTRAFLAGS := \
 
 IMAGE_REGISTRY :=registry.svc.ci.openshift.org
 
+oc: GO_BUILD_PACKAGES:=./cmd/oc
```
soltysh: I like the idea, although I wonder if we want to introduce such granular targets. None of the other repos have it, I don't want oc to be the raisin :wink: 
/hold

tnozicka: I think the one most useful target could be justified, I don't feel strong

soltysh:
/lgtm cancel
@tnozicka how about alternative solution from #51

soltysh:
Closing due to age. 
/close

PR #47, opened 2019-08-08
Bug 1685189 - use server-side printing in kubectl get

soltysh:
/assign @mfojtik @tnozicka 
/hold
until https://github.com/openshift/kubernetes/pull/74 merges first


soltysh:
Replaced with https://github.com/openshift/oc/pull/63

PR #48, opened 2019-08-08
Bug 1704827: Improve error msg when console URL passed to login via CLI

sallyom:
This adds a very specific error message for users who pass the console URL instead of API URL. 

Adds a check for http response with Content-Type "html/text"  (console url) instead of "application/json" (server URL) 

(see output below)
_btw: kubeadmin is not meant to be used via CLI, only to be used for admin login to console to set up a valid identity provider via console, however, I used kubeadmin in the examples below bc I did not set up any other idp_

### DIFF
--- a/pkg/cli/login/error_translation.go
+++ b/pkg/cli/login/error_translation.go
@@ -22,7 +22,15 @@ You can also run this command again providing the path to a config file directly
 `
 	tlsOversizedRecordMsg = `Unable to connect to %[2]s using TLS: %[1]s.
 Ensure the specified server supports HTTPS.`
+	invalidServerURLMsg = `Seems you passed an HTML page (console?) instead of server URL.
+Verify provided address and try again.`
 )
+
+type errInvalidServerURL struct{}
+
+func (e *errInvalidServerURL) Error() string {
+	return invalidServerURLMsg
+}
 
 // GetPrettyMessageForServer prettifys the message of the provided error
 func getPrettyMessageForServer(err error, serverName string) string {
@@ -50,6 +58,7 @@ func getPrettyMessageForServer(err error, serverName string) string {
 
 	case certificateInvalidReason:
 		return fmt.Sprintf("The server is using an invalid certificate: %s", err)
+
 	}
 
 	return err.Error()
--- a/pkg/cli/login/helpers.go
+++ b/pkg/cli/login/helpers.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	restclient "k8s.io/client-go/rest"
@@ -72,6 +73,13 @@ func dialToServer(clientConfig restclient.Config) error {
 	res, err := rt.RoundTrip(req)
 	if err != nil {
 		return err
+	}
+
+	// This is to guide a user who passes the console URL instead of server URL
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1704827
+	contentType := res.Header.Get("Content-Type")
+	if strings.Contains(contentType, "text/html") {
+		return &errInvalidServerURL{}
 	}
 
 	defer res.Body.Close()
--- a/pkg/cli/login/loginoptions.go
+++ b/pkg/cli/login/loginoptions.go
@@ -142,6 +142,11 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 				clientConfig.Insecure = true
 				clientConfig.CAFile = ""
 				clientConfig.CAData = nil
+				// dialToServer was called above but in case of user choosing insecure,
+				// need to call again for invalidServerURL check
+				if err := dialToServer(*clientConfig); err != nil {
+					return nil, err
+				}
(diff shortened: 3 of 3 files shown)


sallyom:
new output: 
```console
$ oc login -u kubeadmin -p blahh-blahh-blahh-blahh https://console-openshift-console.apps.sotest871.devcluster.openshift.com
The server uses a certificate signed by an unknown authority.
You can bypass the certificate check, but any data you send to the server could be intercepted by others.
Use insecure connections? (y/n): y

error: Response Content-type: found text/html
Seems you passed an HTML page (console?) instead of Kubernetes API URL.
Verify you provided Kubernetes server URL and try again.
```

pkg/cli/login/error_translation.go
```diff
@@ -22,6 +23,8 @@ You can also run this command again providing the path to a config file directly
 `
 	tlsOversizedRecordMsg = `Unable to connect to %[2]s using TLS: %[1]s.
 Ensure the specified server supports HTTPS.`
+	invalidServerURLMsg = `Seems you passed an HTML page (console?) instead of Kubernetes API URL.
```
soltysh: Let's keep it consistent, since we're using `server URL` upon login let's call it like so in here. So proposed message:
`Seems you passed an HTML page (console?) instead of server URL, verify provided address and try again.`

pkg/cli/login/helpers.go
```diff
@@ -74,6 +75,13 @@ func dialToServer(clientConfig restclient.Config) error {
 		return err
 	}
 
+	// This is to guide a user who passes the console URL instead of server URL
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1704827
+	contentType := res.Header.Get("Content-Type")
+	if strings.Contains(contentType, "text/html") {
+		return fmt.Errorf("Response Content-type: found text/html")
```
soltysh: Errors always lower case, don't forget to update the other place too. Also consider using an error constant for this.

sallyom: whoops.  fixed, and added `errInvalidServerURL` type

pkg/cli/login/error_translation.go
sallyom: updated, 
```
error: response content-type: found text/html
Seems you passed an HTML page (console?) instead of server URL.
Verify provided address and try again.
```

pkg/cli/login/error_translation.go
```diff
@@ -22,8 +27,14 @@ You can also run this command again providing the path to a config file directly
 `
 	tlsOversizedRecordMsg = `Unable to connect to %[2]s using TLS: %[1]s.
 Ensure the specified server supports HTTPS.`
+	invalidServerURLMsg = `Seems you passed an HTML page (console?) instead of server URL.
+Verify provided address and try again.`
 )
 
+func (e *errInvalidServerURL) Error() string {
```
soltysh: Nit, let's keep it next to the type definition next time.

pkg/cli/login/loginoptions.go
```diff
@@ -174,6 +175,13 @@ func (o *LoginOptions) gatherAuthInfo() error {
 	// make a copy and use it to avoid mutating the original
 	t := *directClientConfig
 	clientConfig := &t
+	// dialToServer was called above but in case of user choosing insecure,
+	// need to call again for invalidServerURL
+	if clientConfig.Insecure {
+		if err := dialToServer(*clientConfig); err != nil {
```
enj: Move this into `getClientConfig` where `clientConfig.Insecure = true` is set.

pkg/cli/login/error_translation.go
```diff
@@ -7,13 +7,18 @@ import (
 	"strings"
 )
 
+type errInvalidServerURL struct {
+	message string
```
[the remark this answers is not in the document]
enj: (i.e. drop the `error: response content-type: found text/html` bit)

pkg/cli/login/error_translation.go
sallyom: done, had to address mo's comments so fixed : )

pkg/cli/login/error_translation.go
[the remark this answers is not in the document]
sallyom: done, thanks, much better.

pkg/cli/login/loginoptions.go
sallyom: Moved, still need the 2nd call to dialToServer in case of clientConfig.Insecure modified mid-function, unless this check for text/html content-type belongs in the tokencmd.RequestToken() fn?  If so, we don't need the 2nd call, but flow continues to gatherAuthInfo().

pkg/cli/login/loginoptions.go
```diff
@@ -155,8 +155,16 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 			}
 			return nil, err
 		}
+
 	}
 
+	// dialToServer was called above but in case of user choosing insecure,
+	// need to call again for invalidServerURL check
+	if clientConfig.Insecure {
```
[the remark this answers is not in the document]
sallyom: done! thanks for the review

PR #54, opened 2019-08-13
Bug 1740596: Don't install CRD authorization types into global scheme

mfojtik:
Authorization are custom resources. Having them in the global oc scheme leads to strange effects like broken oc edit and oc patch.


sttts:
/lgtm

/hold

Maybe a manual test? Or do we trust e2e? Unhold whenever you want.

mfojtik:
This was wrong, rolebindings are not CRD.

PR #64, opened 2019-08-19
Bug 1734504: Parts of oc adm must-gather do not respect --config CLI option

sanchezl:


mfojtik:
/lgtm
/hold

@soltysh can you have second look?

deads2k:
no more commands should honor --config.  The standard arg is --kubeconfig

deads2k:
oh, I see, this is in our already existing mistake. :(

Alright, it should honor --config.

soltysh:
> no more commands should honor --config. The standard arg is --kubeconfig

yeah, it's not about `--config` per se, but rather passing flags properly down the invocation tree.

PR #67, opened 2019-08-22
Introduce legal notice for login and debug commands

JAORMX:
Due to legal requirements, some deployments need to display a legal
notice for users "entering" a system.

This legal notice can range from telling the user that they are entering
a government system and are being monitored, to just displaying the
rules of usage in the system.

This is required by law by several countries, with the intent that in
case the user attacks the system and needs to be prosecuted, the message
can be used as proof that the attacker knew against whom the attack was being
executed.

In this PR, the "oc login" and "oc debug" commands will search for the
legal-notice ConfigMap, which would ideally be configured in the
openshift-config namespace.

If the notice is not present, or the user has no rights to access that
ConfigMap, the command will not fail.

As a separate PR, the (empty) ConfigMap and the needed ClusterRole and
ClusterRoleBinding will be populated by default in the deployment.


JAORMX:
CC @sttts

sttts:
Leaving details to @soltysh. But looks like the right approach.

JAORMX:
@sttts @soltysh I'm thinking about including a sample (empty) legal notice and the appropriate role binding. But I'm not sure in what repo should I include these so they get created in the initial deployment. Any pointers?

soltysh:
> @sttts @soltysh I'm thinking about including a sample (empty) legal notice and the appropriate role binding. But I'm not sure in what repo should I include these so they get created in the initial deployment. Any pointers?

I'd suggest just documenting this, empty legal notice is pointless, imo.

pkg/helpers/legalnotice/legalnotice.go
```diff
@@ -0,0 +1,42 @@
+package legalnotice
```
soltysh: Call it just notice, motd, or welcome screen, or something like that. Not everywhere it will be used as legalnotice.

pkg/helpers/legalnotice/legalnotice_test.go
```diff
@@ -0,0 +1,52 @@
...
+		{"No message because of malformed notice (misspelled key)", "", configmap("openshift-config", "legal-notice", "mesage", "This is a legal notice")},
+		{"No message because of misconfigured notice (wrong name)", "", configmap("openshift-config", "bad-name", "message", "This is a legal notice")},
+		{"No message because notice not found in namespace", "", configmap("default", "legal-notice", "mesage", "This is a legal notice")},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			clientSet := fake.NewSimpleClientset(test.obj)
+			var output bytes.Buffer
+			err := DisplayLegalNotice(clientSet.CoreV1(), &output)
```
soltysh: ` if err := DisplayLegalNotice(clientSet.CoreV1(), &output); err != nil {`

pkg/cli/login/loginoptions.go
```diff
@@ -78,12 +80,17 @@ func NewLoginOptions(streams genericclioptions.IOStreams) *LoginOptions {
 // Gather all required information in a comprehensive order.
 func (o *LoginOptions) GatherInfo() error {
 	if err := o.gatherAuthInfo(); err != nil {
+		// Display legal notice even if authentication failed
+		o.prepareAndDisplayLegalNotice()
```
soltysh: If auth failed, how will you be able to get anything from the server?

pkg/cli/debug/debug.go
```diff
@@ -432,6 +433,12 @@ func (o *DebugOptions) RunDebug() error {
 			fmt.Fprintf(o.ErrOut, "To use host binaries, run `chroot /host`\n")
 		}
 
+		err = legalnotice.DisplayLegalNotice(o.CoreClient, o.ErrOut)
```
soltysh: `if err := legalnotice.DisplayLegalNotice(o.CoreClient, o.ErrOut); err != nil {`

pkg/cli/debug/debug.go
```diff
@@ -432,6 +433,12 @@ func (o *DebugOptions) RunDebug() error {
...
+		if err != nil {
+			fmt.Fprintf(o.ErrOut, "Unable to display legal notice\n")
```
soltysh: `return fmt.Errorf("Unable to display legal notice: %v", err)`

pkg/helpers/legalnotice/legalnotice.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// Display the necessary legal notice if it's configured in the system.
+// Note that the legal notice comes from the "legal-notice" configMap, which
+// should be in the "openshift-config" namespace. This needs to be configured
+// by the deployer.
+func DisplayLegalNotice(coreClient corev1client.CoreV1Interface, out io.Writer) error {
+	legalnoticecm, err := coreClient.ConfigMaps("openshift-config").Get("legal-notice", metav1.GetOptions{})
+	if kerrors.IsNotFound(err) || kerrors.IsForbidden(err) {
+		// NOTE(jaosorior): We shouldn't get here given that a clusterRole and
```
soltysh: Not quite sure what you're talking here about? Just say it's fine if the CM does not exists. Forbidden though, is something unexpected, imo.

soltysh:
Also, what about other commands such as `exec`, `rsh`, etc. ?

pkg/helpers/legalnotice/legalnotice_test.go
JAORMX: Thanks for the tip!

pkg/cli/login/loginoptions.go
JAORMX: That would be handled by RBAC. This is the role I created for it:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: legalnotice-reader
  namespace: openshift-config
rules:
- apiGroups: [""]
  resources: ["configmaps"]
  resourceNames: ["legal-notice"]
  verbs: ["get"]
```

And this is the role binding:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: read-legalnotice-global
subjects:
- kind: Group
  name: system:authenticated
  apiGroup: rbac.authorization.k8s.io
- kind: Group
  name: system:unauthenticated
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: legalnotice-reader
  apiGroup: rbac.authorization.k8s.io
```

As you can see, all users (authenticated and not authenticated) can read that message. Which is necessary for legal reasons.

pkg/helpers/legalnotice/legalnotice.go
JAORMX: I had an idea initially... but I don't think it's relevant anymore. So I'll remove the forbidden check.

JAORMX:
> I'd suggest just documenting this, empty legal notice is pointless, imo.

That's alright with me. Though I still think that the clusterrole and the clusterrolebinding should be there by default, for usability.

JAORMX:
Found a solution to display the message on exec as well. Updated the patch.

smarterclayton:
/hold

Is there a design for this?

smarterclayton:
I don’t think you can display this for rsh or exec without changing their scripting behavior which means automation may change function.

What do other clients do?  Console?  Kubectl?  What do third party applications do?

Displaying this notification on login is one thing.  Doing it on an arbitrary set of commands is not as clear cut.

smarterclayton:
I have concerns about this and want to see more justification before it goes further.  I probably won’t have time to respond this week but will have some time next week.

JAORMX:
@smarterclayton yes, there is a design document for this.

JAORMX:
> Displaying this notification on login is one thing. Doing it on an arbitrary set of commands is not as clear cut.

What about only displaying the message when the commands are set for interactive mode?

JAORMX:
> I don’t think you can display this for rsh or exec without changing their scripting behavior which means automation may change function.

on the other hand... it's displayed to stderr. So it might not be as much of an issue really.

smarterclayton:
> on the other hand... it's displayed to stderr. So it might not be as much of an issue really.

It's an API change to return different output (the current behavior is the output is only the server's output).

smarterclayton:
> This would replicate the motd functionality, which only displays when you actually initiate a session in SSH. It doesn't get display when you run a command via ssh. So... that would be fairly equivalent here; and thus compliant with federal regulations.

Why is `oc login` not sufficient for this?  That is when you are granted access to those systems, and creating pods that run things is effectively the same thing as exec, and pods are created all over the place.

JAORMX:
> > on the other hand... it's displayed to stderr. So it might not be as much of an issue really.
> 
> It's an API change to return different output (the current behavior is the output is only the server's output).

sure. I changed the patch to only display it when rsh and exec are executed on interactive mode.

JAORMX:
> > This would replicate the motd functionality, which only displays when you actually initiate a session in SSH. It doesn't get display when you run a command via ssh. So... that would be fairly equivalent here; and thus compliant with federal regulations.
> 
> Why is `oc login` not sufficient for this? That is when you are granted access to those systems, and creating pods that run things is effectively the same thing as exec, and pods are created all over the place.

What I've gathered from feedback of folks that have worked in compliance (external and internal) is that when you do rsh, debug or exec, you are opening a session in the node (even if it's inside a container); so from a legal perspective, the user should still be warned that it's a government system in case legal action needs to be taken.

If this is not convincing, I can do split this PR in two:

- One PR that only addresses "oc login"

- Another PR that subsequently adds support for the rest of the commands.

This way, we can discuss further for the other commands, as it seems that "oc login" is a clear case and would be easier to merge already.

@smarterclayton how does this sound?

smarterclayton:
Splitting this into login only for the near term is more actionable.  There are a number of Kubernetes components that automatically use exec and rsh, including the console and other tools like Jenkins, and I'm not sure the approach here scales beyond login.

JAORMX:
@smarterclayton thanks for the feedback. I'll split the PR so this can move forward.

JAORMX:
@smarterclayton done

JAORMX:
@smarterclayton alright! separated the PR, and now this only involves "oc login" what do you think?

JAORMX:
Any chance to get another review for this PR?

pkg/helpers/motd/motd_test.go
```diff
@@ -0,0 +1,51 @@
...
+func TestDisplayMOTD(t *testing.T) {
+	var tests = []struct {
+		description string
+		expected    string
+		obj         runtime.Object
+	}{
+		{"Correctly displays message", "This is a legal notice\n", configmap("openshift-config", "motd", "message", "This is a legal notice")},
+		{"No message because of malformed notice (misspelled key)", "", configmap("openshift-config", "motd", "mesage", "This is a legal notice")},
+		{"No message because of misconfigured notice (wrong name)", "", configmap("openshift-config", "bad-name", "message", "This is a legal notice")},
+		{"No message because notice not found in namespace", "", configmap("default", "motd", "mesage", "This is a legal notice")},
```
jhrozek: The key should probably be called "message" in the last test, because the purpose of the test is that the configmap is not loaded from the default namespace, not that the key is misspelled. It's not a big deal, because the code fails where it's expected to fail, it's made me go "hmm?" for a bit.

JAORMX: Thanks! will fix.

JAORMX:
@soltysh @smarterclayton could you revisit this PR?

pkg/helpers/motd/motd.go
```diff
@@ -0,0 +1,43 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog"
+)
+
+// Display the necessary MOTD if it's configured in the system.
+// Note that the MOTD comes from the "motd" configMap, which
+// should be in the "openshift-config" namespace. This needs to be configured
+// by the deployer.
+func DisplayMOTD(coreClient corev1client.CoreV1Interface, out io.Writer) error {
+	motdcm, err := coreClient.ConfigMaps("openshift-config").Get("motd", metav1.GetOptions{})
```
soltysh: `openshift-config` might not be a good place, since it's meant more for configuring operators and cluster. I'd suggest moving this to `openshift` since that's always been a place readable by every user of the cluster. /cc @smarterclayton

JAORMX: Thanks for the feedback!

pkg/helpers/motd/motd.go
```diff
@@ -0,0 +1,43 @@
...
+// Display the necessary MOTD if it's configured in the system.
+// Note that the MOTD comes from the "motd" configMap, which
+// should be in the "openshift" namespace. This needs to be configured
+// by the deployer.
+func DisplayMOTD(coreClient corev1client.CoreV1Interface, out io.Writer) error {
+	motdcm, err := coreClient.ConfigMaps("openshift").Get("motd", metav1.GetOptions{})
+	if kerrors.IsNotFound(err) || kerrors.IsForbidden(err) {
+		// NOTE(jaosorior): If no motd is configured, it's fine. No need to
+		// print anything. If we get a "Forbidden" error, this is because no
+		// role binding has been configured yet. So we need to ignore the
```
soltysh: Since you're working with `openshift` namespace that shouldn't be a problem for logged-in users, but you never know :wink:

pkg/helpers/motd/motd.go
```diff
@@ -0,0 +1,43 @@
...
+	if !ok {
+		klog.V(4).Infof("Unable to display MOTD. It exists but is misconfigured.")
+		return nil
+	}
+
+	// Add newline if needed
+	if !strings.HasSuffix(motd, "\n") {
+		motd = motd + "\n"
+	}
+	fmt.Fprintf(out, "%s", motd)
```
soltysh: Maybe it's worth also adding a new line before that MOTD, I've tried this just now and I got my `hello world` right after the default welcome message, which makes it disappear a bit.

JAORMX: Done

PR #75, opened 2019-08-28
Bug 1709239: Remove Git Repo option from 'oc set volumes' cmd

sallyom:
This PR:
1) Removes Git Repo examples from `oc set volume` help menu, since git volume source is no longer supported.
2) updates examples in oc `dc/registry` to `dc/myapp`  
    (3.x dc/registry vs 4.x deployment/image-registry), might cause confusion if left as/is.

### DIFF
--- a/pkg/cli/set/env.go
+++ b/pkg/cli/set/env.go
@@ -49,8 +49,8 @@ var (
 		syntax.`)
 
 	envExample = templates.Examples(`
-		# Update deployment 'registry' with a new environment variable
-	  %[1]s env dc/registry STORAGE_DIR=/local
+	  # Update deployment config 'myapp' with a new environment variable
+	  %[1]s env dc/myapp STORAGE_DIR=/local
 
 	  # List the environment variables defined on a build config 'sample-build'
 	  %[1]s env bc/sample-build --list
@@ -78,7 +78,7 @@ var (
 	  %[1]s env -f dc.json ENV-
 
 	  # Set some of the local shell environment into a deployment config on the server
-	  env | grep RAILS_ | %[1]s env -e - dc/registry`)
+	  env | grep RAILS_ | %[1]s env -e - dc/myapp`)
 )
 
 type EnvOptions struct {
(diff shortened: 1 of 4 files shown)


pkg/cli/set/env.go
```diff
@@ -50,7 +50,7 @@ var (
 
 	envExample = templates.Examples(`
 		# Update deployment 'registry' with a new environment variable
-	  %[1]s env dc/registry STORAGE_DIR=/local
+	  %[1]s env deployment/image-registry STORAGE_DIR=/local
```
soltysh: dc is not the same as deployment, so you can't change like that. I think we should have examples covering both dc and deployments, still

sallyom: ok, switched back to dc, but modified to `dc/myapp` since `dc/registry` will cause confusion with users, imo.  3.x we deployed a dc/registry, 4.x we don't, (it's deployment/image-registry).

pkg/cli/set/triggers.go
```diff
@@ -60,14 +60,14 @@ var (
 		trigger for a build config will only trigger the first build.`)
 
 	triggersExample = templates.Examples(`
-		# Print the triggers on the registry
-	  %[1]s triggers dc/registry
+		# Print the triggers on the deployment 'myapp'
```
soltysh: Can you fix the indentation here, while at it, also this says `deployment`, but should `deployment config`. We need to be extra careful not to confuse users.

sallyom: done

pkg/cli/set/env.go
```diff
@@ -49,8 +49,8 @@ var (
 		syntax.`)
 
 	envExample = templates.Examples(`
-		# Update deployment 'registry' with a new environment variable
-	  %[1]s env dc/registry STORAGE_DIR=/local
+	  # Update deployment config 'myapp' with a new environment variable
+	  %[1]s env dc/myapp STORAGE_DIR=/local
```
sallyom: fixed indentation, deployment/deployment config here, also : )

PR #76, opened 2019-08-28
Added the .git extension to the examples that had a github repo for the

joellord:
oc new-app --help

command.

I've had some errors and only realized later that I needed to add the .git extension to my repositories in order for new-app to work.
This does not apply to new-build which will work with or without the .git extension.


adambkaplan:
/ok-to-test

@joellord Can you please report the errors you received when you didn't use the `.git` extension with `oc new-app`?

PR #78, opened 2019-08-29
Bug 1745532: FlattenListVisitor now continues traversal on errors and returns an aggregate

soltysh:
/assign @tnozicka


soltysh:
Picked in https://github.com/openshift/oc/pull/79.

PR #81, opened 2019-08-30
release: Fix verification errors when excluding components

smarterclayton:
Passing --exclude resulted in a verification error because changing
those rules didn't correctly detect that the output is expected to
differ. Instead, we should skip if a user specifies any variants.

Also remove the need to declare "AlwaysInclude", manually verified
that it created the same output.

Needed for creating fake payloads during upgrade testing that lack
machine-config-operator.

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -47,11 +47,7 @@ func NewNewOptions(streams genericclioptions.IOStreams) *NewOptions {
 	return &NewOptions{
 		IOStreams:       streams,
 		ParallelOptions: imagemanifest.ParallelOptions{MaxPerRegistry: 4},
-		// TODO: only cluster-version-operator and maybe CLI should be in this list,
-		//   the others should always be referenced by the cluster-bootstrap or
-		//   another operator.
-		AlwaysInclude:  []string{"cluster-version-operator", "cli", "installer"},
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/release/new.go
```diff
@@ -47,11 +47,7 @@ func NewNewOptions(streams genericclioptions.IOStreams) *NewOptions {
 	return &NewOptions{
 		IOStreams:       streams,
 		ParallelOptions: imagemanifest.ParallelOptions{MaxPerRegistry: 4},
-		// TODO: only cluster-version-operator and maybe CLI should be in this list,
-		//   the others should always be referenced by the cluster-bootstrap or
-		//   another operator.
-		AlwaysInclude:  []string{"cluster-version-operator", "cli", "installer"},
-		ToImageBaseTag: "cluster-version-operator",
+		ToImageBaseTag:  "cluster-version-operator",
```
abhinavdahiya: Hmm, i am a little surprised as to what changed that the installer always include is not required anymore.. which operator image is requesting this image... _looking_

abhinavdahiya: Ahh, https://github.com/openshift/cluster-samples-operator/blob/8e5cd0e8149e923c675d11dcbe8e520c337f5c50/manifests/image-references#L17-L24

PR #88, opened 2019-09-10
Bug 1715001: If extracted from payload, 'oc version' reports payload version

sallyom:
This PR refactors `oc adm release extract` to report payload version for `oc version`.
In case of `oc version -o yaml` full output will print the version struct plus a new field `releaseClientVersion: payloadversion` 

See output below. 
`make build` w/ this PR checked out, then
with `oc adm release extract --tools quay.io/sallyom/release:test -v=3` access a test release image that includes:
* registry.svc.ci.openshift.org/ocp/release:4.3.0-0.ci-2019-10-07-053008
* cli image built from this PR 

/assign @soltysh  

cc @wking @tnozicka @smarterclayton

### DIFF
--- a/pkg/cli/admin/release/extract_tools.go
+++ b/pkg/cli/admin/release/extract_tools.go
@@ -150,27 +150,30 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli-artifacts", From: "usr/share/openshift/mac/oc"},
 
-			LinkTo:        []string{"kubectl"},
-			Readme:        readmeCLIUnix,
-			ArchiveFormat: "openshift-client-mac-%s.tar.gz",
+			LinkTo:             []string{"kubectl"},
+			Readme:             readmeCLIUnix,
+			InjectReleaseImage: true,
+			ArchiveFormat:      "openshift-client-mac-%s.tar.gz",
 		},
 		{
 			OS:      "linux",
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli", From: "usr/bin/oc"},
 
-			LinkTo:        []string{"kubectl"},
-			Readme:        readmeCLIUnix,
-			ArchiveFormat: "openshift-client-linux-%s.tar.gz",
+			LinkTo:             []string{"kubectl"},
+			Readme:             readmeCLIUnix,
+			InjectReleaseImage: true,
+			ArchiveFormat:      "openshift-client-linux-%s.tar.gz",
 		},
 		{
 			OS:      "windows",
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli-artifacts", From: "usr/share/openshift/windows/oc.exe"},
 
-			Readme:        readmeCLIWindows,
-			ArchiveFormat: "openshift-client-windows-%s.zip",
-			AsZip:         true,
+			Readme:             readmeCLIWindows,
+			InjectReleaseImage: true,
+			ArchiveFormat:      "openshift-client-windows-%s.zip",
+			AsZip:              true,
 		},
 		{
 			OS:      "darwin",
@@ -489,10 +492,19 @@ func (o *ExtractOptions) extractCommand(command string) error {
 
 		// copy the input to disk
 		if target.InjectReleaseImage {
+			releaseInfo := exactReleaseImage
 			var matched bool
-			matched, err = copyAndReplaceReleaseImage(w, r, 4*1024, exactReleaseImage)
+			errmsg := ""
+			switch target.Command {
+			case "oc":
+				releaseInfo = releaseName
+				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, could not determine version", target.Command)
+			default:
+				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, installer will not be locked to the correct image", target.Command)
+			}
+			matched, err = copyAndReplaceReleaseInfo(w, r, 4*1024, releaseInfo)
 			if !matched {
-				fmt.Fprintf(o.ErrOut, "warning: Unable to replace release image location into %s, installer will not be locked to the correct image\n", target.TargetName)
+				fmt.Fprintf(o.ErrOut, "%s\n", errmsg)
 			}
 		} else {
 			_, err = io.Copy(w, r)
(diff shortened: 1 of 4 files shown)


sallyom:
output when extracting tools from release image w/ this PR's cli image:
```console
$ ./openshift-install version
./openshift-install unreleased-master-1708-gedbfd3e343e4b6b36c3bc2c733c50a427146728b-dirty
built from commit edbfd3e343e4b6b36c3bc2c733c50a427146728b
release image 4.2.0-0.ci-2019-09-10-003223
```
```console
$ ./oc version
Client Version: 4.2.0-0.ci-2019-09-10-003223
Server Version: 4.2.0-0.ci-2019-09-09-042520
Kubernetes Version: v1.14.6+6492ec2
```
full output  (gitVersion is misleading here, since it's not really gitCommit but it's payload string):
```console
$ ./oc version -o yaml
clientVersion:
  buildDate: "2019-09-10T12:09:26Z"
  compiler: gc
  gitCommit: f14c366
  gitTreeState: clean
  gitVersion: 4.2.0-0.ci-2019-09-10-003223
  goVersion: go1.12.9
  major: ""
  minor: ""
  platform: linux/amd64
```
oc built from this PR (not extracted):
```console
$ oc version 
Client Version: v4.2.0-alpha.0-61-g9e4c596
```

sallyom:
here's a payload image w/ the CLI built from PR + `registry.svc.ci.openshift.org/ocp/release:4.2.0-0.ci-2019-09-10-003223`: `quay.io/sallyom/release:local`  and the cli image is `quay.io/sallyom/testcli:local`

sallyom:
_hm_ this isn't behaving as expected with this release: `registry.svc.ci.openshift.org/ci-op-v5smxs7s/release:latest` 
```console
$ ./oc version 
error: release image location is empty, this binary was incorrectly generated
```
I'm looking at that now

/hold

wking:
> gitVersion is misleading here, since it's not really gitCommit but it's payload string

This should be baked in at build time, because renaming a release doesn't change it.  The only things that release renaming should change (which would require extract-time injection) are the release pullspec and release name.

sallyom:
> > gitVersion is misleading here, since it's not really gitCommit but it's payload string
> 
> This should be baked in at build time, because renaming a release doesn't change it. The only things that release renaming should change (which would require extract-time injection) are the release pullspec and release name.

agreed, didn't feel quite right.  I'll re-examine.

wking:
```
$ ./oc version
Client Version: 4.2.0-0.ci-2019-09-10-003223
...
```

Is this constructed on the fly from that version structure?  Which property does the `ci` come from?  I'd expected you'd need to inject both the pullspec and release name, but don't see the release name (which I'd expect to be used as an opaque string here) getting injected.

sallyom:
> Is this constructed on the fly from that version structure? 

yes, it's from `k8s.io/apimachinery/pkg/version.Info`, reports code a binary was built from. 
> Which property does the `ci` come from?

I built the release using `registry.svc.ci.openshift.org/ocp/release:4.2.0-0.ci-2019-09-10-003223` + the cli image from the ci-opxxx ns of ci cluster. `oc adm release new --from-release ...`
>I'd expected you'd need to inject both the pullspec and release name, but don't see the release name (which I'd expect to be used as an opaque string here) getting injected.

not sure what you mean, `oc version` reports same version as 
the Release Metadata: Version ` 4.2.0-0.ci-2019-09-10-003223` (bc I built from that) here's the info on the release:
```console
$ oc adm release info quay.io/sallyom/release:ciop --pullspecs=true
Name:          4.2.0-0.ci-2019-09-10-003223
Digest:        sha256:bc29dbd77a266898e69628bbea714e6efc01f22dcf04953266531573476fe3fe
Created:       2019-09-10T13:40:53Z
OS/Arch:       linux/amd64
Manifests:     318
Unknown files: 3

Pull From: quay.io/sallyom/release@sha256:bc29dbd77a266898e69628bbea714e6efc01f22dcf04953266531573476fe3fe

Release Metadata:
  Version:  4.2.0-0.ci-2019-09-10-003223
  Upgrades: <none>

Component Versions:
  Kubernetes 1.14.6

Images:
  NAME                                          PULL SPEC
------
  baremetal-operator                            registry.svc.ci.openshift.org/ocp/4.2-2019-09-10-003223@sha256:3078c3afce677dfc538a8c412a5841764aa09fa2186580af453176e8dd799f88
  baremetal-runtimecfg                          registry.svc.ci.openshift.org/ocp/4.2-2019-09-10-003223@sha256:6ec5a6209e19762840d1932f7fb1338583d4d994797ebb88f3df7f0b03c3a5fb
  cli                                           registry.svc.ci.openshift.org/ci-op-v5smxs7s/pipeline:cli
  cli-artifacts                                 registry.svc.ci.openshift.org/ocp/4.2-2019-09-10-003223@sha256:ab05fcbfe48a90381b25b1804901647f61141c6b4e83a577a769a8946ecf3b8a
  cloud-credential-operator                     registry.svc.ci.openshift.org/ocp/4.2-2019-09-10-003223@sha256:6a28f70962578b72f940c549e84ecca8c176cd8f78f78dfb4cbefea177dda015
-------
```

pkg/cli/admin/release/extract_tools.go
```diff
@@ -490,7 +492,7 @@ func (o *ExtractOptions) extractCommand(command string) error {
 		// copy the input to disk
 		if target.InjectReleaseImage {
 			var matched bool
-			matched, err = copyAndReplaceReleaseImage(w, r, 4*1024, exactReleaseImage)
+			matched, err = copyAndReplaceReleaseImage(w, r, 4*1024, releaseName)
```
sallyom: I believe we want to report releaseName here instead of exactReleaseImage? sha is reported in sha256sum.txt 
For example, `openshift-install version`, output is 
`release image 4.2.0-0.ci-2019-09-09-042520` instead of
`release image registry.svc.ci.openshift.org/ocp/release@sha256:371460b3801ad48b58a6456464.......`

sallyom:
> > gitVersion is misleading here, since it's not really gitCommit but it's payload string
> 
> This should be baked in at build time, because renaming a release doesn't change it. The only things that release renaming should change (which would require extract-time injection) are the release pullspec and release name.

I've updated this, no longer replaces the build-time field, adds a new  field if/when extracting from payload.

pkg/cli/admin/release/extract_tools.go
wking: > For example, openshift-install version, output is
> release image 4.2.0-0.ci-2019-09-09-042520 instead of
> release image registry.svc.ci.openshift.org/ocp/release@sha256:371460b3801ad48b58a6456464.......

We want the by-digest release pullspec in the installer, because then a signed installer is secure against attacks on the mutable tag.  The short version name is nice to display as well, but we want both the short version name and the by-digest pullspec in the installer, not just the short version name.  And the short version name doesn't have to be part of a pullspec.  You could push to registry.example.com/what/ever:you-want, and you'd still want to inject the `version` field in the image's `release-metadata` as the version string, not whatever the pullspec tag was.

sallyom: and I see now when I try to create cluster the installer needs the by-digest. : )

soltysh: Yes, but I don't think `oc` itself needs that by-digest. For oc only the short is sufficient.

soltysh: So you'll have to have a one for oc and a different for installer.

sallyom: right, I'll simply do an if on the targetName :) got it

wking: I think the installer wants *both* the version and by-digest pullspec.  So either sub in for two placeholders, or sub something structured into a single placeholder.

sallyom: that's what I was thinking.. it'll require a change in installer code (i have that worked out already), so installer wants to print short form for version cmd? then needs pullspec to install, right?

wking: > ... installer wants to print short form for version cmd? then needs pullspec to install, right?

Yup.  And it actually prints both for the `version` command.

sallyom:
@wking, I've updated this PR to enable installer to print short form with[ this PR:](https://github.com/openshift/installer/pull/2345) 

For your convenience :) I've pushed a release image that includes the following:
* registry.svc.ci.openshift.org/ocp/release:4.2.0-0.ci-2019-09-11-114219
* cli image built from this PR
* installer image built from[ this PR:](https://github.com/openshift/installer/pull/2345) 

You can access it as
`oc adm release extract --tools quay.io/sallyom/release:testinstaller -v=3`

pkg/cli/admin/release/extract_tools.go
```diff
@@ -490,9 +493,10 @@ func (o *ExtractOptions) extractCommand(command string) error {
 		// copy the input to disk
 		if target.InjectReleaseImage {
 			var matched bool
-			matched, err = copyAndReplaceReleaseImage(w, r, 4*1024, exactReleaseImage)
+			releaseInfo := fmt.Sprintf("%s,%s", exactReleaseImage, releaseName)
```
abhinavdahiya: Hmm.. this change look like this is backwards incompatible...??

previously only `release-image-pullspec` was being set, now `release-image-pullspec,release-image-name` is being set.

So if I use this new oc to extract installer from an old release image, that installer binary will have invalid contents right???

sallyom: yes, i'll rework that

pkg/cli/admin/release/extract_tools.go
```diff
@@ -489,10 +492,19 @@ func (o *ExtractOptions) extractCommand(command string) error {
 
 		// copy the input to disk
 		if target.InjectReleaseImage {
+			releaseInfo := exactReleaseImage
 			var matched bool
-			matched, err = copyAndReplaceReleaseImage(w, r, 4*1024, exactReleaseImage)
+			errmsg := ""
+			switch target.Command {
```
soltysh: Honestly, this switch could be reversed since only installer relies on full release info and everything else (`default`) should be using the other, but that can be fixed later on when we'll be extending this mechanism at some point.

PR #90, opened 2019-09-12
cli: Add `oc set data` that makes it easy to manipulate secrets/config maps

smarterclayton:
Working with structured data inside of secrets and config maps can
be frustrating, and this command makes it easy to update one or more
keys as you go.


pkg/cli/set/data.go
```diff
@@ -0,0 +1,443 @@
...
+	data := make(map[string]string)
+	for _, fileSource := range fileSources {
+		keyName, filePath, err := util.ParseFileSource(fileSource)
+		if err != nil {
+			return nil, err
+		}
+		info, err := os.Stat(filePath)
+		if err != nil {
+			switch err := err.(type) {
+			case *os.PathError:
```
soltysh: Is there such a big difference between these two errors we need this switch?

pkg/cli/set/data.go
```diff
@@ -0,0 +1,443 @@
...
+			if strings.Contains(fileSource, "=") {
+				return nil, fmt.Errorf("cannot give a key name for a directory path")
+			}
+			fileList, err := ioutil.ReadDir(filePath)
+			if err != nil {
+				return nil, fmt.Errorf("error listing files in %s: %v", filePath, err)
+			}
+			for _, item := range fileList {
+				itemPath := filepath.Join(filePath, item.Name())
+				if item.Mode().IsRegular() {
```
soltysh: \<trollmode\>Oh, that's nice, I wonder when someone will start asking for link support :wink:\</trollmode\>

pkg/cli/set/data.go
```diff
@@ -0,0 +1,443 @@
...
+			}
+			return nil
+		})
+		if valid && !changed {
+			if o.Local || o.DryRun {
+				if err := o.Printer.PrintObj(info.Object, o.Out); err != nil {
+					allErrs = append(allErrs, err)
+				}
+			} else {
+				fmt.Fprintf(o.ErrOut, "info: %s was not changed\n", info.Name)
```
soltysh: We don't print those explicitly anywhere else, only with `V(1)` just like below. With this you might end up with 2 infos about not changed, though.

pkg/cli/set/data.go
```diff
@@ -0,0 +1,443 @@
...
+	if err != nil {
+		return err
+	}
+	for k, v := range kv {
+		if _, ok := o.SetData[k]; ok && k != v {
+			return fmt.Errorf("cannot set key %q in both argument and flag", k)
+		}
+		o.SetData[k] = v
+	}
+	kv, err = keyValuesFromFileSources(o.LiteralSources)
```
soltysh: `keyValuesFromLiteralSources`

pkg/cli/set/data.go
```diff
@@ -0,0 +1,443 @@
...
+func updateDataForObject(obj runtime.Object, fn func(data map[string][]byte) error) (bool, error) {
+	switch t := obj.(type) {
+	case *corev1.Secret:
+		if t.Data == nil {
+			t.Data = make(map[string][]byte)
+		}
+		return true, fn(t.Data)
+		// ReplicationController
+	case *corev1.ConfigMap:
+		if t.BinaryData == nil {
```
soltysh: Can you add a bit of comment about the process here, pls.

soltysh:
And don't forget to update completions :wink:

soltysh:
/assign @sallyom 
We'll want to upstream this eventually, we need to find the proposal from Juan from some time ago, refresh this and get it upstream after gathering some feedback.

PR #91, opened 2019-09-13
WIP: UPSTREAM: <drop>: don't merge

soltysh:
/hold


soltysh:
This proofed that oc from the current PR is being taken into consideration when running e2e-cmd:
```
Running test/cmd/convert.sh:22: executing 'oc convert -f /var/tests/test/cmd/testdata/convert/job-v1.yaml | grep 'apiVersion: batch/v1'' expecting success...
FAILURE after 0.000s: test/cmd/convert.sh:22: executing 'oc convert -f /var/tests/test/cmd/testdata/convert/job-v1.yaml | grep 'apiVersion: batch/v1'' expecting success: the command returned the wrong error code
There was no output from the command.
Standard error from the command:
error: this is borked on purpose
[ERROR] hack/lib/cmd.sh:241: `return "${return_code}"` exited with status 1.
```

PR #95, opened 2019-09-18
pkg/cli/admin/release/mirror: Conditional printImageContentInstructions

wking:
Don't call this unless both `o.From` and `o.To` are set, to avoid dying with:

```
error: Error creating mirror usage instructions: Unable to parse image reference '': repository name must have at least one component
```

when either the source or the target was actually an image-stream.

### DIFF
--- a/pkg/cli/admin/release/mirror.go
+++ b/pkg/cli/admin/release/mirror.go
@@ -104,7 +104,8 @@ type MirrorOptions struct {
 	ToRelease   string
 	SkipRelease bool
 
-	DryRun bool
+	DryRun                        bool
+	PrintImageContentInstructions bool
 
 	ClientFn func() (imageclient.Interface, string, error)
 
@@ -138,6 +139,7 @@ func (o *MirrorOptions) Complete(cmd *cobra.Command, f kcmdutil.Factory, args []
 		}
 		return client, ns, nil
 	}
+	o.PrintImageContentInstructions = true
 	return nil
 }
 
@@ -453,8 +455,10 @@ func (o *MirrorOptions) Run() error {
 		fmt.Fprintf(o.Out, "\nSuccess\nUpdate image:  %s\nMirrored to: %s\n", to, o.To)
 	}
 
-	if err := printImageContentInstructions(o.Out, o.From, o.To, repositories); err != nil {
-		return fmt.Errorf("Error creating mirror usage instructions: %v", err)
+	if o.PrintImageContentInstructions {
+		if err := printImageContentInstructions(o.Out, o.From, o.To, repositories); err != nil {
+			return fmt.Errorf("Error creating mirror usage instructions: %v", err)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


abhinavdahiya:
If we are mirroring from or to image stream, what are the expected instructions for the user?

I would rather fix that than skip..

wking:
> If we are mirroring from or to image stream, what are the expected instructions for the user?

I dunno.  Can you install from an image stream?  Maybe just if you run the installer from a Pod in that cluster?

In the linked CI PR, the "from stream" case is actually an `oc adm release new --mirror ...` call which builds an internal `ImageStream` and [passes that to the mirror logic][1].  Folks don't need `printImageContentInstructions` out of `oc adm release new ...`.

[1]: https://github.com/openshift/oc/blob/32cc702ffbb91c2bc04accaf7ca6c99a886f3269/pkg/cli/admin/release/new.go#L1055

smarterclayton:
--to-image-stream is the "porcelain" flow, if you do that you're assumed to know what you're doing.

/lgtm

can be improved in the future

wking:
[Upgrade failure][1] is [1690087][2].

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_oc/95/pull-ci-openshift-oc-master-e2e-aws-upgrade/274
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1690087

pkg/cli/admin/release/mirror.go
```diff
@@ -453,8 +453,10 @@ func (o *MirrorOptions) Run() error {
 		fmt.Fprintf(o.Out, "\nSuccess\nUpdate image:  %s\nMirrored to: %s\n", to, o.To)
 	}
 
-	if err := printImageContentInstructions(o.Out, o.From, o.To, repositories); err != nil {
-		return fmt.Errorf("Error creating mirror usage instructions: %v", err)
+	if len(o.From) > 0 && len(o.To) > 0 {
```
smarterclayton: The code in printimageContentInstructions is wrong, o.From is not required to be set.  So this PR should change the method to not error out if from is empty in printImageContentInstructions, and then we should add a boolean that says whether to print this thing, and set it true here and it can remain default false in new.go

smarterclayton:
Removed hold and lgtm, left another comment as we talked through it.

pkg/cli/admin/release/mirror.go
wking: >  So this PR should change the method to not error out if from is empty...

Done with a4d060b77 -> da15a34ed.

> ... and then we should add a boolean that says whether to print this thing, and set it true here and it can remain default false in new.go.

I'm still not clear on why we want this.  If `from` is empty, what would you say?

soltysh: In the regular flow, `from` will always be set, b/c that's how mirror command works and w/o it it fails `Complete` invocation. The fact that you have the `from` being empty check is just to ensure we don't do stupid things. The boolean is to completely skip `printImageContentInstructions` invocation in here.

wking: I've pushed da15a34ed -> 4baf093e6 with a new `PrintImageContentInstructions`, now that I understand the non-empty-`repositories` angle.

wking:
Should have addressed the outstanding request with [this][1].  Let me know if you still need more changes.

[1]: https://github.com/openshift/oc/pull/95#discussion_r325908375

PR #99, opened 2019-09-19
Test vendored kubectl too

soltysh:
This builds on top of https://github.com/openshift/oc/pull/98, only last commit matters here. 

/assign @tnozicka


Makefile
```diff
@@ -20,6 +20,10 @@ GO_BUILD_PACKAGES :=$(strip \
 	./cmd/... \
 	$(wildcard ./tools/*) \
 )
+GO_TEST_PACKAGES :=$(strip \
+	./cmd/... \
```
tnozicka: missing `./pkg/...`

soltysh: Yeah, that was meant to be `./...` actually, like in `GO_PACKAGES ?=./...`

Makefile
```diff
@@ -20,6 +20,10 @@ GO_BUILD_PACKAGES :=$(strip \
 	./cmd/... \
 	$(wildcard ./tools/*) \
 )
+GO_TEST_PACKAGES :=$(strip \
```
tnozicka: I'd do a separate target for those given how isolated they are and how much time they take
```
test-unit-kubectl: GO_TEST_PACKAGES:=./vendor/k8s.io/kubernetes/pkg/kubectl/...
test-unit-kubectl: test-unit
.PHONY: test-unit-kubectl
```

soltysh: No need for separate if the time diff between with and w/o is small:
with kubectl:
```
real	4m58.919s
user	13m48.102s
sys	1m26.500s
```
w/o kubectl:
```
real	4m49.915s
user	10m16.252s
sys	1m6.698s
```

tnozicka: I can go for unified if the difference is negligible, the original estimate from you seemed much worse

Makefile
```diff
@@ -20,6 +20,10 @@ GO_BUILD_PACKAGES :=$(strip \
...
+	./... \
+	./vendor/k8s.io/kubernetes/pkg/kubectl/... \
```
tnozicka: let's do it properly since the beginning then
```
$(shell $(GO) list -f '{{range .Deps}}{{printf "%s\n" .}}{{end}}' ./cmd/... | grep --color=never -E '^github.com/openshift/oc/vendor/(sigs\.)?k8s\.io/')
```

tnozicka: but i'd verify the timing

tnozicka:
/lgtm
for now (starting small)

PR #102, opened 2019-09-22
Bump kubernetes to 1.16.0-rc.2

soltysh:
/assign @sallyom @mfojtik @deads2k


soltysh:
Only failure is coming from the fact that `kubectl log` was removed (it was deprecated since before k8s 1.0) and tests should be using `kubectl logs`. I've synced with @mfojtik and we'll wait for origin bump to merge first then we will re-kick this and merge only after.

soltysh:
https://github.com/openshift/release/pull/5017 will drop verify-deps which won't pass since we switched to go.mod.

soltysh:
Previous merged, not sure about those numerous aws failures. 
/retest

tnozicka:
the command is `/override` but you should disable the check in release repo when bringing gomod

PR #105, opened 2019-09-25
release: Strip leading bracket indicators from changelogs

smarterclayton:
Prefixes like `[release-4.3]`, `[4.2]`, `[blocker]` are typically
informational only - to allow a little more flexibility from
contributors in PR titles strip those prefixes. Also support lists
of bugs and mixed case for `Bug` or the use of `Bugs`.

/assign @eparis
/cherrypick release-4.2
/cherrypick release-4.1

### DIFF
--- a/pkg/cli/admin/release/git.go
+++ b/pkg/cli/admin/release/git.go
@@ -13,6 +13,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -25,6 +26,10 @@ import (
 // calls. Consider improving pkg/git in the future.
 type git struct {
 	path string
+}
+
+type gitInterface interface {
+	exec(command ...string) (string, error)
 }
 
 var noSuchRepo = errors.New("location is not a git repo")
@@ -127,7 +132,7 @@ type MergeCommit struct {
 	ParentCommits []string
 
 	PullRequest int
-	Bug         int
+	Bugs        []int
 
 	Subject string
 }
@@ -143,18 +148,15 @@ func gitOutputToError(err error, out string) error {
 	return fmt.Errorf(out)
 }
 
-func mergeLogForRepo(g *git, repo string, from, to string) ([]MergeCommit, error) {
+var (
+	rePR     = regexp.MustCompile(`^Merge pull request #(\d+) from`)
+	reBug    = regexp.MustCompile(`^[B|b]ug(?:s)?\s([\d\s,]+)(?:-|:)\s*`)
+	rePrefix = regexp.MustCompile(`^(\[[\w\.\-]+\]\s*)+`)
+)
+
+func mergeLogForRepo(g gitInterface, repo string, from, to string) ([]MergeCommit, error) {
 	if from == to {
 		return nil, nil
-	}
-
-	rePR, err := regexp.Compile(`^Merge pull request #(\d+) from`)
-	if err != nil {
-		return nil, err
-	}
-	reBug, err := regexp.Compile(`^Bug (\d+)\s*(-|:)\s*`)
-	if err != nil {
-		return nil, err
 	}
 
 	args := []string{"log", "--merges", "--topo-order", "-z", "--pretty=format:%H %P%x1E%ct%x1E%s%x1E%b", fmt.Sprintf("%s..%s", from, to)}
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/release/git.go
```diff
@@ -143,20 +148,17 @@ func gitOutputToError(err error, out string) error {
 	return fmt.Errorf(out)
 }
 
-func mergeLogForRepo(g *git, repo string, from, to string) ([]MergeCommit, error) {
+var (
+	rePR     = regexp.MustCompile(`^Merge pull request #(\d+) from`)
+	reBug    = regexp.MustCompile(`^\s*[B|b]ug(?:s)?\s([\d\s,]+)(?:-|:)\s*`)
```
tnozicka: will `([\d\s,]+)` match if the number is missing and you put just a `,` or second whitespace?

I guess that's fine if these regexes are not to enforce the schema but only to parse as much as they can from what was already committed

pkg/cli/admin/release/git.go
```diff
@@ -205,15 +207,26 @@ func mergeLogForRepo(g *git, repo string, from, to string) ([]MergeCommit, error
...
-				return nil, fmt.Errorf("could not extract bug number from %q: %v", msg, err)
+			for _, part := range strings.Split(m[1], ",") {
+				for _, subpart := range strings.Split(part, " ") {
+					subpart = strings.TrimSpace(subpart)
+					if len(subpart) == 0 {
+						continue
+					}
+					bug, err := strconv.Atoi(subpart)
+					if err != nil {
+						continue
```
tnozicka: log the error at least at high level?

pkg/cli/admin/release/git_test.go
```diff
@@ -0,0 +1,135 @@
...
+			input: "abc\x1e1\x1eMerge pull request #145 from\x1e [release-4.1] Bugs , 17 43,564,: test",
+			want: []MergeCommit{
+				{
+					ParentCommits: []string{}, Commit: "abc", PullRequest: 145, CommitDate: time.Unix(1, 0).UTC(),
+					Bugs: []int{17, 43, 564}, Subject: "test",
+				},
+			},
+		},
+		{
+			input: "abc\x1e1\x1eMerge pull request #145 from\x1e [release-4.1] bugs , 17 43,564,: test",
```
tnozicka: not sure if we don't misinterpret typos where the bug was meant to be `1743`

pkg/cli/admin/release/git_test.go
```diff
@@ -0,0 +1,135 @@
...
+					Bugs: []int{17, 43, 564}, Subject: "test",
+				},
+			},
+		},
+		{
+			input: "abc\x1e1\x1eMerge pull request #145 from\x1e [release 4.1] bugs , 17 43,564,: test",
+			want: []MergeCommit{
+				{
+					ParentCommits: []string{}, Commit: "abc", PullRequest: 145, CommitDate: time.Unix(1, 0).UTC(),
+					Bugs: nil, Subject: "[release 4.1] bugs , 17 43,564,: test",
```
tnozicka: don't we want `[release 4.1]` striped? is the regex intentionally missing the space?

pkg/cli/admin/release/git.go
eparis: I don't like that we are striping leading whitespace, especially since you trimmed the string before you used this regexp.  Can we kill the first `\s*` ?

eparis: reBug    = regexp.MustCompile(`^\s*[B|b]ug(?:s)?`\s`([\d\s,]+)(?:-|:)\s*`)
  Does this mean that `Bug:123456` wouldn't work? Since there is no space?

pkg/cli/admin/release/git_test.go
smarterclayton: Yes, we decided not to strip those because it gets much more complex.

pkg/cli/admin/release/git_test.go
smarterclayton: We have some examples of people doing `Bug A B C D:` and rather than the code being too complicated (too many rules) basically just broke this into separator.

pkg/cli/admin/release/git.go
smarterclayton: Can relax and tighten, yes.

smarterclayton: Bug:153245 hasn't worked no.  Because we only allow dash and colon after the bug number

smarterclayton: Do you want to allow ` Bug 123:` at all (leading whitespace)?  That's what the `\s*` grants

smarterclayton: The bot allows leading whitespace (is what I mean)

pkg/cli/admin/release/git.go
smarterclayton: Technically this is impossible with the current regex.

pkg/cli/admin/release/git.go
eparis: `^[B|b]ugs?\s([\d\s,]+)[:-]\s*`   seems correct and more readable to me.  (no leaving whitespace since that gets cleaned https://github.com/openshift/oc/pull/105/files#diff-17e2ea2595b28107c29a952fdfe5d9cdR210 no?

smarterclayton: I agree it's more readable, but does the bot block that?  The issue here is the bot being a bit promiscuous and just letting any old PR merge, proper English or no.

PR #106, opened 2019-09-26
oc.spec: Update to Go 1.13 and make it consumable by 3rd party build services

LorbusChris:
oc.spec: Update Go to version 1.13

Also moves the Source0 definition below the Name definition so it can be
evaluated correctly by build tools like rpkg.

Also adds versions to Obsoletes and Provides declarations.

### DIFF
--- a/oc.spec
+++ b/oc.spec
@@ -6,8 +6,7 @@
 %global gopath      %{_datadir}/gocode
 %global import_path github.com/openshift/oc
 
-%global golang_version 1.12
-%global product_name OpenShift
+%global golang_version 1.13
 
 %{!?version: %global version 0.0.1}
 %{!?release: %global release 1}
@@ -28,15 +27,16 @@
 %global make %{os_git_vars} && make SOURCE_GIT_TAG:="${OS_GIT_VERSION}" SOURCE_GIT_COMMIT:="${OS_GIT_COMMIT}" SOURCE_GIT_MAJOR:="${OS_GIT_MAJOR}" SOURCE_GIT_MINOR:="${OS_GIT_MINOR}" SOURCE_GIT_TREE_STATE:="${OS_GIT_TREE_STATE}"
 %endif
 
-%if ! 0%{?local_build:1}
-Source0:        https://%{import_path}/archive/%{commit}/%{name}-%{version}.tar.gz
-%endif
 Name:           openshift-clients
 Version:        %{version}
 Release:        %{release}%{dist}
 Summary:        OpenShift client binaries
 License:        ASL 2.0
 URL:            https://%{import_path}
+
+%if ! 0%{?local_build:1}
+Source0:        https://%{import_path}/archive/%{commit}/%{name}-%{version}.tar.gz
+%endif
 
 # If go_arches not defined fall through to implicit golang archs
 %if 0%{?go_arches:1}
@@ -45,13 +45,12 @@ ExclusiveArch:  %{go_arches}
 ExclusiveArch:  x86_64 aarch64 ppc64le s390x
 %endif
 
-#BuildRequires:  bsdtar
 BuildRequires:  golang >= %{golang_version}
 BuildRequires:  krb5-devel
 BuildRequires:  rsync
 
-Provides:       atomic-openshift-clients
-Obsoletes:      atomic-openshift-clients
+Provides:       atomic-openshift-clients = %{version}
+Obsoletes:      atomic-openshift-clients <= %{version}
 Requires:       bash-completion
 
 %description
@@ -59,14 +58,13 @@ Requires:       bash-completion
 
 %package redistributable
 Summary:        OpenShift Client binaries for Linux, Mac OSX, and Windows
-Provides:       atomic-openshift-clients-redistributable
-Obsoletes:      atomic-openshift-clients-redistributable
+Provides:       atomic-openshift-clients-redistributable = %{version}
+Obsoletes:      atomic-openshift-clients-redistributable <= %{version}
 
 %description redistributable
 %{summary}
 
 %prep
-
 %if ! 0%{?local_build:1}
 %setup -q
 %endif


Makefile
```diff
@@ -25,14 +25,16 @@ GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi'
...
 OUTPUT_DIR :=_output
 CROSS_BUILD_BINDIR :=$(OUTPUT_DIR)/bin
 RPM_VERSION :=$(shell set -o pipefail && echo '$(SOURCE_GIT_TAG)' | sed -E 's/v([0-9]+\.[0-9]+\.[0-9]+)-.*/\1/')
 RPM_EXTRAFLAGS := \
-	--define 'local_build true' \
+	--define 'local_build $(LOCAL_BUILD)' \
+	--define 'ose_build $(OSE_BUILD)' \
 	--define 'os_git_vars ignore' \
 	--define 'version $(RPM_VERSION)' \
-	--define 'dist .el7' \
```
LorbusChris: is it safe to remove this override here?

oc.spec
```diff
@@ -7,17 +7,24 @@
 %global import_path github.com/openshift/oc
 
 %global golang_version 1.12
+
+%if 0%{?ose_build}
```
LorbusChris: ose or ocp build?

oc.spec
```diff
@@ -28,51 +35,56 @@
...
 BuildRequires:  rsync
 
-Provides:       atomic-openshift-clients
-Obsoletes:      atomic-openshift-clients
+Provides:       atomic-openshift-clients = %{version}
+Obsoletes:      atomic-openshift-clients = %{version}
 Requires:       bash-completion
 
+%if 0%{?ose_build}
+Conflicts:      okd-clients
```
LorbusChris: be explicit and add this, or leave this out?

LorbusChris: ref discussion at https://github.com/openshift/origin/pull/23747/files#r321980453

smarterclayton: `ocp_build`

smarterclayton: I would want the package names to stay the same `openshift-clients`

oc.spec
```diff
@@ -28,51 +35,56 @@
...
+%endif
+
 %description
 %{summary}
 
 %package redistributable
-Summary:        OpenShift Client binaries for Linux, Mac OSX, and Windows
-Provides:       atomic-openshift-clients-redistributable
-Obsoletes:      atomic-openshift-clients-redistributable
+Summary:        %{product_name} Client binaries for Linux, Mac OSX, and Windows
```
smarterclayton: Hrm, technically no matter what this is still openshift - the binaries say openshift in them.  I don't think you need to change this, the same code works for all versions.

LorbusChris:
Version 0.0.1 is currently hardcoded in the build-rpms-form-tar CI job (https://github.com/openshift/release/blob/be30696aaa816ae3086d305bdcc53150e3a6cd34/ci-operator/config/openshift/oc/openshift-oc-master.yaml#L75)

Probably best to pull that into a script here in this repo and have the version be updated dynamically. Can be done as a follow up imo.

LorbusChris:
while this is not strictly necessary for the way we build things now, this PR makes it possible to build the RPM on other envs and e.g. on the COPR build service.

Can somebody have another look?

Makefile
LorbusChris: The dist tag is not usually hardcoded but determined by the build env. On the current OCP 4.5 base image `%{?dist}` expands to `.el7`, so this wouldn't change until e.g. UBI8 is used to build the rpm.

tnozicka:
> Version 0.0.1 is currently hardcoded in the build-rpms-from-tar CI job (https://github.com/openshift/release/blob/be30696aaa816ae3086d305bdcc53150e3a6cd34/ci-operator/config/openshift/oc/openshift-oc-master.yaml#L75)
Probably best to pull that into a script here in this repo and have the version be updated dynamically. Can be done as a follow up imo.

I am not sure if that gets used anywhere. This should be just verifying that we can build from tar so we don't break ART builds. so the version is likely fine for test unless you see it propagated somewhere.

oc.spec
```diff
@@ -28,51 +27,48 @@
...
 %else
 ExclusiveArch:  x86_64 aarch64 ppc64le s390x
 %endif
 
-#BuildRequires:  bsdtar
 BuildRequires:  golang >= %{golang_version}
 BuildRequires:  krb5-devel
 BuildRequires:  rsync
 
-Provides:       atomic-openshift-clients
```
tnozicka: Would you mind explaining the change?

Makefile
tnozicka: this is for dev builds (e.g.  if you are on fedora) so I don't think it breaks but I don't see how that helps okd

LorbusChris: It doesn't. The approach has changed and we're currently not doing separate oc rpm builds for OKD.

Makefile
```diff
@@ -25,14 +25,14 @@ GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi'
 GO_BUILD_FLAGS_DARWIN :=-tags 'include_gcs include_oss containers_image_openpgp'
 GO_BUILD_FLAGS_WINDOWS :=-tags 'include_gcs include_oss containers_image_openpgp'
 
+LOCAL_BUILD ?=1
```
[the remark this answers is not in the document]
LorbusChris: not needed anymore

oc.spec
LorbusChris: done

LorbusChris: it's obsolete, but might as well leave it there. Instead I've now added the versions to both Provides and Obsoletes declarations to conform to guidelines.

oc.spec
```diff
@@ -7,7 +7,6 @@
 %global import_path github.com/openshift/oc
 
 %global golang_version 1.12
```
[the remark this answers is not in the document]
LorbusChris: done

LorbusChris:
@tnozicka thanks for the review! I've updated the commit, ptal

PR #42, opened 2019-08-05
Add oc adm inspect command

sanchezl:
Move `inspect` command from `openshift-must-gather` to `oc adm`.

PR #107, opened 2019-09-27
Add oc adm inspect command

soltysh:
This is fixed version of #42. 

Move inspect command from openshift-must-gather to oc adm.

/assign @sanchezl


soltysh:
Self-tagging since this was already approved in the other instance.

PR #109, opened 2019-09-27
Bug 1756453: Separate upgrade flags for safety instead of abusing force

smarterclayton:
The --force flag is dangerous and potentially allows untrusted content to be upgraded to accidentally. Instead, introduce two new flags `--allow-explicit-upgrade` (for upgrading to something not in availableVersions) and `--allow-upgrade-with-warnings` (for upgrading when another upgrade is in progress or the cluster is reporting an error) and remove those checks from `--force`.

While this is an API change, it is necessary to ensure that users do not accidentally get access to untrusted content when performing upgrades across major versions in advance of graph updates, or when they are upgrading in disconnected environments.

/assign @jwforres
@eparis as discussed today

This should be a candidate for backport to 4.2 and 4.1. While it changes flag behavior, it will only break things in a "safe" way. This addresses a significant security challenge as we have introduced new meanings for --force.

```
# install 4.1.15

$ oc adm upgrade --to-image quay.io/openshift-release-dev/ocp-release:4.1.16
error: The requested upgrade image is not one of the available updates, you must pass --allow-explicit-upgrade to continue
$ oc adm upgrade --to-image quay.io/openshift-release-dev/ocp-release:4.1.16 --allow-explicit-upgrade

... // wait

$ oc adm upgrade --to-image quay.io/openshift-release-dev/ocp-release:4.1.7
error: The requested upgrade image is not one of the available updates, you must pass --allow-explicit-upgrade to continue
$ oc adm upgrade --to-image quay.io/openshift-release-dev/ocp-release:4.1.7 --allow-explicit-upgrade
error: Already upgrading, pass --allow-upgrade-with-warnings to override.

  Reason:
  Message: Working towards 4.1.16: 13% complete

$ oc adm upgrade --to-image quay.io/openshift-release-dev/ocp-release:4.1.7 --allow-explicit-upgrade --allow-upgrade-with-warnings
Updating to release image quay.io/openshift-release-dev/ocp-release:4.1.7
```

PR #111, opened 2019-09-27
Bug 1756454: Separate upgrade flags for safety instead of abusing force

smarterclayton:
The --force flag is dangerous and potentially allows untrusted
content to be upgraded to accidentally. Instead, introduce two
new flags `--allow-explicit-upgrade` (for upgrading to something not
in availableVersions) and `--allow-unsafe-upgrade` (for upgrading
when another upgrade is in progress or the cluster is reporting
an error) and remove those checks from `--force`.

While this is an API change, it is necessary to ensure that users
do not accidentally get access to untrusted content when
performing upgrades across major versions in advance of graph
updates, or when they are upgrading in disconnected environments.

Backport of #109


abhinavdahiya:
There also this bug https://bugzilla.redhat.com/show_bug.cgi?id=1713263 that states that `--to-latest` and `--force` don't work together correctly? Is that expected @smarterclayton ?

jwforres:
Given the state of CI on master and today's deadline, and that manual testing has been done to verify this on 4.2, overriding bugzilla bot.

PR #112, opened 2019-09-27
Bug 1754042: Retry 429 and 503 errors in registry client

smarterclayton:
Add general across the board retries on 429 and 503 in registry client
commands that share the context.

Depends on openshift/library-go#534


smarterclayton:
@dmage bumped vendor, updated with retries.  PTAL

pkg/cli/image/manifest/manifest.go
```diff
@@ -87,6 +87,7 @@ func (o *SecurityOptions) Context() (*registryclient.Context, error) {
 	}
 	context, err := o.NewContext()
 	o.CachedContext = context
+	o.CachedContext.Retries = 3
```
dmage: Maybe it's better to do it as the last step in `o.NewContext()`?

smarterclayton: `Context()` is "I want the defaults for this command" while `NewContext` is "read from options only".  So Context

PR #113, opened 2019-09-30
Bug 1756107: Make container endpoint collection errors non-fatal

mfojtik:
Makes the endpoint collection non-fatal.

### DIFF
--- a/pkg/cli/admin/inspect/pod.go
+++ b/pkg/cli/admin/inspect/pod.go
@@ -75,7 +75,7 @@ func (o *InspectOptions) gatherContainerInfo(destDir string, pod *corev1.Pod, co
 	}
 
 	if err := o.gatherContainerEndpoints(path.Join(destDir, "/"+container.Name), pod, &container, port); err != nil {
-		return err
+		log.Printf("        Skipping one or more container endpoint collection for pod %q container %q: %v\n", pod.Name, container.Name, err)
 	}
 
 	return nil


pkg/cli/admin/inspect/pod.go
```diff
@@ -75,7 +75,8 @@ func (o *InspectOptions) gatherContainerInfo(destDir string, pod *corev1.Pod, co
 	}
 
 	if err := o.gatherContainerEndpoints(path.Join(destDir, "/"+container.Name), pod, &container, port); err != nil {
-		return err
+		log.Printf("        Skipping container endpoint collection for pod %q container %q: %v\n", pod.Name, container.Name, err)
```
tnozicka: `skipping` might sound misleading given it could have already collected some of the endpoints

pkg/cli/admin/inspect/pod.go
```diff
@@ -75,7 +75,8 @@ func (o *InspectOptions) gatherContainerInfo(destDir string, pod *corev1.Pod, co
...
+		return nil
```
tnozicka: nit: looking at endpoints collection it doesn't look fatal so I'd not return here and let it continue, although that doesn't matter much when it is currently at the end of the function (can change over time)

pkg/cli/admin/inspect/pod.go
mfojtik: "skipping one or more" ;-)

mfojtik:
@tnozicka fixed

PR #114, opened 2019-09-30
inspect: collect /debug/pprof data

mfojtik:
This will teach inspect to download /debug/pprof/heap, profile and trace for containers that expose that endpoint using the metrics port (usually the apiserver).

/cc @deads2k 

(do not merge, pending manual testing)

### DIFF
--- a/pkg/cli/admin/inspect/pod.go
+++ b/pkg/cli/admin/inspect/pod.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


deads2k:
This is a clever idea, but I'll leave the choice to @soltysh.  You should update the OEP with this functionality, describing what it does.

soltysh:
> This is a clever idea, but I'll leave the choice to @soltysh. You should update the OEP with this functionality, describing what it does.

:+1: and yes updated OEP is appreciated.

PR #115, opened 2019-10-01
inspect: capture events for each namespace

mfojtik:
This enhance the `oc adm inspect` by adding events collection per namespace. The events are sorted by timestamp and formatted as one-line log and stored in `./events` file next to the namespace YAML file.

/cc @deads2k

### DIFF
--- a/pkg/cli/admin/inspect/events.go
+++ b/pkg/cli/admin/inspect/events.go
@@ -1,0 +1,28 @@
+package inspect
+
+import (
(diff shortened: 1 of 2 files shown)


deads2k:
I'm not a fan of this.  We already gather events: https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/namespace.go#L22 and we should focus on tools that slice events from any input source.

/close

PR #117, opened 2019-10-01
Bug 1540560: fix oc status for daemonsets

soltysh:
/assign @sallyom


sallyom:
/lgtm

confirmed this: https://bugzilla.redhat.com/show_bug.cgi?id=1540560#c16  (all other comments in bz have been verified AFAICT)

PR #120, opened 2019-10-05
mirror: Add --to-dir to release mirror, uses skopeo to archive

smarterclayton:
Each tag is written as a separate docker archive. A user could then 
trivially upload these. Passing --dry-run prints the commands to run. If
skopeo is not present the command reports an error.


smarterclayton:
Fixed verify, reapplied tag.

I may come back to this and iterate based on feedback, since our 6GB payload is 30GB if you don't skip default layers and that kind of sucks.  On rhel/fedora boxes we may be able to load these into container storage (and thus avoid the duplicates).  --to-dir will stay because this is the closest we have to "download" that works on all systems.

PR #122, opened 2019-10-09
Display MOTD on interactive actions (debug, exec, rsh, run, attach)

JAORMX:
Whenever one of the aforementioned commands is executed, and the command
is executed on interactive mode, the MOTD will be displayed. This is
similar to the behavior that SSH provides, where running a single
command on a node will not trigger this behavior, but initiating a
session will.

This work is done to address compliance requirements.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -52,6 +52,7 @@ import (
 	"github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/oc/pkg/helpers/conditions"
 	utilenv "github.com/openshift/oc/pkg/helpers/env"
+	"github.com/openshift/oc/pkg/helpers/motd"
 	generateapp "github.com/openshift/oc/pkg/helpers/newapp/app"
 )
 
@@ -499,6 +500,11 @@ func (o *DebugOptions) RunDebug() error {
 			}
 
 			return false, nil
+		}
+
+		if err = motd.DisplayMOTD(o.CoreClient, o.ErrOut); err != nil {
+			fmt.Fprintf(o.ErrOut, "Unable to display the MOTD: %v", err)
+			return err
 		}
 
 		ctx, cancel := context.WithTimeout(context.Background(), o.Timeout)
--- a/pkg/cli/kubectlwrappers/wrappers.go
+++ b/pkg/cli/kubectlwrappers/wrappers.go
@@ -3,6 +3,7 @@ package kubectlwrappers
 import (
 	"bufio"
 	"flag"
+	"fmt"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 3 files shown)


jhrozek:
Of course I'll leave the review for the core team members, but perhaps it would be useful and speed up the review to see how the code behaves. FWIW, the code looks good to me.

1. `oc debug` always displays the message:
```
oc debug node/ip-10-0-143-135.ec2.internal
Starting pod/ip-10-0-143-135ec2internal-debug ...

Hello world!
Pod IP: 10.0.143.135
sh-4.4# 
```
This is true even if `-T` is passed and no tty is allocated:
```
oc debug -T node/ip-10-0-143-135.ec2.internal
Starting pod/ip-10-0-143-135ec2internal-debug ...

Hello world!

Pod IP: 10.0.143.135

^C
```

2.  `oc exec` only displays the motd only if both `-i` and `-t` are passed, e.g.:
```
oc exec -it ip-10-0-143-135ec2internal-debug ls /

Hello world!
bin   dev  home  lib    media  opt   root  sbin  sys  usr
boot  etc  host  lib64  mnt    proc  run   srv   tmp  var
```
but:
```
oc exec ip-10-0-143-135ec2internal-debug ls / 
bin
boot
dev
...
```

3. `oc rsh` displays the legal notice unless `-T` is selected:
```
oc rsh ip-10-0-143-135ec2internal-debug cat /etc/resolv.conf

Hello world!
search ec2.internal
nameserver 10.0.0.2
```
but:
```
oc rsh -T ip-10-0-143-135ec2internal-debug cat /etc/resolv.conf
search ec2.internal
nameserver 10.0.0.2
```

I'm mentioning all the tty/notty flags explicitly because [a comment](https://github.com/openshift/oc/pull/67#issuecomment-524674640) in the previous related PR raised the concern of changing scripting behaviour of exec or rsh.

JAORMX:
@soltysh any feedback on this?

JAORMX:
@soltysh thanks for the feedback!

-  I don't think `rsync` is applicable, given that it's a non-interactive command, and modifying it would disrupt the output that folks already expect.
- I'll add `run` and `attach` here! good catch!

jhrozek:
btw the latest iteration seems to work for both rsh and attach, too, as long as `i` and `t` are selected:
```
$ oc run -i -t busybox --image=busybox --restart=Never
Hello world
```
and:
```
$ oc attach -it busybox
Hello world
```

akostadinov:
For `debug` there is a bug [1] that everything goes to `stdout` and nothing to `stderr`. If it is not fixed, then displaying the MOTD will certainly break scripts.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1771549

JAORMX:
@akostadinov thanks for chiming in! One thing to consider is that there wouldn't be an MOTD by default. So folks with their default deployments would be just fine. It's something that folks have to specify.

Also, the debug command only outputting to stderr applies only to the output of the container session or the command ran on that container. stderr coming from the `oc` command does indeed go to stderr.

e.g.

```
$ oc debug pod/log-generator-77d66774fb-qt98k  2>  stderr.txt
~ $ exit
$ cat stderr.txt
Starting pod/log-generator-77d66774fb-qt98k-debug ...

This is a legal notice!!!
Stop doing illegal stuff
Pod IP: 10.129.2.72

Removing debug pod ...

```
So if folks rely on stdout for the output of the command (which seems to be the case), this won't affect them.

mrogers950:
LGTM.

PR #123, opened 2019-10-09
registry-login: Allow login with credentials

smarterclayton:
We have no way to login to an arbitrary registry, like quay.io. This
isn't terribly common, but feels like a gap in a command called
`registry login`. This allows `--basic` to provide `user:password`
instead of service account or the user's ID. You can now run:

    $ oc registry login --to=/tmp/config --registry=quay.io --auth-basic foo:bar
    Saved credentials for quay.io

Now outputs a print message.


sallyom:
This is to use `oc registry login ..` with any registry, right? 
similar to `docker login -u="name" -p="blah" quay.io`  
There should be some output upon success, like
`Login Succeeded` 
I've tried this pr and I see my config file is updated, but I'm missing the expected success message.

smarterclayton:
The command has never done that.  I'm not totally opposed to it, but it's not a new thing.

smarterclayton:
Updated to also print a message `Saved credentials for %s`

smarterclayton:
The check the command does verifies that the registry exists, is reachable.  However, there is no global API that we can safely check across all registries that your credentials work.  Docker logins check does not always work for non docker registries.

So this is WAD, and there is potentially room for future improvement.

smarterclayton:
Any other comments?

sallyom:
> The check the command does verifies that the registry exists, is reachable. However, there is no global API that we can safely check across all registries that your credentials work. Docker logins check does not always work for non docker registries.
> 
> So this is WAD, and there is potentially room for future improvement.

I've only tried w/ Docker, Podman logging into quay/docker to compare, but when I pass a wrong password, the bad credentials don't overwrite (possibly) good credentials.   We're ok with oc possibly overwriting good creds w/ bad ones?

smarterclayton:
Yes, we're not changing how the command works before this code landed.  Changing the type of check is more involved (although probably valuable) but it's more involved and impacts other code flows.

pkg/cli/registry/login/login.go
```diff
@@ -126,6 +132,17 @@ func NewRegistryLoginCmd(name string, f kcmdutil.Factory, streams genericcliopti
...
 func (o *LoginOptions) Complete(f kcmdutil.Factory, args []string) error {
+	credentials := 0
+	if len(o.ServiceAccount) > 0 {
+		credentials++
+	}
+	if len(o.AuthBasic) > 0 {
+		credentials++
+	}
+	if credentials > 1 {
+		return fmt.Errorf("You may only specify a single authentication input as -z or --auth-basic")
```
soltysh: lower case error

pkg/cli/registry/login/login.go
```diff
@@ -321,5 +344,6 @@ func (o *LoginOptions) Run() error {
 	if err := f.Close(); err != nil {
 		return err
 	}
+	fmt.Fprintf(o.Out, "Saved credentials for %s\n", o.HostPort)
```
soltysh: `Saved credentials for %s to %s\n`

pkg/cli/registry/login/login.go
smarterclayton: This is an end user message from the top of the command so it is capitalized like other usage errors

PR #127, opened 2019-10-15
Bug 1760670: Warn about idling if network policies are in place to all users

rcarrillocruz:


danwinship:
We had decided against this approach in https://github.com/openshift/origin/pull/22403 because:
1. the bug is specific to openshift-sdn, so we shouldn't print a warning if they are using ovn-kubernetes
2. it means that if you idle first and then create a NetworkPolicy, you don't get any warning

soltysh:
/lgtm cancel
based on what @danwinship said in the previous comment, @rcarrillocruz pls sync with networking team for the proper fix.

PR #129, opened 2019-10-18
Switch inspect to use unstructured when reading .status.relatedObjects

soltysh:
With this PR we should be able to:
1. Gather data from `.status.relatedObject` if passed object has it.
2. Use `-A` for `--all-namespaces`
3. Specify multiple resources, like so: `oc adm inspect deploy/downloads deploy/console -n openshift-console `.

Additionally, this switches code to use `klog` and cleans unnecessary flags that should not have been exposed before.

/assign @deads2k


pkg/cli/admin/inspect/inspect.go
```diff
@@ -251,12 +250,12 @@ func (o *InspectOptions) gatherConfigResourceData(destDir string, ctx *resourceC
...
 	// determine if we've already collected operatorResourceData
 	if ctx.visited.Has(operatorResourceDataKey) {
-		log.Printf("Skipping previously-collected operator.openshift.io resource data")
+		klog.V(1).Infof("Skipping previously-collected operator.openshift.io resource data")
 		return nil
 	}
 	ctx.visited.Insert(operatorResourceDataKey)
 
-	log.Printf("Gathering kubeapiserver.operator.openshift.io resource data...\n")
+	klog.V(1).Infof("Gathering kubeapiserver.operator.openshift.io resource data...\n")
```
deads2k: this message looks like an incorrect carry-over.

pkg/cli/admin/inspect/namespace.go
```diff
@@ -65,10 +65,10 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
 		resourcesToStore[gvr] = list
 	}
 
-	log.Printf("    Gathering pod data for namespace %q...\n", namespace)
+	klog.V(1).Infof("    Gathering pod data for namespace %q...\n", namespace)
```
deads2k: these messages exist because the process of collection takes a long time and we needed status indicators.

pkg/cli/admin/inspect/util.go
```diff
@@ -37,6 +38,15 @@ func objectReferenceToString(ref *configv1.ObjectReference) string {
 	return resource + group + name
 }
 
+func unstructuredToString(obj *unstructured.Unstructured) string {
+	resource := obj.GetKind()
```
deads2k: you should make the kind.group calculation.

deads2k:
/lgtm
/hold

hold in case you want to address nits here instead of a followup.  Your choice.

soltysh:
Updated.
/hold cancel

soltysh:
Self tagging based on previous comment.

PR #131, opened 2019-10-22
Bug 1763728: fix 'oc adm release extract' version injection

sallyom:
Since[ this](https://github.com/openshift/oc/pull/88) merged, when using an extracted oc to extract openshift-install|oc, command fails, because the shared binary replacement location.  This PR is the fix. 

/cc @soltysh

### DIFF
--- a/pkg/cli/admin/release/extract_tools.go
+++ b/pkg/cli/admin/release/extract_tools.go
@@ -40,9 +40,8 @@ type extractTarget struct {
 	Command  string
 	Optional bool
 
-	TargetName string
-
-	InjectReleaseImage bool
+	InjectReleaseImage   bool
+	InjectReleaseVersion bool
 
 	ArchiveFormat string
 	AsArchive     bool
@@ -150,30 +149,30 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli-artifacts", From: "usr/share/openshift/mac/oc"},
 
-			LinkTo:             []string{"kubectl"},
-			Readme:             readmeCLIUnix,
-			InjectReleaseImage: true,
-			ArchiveFormat:      "openshift-client-mac-%s.tar.gz",
+			LinkTo:               []string{"kubectl"},
+			Readme:               readmeCLIUnix,
+			InjectReleaseVersion: true,
+			ArchiveFormat:        "openshift-client-mac-%s.tar.gz",
 		},
 		{
 			OS:      "linux",
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli", From: "usr/bin/oc"},
 
-			LinkTo:             []string{"kubectl"},
-			Readme:             readmeCLIUnix,
-			InjectReleaseImage: true,
-			ArchiveFormat:      "openshift-client-linux-%s.tar.gz",
+			LinkTo:               []string{"kubectl"},
+			Readme:               readmeCLIUnix,
+			InjectReleaseVersion: true,
+			ArchiveFormat:        "openshift-client-linux-%s.tar.gz",
 		},
 		{
 			OS:      "windows",
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli-artifacts", From: "usr/share/openshift/windows/oc.exe"},
 
-			Readme:             readmeCLIWindows,
-			InjectReleaseImage: true,
-			ArchiveFormat:      "openshift-client-windows-%s.zip",
-			AsZip:              true,
+			Readme:               readmeCLIWindows,
+			InjectReleaseVersion: true,
+			ArchiveFormat:        "openshift-client-windows-%s.zip",
+			AsZip:                true,
 		},
 		{
 			OS:      "darwin",
@@ -492,19 +491,16 @@ func (o *ExtractOptions) extractCommand(command string) error {
 
 		// copy the input to disk
 		if target.InjectReleaseImage {
-			releaseInfo := exactReleaseImage
 			var matched bool
-			errmsg := ""
-			switch target.Command {
-			case "oc":
-				releaseInfo = releaseName
-				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, could not determine version", target.Command)
-			default:
-				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, installer will not be locked to the correct image", target.Command)
-			}
-			matched, err = copyAndReplaceReleaseInfo(w, r, 4*1024, releaseInfo)
+			matched, err = copyAndReplaceReleaseInfo(w, r, 4*1024, exactReleaseImage)
 			if !matched {
-				fmt.Fprintf(o.ErrOut, "%s\n", errmsg)
+				fmt.Fprintf(o.ErrOut, "warning: Unable to replace release image location into %s, installer will not be locked to the correct image\n", target.Command)
+			}
+		} else if target.InjectReleaseVersion {
+			var matched bool
+			matched, err = copyAndReplaceReleaseVersion(w, r, 4*1024, releaseName)
+			if !matched {
+				fmt.Fprintf(o.ErrOut, "warning: Unable to replace release version location into %s\n", target.Command)
 			}
 		} else {
 			_, err = io.Copy(w, r)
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/release/extract_tools.go
```diff
@@ -674,6 +686,7 @@ func copyAndReplaceReleaseInfo(w io.Writer, r io.Reader, bufferSize int, release
 		}
 		if err != nil {
 			if err == io.EOF || err == io.ErrUnexpectedEOF {
+				klog.V(2).Infof("GOT HERE")
```
sallyom: oops, will remove but want to see it run thru CI/get a release image from this first.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -492,17 +492,16 @@ func (o *ExtractOptions) extractCommand(command string) error {
 
 		// copy the input to disk
 		if target.InjectReleaseImage {
```
smarterclayton: Why are we using release image and release image name under the same flag?  They are unrelated - oc and Openshift-install should be using this flag for replace, rather than checking on binary name.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -492,17 +492,16 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
-			case "oc":
-				releaseInfo = releaseName
-				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, could not determine version", target.Command)
-			default:
+			case "openshift-install":
 				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, installer will not be locked to the correct image", target.Command)
+				matched, err = copyAndReplaceReleaseInfo(w, r, 4*1024, exactReleaseImage)
+			default:
+				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, could not determine version", target.Command)
+				matched, err = copyAndReplaceReleaseInfo(w, r, 4*1024, releaseName)
```
smarterclayton: Version and image are unrelated and should not be in the same code path.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -624,22 +623,35 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
+	// try for exactReleaseImageBinaryReplacement first, for backwards compatibility
+	if len(releaseInfo)+1 > len(exactReleaseImageBinaryReplacement) {
 		return false, fmt.Errorf("the release image pull spec is longer than the maximum replacement length for the binary")
 	}
-	if bufferSize < len(binaryReplacement) {
-		return false, fmt.Errorf("the buffer size must be greater than %d bytes", len(binaryReplacement))
+	if bufferSize < len(exactReleaseImageBinaryReplacement) {
+		return false, fmt.Errorf("the buffer size must be greater than %d bytes", len(exactReleaseImageBinaryReplacement))
 	}
+	matched, err := copyReplace(w, r, releaseInfo, exactReleaseImageBinaryReplacement, bufferSize)
```
smarterclayton: Calling this twice is not appropriate for very large binaries.  These two code paths (version and image) should be split into separate functions.

pkg/cli/admin/release/extract_tools.go
sallyom: good point, thanks! _reworking_

pkg/cli/admin/release/extract_tools.go
sallyom: i know.. was doing that now..

pkg/cli/admin/release/extract_tools.go
sallyom: ok, will rework

sallyom:
> I'd like some structural changes to separate these out, if you're using a flag on the binary to denote what to replace they should be split by version or image, and you should only replace when denoted.

thanks for taking a look, will do!

sallyom:
bug is fixed w/ this, i'll simplify code b4 merging, there's some repeated code

sallyom:
/retest

to get a release image for testing - still have some work to do w/ this

pkg/cli/admin/release/extract_tools.go
sallyom: pushed an update, added `InjectReleaseVersion` bool

sallyom:
added image built from this PR to latest published release, then used oc built from this pr to extract the release, then used the extracted oc to extract openshift-install:
```console
$ oc adm release extract --command oc quay.io/sallyom/release:test
$ ./oc version
Client Version: 4.3.0-0.ci-2019-10-23-103858
Server Version: 4.3.0-0.ci-2019-10-23-061559
Kubernetes Version: v1.16.0-beta.2+3012553
$ ./oc adm release extract --command openshift-install quay.io/sallyom/release:test
$ ./openshift-install version
./openshift-install unreleased-master-2003-g42a87f05c2e1a42ad815ec8f37b839fd5952cbfc-dirty
built from commit 42a87f05c2e1a42ad815ec8f37b839fd5952cbfc
release image quay.io/sallyom/release@sha256:6ef6426b9181001bee5bc6d693a8205e0bcff55ef4469b99ba7ada66cb7c8811
```

pkg/version/version.go
```diff
@@ -63,40 +64,43 @@ func Get() version.Info {
...
 	}
-	releaseName := defaultReleaseInfoPadded[:nullTerminator]
+	releaseName := defaultVersionPadded[:nullTerminator]
 	if len(releaseName) == 0 {
 		// the binary has been altered, but the replaced release name is empty which is incorrect
-		return version.Info{}, "", fmt.Errorf("release name location is empty, this binary was incorrectly generated")
+		// the oc binary will not be pinned to Release Metadata:Version, instead it will output the version
+		// as it would have without being extracted (the Git version info) 
+		klog.V(2).Infof("DEBUG: %v", defaultVersionPadded)
+		klog.Warningf("oc version was incorrectly replaced during extract")
```
sallyom: I can't figure this out, when I pull the image from this PR, I get an empty releaseName (match found, but len(releaseName)==0.  However, when I create a release from a published release w/ new cli image (built from this PR, but locally), and extract oc from that, releaseName is the expected `4.3.0-0.ci-2019-10-23-blah`  Adding debug to see what's going on, will remove

pkg/version/version.go
```diff
@@ -63,40 +64,42 @@ func Get() version.Info {
...
 		return version.Info{}, "", fmt.Errorf("release name location contains no null-terminator and constant is corrupted")
 	}
-	releaseName := defaultReleaseInfoPadded[:nullTerminator]
+	releaseName := defaultVersionPadded[:nullTerminator]
 	if len(releaseName) == 0 {
 		// the binary has been altered, but the replaced release name is empty which is incorrect
-		return version.Info{}, "", fmt.Errorf("release name location is empty, this binary was incorrectly generated")
+		// the oc binary will not be pinned to Release Metadata:Version, instead it will output the version
+		// as it would have without being extracted (the Git version info) 
+		klog.Warningf("oc version was incorrectly replaced during extract: DEBUG: %v", defaultVersionPadded)
```
sallyom: I can't figure this out, when I extract oc from the release image from this PR, I get an empty releaseName (match found, but len(releaseName)==0. However, when I create a release from a published release w/ new cli image (built from this PR, but locally), and extract oc from that, releaseName is the expected 4.3.0-0.ci-2019-10-23-blah Adding debug to see what's going on, will remove

sallyom: and this should return an error, not only a Warning? The oc binary doesn't depend on the version being pinned to the release, so I can see it as just a warning, but if we're meddling w/ the binary and the result isn't what is expected, it should probably error? `oc` works the same either way, though, so I'm not sure what's best.

pkg/version/version.go
```diff
@@ -63,40 +63,41 @@ func Get() version.Info {
...
-	if strings.HasPrefix(defaultReleaseInfoPadded, defaultReleaseInfoPrefix) {
+	if strings.HasPrefix(defaultVersionPadded, defaultVersionPrefix) {
 		return Get(), "", nil
 	}
-	nullTerminator := strings.IndexByte(defaultReleaseInfoPadded, '\x00')
+	nullTerminator := strings.IndexByte(defaultVersionPadded, '\x00')
 	if nullTerminator == -1 {
 		// the binary has been altered, but we didn't find a null terminator within the release name constant which is an error
-		return version.Info{}, "", fmt.Errorf("release name location was replaced but without a null terminator before %d bytes", defaultReleaseInfoLength)
+		return Get(), "", fmt.Errorf("release name location was replaced but without a null terminator before %d bytes", defaultVersionLength)
```
sallyom: The oc binary doesn't depend on the version being pinned to the release, but if we're meddling w/ the binary and the result isn't what is expected, it should return an error? `oc` works the same either way, though, so I could see returning a warning for these, instead.  Also, setting version to the Git version info rather than nothing, in these.

sallyom: I can't figure this out, when I extract oc from the release image from this PR, I get an empty releaseName (match found, but len(releaseName)==0. However, when I create a release from a published release w/ new cli image (built from this PR, but locally), and extract oc from that, releaseName is the expected 4.3.0-0.ci-2019-10-23-blah - maybe the ci-op/release:lastes image is in flux w/ multiple pushes/test restarts so the Release Metadata: Version is absent? not sure.

pkg/version/version.go
```diff
@@ -63,40 +63,41 @@ func Get() version.Info {
...
 		// the binary has been altered, but the null terminator is *longer* than the constant encoded in the binary
 		return version.Info{}, "", fmt.Errorf("release name location contains no null-terminator and constant is corrupted")
 	}
-	releaseName := defaultReleaseInfoPadded[:nullTerminator]
+	releaseName := defaultVersionPadded[:nullTerminator]
 	if len(releaseName) == 0 {
 		// the binary has been altered, but the replaced release name is empty which is incorrect
-		return version.Info{}, "", fmt.Errorf("release name location is empty, this binary was incorrectly generated")
+		// the oc binary will not be pinned to Release Metadata:Version
+		return version.Info{}, "", fmt.Errorf("oc version was incorrectly replaced during extract")
```
sallyom: The oc binary doesn't depend on the version being pinned to the release, but if we're meddling w/ the binary and the result isn't what is expected, it should return an error? oc works the same either way, though, so I could see return    ing a warning for these, instead.

sallyom: I can't figure this out, when I extract oc from the release image from this PR, I get an empty releaseName (match found, but len(releaseName)==0. However, when I create a release from a published release w/ new cli image (built from this PR, but locally), and extract oc from that, releaseName is the expected 4.3.0-0.ci-2019-10-23-blah - maybe the ci-op/release:latest image is in flux w/ multiple pushes/test restarts so the Release Metadata: Version is absent? not sure.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -492,19 +491,16 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
-			releaseInfo := exactReleaseImage
 			var matched bool
-			errmsg := ""
-			switch target.Command {
-			case "oc":
-				releaseInfo = releaseName
-				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, could not determine version", target.Command)
-			default:
-				errmsg = fmt.Sprintf("warning: Unable to inject release image info into %s, installer will not be locked to the correct image", target.Command)
+			matched, err = copyAndReplaceReleaseInfo(w, r, 4*1024, exactReleaseImage)
```
soltysh: `copyAndReplaceReleaseImage`

pkg/cli/admin/release/extract_tools.go
```diff
@@ -624,25 +620,43 @@ func (o *ExtractOptions) extractCommand(command string) error {
 }
 
 const (
-	// binaryReplacement is the location within the installer binary that we can insert our
-	// release image info string
-	binaryReplacement = "\x00_RELEASE_IMAGE_LOCATION_\x00XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX\x00"
+	// installerReplacement is the location within the installer binary that we can insert our release image info string
+	installerReplacement = "\x00_RELEASE_IMAGE_LOCATION_\x00XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX\x00"
```
soltysh: `releaseImageReplacement`

pkg/version/version.go
soltysh: :+1: for the warning and printing the built-in version.

soltysh: For the 2nd, let's land this and we'll check you still have a few followups to fix. Side note, have you checked with hexeditor that binary extracted from the release image build from this PR?

PR #132, opened 2019-10-23
Clean kubectl wrappers

soltysh:
We have a bunch of unnecessary wrappers where we should not have. This PR removes them and re-uses the kubectl information where possible.

/assign @damemi


damemi:
Looks good to me but your verify has failed, looks like some generated difference needed with the flag change from switching commands?

soltysh:
> Looks good to me but your verify has failed, looks like some generated difference needed with the flag change from switching commands?

I cleaned one bit which was removing some global, removed it so we don't add anything unexpected. Should be good now. Self-tagging based on previous approval.

PR #134, opened 2019-10-25
oc adm catalog command

ecordell:
This adds two commands in accordance with the olm mirroring [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/oc/olm-mirroring.md).

An example of using these commands:

```sh
$ oc adm catalog build --appregistry-org=redhat-operators --to=quay.io/operatorframework/test-opreg:12
< status omitted > 
Uploading ... 387.5kB/s
Pushed sha256:8c8092a596939d206b90f9d84beec8fda944cfafba43fa2e4325cbbe61b3ef65 to quay.io/operatorframework/test-opreg:12

$ oc adm catalog mirror quay.io/operatorframework/test-opreg:12 localhost:5000
I1028 14:25:02.096029   75119 mirror.go:176] wrote database to /var/folders/4m/pt431q9d2fsbk9zz2_vmg3tc0000gn/T/038631746/bundles.db
I1028 14:25:13.765137   75119 mirror.go:271] wrote mirroring manifests to test-opreg-manifests

$ ls test-opreg-manifests
imageContentSourcePolicy.yaml mapping.txt

$ cat test-opreg-manifests/imageContentSourcePolicy.yaml
metadata:
  name: test-opreg
spec:
  repositoryDigestMirrors:
  - mirrors:
    - localhost:5000/fuse7/fuse-online-operator
    source: registry.redhat.io/fuse7/fuse-online-operator
  - mirrors:
    - localhost:5000/container-native-virtualization/node-maintenance-operator
    source: registry.redhat.io/container-native-virtualization/node-maintenance-operator
    # etc

$ cat test-opreg-manifests/mapping.txt
registry.redhat.io/fuse7/fuse-online-operator:1.4=localhost:5000/fuse7/fuse-online-operator:1.4
registry.redhat.io/container-native-virtualization/node-maintenance-operator:v2.0.0-6=localhost:5000/container-native-virtualization/node-maintenance-operator:v2.0.0-6
registry.redhat.io/openshift4/ose-template-service-broker-operator@sha256:9dbb5f8092764e1987731ea3e9c071092c12974c06cde48c637dc7fdfad2fbd1=localhost:5000/openshift4/ose-template-service-broker-operator
```

Here is the help for both commands:

```sh
$ oc adm catalog build -h
Builds a catalog container image from a collection operator manifests.

 Extracts the contents of a collection of operator manifests to disk, and builds them into an operator registry catalog
image.

Usage:
  oc adm catalog build [flags]

Options:
      --appregistry-endpoint='https://quay.io/cnr': Endpoint for pulling from an application registry instance.
  -o, --appregistry-org='': Organization (Namespace) to pull from an application registry instance
      --auth-token='': Auth token for communicating with an application registry.
      --dir='': Local path to cache manifests when downloading.
      --from='quay.io/operator-framework/operator-registry-server:latest': The image to use as a base.
      --to='': The image repository tag to apply to the built catalog image.
  -d, --to-db='': Local path to save the database to.

Use "oc adm options" for a list of global command-line options (applies to all commands).
```
```sh
$ oc adm catalog mirror -h
Mirrors the contents of a catalog into a registry.

 This command will pull down an image containing a catalog database, extract it to disk, query it to find all of the images used in the manifests, mirrors them to a target registry.

 By default, the database is extracted to a temporary directory, but can be saved locally via flags.

 An ImageContentSourcePolicy is written to a file for application to a cluster with access to the target registry. This will configure the cluster to pull from the mirrors instead of the locations referenced in the operator manifests.

 A mapping.txt file is also created that is compatible with "oc image mirror". This may be used to further customize the mirroring configuration, but should not be needed in normal circumstances.

Usage:
  oc adm catalog mirror [flags]

Options:
      --dry-run=false: Print the actions that would be taken and exit without writing to the destinations.
      --manifests-only=false: Calculate the manifests required for mirroring, but don not actually mirror image content.
      --path='': In-container to local path mapping for database.
      --to-manifests='': Local path to store manifests.

Use "oc adm options" for a list of global command-line options (applies to all commands).
```

Note that `path` and `manifests-only` are not in the proposal, but they seemed sensible to add. `path` functions as in `oc image extract --path=/dir:./local/dir`


pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,111 @@
...
+				return a.Run()
+			}
+
+			var extractor mirror.DatabaseExtractorFunc = func(from string) (string, error) {
+				e := imgextract.NewOptions(streams)
+				e.Mappings = []imgextract.Mapping{
+					{
+						ImageRef: srcRef,
+						From:     "/",
+						To:       "./db",
```
ecordell: todo: not hardcoded

pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,111 @@
...
+	"k8s.io/klog"
+	"k8s.io/kubectl/pkg/util/templates"
+
+	imgextract "github.com/openshift/oc/pkg/cli/image/extract"
+	imgmirror "github.com/openshift/oc/pkg/cli/image/mirror"
+)
+
+var (
+	mirrorLong = templates.LongDesc(`
+			Builds a catalog container image from a collection operator manifests.
```
ecordell: todo: desc wrong

pkg/cli/admin/admin.go
```diff
@@ -134,6 +135,10 @@ func NewCommandAdmin(name, fullName string, f kcmdutil.Factory, streams genericc
 		options.NewCmdOptions(streams),
 	)
 
+	cmds.AddCommand(
```
soltysh: Add it to the above section, please.

pkg/cli/admin/catalog/build.go
```diff
@@ -0,0 +1,57 @@
...
+var (
+	buildLong = templates.LongDesc(`
+			Builds a catalog container image from a collection operator manifests.
+
+			Extracts the contents of a collection of operator manifests to disk, and builds them into
+			an operator registry catalog image.
+		`)
+)
+
+func NewBuildImage(streams genericclioptions.IOStreams) *cobra.Command {
```
soltysh: This command is not following the patterns set by all commands, this means **every** command needs to:
1. Have `<command-name>Options` struct with all the necessary flags used by it. Flags should be mapped 1-to-1 in this struct. Additional necessary fields are allowed. 
2. Have these methods:
  * `Complete` - responsible for completing the options struct from 1 (required)
  * `Verify` - responsible for performing validation (optional)
  * `Run` - responsible for actual execution
3. Each should use `kcmd.util.CheckErrr` when running ^
4. Use `Run` from cobra.
See other commands for examples.

pkg/cli/admin/catalog/catalog.go
```diff
@@ -0,0 +1,28 @@
...
+		Use:   "catalog",
+		Short: "Tools for managing the OpenShift OLM Catalogs",
+		Long: templates.LongDesc(`
+			This tool is used to extract and mirror the contents of catalogs for Operator
+			Lifecycle Manager.
+
+			The subcommands allow you to build catalog images from a source (such as appregistry) 
+			and mirror its content across registries.
+			`),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
soltysh: `Run:   kcmdutil.DefaultSubCommandRun(streams.ErrOut),` like all the other commands.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,281 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	"github.com/openshift/library-go/pkg/image/reference"
+	"github.com/operator-framework/operator-registry/pkg/mirror"
+	"github.com/sirupsen/logrus"
```
soltysh: This is not allowed, only `k8s.io/klog`

pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,281 @@
...
+		return fmt.Errorf("must specify path for database")
+	}
+	if o.ManifestDir == "" {
+		return fmt.Errorf("must specify path for manifests")
+	}
+	return nil
+}
+
+func NewMirrorCatalog(streams genericclioptions.IOStreams) *cobra.Command {
+	o := MirrorCatalogOptions{
```
soltysh: We prefer having explicit `NewMirrorCatalogOptions` method.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use:   "mirror",
+		Short: "mirror an operator-registry catalog",
+		Long:  mirrorLong,
+
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if debug, _ := cmd.Flags().GetBool("debug"); debug {
+				logrus.SetLevel(logrus.DebugLevel)
```
soltysh: Nope.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+			return nil
+		},
+
+		RunE: func(cmd *cobra.Command, args []string) error {
```
soltysh: Does not follow any of the standards in oc.

pkg/cli/admin/catalog/mirror.go
ecordell: The libraries I'm calling use `logrus`. Is there a restriction around calling libraries that use different log frameworks?

go.mod
```diff
@@ -5,11 +5,10 @@ go 1.12
 require (
 	github.com/AaronO/go-git-http v0.0.0-20161214145340-1d9485b3a98f
 	github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd
-	github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5 // indirect
 	github.com/RangelReale/osincli v0.0.0-20160924135400-fababb0555f2
 	github.com/alexbrainman/sspi v0.0.0-20180613141037-e580b900e9f5
 	github.com/apcera/gssapi v0.0.0-00010101000000-000000000000
-	github.com/aws/aws-sdk-go v1.16.26
+	github.com/aws/aws-sdk-go v1.17.7
```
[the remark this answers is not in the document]
ecordell: I'll have to pin some of this - is that okay?

pkg/cli/admin/catalog/mirror.go
ecordell: I removed this for now - I can look into klog for the dependency if we need to see those in the future.

umohnani8:
@sallyom @soltysh PTAL, aiming to get this in by tomorrow.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+	}
+
+	// Create an unstructured object for removing creationTimestamp
+	unstructuredObj := unstructured.Unstructured{}
+	unstructuredObj.Object, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&icsp)
+	if err != nil {
+		return fmt.Errorf("ToUnstructured error: %v", err)
+	}
+	delete(unstructuredObj.Object["metadata"].(map[string]interface{}), "creationTimestamp")
```
soltysh: I see this code copied from https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/mirror.go and I'm not that happy about it :cry: but I'll let you land it. Please fix this in a followup.
Since you have the access to the resource just set the creation timestamp to nil and for printing how about using printers? You can force it to produce yaml, always.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	o.DatabaseExtractor = extractor
+	return nil
+}
+
+func (o *MirrorCatalogOptions) Validate() error {
+	if o.DatabasePath == "" {
+		return fmt.Errorf("must specify path for database")
+	}
+	if o.ManifestDir == "" {
```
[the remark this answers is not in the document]
ecordell: it's possible if you don't run `Complete` 😄

PR #144, opened 2019-11-04
bump to kubernetes 1.16.2

mfojtik:


mfojtik:
/hold cancel

* Fixed build failure because of bad level of k8s.io/api

PR #145, opened 2019-11-04
Bug 1713263: pkg/cli/admin/upgrade: allow users to force when updating to latest

abhinavdahiya:
--force is used when
a) the release image is not verified.
b) the operators are blocking upgrades.

For cases when the users are using custom update channel, the release-image is probably not going to be verified by Red Hat signature, but the users should be allowed to update to the latest
available update by skipping the verification step.

And for cases where the users want to upgrade to the latest upgrade, even when the operators are blocking updates should be achievable by setting the force flag.

/assign @smarterclayton

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -180,9 +180,13 @@ func (o *Options) Run() error {
 		}
 
 		sortSemanticVersions(cv.Status.AvailableUpdates)
-		update := cv.Status.AvailableUpdates[len(cv.Status.AvailableUpdates)-1]
-		cv.Spec.DesiredUpdate = &update
 
+		update := cv.Status.AvailableUpdates[len(cv.Status.AvailableUpdates)-1]
+		if o.Force {
+			update.Force = true
+		}
+
+		cv.Spec.DesiredUpdate = &update
 		_, err := o.Client.ConfigV1().ClusterVersions().Update(cv)
 		if err != nil {
 			return fmt.Errorf("Unable to upgrade to latest version %s: %v", update.Version, err)


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -180,9 +180,13 @@ func (o *Options) Run() error {
 		}
 
 		sortSemanticVersions(cv.Status.AvailableUpdates)
+
 		update := cv.Status.AvailableUpdates[len(cv.Status.AvailableUpdates)-1]
-		cv.Spec.DesiredUpdate = &update
+		if o.Force {
```
smarterclayton: Hrm, I kind of would prefer not allowing this.  I think if you have to force you should have to at least give us a version (was the rationale for not having this before).

abhinavdahiya: So I think QE is a situation where they set the upstream to the point to the nightly, and then mark update to the latest version.
but they can't just `--to-latest` because nightly is not signed. and there they need to `--force`
So, hopefully when the nighties get signed, maybe that use-case doesn't apply.

But if the customer is running their own update-server in disconnected env wouldn't have access to verify, so they might want to use `--to-latest --force`
but maybe when we support such an env in a better way we can allow these customers to host our signatures locally, in which case this wouldn't apply too.

So the last one is, i have overrides set, or i have techpreview features on, and i want to update to whtever latest version there is, but that needs a `--to-latest --force`.

abhinavdahiya:
ping @smarterclayton ?

pkg/cli/admin/upgrade/upgrade.go
smarterclayton: OK.

PR #147, opened 2019-11-05
Bug 1725977: admin/release/new: always consider all mappings

crawford:
Previously, additional mappings provided as args weren't considered when
creating a new release image from an ImageStream, an existing release
image, or a directory. The was particularly problematic when attempting
to add new images:

    oc adm release new \
      --from-release quay.io/openshift-release-dev/ocp-release:4.2 \
      --to-image quay.io/crawford/openshift-release \
      cluster-etcd-operator=quay.io/hexfusion/cluster-etcd-operator

Without this patch, the previous command would create a release image
but it wouldn't the cluster-etcd-operator manifests.

Instead of adding new mappings in the default case (when not using an
existing ImageStream, release image, or directory), add them
unconditionally. Clayton thinks that this defect may have been
introduced when the validation code was being reworked.

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -596,15 +596,6 @@ func (o *NewOptions) Run() error {
 			}
 		}
 		fmt.Fprintf(o.ErrOut, "info: Found %d operator manifest directories on disk\n", len(ordered))
-
-	default:
-		for _, m := range o.Mappings {
-			if exclude.Has(m.Source) {
-				klog.V(2).Infof("Excluded mapping %s", m.Source)
-				continue
-			}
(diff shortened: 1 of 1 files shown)


crawford:
Reincarnation of https://github.com/openshift/oc/pull/2.

abhinavdahiya:
/retitle Bug 1768674: admin/release/new: always consider all mappings

abhinavdahiya:
/retitle Bug 1725977: admin/release/new: always consider all mappings

smarterclayton:
/lgtm

I think the only concern that would be new is the "has this changed from the defaults" (and thus we wouldn't perform a revalidation) and that code is up above, so this should be ok.

PR #152, opened 2019-11-07
Update ruby-22-centos7 to ruby-25-centos7

soltysh:
Similarly how in https://github.com/openshift/origin/pull/24100 change deprecated ruby-22-centos7 instances.

/assign @adambkaplan


soltysh:
Fix single place where `openshift/ruby-25-centos7` was specified instead of `centos/ruby-25-centos7` selftagging to unblock queue.

PR #153, opened 2019-11-08
Enable all Linux arches in cli-artifacts

yselkowitz:
In order for console-operator deployment to succeed, cli-artifacts needs to be available on all arches for downloads-openshift-console.  However, in that case, /usr/bin/oc (inherited from cli) is a native binary, and we want to provide all primary Linux architectures to match those on mirror.openshift.com.

In order to do so, this provides cross-compiled Linux binaries for multiple architectures.  Cross-compiling oc fails with gssapi enabled, therefore it is disabled in the cross builds.

Because this affects cluster deployment, we are looking to get this into 4.2 and later branches (we can provide separate PRs for stable branches as needed).

### DIFF
--- a/images/cli-artifacts/Dockerfile.rhel
+++ b/images/cli-artifacts/Dockerfile.rhel
@@ -4,11 +4,11 @@ FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/oc
 COPY . .
 RUN yum install -y --setopt=skip_missing_names_on_install=False gpgme-devel libassuan-devel
-RUN make cross-build-darwin-amd64 cross-build-windows-amd64 --warn-undefined-variables
+RUN make cross-build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.2:cli
-COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/darwin_amd64/oc /usr/share/openshift/mac/oc
-COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/windows_amd64/oc.exe /usr/share/openshift/windows/oc.exe
+COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/ /usr/share/openshift/
+RUN mkdir /usr/share/openshift/mac /usr/share/openshift/windows && ln /usr/share/openshift/darwin_amd64/oc /usr/share/openshift/mac/oc && ln /usr/share/openshift/windows_amd64/oc.exe /usr/share/openshift/windows/oc.exe
 LABEL io.k8s.display-name="OpenShift Clients" \
       io.k8s.description="OpenShift is a platform for developing, building, and deploying containerized applications." \
       io.openshift.tags="openshift,cli"
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ GO_BUILD_PACKAGES :=$(strip \
 GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp gssapi'
 GO_BUILD_FLAGS_DARWIN :=-tags 'include_gcs include_oss containers_image_openpgp'
 GO_BUILD_FLAGS_WINDOWS :=-tags 'include_gcs include_oss containers_image_openpgp'
+GO_BUILD_FLAGS_LINUX_CROSS :=-tags 'include_gcs include_oss containers_image_openpgp'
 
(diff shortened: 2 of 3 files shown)


images/cli-artifacts/Dockerfile.rhel
```diff
@@ -4,11 +4,15 @@ FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/oc
 COPY . .
 RUN yum install -y --setopt=skip_missing_names_on_install=False gpgme-devel libassuan-devel
-RUN make cross-build-darwin-amd64 cross-build-windows-amd64 --warn-undefined-variables
+RUN make cross-build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.2:cli
 COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/darwin_amd64/oc /usr/share/openshift/mac/oc
 COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/windows_amd64/oc.exe /usr/share/openshift/windows/oc.exe
+COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/linux_amd64/oc /usr/share/openshift/linux-x86_64/oc
```
wking: why rename here?  Can't we just stick with the GOARCH?

images/cli-artifacts/Dockerfile.rhel
```diff
@@ -4,11 +4,15 @@ FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/oc
 COPY . .
 RUN yum install -y --setopt=skip_missing_names_on_install=False gpgme-devel libassuan-devel
-RUN make cross-build-darwin-amd64 cross-build-windows-amd64 --warn-undefined-variables
+RUN make cross-build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.2:cli
 COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/darwin_amd64/oc /usr/share/openshift/mac/oc
 COPY --from=builder /go/src/github.com/openshift/oc/_output/bin/windows_amd64/oc.exe /usr/share/openshift/windows/oc.exe
```
wking: Now that we want to include the arch, can we add hardlinks or symlinks or some such to these with explicit arches?  We'd need to keep the old paths around (at least as links) until `oc adm release extract` and the console downloads deployment had been ported to use the new paths.  And personally, I'd be happier if we just stuck to GOOS and used `darwin` instead of `mac`.

images/cli-artifacts/Dockerfile.rhel
yselkowitz: I wasn't sure how to layout these directories so I went by https://mirror.openshift.com/pub/openshift-v4/clients/oc/4.2/ (where uname-style arches are used instead of GOARCH-style names).  As they are new, they can be named as desired, I just need to keep the console-operator patch in sync.

wking: I don't see us porting `oc` from Go to another language in the near future, so I'd rather just:

```Dockerfile
COPY /go/src/github.com/openshift/oc/_output/bin/ /usr/share/openshift/
```

Folks who want a different naming scheme can translate; we don't have to translate for them here.

images/cli-artifacts/Dockerfile.rhel
wking: > I didn't change any of the existing paths, as I wasn't sure where else they may be used. Can we leave that for a separate commit?

I think we should leave *dropping* the existing paths to future work.  But openshift/console-operator#343 can be simplified if we add the consistent paths in this PR.  Something like:

```Dockerfile
COPY /go/src/github.com/openshift/oc/_output/bin/ /usr/share/openshift/
RUN mkdir /usr/share/openshift/mac /usr/share/openshift/windows && ln /usr/share/openshift/darwin_amd64/oc /usr/share/openshift/mac/oc && ln /usr/share/openshift/windows_amd64/oc.exe /usr/share/openshift/windows/oc.exe
```

yselkowitz:
@wking reworked the patch per your comments, anything else needed here?

crawford:
/approve

I'll let @wking reply with the LGTM once he's happy with the changes.

wking:
/lgtm

(but I'm not an approver).  Dunno how robust our CI coverage is here, but we can always adjust later if we're somehow breaking backwards compat here (although I'd expect the console downloads pod to crashloop if we did).

PR #161, opened 2019-11-13
Bug 1771329: only build adm catalog command for linux

njhale:
Constrain `oc adm catalog` subcommand to linux builds. 

This temporarily addresses an [issue with cross-compiling the subcommand](https://bugzilla.redhat.com/show_bug.cgi?id=1771329 ) for mac and windows in 4.3.


soltysh:
This approach is concerning to start with, I'm inclined to revert the entire command because as is, it is putting this part of the project at significant risk of incompatibility and problems we don't want to resolve.

njhale:
> This approach is concerning to start with, I'm inclined to revert the entire command because as is, it is putting this part of the project at significant risk of incompatibility and problems we don't want to resolve.

@soltysh Are you referring to the build tag or sqlite use?

PR #163, opened 2019-11-14
Bug 1772463: Set TypeSetter in get print flags, like we do everywhere elsewhere

soltysh:
/assign @sallyom


sallyom:
@soltysh units failing because of ruby-22-25 hello-world issue

soltysh:
/retest
@sallyom 4.2 should be back in shape, ptal

PR #166, opened 2019-11-17
Bug 1779331: Fix a number of issues around debugging daemonset pods and init containers

smarterclayton:
A bunch of small minor fixes. After these, I was able to debug the init container of the node exporter daemonset host network pod.


pkg/cli/debug/debug.go
```diff
@@ -362,6 +364,9 @@ func (o *DebugOptions) RunDebug() error {
 	if len(ns) == 0 {
 		ns = o.Namespace
 	}
+	if len(o.ToNamespace) > 0 {
```
soltysh: How is this different from explicitly using `--namespace`?

smarterclayton: example is:

oc debug istag/cli:latest -n openshift —to-namespace=my-test-ns

For when you don’t have write access in the namespace that contains the thing you want to debug.

smarterclayton:
Any other comments?

smarterclayton:
@soltysh I want this in 4.3, so if you have any other comments let me know soon

soltysh:
> @soltysh I want this in 4.3, so if you have any other comments let me know soon

nope, this is good
/cherry-pick release-4.3

PR #167, opened 2019-11-19
Bug 1775875: pkg/cli/admin/release/extract_tools: Pass []replacements to copyAndReplace

wking:
This is everything from #165 except setting `InjectReleaseVersion` for the installers, because @abhinavdahiya [wants some time][1] to look into the ecosystem around that.  @sallyom, this also lets us fix the existing bugs described in #165 while avoiding [your concerns][2] about warnings when running code that expects to inject release versions into the installer before the installer has the marker to recieve them.

/assign @soltysh

[1]: https://github.com/openshift/installer/pull/2682#issuecomment-554725150
[2]: https://github.com/openshift/oc/pull/165#issuecomment-555026307

### DIFF
--- a/pkg/cli/admin/release/extract_tools.go
+++ b/pkg/cli/admin/release/extract_tools.go
@@ -492,21 +492,22 @@ func (o *ExtractOptions) extractCommand(command string) error {
 		}
 
 		// copy the input to disk
+		replacements := []replacement{}
 		if target.InjectReleaseImage {
-			var matched bool
-			matched, err = copyAndReplaceReleaseInfo(w, r, 4*1024, exactReleaseImage)
-			if !matched {
-				fmt.Fprintf(o.ErrOut, "warning: Unable to replace release image location into %s, installer will not be locked to the correct image\n", target.Command)
-			}
-		} else if target.InjectReleaseVersion {
-			var matched bool
-			matched, err = copyAndReplaceReleaseVersion(w, r, 4*1024, releaseName)
-			if !matched {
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/release/extract_tools.go
```diff
@@ -622,88 +623,113 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
-				return matched, nil
+				remainingNames := make([]string, 0, len(replacements))
+				for i, matched := range matches {
+					if !matched {
+						remainingNames = append(remainingNames, replacements[i].name)
+					}
+				}
+				sort.Strings(remainingNames)
+				if len(remainingNames) > 0 {
+					klog.Warningf("Unable to make all expected replacements in %s.  Remaining: %s", name, strings.Join(remainingNames, ", "))
```
wking: Hmm.  I see a number of these in master, but I also see things like `fmt.Fprintf(o.ErrOut, "warning...`.  What is the recommended approach to writing warnings from these CLI packages?

smarterclayton: In these commands warnings are printed with fprintf.

Klog is for debugging and servers, not end users.

wking: > In these commands warnings are printed with fprintf.

Seems like a pain to pass around an error-output `Writer`, but :man_shrugging:, updated to use `Fprintf` with 1604d6d28 -> cbcc1fd6c.

wking:
Now that 4.4 has forked off, I've created a new bug to hang this PR and subsequent backports on.

PR #174, opened 2019-11-20
pkg/cli/admin/release: fmt.Errorf -> errors.New when there's no formatting

wking:
Generated with:

```console
$ sed -i 's/fmt\.Errorf(\("[^"%]*"\))/errors.New(\1)/g' pkg/cli/admin/release/*.go
```

Followed by manually:

* Adding errors imports to the touched files.
* Using `apierrors` for `k8s.io/apimachinery/pkg/api/errors` to distinguish from the standard library package.
* Removing the `fmt` import from `signature.go`, since there were no longer any consumers there.

### DIFF
--- a/pkg/cli/admin/release/extract.go
+++ b/pkg/cli/admin/release/extract.go
@@ -2,6 +2,7 @@ package release
 
 import (
 	"archive/tar"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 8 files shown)


wking:
Pushed 79575cd3a -> 3a3154e18 to convert an old `fmt.Errorf(out)` and `fmt.Errorf("supported reference modes are: \"public\" (default) and \"source\"")` as well.  The `fmt.Errorf(out)` is more than cosmetic, because the old approach would crash if the output contained any `%` formatting directives.

wking:
Will rebase if/when there is any maintainer interest ;)

soltysh:
> Will rebase if/when there is any maintainer interest ;)

Sure, but please ping me on slack, preferably ;)

PR #177, opened 2019-11-21
Bug 1748061: copy gotdotenv and add ability to override underlying scanner buffer …

gabemontero:
…for big files

/assign @adambkaplan 

this is what would be required to fix https://bugzilla.redhat.com/show_bug.cgi?id=1748061

or alternatively, we submit an PR to github.com/joho/godotenv to add this capability and see if they accept it

Per my question in the bug, do we want to take this on given there is a customer case, or close as wontfix ... still waiting from feedback in bug on their feelings wrt to the workaround of using `--param` instead of `--param-file`

### DIFF
--- a/pkg/helpers/newapp/app/env.go
+++ b/pkg/helpers/newapp/app/env.go
@@ -1,14 +1,15 @@
 package app
 
 import (
+	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
-
-	"github.com/joho/godotenv"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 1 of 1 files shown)


gabemontero:
terraform / aws flake e2e-aws-builds

/test e2e-aws-builds

gabemontero:
per discussion in bugzilla removed WIP and associated with bug @adambkaplan 

bump / PTAL

gabemontero:
Presumably yes @adambkaplan as I noted in the description ... merely a question of expediency.

Note, it would require a new "API", as the parameters for size would have to be passed down somehow ... so I could envision some time spent going back and forth nailing that down.   Is it worth spending that time at this point?

adambkaplan:
/lgtm

/hold

@gabemontero since this is for 4.4 with no plan to backport, I'd rather us be good citizens and propose a patch. If it takes too long we can merge this as is.

gabemontero:
/hold cancel

per discussion with @adambkaplan an upstream contribution would require changes to their public API (as public method signatures would have to change), or we would need klunky `ReadWithSize` and `ParseWithSize` methods for the existing `Read` and `Parse` methods.

I've updated comments in code to explain the situation

@adambkaplan bump / ptal

PR #181, opened 2019-11-25
Bug 1772942: Fix ignored `--manifest-dir` option in `oc adm catalog build` 

njhale:
Bumps the `operator-registry` dependency to a version that handles the cache directory option correctly.

~Also fixes a bad pseudoversion for the module replacement of `openshift/api` which was causing module operations to fail.~ Looks like this was fixed in another PR.


njhale:
@soltysh IIRC I bumped the operator-registry dependency by modifying the minor version in the `go.mod` followed by `go mod tidy` and `go vendor`. However, before I could run these commands, I needed to alter the `replace` directive for `openshift/api`, since the given pseudoversion (`v0.0.0-0.20191112184635-86def77f6f`) was causing the commands to fail. Hence the PR description:

> Also fixes a bad pseudoversion for the module replacement of openshift/api which was causing module operations to fail.

Other than that, are there additional changes you see that you weren't expecting?

I noticed a `goodmod.yaml` file, but didn't see any make rules using the related tool for vendoring. If there is a preferred process for updating `oc` deps, I'd be happy to follow it -- just point me in the right direction.

njhale:
@soltysh Just following up on my previous post; any thoughts, questions?

soltysh:
@njhale so the process you described seems legit (`go mod vendor` + `go mod tidy`) but I just did the same on top of your PR and files such as `vendor/golang.org/x/sys/unix/mkasm_darwin.go` and others I didn't expect to be modified are not in my local copy, whereas somehow they are in yours? Which version of go you used, I've tried both go1.12 and 1.13 and both did not clean those files for me.

njhale:
@soltysh Thanks for getting back to me. I used 1.13: `go version go1.13.4 darwin/amd64`.

It may have something to do with the OS deps are being pulled onto (`mkasm_darwin.go` is suspect), or maybe not -- either way, I'll try again and see If I come away without the changes.

njhale:
@soltysh @mfojtik There are several other PRs fixing the pseudo-version issue. If either of those merge first, I'll drop the changes here and try again.

soltysh:
@njhale some changes merged, can you rebase?

njhale:
@soltysh After rebasing, things are looking much better. PTAL.

njhale:
/hold

@soltysh is there a new requirement for test coverage that's failing here?

soltysh:
This was fixed in https://github.com/openshift/oc/pull/216
/hold cancel
/retest

PR #185, opened 2019-11-26
Bug 1777030: Update support tools container image

crawford:
The RHEL7 support tools image is only built for AMD64, but there will
soon be a need for an image that works on System Z (and eventually
Power). Changing to the RHEL8 image gets us the additional
architectures. Additionally, RHCOS has moved to RHEL8 content, so it's
best to keep the support tools in sync.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -765,7 +765,7 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 			}
 		}
 		if len(o.Image) == 0 {
-			image = "registry.redhat.io/rhel7/support-tools"
+			image = "registry.redhat.io/rhel8/support-tools"
 		}
 		zero := int64(0)
 		isTrue := true


crawford:
/retitle Bug: 1777030: Update support tools container image

crawford:
/retitle Bug 1777030: Update support tools container image

smarterclayton:
Will this fail when customers run on rhel7 worker nodes?  Or work but with issues?

yselkowitz:
If there is a need, `rhel7/support-tools` could be multi-arch'ed.  Is there a way to use `rhel7/support-tools` on RHEL7 worker nodes and `rhel8/support-tools` with RHCOS?

crawford:
@smarterclayton The utilities that I tried seemed to work. I was able to use `dnf` to install and test `ip`, `top`, `tcpdump`, and `systemd-cgtop`. In the event that there is a problematic utility, we can always fall back to using `oc debug --image=registry.redhat.io/rhel7/support-tools`.

PR #193, opened 2019-11-28
fix(oc): fixes invalid compdef reference to kubectl for zsh

ashishranjan738:
This commit fixes invalid compdef reference to kubectl for zsh
completion script. This prevents zsh to autoload the completions from
FPATH.

Signed-off-by: Ashish Ranjan <ashishranjan738@gmail.com>

### DIFF
--- a/contrib/completions/zsh/oc
+++ b/contrib/completions/zsh/oc
@@ -1,4 +1,4 @@
-#compdef kubectl
+#compdef oc
 
 
 __kubectl_bash_source() {


contrib/completions/zsh/oc
```diff
@@ -1,4 +1,4 @@
-#compdef kubectl
+#compdef oc
```
soltysh: This is generated file, so you can't switch this, you need to find a different place where that can be fixed.

ashishranjan738: yeah just realized.

ashishranjan738: is there any template from where this file is generated. I tried to look around but didn't find any.

PR #194, opened 2019-12-02
Fix build on go 1.13

isimluk:
Addressing:
```
$ go version
go version go1.13 linux/amd64
$ make build
go: github.com/openshift/api@v0.0.0-0.20191112184635-86def77f6f: invalid pseudo-version: revision is shorter than canonical (86def77f6f90)
make: Nothing to be done for 'build'.
```

When extracting a module from a version control system, the go command now performs additional validation on the requested version string. More information can be found at https://tip.golang.org/doc/go1.13#version-validation


mfojtik:
i propose https://github.com/openshift/oc/pull/206 as better fix :-)

soltysh:
This is already fixed.
/close

PR #202, opened 2019-12-04
Revert "Bug 1777031: Update support tools container image"

crawford:
Reverts openshift/oc#196

Apparently we can't support RHEL8 container images on RHEL7 nodes, so this isn't going to work on RHEL7 workers.


crawford:
This is a revert, so we don't need to wait on end-to-end tests.

/override ci/prow/e2e-aws

PR #209, opened 2019-12-11
Bug 1783055: Allow 'machine-os' to register itself and provide display name

smarterclayton:
We want to show in changelog the FCoS or RHCoS version for the 'machine-os'
content and to give them a display label.

This PR changes the changelog to support that function.


smarterclayton:
@abhinavdahiya this will allow us to have labels for the OS version and for release controller to link to the rhcos or fcos dashboards for code that was in the release.

pkg/cli/admin/release/image_mapper.go
```diff
@@ -435,7 +515,10 @@ func parseComponentVersionsLabel(label string) (ComponentVersions, error) {
 			return nil, fmt.Errorf("the version pair %q must have a valid semantic version: %v", pair, err)
 		}
 		v.Build = nil
-		labels[parts[0]] = v.String()
+		labels[parts[0]] = ComponentVersion{
```
ashcrow: nit: Do we need to worry about having the same version name listed twice causing the last found to take precedence? I tend to think just noting "don't do that please" is enough.

abhinavdahiya:
took a quick pass, and looked good.

But my question is machine-os is OS version carrier so having a version 4.4 but name being FCOS or RHCOS seems a bit weird.
fcos-4.4 or rhoas-4.4 like versions are clearer and don't require extra knowledge in clients that display name is going to be different.

If we want to use the display name to provide human friendly data that seems more appropriate. like the version could be rhaos-43.2312 but the display name is RHEL CoreOS 4.3

pkg/cli/admin/release/image_mapper.go
smarterclayton: there's logic in the build that errors if that happens (if you try to create a release payload that has image A with component foo @ 1.2.0 and image B with component foo @1.2.1 you can't build the payload).  The logic is copied for display name, see the tests I added that use an image stream.

smarterclayton:
> But my question is machine-os is OS version carrier so having a version 4.4 but name being FCOS or RHCOS seems a bit weird.

So the rules of components are that your name (e.g. `kubernetes`) has to have a semantic version (which rhcos and fcos do, I believe), and then the display name overrides the component name.  We want to display the kube sem ver (1.16.2) as well as the full rhcos string (43.xxxxx.xxxx) so that we can link that back to a display page from release controller and see what changed from a previous version.  

Now, as I look at rhcos it doesn't actually have a semantic version `44.81.201912120251.0`, so we probably need to fix that so we can link from release controller details directly.  @ashcrow, do you foresee a problem changing that to be `44.81.201912120251`?

pkg/cli/admin/release/image_mapper.go
smarterclayton: Test is https://github.com/openshift/oc/pull/209/files#diff-a42d49db6f7580ec9dea5398fe27b8deR62

cgwalters:
We can change it.  This is a complex topic though.  Today the `.0` is obviously redundant because we have the full timestamp in there.

But the intention of the design originally was that we'd use `YMD.SERIAL` - as we're doing in 4.1/4.2.  But we need to rework our build system to [avoid reusing version numbers](https://github.com/coreos/coreos-assembler/pull/885) in order to go back to that.

If we're trying to match semver though...hm, doesn't it need to be e.g.

`4.3-81.20191212-0` ?

Where the `MAJOR` here is `4` - we're not breaking it.  The `3` is the compatible part of the OpenShift version, and the `81` is the RHEL version, and the rest is `DATESTAMP-SERIAL`?

cgwalters:
I guess what I'm really asking is: Do you care just that it's *formatted* as semver, or that we actually try to match the *meaning* of semver?

smarterclayton:
It needs to be formatted as semver.  Technically everything is a prerelease anyway so x.y.z-PRE is totally acceptable for the version.  We only allow dashes too - you can't do the + semantics for "good but annoying reasons".

I would say starting as 43.81.DATE-INC is totally a good place and you can change at any point in the future.  obv if you can change the version reported that's best, but we can work around in the short term any gap.

smarterclayton:
This is a prerequisite to the other one - taking abhinav's review (since this is opt in, it's low risk) and merging now.

smarterclayton:
Minor fix, made the unit test deterministic (was a bug in previous code)

abhinavdahiya:
>  I'd prefer not to put version in display name - that potentially means duplicates, and the version of RHCoS is implicitly tied to the product, whereas FCoS is implicitly not.

hmm, i don't think i'm following this anymore..

a concrete example wrt RHCOS vs FCOS maybe would help?

pkg/cli/admin/release/annotations.go
```diff
@@ -22,6 +22,10 @@ const (
 	// alphanumeric characters and dashes only. The value `0.0.1-snapshot-key` in a
 	// manifest will be substituted with the version value for key.
 	annotationBuildVersions = "io.openshift.build.versions"
+	// This LABEL is a comma-delimited list of key=version pairs that can be consumed
```
abhinavdahiya: `key=version` -> `key=display name`

pkg/cli/admin/release/image_mapper.go
```diff
@@ -166,25 +166,30 @@ func NewTransformFromImageStreamFile(path string, input *imageapi.ImageStream, a
 		if !ok {
 			continue
 		}
-		klog.V(4).Infof("Found build versions from %s: %s", tag.Name, value)
-		items, err := parseComponentVersionsLabel(value)
+		displayNameValue, ok := tag.Annotations[annotationBuildVersionsDisplayNames]
+		if !ok {
+			continue
+		}
```
abhinavdahiya: will this skip our current images that don't have display names configured yet, only the versions ?

smarterclayton: Yikes, good catch

smarterclayton: Going to unit test this pretty thoroughly

smarterclayton: Sent a fix

smarterclayton:
> a concrete example wrt RHCOS vs FCOS maybe would help?

```
LABEL io.openshift.build.versions="machine-os=43.81.20191201101213-0"
LABEL io.openshift.build.version-display-name="machine-os=RHEL CoreOS"
```

```
LABEL io.openshift.build.versions="machine-os=31.10.20191201101213-10"
LABEL io.openshift.build.version-display-name="machine-os=Fedora CoreOS"
```

smarterclayton:
/retest

Further review comments?

cgwalters:
https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/776

smarterclayton:
Tagging, will watch for anomalies on release builds.

ashcrow:
> Further review comments?

No objection from me. I didn't test it locally but following the code it looks :+1:

cgwalters:
https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/776 merged which means the RHCOS version will be semver-compatible starting with 4.4 currently.

PR #210, opened 2019-12-11
Bug 1779351: Verify outbound manifests now that schema1 support is dropped

smarterclayton:


soltysh:
We have up-to-date library-go by now, so this not relevant anymore.
/close

PR #142, opened 2019-11-01
oc catalog adm docs

ecordell:
- Adds examples for `oc catalog adm`
- Fixes mapping for mirroring images

PR #212, opened 2019-12-12
Add security options to `oc adm catalog` commands

ecordell:
This brings them in line with other `oc` commands that communicate with registries, and give control over tls verification, etc. 

This also rebases and replaces #142


soltysh:
The e2e failures should be solved by now.
/retest

PR #214, opened 2019-12-13
[release-4.2] Bug 1776495: don't install quota CRDs

soltysh:
Pick https://github.com/openshift/oc/pull/184 since the other reached context limits and I can't repush it.


tnozicka:
/lgtm

/hold
to avoid CI loop (feel free to remove after assessing potentional failures)

soltysh:
/hold cancel
since we're green here

PR #218, opened 2019-12-14
Bump kube 1.17

sallyom:
/assign @soltysh 

also this PR is removing k8s.io/kubernetes dependencies: https://github.com/openshift/oc/pull/221


sallyom:
@soltysh the 'remove k8s.io/kubernetes/pkg/printers dependency' was necessary here because:
a) `type PrintOptions` moved to `k8s.io/cli-runtime/pkg/printers/interface.go`
b) fixing that, only k8s.io/kubernetes/pkg/printers deps left were 
* `AddKubeHandler` - that is a carry patch and only called in `pkg/printers/deprecated_printer.go` -do we need it still? copied to pkg/helpers/printers
* `PrintHandler` - that brings in no other k8s.io/kubernetes deps, copied to pkg/helpers/printers.

so rather than keep imports for k8s.io/cli-runtime/pkg/printers _and_ k8s.io/kubernetes/pkg/printers, I am removing the k8s.io/kubernetes dep.  Make sense?

soltysh:
@sallyom actually we should be able to drop that entirely, we should not need it, we can also drop that carry patch in the next k8s update (either you or @damemi will do that), I'll try to remember too :wink:

sallyom:
> @sallyom actually we should be able to drop that entirely, we should not need it, we can also drop that carry patch in the next k8s update (either you or @damemi will do that), I'll try to remember too

I dropped KubeHandler and removed pkg/printers/deprecated_printers* - sound good?

soltysh:
The builds failure is a known flake not related to this PR.

/override ci/prow/e2e-aws-builds

PR #222, opened 2019-12-18
Bug 1785194: don't gather pod endpoints using tokens

deads2k:
Updates `oc adm inspect` to only gather endpoint information with a certificate to protect against replays.  Also uses cluster-reader permission instead of cluster-admin.

/assign @soltysh


soltysh:
The must-gather e2e failure is interesting, the cluster does not have any secrets :scream: 
/retest

PR #230, opened 2019-12-23
Set path based on requestURI, not just path

soltysh:
Fixes openshift/origin#24231
/assign @sallyom 
/cc @bszeti @xingxingxia


xingxingxia:
/lgtm
```
$ go run test.go
when url is: http://:8080/health?param=1&param=2
  url_parsed.RequestURI() is: /health?param=1&param=2
when url is: http://:8080/health%3Fparam=1&param=2
  url_parsed.RequestURI() is: /health%3Fparam=1&param=2
```

soltysh:
The offending tests will be removed in https://github.com/openshift/origin/pull/24357

PR #231, opened 2019-12-23
Bug 1774424: set builder flags so it prints resources when nothing is specified

soltysh:
/assign @sallyom


sallyom:
/lgtm

i would've changed the example according to the bz, that's why you're awesome and i'm not :smiley:

PR #234, opened 2019-12-28
Bug 1778019: If required flag is not passed, don't print usage menu, only print error

sallyom:
Don't print usage menu upon error, only print error message. 

/cc @soltysh


sallyom:
> You also need the upstream commit from upstream 86693 b/c when we replace the name and act as kubectl you want the problem fixed there too.

opened https://github.com/openshift/kubernetes-kubectl/pull/9  will pull it in when it merges

PR #240, opened 2020-01-03
Verify imports

soltysh:
This should help us track k/k imports ensuring nobody adds new.
If you look at `hack/verify-imports.sh` script you'll notice there's only 2 big offenders:
- `k8s.io/kubernetes/pkg/credentialprovider`
- `k8s.io/kubernetes/pkg/apis/rbac/v1`

But IIRC those need to be solved upstream too, I'll need to look into it. The next 2 imports should disappear.

/assign @sallyom

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ image-ocp-recycler: image-ocp-cli
 update: update-generated-completions
 .PHONY: update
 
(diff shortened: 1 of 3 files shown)


sallyom:
nice! 

/lgtm

Issue #223, opened 2019-12-19
oc start-build should allow overriding tar default exclusion pattern


tomconte:
It looks like `oc start-build` always use `DefaultExclusionPattern` when running `tar`. I would need to override that pattern, so that the `.git` directory is included when using `--from-dir=.`.

PR #242, opened 2020-01-06
Add --exclude flag to start-build

tomconte:
Fix for #223


adambkaplan:
@tomconte since you are adding a new CLI flag, you need to run `make update` to add the new bash completions.

pkg/cli/startbuild/startbuild_test.go
```diff
@@ -258,7 +259,9 @@ func TestHttpBinary(t *testing.T) {
 			fromDir = server.URL + tc.urlPath
 		}
 
-		build, err := streamPathToBuild(nil, stdin, stdout, &FakeBuildConfigs{t: t, expectAsFile: tc.fromFile}, fromDir, fromFile, "", &options)
+		var defaultExclusionPattern = tar.DefaultExclusionPattern.String()
```
adambkaplan: nit: 
```suggestion
	 defaultExclusionPattern := tar.DefaultExclusionPattern.String()
```

gabemontero:
> /approve
> 
> I like this feature.
> 
> /cc @soltysh
> 
> fyi @openshift/openshift-team-developer-experience . I'm wondering if this flag should also be used in `new-app` and `new-build`, since these can trigger builds from local source.

In the past @adambkaplan there was some angst about adding ignore/exclude options that overlapped with existing ones like gitignore and dockerignore ... in fact I *may* recall some regret around adding the s2i ignore stuff (though @bparees or @csrwng might remember better than me).

But on the other hand I don't see how those existing items can deal git metadata (i.e. .git ... maybe that is why we added s2i ignore ??), and anything to lessen the burden on our tar, the better.

So other than dusting off memory cells from the folks I added who could have a better grasp of them than I ... I like it too

bparees:
it looks like this flag only applies to binary builds?  if so it needs logic to reject attempts to use it when --from-* isn't specified.

(and actually it seems like it only applies to --from-dir so it should reject the flag if --from-dir is not specified).

otherwise i'm ok with this as a feature given the relatively narrow scope.

pkg/cli/startbuild/startbuild.go
```diff
@@ -171,6 +173,7 @@ func NewCmdStartBuild(fullName string, f kcmdutil.Factory, streams genericcliopt
 	cmd.Flags().StringVar(&o.FromArchive, "from-archive", o.FromArchive, "An archive (tar, tar.gz, zip) to be extracted before the build and used as the binary input.")
 	cmd.Flags().StringVar(&o.FromRepo, "from-repo", o.FromRepo, "The path to a local source code repository to use as the binary input for a build.")
 	cmd.Flags().StringVar(&o.Commit, "commit", o.Commit, "Specify the source code commit identifier the build should use; requires a build based on a Git repository")
+	cmd.Flags().StringVarP(&o.ExcludeRegExp, "exclude", "", tar.DefaultExclusionPattern.String(), "Regular expression for selecting files from the source tree to exclude from the build, where the default excludes the '.git' directory (see https://golang.org/pkg/regexp for syntax, but note that \"\" will be interpreted as allow all files and exclude no files)")
```
bparees: text needs to make it clear this only applies to builds started with --from-dir.

pkg/cli/startbuild/startbuild_test.go
tomconte: 👍🏻

pkg/cli/startbuild/startbuild.go
tomconte: Fixed the usage test and added a test to ensure --exclude is only used with --from-dir.

pkg/cli/startbuild/startbuild.go
```diff
@@ -171,6 +173,7 @@ func NewCmdStartBuild(fullName string, f kcmdutil.Factory, streams genericcliopt
 	cmd.Flags().StringVar(&o.FromArchive, "from-archive", o.FromArchive, "An archive (tar, tar.gz, zip) to be extracted before the build and used as the binary input.")
 	cmd.Flags().StringVar(&o.FromRepo, "from-repo", o.FromRepo, "The path to a local source code repository to use as the binary input for a build.")
 	cmd.Flags().StringVar(&o.Commit, "commit", o.Commit, "Specify the source code commit identifier the build should use; requires a build based on a Git repository")
+	cmd.Flags().StringVarP(&o.ExcludeRegExp, "exclude", "", tar.DefaultExclusionPattern.String(), "When using the --from-dir option: regular expression for selecting files from the source tree to exclude from the build; the default excludes the '.git' directory (see https://golang.org/pkg/regexp for syntax, but note that \"\" will be interpreted as allow all files and exclude no files)")
```
bparees: is this changing the default behavior? (was `.git` excluded by default before?)  we can't change default behaviors.

tomconte: https://github.com/openshift/oc/blob/master/pkg/helpers/source-to-image/tar/tar.go#L138

tomconte: Yes `.git` was excluded by default, per the default exclusion pattern in `tar.go`:

https://github.com/openshift/oc/blob/master/pkg/helpers/source-to-image/tar/tar.go#L29

tomconte: My use case was actually that I needed to *include* `.git` and thus needed a flag to control that pattern.

bparees: ack, thanks

bparees:
looks ok to me.  @soltysh @adambkaplan should be the approvers on it though.

PR #245, opened 2020-01-09
Fixes wrong placement of waitgroup.Add function

ashishranjan738:
Signed-off-by: Ashish Ranjan <ashishranjan738@gmail.com>

This commit fixes wrong placement of `waitgroup.Add` function in log collection method for `oc adm inspect` command. The `waitgroup.Add` function should be called before the goroutine is called.


ashishranjan738:
@soltysh @deads2k @mfojtik we need this PR to get backported to 4.3 branch

PR #250, opened 2020-01-13
Bug 1779519: fix version ldflags

soltysh:
/assign @sallyom @mfojtik


soltysh:
I'll put together fixed for 4.3, but it'll probably be slightly different, since 4.3 was build with different version of go which required different way of injecting that version.

sallyom:
With this PR I see 
```
   24066 round_trippers.go:431]     User-Agent: oc/v4.2.0 (linux/amd64) kubernetes/ad961d5
```
/lgtm

PR #257, opened 2020-01-15
Bug 1790978: Zeroing BearerTokenFile for prune.

ricardomaraschini:
Kubernetes client-go requires only one authentication method to be present therefore this patch zeroes the content of BearerTokenFile as basic auth is used during prune.
    
In a nutshell: or we use token based authentication or basic auth, never both at the same time.

Note: This patch is also necessary to fix [Bug 1762837](https://bugzilla.redhat.com/show_bug.cgi?id=1762837)

### DIFF
--- a/pkg/cli/admin/prune/images/images.go
+++ b/pkg/cli/admin/prune/images/images.go
@@ -834,6 +834,7 @@ func getRegistryClient(clientConfig *restclient.Config, registryCABundle string,
 
 	// zero out everything we don't want to use
 	registryClientConfig.BearerToken = ""
+	registryClientConfig.BearerTokenFile = ""
 	registryClientConfig.CertFile = ""
 	registryClientConfig.CertData = []byte{}
 	registryClientConfig.KeyFile = ""


pkg/cli/admin/prune/images/images.go
```diff
@@ -852,8 +852,12 @@ func getRegistryClient(clientConfig *restclient.Config, registryCABundle string,
 		registryCABundleIncluded = true
 	}
 
-	// we have to set a username to something for the Docker login but it's not actually used
-	registryClientConfig.Username = "unused"
+	// Docker client-go library requires only one authentication method to be present. If we
+	// don't have a token file set we need to define at least an username so basic auth will
+	// be the chosen authentication method.
+	if registryClientConfig.BearerTokenFile == "" {
+		registryClientConfig.Username = "unused"
```
dmage: @ricardomaraschini I guess you want to zero BearerTokenFile in the `// zero out everything we don't want to use` block above.

ricardomaraschini: Oh yes, thanks!

PR #261, opened 2020-01-17
Bug 1792330: collect endpoints for namespaces

deads2k:
we want this so we can see if things like etcd have valid endpoints.

@soltysh needs a pick into 4.3, so you'll want to do a bug dance.  Also, I suspect that you really want to find a way to easily wire this into a service get, not a general namespace blast.

/assign @soltysh


soltysh:
/hold
for debugging

soltysh:
I reported the upgrade failures in https://bugzilla.redhat.com/show_bug.cgi?id=1792906

/retest

PR #262, opened 2020-01-17
Bug 1792330: pods that are no longer running can still have logs

deads2k:
When pods are stopped or crashlooped, they can still have logs.  We want to collect those logs.

/assign @soltysh 

You'll need to pick this one too.

### DIFF
--- a/pkg/cli/admin/inspect/pod.go
+++ b/pkg/cli/admin/inspect/pod.go
@@ -17,11 +17,6 @@ import (
 )
 
 func (o *InspectOptions) gatherPodData(destDir, namespace string, pod *corev1.Pod) error {
-	if pod.Status.Phase != corev1.PodRunning {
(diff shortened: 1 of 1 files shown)


ashishranjan738:
I guess this statement https://github.com/openshift/oc/blob/8145eb2c8a0e07eab9ccb299db033843febfb783/pkg/cli/admin/inspect/pod.go#L20  also needs to be removed @soltysh @deads2k

ashishranjan738:
> I guess this statement
> 
> https://github.com/openshift/oc/blob/8145eb2c8a0e07eab9ccb299db033843febfb783/pkg/cli/admin/inspect/pod.go#L20
> 
> also needs to be removed @soltysh @deads2k

there can be some pods in Completed state, and they can contain logs.

deads2k:
updated

PR #275, opened 2020-01-30
Bug 1746149: fix tag creation when imagestreamtags are not accessible

dmage:


dmage:
@adambkaplan I created the PR https://github.com/openshift/origin/pull/24505 to add tests.

PR #277, opened 2020-01-31
Bug 1771552: Bug 1771553: Bug 1771549: Bug 1798549: oc debug

sallyom:
/assign @soltysh 

1 commit:
* [Bug 1771552](https://bugzilla.redhat.com/show_bug.cgi?id=1771552)
In case of bad command passed to oc debug, user will now get information about why the command failed to run, and also will not have to wait the Timeout (15 min)  to get the output.  

2 commit:
* [Bug 1771553](https://bugzilla.redhat.com/show_bug.cgi?id=1771553)
If container exit code != 0, debug cmd should report non-zero
* [Bug 1771549](https://bugzilla.redhat.com/show_bug.cgi?id=1771549)
oc debug pod logs should go to stderr if container exit code != 0
* [Bug 1798549](https://bugzilla.redhat.com/show_bug.cgi?id=1798549)
oc debug node/nodename should fail quickly if debug pod image pull fails

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -444,7 +444,7 @@ func (o *DebugOptions) RunDebug() error {
 
 		ctx, cancel := context.WithTimeout(context.Background(), o.Timeout)
 		defer cancel()
-		switch containerRunningEvent, err := watchtools.UntilWithoutRetry(ctx, w, conditions.PodContainerRunning(o.Attach.ContainerName)); {
+		switch containerRunningEvent, err := watchtools.UntilWithoutRetry(ctx, w, conditions.PodContainerRunning(o.Attach.ContainerName, o.CoreClient)); {
 		// api didn't error right away but the pod wasn't even created
 		case kapierrors.IsNotFound(err):
 			msg := fmt.Sprintf("unable to create the debug pod %q", pod.Name)
@@ -453,20 +453,17 @@ func (o *DebugOptions) RunDebug() error {
 			}
 			return fmt.Errorf(msg)
 			// switch to logging output
-		case err == krun.ErrPodCompleted, err == conditions.ErrContainerTerminated, !o.Attach.Stdin:
-			return logs.LogsOptions{
-				Object: pod,
-				Options: &corev1.PodLogOptions{
-					Container: o.Attach.ContainerName,
-					Follow:    true,
-				},
-				RESTClientGetter: o.RESTClientGetter,
-				ConsumeRequestFn: logs.DefaultConsumeRequest,
-				IOStreams:        o.IOStreams,
-				LogsForObject:    o.LogsForObject,
-			}.RunLogs()
+		case err == krun.ErrPodCompleted, err == conditions.ErrContainerTerminated:
+			return o.getLogs(pod)
+		case err == conditions.ErrNonZeroExitCode:
+			if err = o.getLogs(pod); err != nil {
+				return err
+			}
+			return conditions.ErrNonZeroExitCode
 		case err != nil:
 			return err
+		case !o.Attach.Stdin:
+			return o.getLogs(pod)
 		default:
 			// TODO this doesn't do us much good for remote debugging sessions, but until we get a local port
 			// set up to proxy, this is what we've got.
@@ -960,6 +957,20 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 	}
 
 	return nil, fmt.Errorf("unable to extract pod template from type %v", reflect.TypeOf(object))
+}
+
+func (o *DebugOptions) getLogs(pod *corev1.Pod) error {
+	return logs.LogsOptions{
+		Object: pod,
+		Options: &corev1.PodLogOptions{
+			Container: o.Attach.ContainerName,
+			Follow:    true,
+		},
+		RESTClientGetter: o.RESTClientGetter,
+		ConsumeRequestFn: logs.DefaultConsumeRequest,
+		IOStreams:        o.IOStreams,
+		LogsForObject:    o.LogsForObject,
+	}.RunLogs()
 }
 
 func setNodeName(template *corev1.PodTemplateSpec, nodeName string) *corev1.PodTemplateSpec {
--- a/pkg/helpers/conditions/conditions.go
+++ b/pkg/helpers/conditions/conditions.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
 	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/klog"
 	krun "k8s.io/kubectl/pkg/cmd/run"
 )
 
@@ -27,6 +28,12 @@ func PodContainerRunning(containerName string) watchtools.ConditionFunc {
 		case *corev1.Pod:
 			switch t.Status.Phase {
 			case corev1.PodRunning, corev1.PodPending:
+				if t.Status.ContainerStatuses[0].State.Waiting != nil {
+					if t.Status.ContainerStatuses[0].State.Waiting.Reason == "CreateContainerError" {
+						klog.V(0).Info(t.Status.ContainerStatuses[0].State.Waiting.Message)
+						return false, fmt.Errorf(t.Status.ContainerStatuses[0].State.Waiting.Message)
+					}
+				}
 			case corev1.PodFailed, corev1.PodSucceeded:
 				return false, krun.ErrPodCompleted
 			default:
--- a/pkg/helpers/conditions/conditions.go
+++ b/pkg/helpers/conditions/conditions.go
@@ -7,18 +7,19 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	watchtools "k8s.io/client-go/tools/watch"
-	"k8s.io/klog"
 	krun "k8s.io/kubectl/pkg/cmd/run"
 )
 
 // ErrContainerTerminated is returned by PodContainerRunning in the intermediate
 // state where the pod indicates it's still running, but its container is already terminated
 var ErrContainerTerminated = fmt.Errorf("container terminated")
+var ErrNonZeroExitCode = fmt.Errorf("non-zero exit code from debug container")
 
 // PodContainerRunning returns false until the named container has ContainerStatus running (at least once),
 // and will return an error if the pod is deleted, runs to completion, or the container pod is not available.
-func PodContainerRunning(containerName string) watchtools.ConditionFunc {
+func PodContainerRunning(containerName string, coreClient corev1client.CoreV1Interface) watchtools.ConditionFunc {
 	return func(event watch.Event) (bool, error) {
 		switch event.Type {
 		case watch.Deleted:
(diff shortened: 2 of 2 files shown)


pkg/cli/debug/debug.go
```diff
@@ -157,7 +157,7 @@ func NewDebugOptions(streams genericclioptions.IOStreams) *DebugOptions {
 	return &DebugOptions{
 		PrintFlags:         genericclioptions.NewPrintFlags("").WithTypeSetter(scheme.Scheme),
 		IOStreams:          streams,
-		Timeout:            15 * time.Minute,
+		Timeout:            10 * time.Minute,
```
soltysh: I'd prefer we don't change this timeout, unless we have valid proof we should.

pkg/helpers/conditions/conditions.go
```diff
@@ -27,6 +28,12 @@ func PodContainerRunning(containerName string) watchtools.ConditionFunc {
 		case *corev1.Pod:
 			switch t.Status.Phase {
 			case corev1.PodRunning, corev1.PodPending:
+				if t.Status.ContainerStatuses[0].State.Waiting != nil {
+					if t.Status.ContainerStatuses[0].State.Waiting.Reason == "CreateContainerError" {
+						klog.V(0).Info(t.Status.ContainerStatuses[0].State.Waiting.Message)
```
soltysh: More info what how and why, same applies for the next error.

pkg/helpers/conditions/conditions.go
```diff
@@ -28,13 +31,28 @@ func PodContainerRunning(containerName string) watchtools.ConditionFunc {
...
+				for _, s := range t.Status.ContainerStatuses {
+					if s.State.Terminated != nil {
+						exitCode := s.State.Terminated.ExitCode
+						if exitCode != 0 {
+							klog.V(0).Infof("non-zero exit code(%d) from container %s", exitCode, s.Name)
+							// delete the existing pod
+							if err := coreClient.Pods(t.Namespace).Delete(t.Name, metav1.NewDeleteOptions(0)); err != nil && !errors.IsNotFound(err) {
+								return false, fmt.Errorf("unable to delete existing debug pod %q: %v", s.Name, err)
+							}
+							os.Exit(int(exitCode))
```
soltysh: Helpers can't and should never `os.Exit` it should return error and only there should be a proper interpretation of the error and `os.Exit`

pkg/cli/debug/debug.go
```diff
@@ -453,20 +456,40 @@ func (o *DebugOptions) RunDebug() error {
...
+					Object: pod,
+					Options: &corev1.PodLogOptions{
+						Container: o.Attach.ContainerName,
+						Follow:    true,
+					},
+					RESTClientGetter: o.RESTClientGetter,
+					ConsumeRequestFn: logs.DefaultConsumeRequest,
+					IOStreams:        streams,
+					LogsForObject:    o.LogsForObject,
+				}.RunLogs())
```
soltysh: Why can't you just re-use `logsOptions` here passing streams accordingly?

pkg/cli/debug/debug.go
```diff
@@ -453,20 +456,40 @@ func (o *DebugOptions) RunDebug() error {
...
+				w.Close()
+			}()
+			buf := new(bytes.Buffer)
+			buf.ReadFrom(r)
+			trimmed := strings.TrimSuffix(buf.String(), "<nil>")
```
soltysh: What's this?

pkg/cli/debug/debug.go
```diff
@@ -453,20 +456,40 @@ func (o *DebugOptions) RunDebug() error {
...
+			o.logsOptions(pod)
+		case err == conditions.ErrNonZeroExitCode:
+			r, w := io.Pipe()
+			streams := genericclioptions.IOStreams{
+				In:     r,
+				Out:    w,
+				ErrOut: ioutil.Discard,
+			}
+			go func() {
+				fmt.Fprintf(w, "%v", logs.LogsOptions{
```
soltysh: I see what you're trying to do here, but frankly said I don't like it. It has to be cleaner.

pkg/cli/debug/debug.go
sallyom: the only way I found to redirect the output (non-error) to an error (which is what we want whenever container exits non-zero) have to capture the output and create the error. the output of `fmt.Fprintf(w, "%v", logs.LogsOptions... ` is 
"the logs string <nil>" the <nil> comes from the fact that the function `LogsOptions` returns an error (which is nil, even when the logs indicate the container errored out) so that is trimmed.

pkg/cli/debug/debug.go
sallyom: yea, i'm capturing the stdout and creating the error when appropriate (if container exits non-zero)

pkg/cli/debug/debug.go
sallyom: ok, i have no idea why it's 15 min, but i'll change it back

pkg/cli/debug/debug.go
sallyom: I don't like it much either, but the LogsOptions will return nil error as long as it completes, so the cmd will never return non-zero unless LogsOptions fails. I think the goal is for oc debug to return non-zero if the container returns non-zero

pkg/helpers/conditions/conditions.go
sallyom: that was from a previous commit, it was removed in subsequent commit - i'll try to clear the history there.

pkg/helpers/conditions/conditions.go
sallyom: I've added info

pkg/cli/debug/debug.go
sallyom: because you have to capture the output and push to error, otherwise the cmd output will _always_ return 0, even if container exits non-zero.

pkg/helpers/conditions/conditions.go
```diff
@@ -27,7 +30,32 @@ func PodContainerRunning(containerName string) watchtools.ConditionFunc {
 		case *corev1.Pod:
 			switch t.Status.Phase {
 			case corev1.PodRunning, corev1.PodPending:
+				for _, s := range t.Status.ContainerStatuses {
+					if s.State.Waiting != nil {
+						// Return error here if pod is pending and container status indicates a failure
+						// otherwise, user will have to wait the timeout period (15 min)
+						// for pod to exit.
+						if s.State.Waiting.Reason == "CreateContainerError" {
+							if s.State.Waiting.Message != "" {
```
soltysh: Drop this condition entirely, just return the error.

pkg/helpers/conditions/conditions.go
```diff
@@ -27,7 +30,32 @@ func PodContainerRunning(containerName string) watchtools.ConditionFunc {
...
+							return false, fmt.Errorf(s.State.Waiting.Message)
+						}
+					}
+				}
 			case corev1.PodFailed, corev1.PodSucceeded:
+				for _, s := range t.Status.ContainerStatuses {
+					if s.State.Terminated != nil {
+						exitCode := s.State.Terminated.ExitCode
+						if exitCode != 0 {
+							klog.V(0).Infof("non-zero exit code(%d) from container %s", exitCode, s.Name)
```
soltysh: Drop the log here.

pkg/cli/debug/debug.go
soltysh: Let's split this PR, this bit that's questionable - open a separate PR for it, the rest ([BZ#1771552](https://bugzilla.redhat.com/show_bug.cgi?id=1771552) and [BZ#1771553](https://bugzilla.redhat.com/show_bug.cgi?id=1771553)) leave here.

pkg/helpers/conditions/conditions.go
soltysh: Add `ImagePullBackOff` condition here and we'll be able to close https://bugzilla.redhat.com/show_bug.cgi?id=1798549 too

pkg/cli/debug/debug.go
sallyom: @soltysh I've reworked the fn to return logs if container fails, i think it's much better now, take a look - perhaps we can kill all 4 bzs now

pkg/cli/debug/debug.go
sallyom: see my comment above, please take another look : )

pkg/helpers/conditions/conditions.go
sallyom: done, lol not w/out failing for awhile because I was checking for `ImagePullBackoff` not `ImagePullBackOff`  :woman_shrugging:

pkg/helpers/conditions/conditions.go
sallyom: done

pkg/cli/debug/debug.go
```diff
@@ -962,6 +957,40 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
...
+			Container: o.Attach.ContainerName,
+			Follow:    true,
+		},
+		RESTClientGetter: o.RESTClientGetter,
+		ConsumeRequestFn: logs.DefaultConsumeRequest,
+		IOStreams:        o.IOStreams,
+		LogsForObject:    o.LogsForObject,
+	}.RunLogs()
+}
+
```
sallyom: No change in the above, I just moved it to a function so as not to repeat
Below, I've added the `getPodLogsForError` to return the podlogs as error in case of non-zero container exit.

pkg/cli/debug/debug.go
sallyom: I've added a new getPodLogsForError fn, ptal

sallyom:
new output:
```console
localhost:oc(node-debug)$ oc debug node/nodename -- false
Starting pod/sotest-lf2xg-w-b-hwwj9copenshift-gce-develinternal-debug ...

Removing debug pod ...
error: non-zero exit code from debug container
localhost:oc(node-debug)$ echo $?
1
localhost:oc(node-debug)$ oc debug node/nodename -- /bin/foo
Starting pod/sotest-lf2xg-w-b-hwwj9copenshift-gce-develinternal-debug ...

Removing debug pod ...
error: container create failed: time="2020-02-11T19:31:53Z" level=error msg="container_linux.go:346: starting container process caused \"exec: \\\"/bin/foo\\\": stat /bin/foo: no such file or directory\""
container_linux.go:346: starting container process caused "exec: \"/bin/foo\": stat /bin/foo: no such file or directory"
localhost:oc(node-debug)$ echo $?
1
localhost:oc(node-debug)$ oc debug node/nodename -- ls -al /foo
Starting pod/sotest-lf2xg-w-b-hwwj9copenshift-gce-develinternal-debug ...

Removing debug pod ...
error: ls: cannot access /foo: No such file or directory
localhost:oc(node-debug)$ echo $?
1
localhost:oc(node-debug)$ oc debug node/nodename -- ls
Starting pod/sotest-lf2xg-w-b-hwwj9copenshift-gce-develinternal-debug ...

Removing debug pod ...
error: Back-off pulling image "registry.redhat.io/rhel7/foo"
localhost:oc(node-debug)$ echo $?
1
localhost:oc(node-debug)$ oc debug node/nodename -- true
Starting pod/sotest-lf2xg-w-b-hwwj9copenshift-gce-develinternal-debug ...

Removing debug pod ...
localhost:oc(node-debug)$ echo $?
0
```

pkg/cli/debug/debug.go
sallyom: No change in the above `logsOptions`, I  moved it to a function so as not to repeat
Below, I've added the getPodLogsForError to return the podlogs as error in case of non-zero container exit.

pkg/cli/debug/debug.go
```diff
@@ -453,20 +453,14 @@ func (o *DebugOptions) RunDebug() error {
 			}
 			return fmt.Errorf(msg)
 			// switch to logging output
-		case err == krun.ErrPodCompleted, err == conditions.ErrContainerTerminated, !o.Attach.Stdin:
```
soltysh: My proposal for this switch:
```go
case err == krun.ErrPodCompleted, err == conditions.ErrContainerTerminated, !o.Attach.Stdin:
	return o.logsOptions(pod)
case err == conditions.ErrNonZeroExitCode:
	o.logsOptions(pod)
	return conditions.ErrNonZeroExitCode
case err != nil:
	return err
case !o.Attach.Stdin:
	return o.logsOptions(pod)
```
oh, and please change `logsOptions` to `getLogs` or something like that.

pkg/cli/debug/debug.go
soltysh: `getPodLogsForError` is not needed, see my above suggestion.

sallyom: great, thanks, updated : )

pkg/cli/debug/debug.go
sallyom: updated - but handled the error from geLogs(pod) - or should I not? :
```
case err == conditions.ErrNonZeroExitCode:
	if err = o.getLogs(pod); err != nil {
            return err
         }
	return conditions.ErrNonZeroExitCode
```

soltysh: Perfect :+1:

PR #281, opened 2020-02-04
Bug 1798622: oc adm release info: escape PR titles which might contain html

eparis:
Some PRs include things like <title> which break the release viewer
We also escape any instance of ` to prevent odd formatting

### DIFF
--- a/pkg/cli/admin/release/git.go
+++ b/pkg/cli/admin/release/git.go
@@ -6,6 +6,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"html"
 	"io"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/release/git.go
```diff
@@ -245,6 +246,13 @@ func mergeLogForRepo(g gitInterface, repo string, from, to string) ([]MergeCommi
 			mergeCommit.Subject = "Merge"
 		}
 
+		// Clean up the string so backticks in pr titles don't get formatted
+		// in the markdown output
+		mergeCommit.Subject = strings.Replace(mergeCommit.Subject, "`", "\\`", -1)
```
smarterclayton: Don't we want that?  I use this all the time:

> Bug 30498: Improve output of `oc status`

eparis:
for example https://github.com/openshift/console/pull/3255 breaks all subsequent PRs in the release viewer

pkg/cli/admin/release/git.go
smarterclayton: Or are you concerned about unbalanced backticks?

eparis: I believe when I found this a long time ago I found unbalanced backticks.  But right now i'd be happy just with the html escape. So I dropped this line.

PR #282, opened 2020-02-04
adm release info: refactor to build messages in order

eparis:
we did things like build the suffix then had lots of other strings
to stuff the suffix into. This builds and write the lines to out as
they are needed with no intermediate steps.


eparis:
/retest
The failures were yum repo mirroring problems. not ours

PR #288, opened 2020-02-06
Bug 1790785: `oc adm catalog build` should immediately error out if it can't find any data in appregistry 

Bowenislandsong:
This commit updates operator-registry version in go.mod inorder to include operator-registry bug fix to oc. this commit updates the operator-registry version to v1.5.9.


Bowenislandsong:
This is bumped by https://github.com/openshift/oc/pull/316
/close

PR #298, opened 2020-02-14
Bug 1800674: Option to force RepoDigest mirror

kevinrizza:
Currently `oc image mirror`, when pointed at a
manifest list digest, checks the contents of the manifest and mirrors
just the underlying image if it contains only a single image.

This change adds an option to mirroring to force the repodigest to
mirror in addition to the underlying image digest. Additionally, it
forces `oc image mirror` and `oc adm catalog mirror` to always
enable this option

https://bugzilla.redhat.com/show_bug.cgi?id=1800674

### DIFF
--- a/pkg/cli/image/mirror/mirror.go
+++ b/pkg/cli/image/mirror/mirror.go
@@ -100,6 +100,7 @@ type MirrorImageOptions struct {
 	SkipMultipleScopes bool
 	SkipMissing        bool
 	Force              bool
+	ForceManifestList  bool
 
 	MaxRegistry     int
 	ParallelOptions imagemanifest.ParallelOptions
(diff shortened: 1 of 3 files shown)


benluddy:
Kevin's out today, so I went ahead and regenerated the shell completions in https://github.com/openshift/oc/pull/301. If this change doesn't need to merge today, I'll close my PR.

pkg/cli/image/mirror/mirror.go
```diff
@@ -149,6 +150,7 @@ func NewCmdMirrorImage(name string, streams genericclioptions.IOStreams) *cobra.
 	flag.BoolVar(&o.SkipMount, "skip-mount", o.SkipMount, "Always push layers instead of cross-mounting them")
 	flag.BoolVar(&o.SkipMultipleScopes, "skip-multiple-scopes", o.SkipMultipleScopes, "Some registries do not support multiple scopes passed to the registry login.")
 	flag.BoolVar(&o.Force, "force", o.Force, "Attempt to write all layers and manifests even if they exist in the remote repository.")
+	flag.BoolVar(&o.ForceManifestList, "force-manifest-list", o.ForceManifestList, "Always attempt to mirror manifestlist even if manifestlist contains a single image digest")
```
smarterclayton: I am a bit concerned here that the default behavior of image mirror is wrong when users specify `--filter-by-os=*`, regardless of this flag. Mirror doesn't have any issue dealing with manifest lists, so I think ProcessManifestList within mirror should always have the equivalent of `force=true`.  That being said, mirror for release payloads *should* resolve image streams, so it can't be the default.

That may be a behavior change, but at least on the surface this looks like a bug that mirror by default strips manifest lists.

soltysh: @smarterclayton so you're basically saying the fix is correct, but we should not expose that outside for users, @kevinrizza would you mind updating?

kevinrizza:
@smarterclayton @soltysh Updated the PR to remove the command line flag from `oc image mirror`. Let me know if you have any other thoughts. Thanks!

PR #299, opened 2020-02-17
Bug 1801064: Add '\n' for pipeline deprecate info

wewang58:
/assign @adambkaplan

### DIFF
--- a/pkg/cli/newapp/newapp.go
+++ b/pkg/cli/newapp/newapp.go
@@ -458,7 +458,7 @@ func (o *AppOptions) RunNewApp() error {
 			}
 		case *buildv1.BuildConfig:
 			if t.Spec.Strategy.Type == buildv1.JenkinsPipelineBuildStrategyType {
-				fmt.Fprintf(o.Action.ErrOut, "JenkinsPipeline build strategy is deprecated. Use Jenkinsfiles directly on Jenkins or OpenShift Pipelines instead")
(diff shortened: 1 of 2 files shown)


adambkaplan:
@soltysh is there a reason the newline is not needed in https://github.com/openshift/oc/blob/master/pkg/cli/rollout/latest.go#L114 ?

pkg/cli/newapp/newapp.go
```diff
@@ -458,7 +458,7 @@ func (o *AppOptions) RunNewApp() error {
 			}
 		case *buildv1.BuildConfig:
 			if t.Spec.Strategy.Type == buildv1.JenkinsPipelineBuildStrategyType {
-				fmt.Fprintf(o.Action.ErrOut, "JenkinsPipeline build strategy is deprecated. Use Jenkinsfiles directly on Jenkins or OpenShift Pipelines instead")
+				fmt.Fprintf(o.Action.ErrOut, "JenkinsPipeline build strategy is deprecated. Use Jenkinsfiles directly on Jenkins or OpenShift Pipelines instead\n")
```
soltysh: Maybe just use `fmt.Fprintln` instead of manually adding the newline.

soltysh:
> @soltysh is there a reason the newline is not needed in https://github.com/openshift/oc/blob/master/pkg/cli/rollout/latest.go#L114 ?

With `Fprintf` you're responsible for formatting, that's why I propose to use `Fprintln` my bad I didn't catch it in your PR.

pkg/cli/newapp/newapp.go
wewang58: @soltysh Updated it, thanks

PR #300, opened 2020-02-17
Bug 1779675: [release-4.3] Fix ignored `--manifest-dir` option in `oc adm catalog build`

njhale:
Backports to 4.3 a bump of the operator-registry dependency to a version that handles the cache directory option correctly.

Also fixes a bad pseudoversion for the module replacement of openshift/api which was causing module operations to fail. I can cherrypick that change in a separate PR if that's preferable, LMK.


njhale:
Superseded by #288

njhale:
Spoke too soon. This is the closest to making into 4.3.

PR #301, opened 2020-02-17
Bug 1800674: Option to force RepoDigest mirror

benluddy:
Currently `oc image mirror`, when pointed at a manifest list digest,
checks the contents of the manifest and mirrors just the
underlying image if it contains only a single image.

This change adds an option to mirroring to force the repodigest to
mirror in addition to the underlying image digest. Additionally, it
forces `oc adm catalog mirror` to always enable this option since
the repodigest is what is encoded into the OLM manifests.

https://bugzilla.redhat.com/show_bug.cgi?id=1800674


benluddy:
This can be closed if the change in #298 can wait another day.

https://github.com/openshift/oc/pull/298#issuecomment-587026530

soltysh:
There's one additional comment that needs fixing in the other PR. 
/close

PR #303, opened 2020-02-19
Bug 1800464: fix(catalog): use registry image from openshift quay namespace

njhale:
Updates the default image used in the `--from` option of `oc adm catalog build` to the official mirror.

### DIFF
--- a/pkg/cli/admin/catalog/build.go
+++ b/pkg/cli/admin/catalog/build.go
@@ -42,9 +42,15 @@ type BuildImageOptions struct {
 	FileDir     string
 }
 
+const defaultFromImage = "quay.io/openshift/origin-operator-registry:latest"
+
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/catalog/build.go
```diff
@@ -43,8 +43,12 @@ type BuildImageOptions struct {
 }
 
 func NewBuildImageOptions(streams genericclioptions.IOStreams) *BuildImageOptions {
+	// Adjust default build opts to use an OpenShift base "from" image
+	defaults := appregistry.DefaultAppregistryBuildOptions()
+	defaults.From = "quay.io/openshift/origin-operator-registry:latest"
```
ecordell: should be a constant at the top?

njhale: Sure, I'll go ahead and update that.

It seems like `e2e-aws-upgrade is consistently failing -- think it might be related?

PR #316, opened 2020-02-21
Bug 1790785: bump(operator-framework/operator-registry)

ecordell:
Bumps operator-registry, which fixes two BZs:

https://bugzilla.redhat.com/show_bug.cgi?id=1790785
https://bugzilla.redhat.com/show_bug.cgi?id=1772942 (already fixed in master, but this needs to propagate to 4.3)


go.mod
```diff
@@ -33,11 +33,11 @@ require (
 	github.com/moby/buildkit v0.0.0-20181107081847-c3a857e3fca0
 	github.com/mtrmac/gpgme v0.1.2 // indirect
 	github.com/opencontainers/go-digest v1.0.0-rc1
-	github.com/openshift/api v0.0.0-20200205133042-34f0ec8dab87
+	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
```
soltysh: This is unexpected change.

go.mod
```diff
@@ -55,18 +55,19 @@ require (
 	k8s.io/apimachinery v0.17.1
 	k8s.io/apiserver v0.17.1
 	k8s.io/cli-runtime v0.17.0
-	k8s.io/client-go v0.17.1
+	k8s.io/client-go v8.0.0+incompatible
```
soltysh: Same this one.

go.mod
```diff
@@ -55,18 +55,19 @@ require (
...
 	k8s.io/component-base v0.17.1
 	k8s.io/klog v1.0.0
 	k8s.io/kubectl v0.0.0
-	k8s.io/kubernetes v0.0.0-00010101000000-000000000000
+	k8s.io/kubernetes v1.16.0
```
soltysh: Ditto.

go.mod
```diff
@@ -55,18 +55,19 @@ require (
...
 	k8s.io/utils v0.0.0-20191114184206-e782cd3c129f
 	sigs.k8s.io/yaml v1.1.0
 )
 
 replace (
 	github.com/apcera/gssapi => github.com/openshift/gssapi v0.0.0-20161010215902-5fb4217df13b
 	github.com/containers/image => github.com/openshift/containers-image v0.0.0-20190130162819-76de87591e9d
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20200205133042-34f0ec8dab87
```
soltysh: This is not needed.

go.mod
Bowenislandsong: I do want to point out that this is replaced later at https://github.com/openshift/oc/blob/f44732172c3dcf854f578aabd4465ff7e517ef2d/go.mod#L77. Therefore it does not matter that is it changed at this stage.

go.mod
Bowenislandsong: Same as above at https://github.com/openshift/oc/blob/f44732172c3dcf854f578aabd4465ff7e517ef2d/go.mod#L90.

go.mod
Bowenislandsong: This replaces https://github.com/openshift/oc/pull/316/files#r383280617, therefore, it does not change dependencies at all.

go.mod
Bowenislandsong: I will try to resolve this, but the other 2 are replaced anyway. They should be harmless. Please let me know what you think. Thank you.

soltysh: From what I checked this is somehow a transitive dependency that's coming only when I change operator-registry, if I bump openshift deps this will work just fine picking newer version, which makes me worried a lot about that bump.

soltysh:
/hold
I'll have another look at it

soltysh:
I'll keep this on hold, I have agreement with @ecordell he'll look into fixing some of deps in operator-registry.

Bowenislandsong:
I want to mention that this solution makes the least amount of effort in changes and it is easy to backport.

PR #318, opened 2020-02-24
override FileDir with FromFileDir if set in Complete

pweil-:
`Complete` gives error `error: you must specify at least one source image to pull and the destination to push to as SRC=DST or SRC DST [DST2 DST3 ...]` if only `--from-dir` is specified even though it is supposed to override `--dir`.

This adds the override logic to Complete like what happens during `Run().plan.Repository`.


pweil-:
for any interested party, this was found testing air-gapped mirroring with the steps here: https://gist.github.com/pweil-/e7b156476c6171f04140370708f0cd56

The `oc image mirror` command started with `--from-dir` which is how I ran into the bug.  Using `--dir` works fine so I don't think this is necessarily a backport candidate.

PR #320, opened 2020-02-25
Bug 1806779: Allow a file:// release argument to release mirror

smarterclayton:
When a file:// argument is provided as the source for oc adm release
mirror, do a digest lookup within that repo for any image referenced
by the payload, and if found use the local directory as the source.
This allows releases to be mirrored to disk and then oc adm release
mirror to be used to push it back with:

    oc adm release mirror file://openshift/release:4.3.1 --to REPO

where before it would fail with "invalid image reference".

Ensure that when file:// is used the correct instructions are printed
for mirroring.

Fix the mirror command provided to users to put the wildcard in
quotes (to avoid bash shell expansion).


smarterclayton:
/assign @wking 

Found these while reviewing the documentation and fielding questions from folks trying the offline mode.

smarterclayton:
4.4 bug is 1806780 and 4.3 bug is 1806782

pkg/cli/admin/release/mirror.go
```diff
@@ -307,27 +322,64 @@ func (o *MirrorOptions) Run() error {
...
+	// sourceFn is given a chance to rewrite source mappings
+	sourceFn := func(ref imagesource.TypedImageReference) imagesource.TypedImageReference {
+		return ref
+	}
 	var mappings []mirror.Mapping
-	if len(o.From) > 0 && !o.SkipRelease {
+	if len(o.From) > 0 {
 		src := o.From
-		srcRef, err := imagereference.Parse(src)
+		srcRef, err := imagesource.ParseReference(src)
```
soltysh: This will also allow `s3://` reference, not sure if you want to allow that?

smarterclayton: I think we do at some point, just haven't gotten a chance to get back to it.  We're keeping the s3 in our pocket so we can mirror to static stuff like file.  In theory s3 and file should always work everywhere.

PR #329, opened 2020-02-27
Bug 1790785: bump(operator-framework/operator-registry)

Bowenislandsong:
Bumps operator-registry, which fixes two BZs:

https://bugzilla.redhat.com/show_bug.cgi?id=1790785
https://bugzilla.redhat.com/show_bug.cgi?id=1772942 (already fixed in master, but this needs to propagate to 4.3)

This commit supersedes https://github.com/openshift/oc/pull/316.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/alexbrainman/sspi v0.0.0-20180613141037-e580b900e9f5
 	github.com/apcera/gssapi v0.0.0-00010101000000-000000000000
 	github.com/aws/aws-sdk-go v1.17.7
-	github.com/blang/semver v3.5.0+incompatible
+	github.com/blang/semver v3.5.1+incompatible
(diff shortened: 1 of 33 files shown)


go.mod
```diff
@@ -55,7 +55,7 @@ require (
 	k8s.io/apimachinery v0.17.1
 	k8s.io/apiserver v0.17.1
 	k8s.io/cli-runtime v0.17.0
-	k8s.io/client-go v0.17.1
+	k8s.io/client-go v8.0.0+incompatible
```
Bowenislandsong: I do want to point out that this is replaced later at https://github.com/openshift/oc/blob/f44732172c3dcf854f578aabd4465ff7e517ef2d/go.mod#L77. Therefore it does not matter that is it changed at this stage.

tnozicka: it matters for transitive dependencies

PR #332, opened 2020-02-28
`set volume` does not work with `--local`

smarterclayton:
When `--local` is set `info.Mapping` is nil, which causes panics in the
command. Use the external patches method and remove the old helper that no
longer works, and fix places in the error path where we assume mapping is
present. This was missed when the upstream changed.

Also fix bug where --local causes the object to be printed twice.

Needs a test/cmd test


pkg/cli/set/volume.go
```diff
@@ -494,6 +494,7 @@ func (o *VolumeOptions) RunVolume() error {
 				return err
 			}
 		}
+		return nil
```
soltysh: We should drop similar lines few lines below, I'll handle that in a followup.

soltysh: Done in https://github.com/openshift/oc/pull/349

PR #333, opened 2020-02-28
Fix oc new-app example

drpaneas:
Fixes a cosmetic issue introduced with the replacement of the
example from Django to Ruby-ex.


drpaneas:
I forgot about it but github remembers ;) Any news on this?

PR #338, opened 2020-03-04
Bug 1810249: 'oc debug: '--to-namespace' only supported with 'oc debug node' not w/ other resources

sallyom:
/assign @soltysh 

this is to fix 4.4: https://bugzilla.redhat.com/show_bug.cgi?id=1782260 

Volumes/VolumeMounts are dependent on the ns, as are ServiceAccountName/DeprecatedServiceAccount `oc debug anything-but-node/name --to-namespace` will fail bc of this (most of the time, or whenever there are volumemounts/serviceaccounts).  InitContainers also cause debug to fail when --to-namespace.  

I've added logic to clear Volumes,VolumeMounts,InitContainers,ServiceAccountName,DeprecatedServiceAccount to allow passing `--to-namespace` when running `oc debug anything-but-node`

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -594,6 +594,25 @@ func (o *DebugOptions) getContainerImageCommand(pod *corev1.Pod, container *core
 func (o *DebugOptions) transformPodForDebug(annotations map[string]string) (*corev1.Pod, []string) {
 	pod := o.Attach.Pod
 
+	if len(o.ToNamespace) > 0 {
+		svcAcctOrig := pod.Spec.ServiceAccountName
+		if !o.IsNode {
+			if len(pod.Spec.Volumes) > 0 {
+				fmt.Fprintf(o.ErrOut, "Volumes and VolumeMounts were cleared in debug pod to run with '--to-namespace', run without '--to-namespace' to keep Volumes,VolumeMounts\n")
+				clearVolumes(pod)
+			}
+		}
+		if len(pod.Spec.ServiceAccountName) != 0 {
+			pod.Spec.ServiceAccountName = ""
+		}
+		if len(pod.Spec.DeprecatedServiceAccount) != 0 {
+			pod.Spec.DeprecatedServiceAccount = ""
+		}
+		if svcAcctOrig != pod.Spec.ServiceAccountName {
+			fmt.Fprintf(o.ErrOut, "ServiceAccountName was cleared in debug pod to run with '--to-namespace', run without '--to-namespace' to keep ServiceAccountName\n")
+		}
+	}
+
 	// reset the container
 	container := containerForName(pod, o.Attach.ContainerName)
 
@@ -658,6 +677,12 @@ func (o *DebugOptions) transformPodForDebug(annotations map[string]string) (*cor
 		container.SecurityContext.RunAsNonRoot = nil
 	}
 
+	// When run in another namespace than original pod, any initContainers that run a command
+	// that depends on VolumeMounts will fail, so unless the initContainer is explicitly being
+	// debugged, remove the initContainers.
+	if len(o.ToNamespace) > 0 && !isInitContainer(pod, o.Attach.ContainerName) {
+		o.KeepInitContainers = false
+	}
 	switch {
 	case o.OneContainer:
 		pod.Spec.InitContainers = nil
@@ -786,6 +811,16 @@ func clearHostPorts(pod *corev1.Pod) {
 				pod.Spec.Containers[i].Ports[j].HostPort = 0
 			}
 		}
+	}
+}
+
+func clearVolumes(pod *corev1.Pod) {
+	pod.Spec.Volumes = []corev1.Volume{}
+	for i := range pod.Spec.Containers {
+		pod.Spec.Containers[i].VolumeMounts = []corev1.VolumeMount{}
+	}
+	for i := range pod.Spec.InitContainers {
+		pod.Spec.InitContainers[i].VolumeMounts = []corev1.VolumeMount{}
 	}
 }
 


pkg/cli/debug/debug.go
```diff
@@ -859,17 +864,6 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
...
-		// create a minimal pod spec that uses the image referenced by the istag without any introspection
-		// it possible that we could someday do a better job introspecting it
-		return setNodeName(&corev1.PodTemplateSpec{
-			Spec: corev1.PodSpec{
-				RestartPolicy: corev1.RestartPolicyNever,
-				Containers: []corev1.Container{
-					{Name: "container-00", Image: t.Image.DockerImageReference},
-				},
-			},
-		}, o.NodeName), nil
```
sallyom: this is a dupe, cleanup

pkg/cli/debug/debug.go
```diff
@@ -848,6 +848,11 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 				},
 			},
 		}, nil
+	}
+	if len(o.ToNamespace) > 0 {
```
soltysh: Nope, we need this for everything else too.

pkg/cli/debug/debug.go
soltysh: It's not dupe, the above is ImageStream**Tag** this is IS**Image**

sallyom: :woman_facepalming: I swear I stared at that/quadruple-checked!

pkg/cli/debug/debug.go
sallyom: ok, my first go was to clear the Volumes/VMounts, but then still saw fails w/ ServiceAccounts, so I'll clear both when ToNamespace.

pkg/cli/debug/debug.go
```diff
@@ -594,6 +594,30 @@ func (o *DebugOptions) getContainerImageCommand(pod *corev1.Pod, container *core
...
+		if svcAcctOrig != pod.Spec.ServiceAccountName {
+			klog.V(4).Infof("PodSpec ServiceAccountName was cleared to run with '--to-namespace', run without '--to-namespace' to keep ServiceAccountName")
+		}
+		if len(pod.Spec.Volumes) != len(volumesOrig) {
+			if volumesOrig[0] != pod.Spec.Volumes[0] {
+				klog.V(4).Infof("PodSpec Volumes and VolumeMounts were cleared to run with '--to-namespace', run without '--to-namespace' to keep Volumes,VolumeMounts")
+			}
+		}
+	}
+
```
sallyom: @soltysh 
These all cause debug to fail  when run in a different namespace.
oc debug node: no change
oc debug everything-else: at least 1 of initContainers/Volumes/Mounts/ServiceAcct/ServiceAcctName cleared

sallyom:
test infrastructure errs

/retest

pkg/cli/debug/debug.go
sallyom: I've updated

pkg/cli/debug/debug.go
```diff
@@ -594,6 +594,30 @@ func (o *DebugOptions) getContainerImageCommand(pod *corev1.Pod, container *core
...
+		volumesOrig := pod.Spec.Volumes
+		clearVolumes(pod)
+		if len(pod.Spec.ServiceAccountName) != 0 {
+			pod.Spec.ServiceAccountName = ""
+		}
+		if len(pod.Spec.DeprecatedServiceAccount) != 0 {
+			pod.Spec.DeprecatedServiceAccount = ""
+		}
+		if len(pod.Spec.InitContainers) != 0 {
+			klog.V(4).Infof("PodSpec InitContainers cleared to run with '--to-namespace', run without '--to-namespace' to keep InitContainers")
```
soltysh: Why initcontainers?

pkg/cli/debug/debug.go
```diff
@@ -594,6 +594,30 @@ func (o *DebugOptions) getContainerImageCommand(pod *corev1.Pod, container *core
...
+			pod.Spec.InitContainers = nil
+		}
+		if svcAcctOrig != pod.Spec.ServiceAccountName {
+			klog.V(4).Infof("PodSpec ServiceAccountName was cleared to run with '--to-namespace', run without '--to-namespace' to keep ServiceAccountName")
```
soltysh: Make it more explicit `fmt.Fprintf(o.ErrOut, "ServiceAccountName was cleared to run with...`

pkg/cli/debug/debug.go
```diff
@@ -594,6 +594,30 @@ func (o *DebugOptions) getContainerImageCommand(pod *corev1.Pod, container *core
...
+		}
+		if len(pod.Spec.Volumes) != len(volumesOrig) {
+			if volumesOrig[0] != pod.Spec.Volumes[0] {
+				klog.V(4).Infof("PodSpec Volumes and VolumeMounts were cleared to run with '--to-namespace', run without '--to-namespace' to keep Volumes,VolumeMounts")
```
soltysh: Same as above, and skip `PodSpec` from the beginning of that message.

pkg/cli/debug/debug.go
```diff
@@ -789,6 +813,29 @@ func clearHostPorts(pod *corev1.Pod) {
...
+	}
+	for i := range pod.Spec.Containers {
+		pod.Spec.Containers[i].VolumeMounts = []corev1.VolumeMount{
+			{
+				Name:      "host",
+				MountPath: "/host",
+			},
+		}
+	}
+}
```
soltysh: This doesn't clear volumes from what I see. I assume you wanted to address the case when you invoke `oc debug node`, but with this you always inject hostpath mount. I think you should filter out hostpathmount when clearing, rather.

pkg/cli/debug/debug.go
sallyom: I empirically found it to be true, didn't pinpoint why yet - I'll get back to you on that.

sallyom: I see why now, most initContainers are configured to run commands that depend on VolumeMounts, so if not debugging the initContainer, that cmd will try to run as pod starts, then will leave debug pod in error state w/ PodInitializing.  Now we'll clear InitContainers only if not explicitly debugging the initContainer (if you pass -c initContainerName, though, the command will become `/bin/sh` in the debug pod, so that will not fail when debugging initContainers.

pkg/cli/debug/debug.go
sallyom: got it

pkg/cli/debug/debug.go
sallyom: got it

pkg/cli/debug/debug.go
sallyom: right, fixed : ) thanks

pkg/cli/debug/debug.go
sallyom: initContainers need to be cleared (almost always) because their commands almost always fail if not in original namespace - the initContainer cmd usually depends on VolumeMounts.  However, when debugging initContainers themselves, with `oc debug pod/name -c initContainerName` we will not clear the initContainers, bc in that case the cmd will be simply `/bin/sh`

pkg/cli/debug/debug.go
```diff
@@ -594,6 +594,25 @@ func (o *DebugOptions) getContainerImageCommand(pod *corev1.Pod, container *core
 func (o *DebugOptions) transformPodForDebug(annotations map[string]string) (*corev1.Pod, []string) {
 	pod := o.Attach.Pod
 
+	if len(o.ToNamespace) > 0 {
+		svcAcctOrig := pod.Spec.ServiceAccountName
+		if !o.IsNode {
```
smarterclayton: I don't think this is appropriate..  Instead, have ToNamespace trigger defaulting on specific skips - --keep-volumes, --keep-service-account.  Those would be defaults, if the user then specifies the flag that should override the default.

You should not have to skip volumes that are not namespace scoped, so that's a further nuance.

The warning should be when we clear things without you explicitly giving us the flag (and warnings should always have a prefix `warning: `) - i.e.:

```
warning: Volumes that referenced the source namespace have been disabled, use --keep-volumes to override this
```

and 

```
warning: Service account name cleared, pass --keep-service-account to preserve this value
```

pkg/cli/debug/debug.go
```diff
@@ -658,6 +677,12 @@ func (o *DebugOptions) transformPodForDebug(annotations map[string]string) (*cor
 		container.SecurityContext.RunAsNonRoot = nil
 	}
 
+	// When run in another namespace than original pod, any initContainers that run a command
+	// that depends on VolumeMounts will fail, so unless the initContainer is explicitly being
+	// debugged, remove the initContainers.
+	if len(o.ToNamespace) > 0 && !isInitContainer(pod, o.Attach.ContainerName) {
```
smarterclayton: This is too broad a check.  Only init containers that depend on volumes that were stripped are impacted.  Many init containers use empty dir, none of those are affected.

So the logic should be when stripping volumes is to identify the containers that referenced those mounts, and only remove those.

smarterclayton:
Honestly, I'm not even sure this is a valid bug.  It's really more working as designed.  Debug tries to help you, it's not doing magic.  I would probably be inclined to say this is working as designed, not everything can be debugged.

sallyom:
> Honestly, I'm not even sure this is a valid bug. It's really more working as designed. Debug tries to help you, it's not doing magic. I would probably be inclined to say this is working as designed, not everything can be debugged.

we can at least move it to 4.5  @soltysh  : )

sallyom:
no longer needed, will add msg to help menu that debug might not work for all pods with --to-namespace.

PR #339, opened 2020-03-04
Bug 1810256: Don't default to always mirror manifestlists

kevinrizza:
A previous commit introduced new behavior to always mirror manifest list
digests when running `oc image mirror`. This commit reverts that
behavior, and instead adds a flag `--force-manifest-list` to allow that
option. Additionally, the force behavior occurs when `--filter-by-os` is
set to a wildcard.


Bowenislandsong:
/lgtm
It is exactly what was discussed. Great!

mfojtik:
/lgtm
/hold
@smarterclayton or @soltysh please double check

pkg/cli/image/mirror/mirror.go
```diff
@@ -145,14 +145,12 @@ func NewCmdMirrorImage(name string, streams genericclioptions.IOStreams) *cobra.
...
 
-	// Always mirror manifestlist if available
-	o.ForceManifestList = true
-
 	flag.BoolVar(&o.DryRun, "dry-run", o.DryRun, "Print the actions that would be taken and exit without writing to the destinations.")
 	flag.BoolVar(&o.SkipMissing, "skip-missing", o.SkipMissing, "If an input image is not found, skip them.")
 	flag.BoolVar(&o.SkipMount, "skip-mount", o.SkipMount, "Always push layers instead of cross-mounting them")
 	flag.BoolVar(&o.SkipMultipleScopes, "skip-multiple-scopes", o.SkipMultipleScopes, "Some registries do not support multiple scopes passed to the registry login.")
 	flag.BoolVar(&o.Force, "force", o.Force, "Attempt to write all layers and manifests even if they exist in the remote repository.")
+	flag.BoolVar(&o.ForceManifestList, "force-manifest-list", o.ForceManifestList, "Force mirroring the manifest list if it exists.")
```
smarterclayton: Would probably prefer `include-manifest-list` and the description be `If an image is part of a manifest list, always mirror the list even if only one image is found. Defaults to false unless --filter-by-os is '.*'.`.

smarterclayton: The reason why I don't want to use force is because in our CLI lexicon force means "override the safe behavior with an unsafe behavior that may still be what you want", where as this is more "avoid an optimization".

Actually, we should probably use `--keep-manifest-list` instead of `--include-manifest-list` (keep and preserve are our "softer" verbs)

kevinrizza: @smarterclayton Makes sense. I just updated the flag to `--keep-manifest-list` and updated the description. ptal

smarterclayton:
/hold cancel
/lgtm

thanks

PR #346, opened 2020-03-10
oc debug: add help for --to-namespace

sallyom:
/cc @soltysh 
/cc @smarterclayton

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -99,7 +99,13 @@ var (
 	  %[1]s daemonset/test -c second -- /bin/env
 
 	  # See the pod that would be created to debug
-	  %[1]s mypod-9xbc -o yaml`)
+	  %[1]s mypod-9xbc -o yaml
+
+	  # Debug a resource but launch the debug pod in another namespace.
+	  # Note: Not all resources can be debugged using --to-namespace without modification. For example,
+	  # volumes and serviceaccounts are namespace-dependent. Add '-o yaml' to output the debug pod definition
+	  # to disk.  If necessary, edit the definition then run 'oc debug -f -' or run without --to-namespace.
+	  %[1]s mypod-9xbc --to-namespace testns`)
 )
 
 type DebugOptions struct {


pkg/cli/debug/debug.go
```diff
@@ -99,7 +99,13 @@ var (
...
 	  # See the pod that would be created to debug
-	  %[1]s mypod-9xbc -o yaml`)
+	  %[1]s mypod-9xbc -o yaml
+
+	  # Debug a resource but launch the debug pod in another namespace.
+	  # Note: Not all resources can be debugged using --to-namespace without modification. For example,
+	  # volumes and serviceaccounts are namespace-dependent. Add '-o yaml' to output the debug pod definition
+	  # to disk.  If necessary, edit the definition then run 'oc debug -f -' or run without --to-namespace.
+	  %[1]s node/node-m-0.openshift.example.internal --to-namespace testns`)
 )
```
sallyom: might be a bit wordy.  from what I see, --to-namespace is the only default flag that causes issues (that a user might stumble across). `--node-name` causes issues if you try to run a `node-role.kubernetes.io/master` pod on a worker node :woman_shrugging:

pkg/cli/debug/debug.go
```diff
@@ -99,7 +99,13 @@ var (
...
 
 	  # See the pod that would be created to debug
-	  %[1]s mypod-9xbc -o yaml`)
+	  %[1]s mypod-9xbc -o yaml
+
+	  # Debug a resource but launch the debug pod in another namespace.
+	  # Note: Not all resources can be debugged using --to-namespace without modification. For example,
+	  # volumes and serviceaccounts are namespace-dependent. Add '-o yaml' to output the debug pod definition
+	  # to disk.  If necessary, edit the definition then run 'oc debug -f -' or run without --to-namespace.
+	  %[1]s node/node-m-0.openshift.example.internal --to-namespace testns`)
```
soltysh: User simpler example `%[1]s mypod-9xbc --to-namespace testns`, debugging node is a heavy gun :wink:

sallyom: ok, i updated to use pod instead.

PR #347, opened 2020-03-13
Fix a number of bugs in debug

smarterclayton:
While trying to debug a specific pod, I ran into a number of issues around the command that complicated debugging.

* Add debugging output to the event loop
* Support `--preserve-pod` for debugging the result and other user interactions
* Support `--quiet` so very long command strings don't blow up user output
* Ensure `--node-name ''` correctly sets pod nodeName to empty string
* Piping a pod to `oc debug -f -` was failing because `-` is implicitly a list
* A pod that is in pending may not have reported container status and it is not appropriate to the container loop
* Use watchtools.UntilWithSync instead of watchtools.UntilWithoutRetry - this requires LIST permissions but the outcome is superior.


pkg/helpers/conditions/conditions.go
```diff
@@ -39,6 +39,10 @@ func PodContainerRunning(containerName string, coreClient corev1client.CoreV1Int
 						}
 					}
 				}
+				// until the pod is marked as running, we we have to loop
+				if t.Status.Phase == corev1.PodPending {
+					return false, nil
+				}
```
sallyom: should be this (i should have separated Running from Pending in previous PR)? 
```shell
switch t.Status.Phase {
			case corev1.PodRunning:
			case corev1.PodPending:
				for _, s := range t.Status.ContainerStatuses {
					if s.State.Waiting != nil {
						// Return error here if pod is pending and container status indicates a failure
						// otherwise, user would have to wait the timeout period (15 min)
						// for pod to exit.
						if s.State.Waiting.Reason == "CreateContainerError" || s.State.Waiting.Reason == "ImagePullBackOff" {
							return false, fmt.Errorf(s.State.Waiting.Message)
						}
					}
				}
				return false, nil
```

smarterclayton: I'm not sure whether that loop is safe.  If *any* container is waiting doesn't mean your current container can't start.  It's only your container or a previous init container.  And you can get CreateContainerError and ImagePullBackoff very normally.

This might need to be a print instead `Pod is waiting for image pull for 15s` (which is why this loop may not be in the right place)

soltysh: I'm inclined to say this is a general-purpose helper that should only wait for Running, but we need a smarter one inside debug to prevent the problems that the fixes @sallyom added. Iow. when you have `CreateContainerError` or `ImagePullBackoff` or any other error that I'd expect debug to return early.

pkg/cli/debug/debug.go
```diff
@@ -460,9 +460,35 @@ func (o *DebugOptions) RunDebug() error {
...
+				return true, errors.NewNotFound(corev1.Resource("pods"), pod.Name)
+			}
+
+			return false, nil
+		}
+
 		ctx, cancel := context.WithTimeout(context.Background(), o.Timeout)
 		defer cancel()
-		containerRunningEvent, err := watchtools.UntilWithoutRetry(ctx, w, conditions.PodContainerRunning(o.Attach.ContainerName, o.CoreClient))
+		containerRunningEvent, err := watchtools.UntilWithSync(ctx, lw, &corev1.Pod{}, preconditionFunc, conditions.PodContainerRunning(o.Attach.ContainerName, o.CoreClient))
```
soltysh: :+1:

pkg/cli/debug/debug.go
```diff
@@ -336,13 +336,13 @@ func (o *DebugOptions) RunDebug() error {
...
-	one := false
-	infos, err := b.Do().IntoSingleItemImplied(&one).Infos()
+	infos, err := b.Do().Infos()
 	if err != nil {
 		return err
 	}
-	if !one {
-		return fmt.Errorf("you must identify a resource with a pod template to debug")
+	if len(infos) != 1 {
+		klog.V(4).Infof("Objects: %#v", infos)
```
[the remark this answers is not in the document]
smarterclayton: In general singletons that work on a single object we don't name the total (because it could be hundreds).  I had the debugging line because I didn't trust infos when I was piping input (that's what triggered the bug this fixes, because stdin is not SingleItemImplied, so you couldn't pipe a single pod to this).

pkg/cli/debug/debug.go
```diff
@@ -370,10 +371,12 @@ func (o *DebugOptions) RunDebug() error {
 	o.Attach.Pod = pod
 
 	if len(o.Attach.ContainerName) == 0 && len(pod.Spec.Containers) > 0 {
-		if len(pod.Spec.Containers) > 1 && len(o.FullCmdName) > 0 {
-			fmt.Fprintf(o.ErrOut, "Defaulting container name to %s.\n", pod.Spec.Containers[0].Name)
-			fmt.Fprintf(o.ErrOut, "Use '%s describe pod/%s -n %s' to see all of the containers in this pod.\n", o.FullCmdName, pod.Name, pod.Namespace)
-			fmt.Fprintf(o.ErrOut, "\n")
+		if !o.Attach.Quiet {
```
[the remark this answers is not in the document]
smarterclayton: If you have stderr in the thing you are running those then get interspersed with your debug output.  I.e. the command can be run `oc debug ... -- tar xvzf ... > /foo`.  I don't want the helpful info message text showing up unless the actual execution of the command fails.  The behavior is like RSH and SSH which have silent modes for remote command execution.

pkg/helpers/conditions/conditions.go
smarterclayton: Note that ImagePullBackoff can happen in normal operation and eventually succeed (such as if you create a second and debug pod at the same time).

soltysh: Yeah, but the idea here was to provide fast feedback to `oc debug` author, instead of him going to figure out what has happened only after waiting for indefinite amount of time.

PR #354, opened 2020-03-22
Bug 1825809: create: Add the bare minimum `oc create build`

smarterclayton:
This is a useful one liner for one off builds that starts to move
people away from `oc new-app` for all use cases. Supports only a
bare minimum of flags sufficient to launch source, docker, and
common docker build actions. We would gradually add flags as
necessary.

```
oc create build test --dockerfile-contents $'FROM centos:7\nRUN true' --to-image-stream my file:latest
```


smarterclayton:
/assign @soltysh 

or whomever.  This was useful enough for me to quickly iterate on docker files and s2i builds with a few bash commands.  It replaced docker build (which build config doesn't do).

pkg/cli/create/build.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	if len(o.Strategy) > 0 {
+		switch build.Spec.Strategy.Type {
+		case buildv1.DockerBuildStrategyType:
+			build.Spec.Strategy.DockerStrategy = &buildv1.DockerBuildStrategy{}
+		case buildv1.SourceBuildStrategyType:
+			build.Spec.Strategy.SourceStrategy = &buildv1.SourceBuildStrategy{}
+		case buildv1.CustomBuildStrategyType:
+			build.Spec.Strategy.CustomStrategy = &buildv1.CustomBuildStrategy{}
+		}
```
soltysh: Add `default` which either fails, or screams it defaulted strategy, w/o it we're missing clean information for user what's wrong.

pkg/cli/create/build.go
```diff
@@ -0,0 +1,223 @@
...
+	buildv1client "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
+	"github.com/openshift/library-go/pkg/image/reference"
+	"github.com/openshift/oc/pkg/helpers/env"
+)
+
+const BuildRecommendedName = "build"
+
+var (
+	buildLong = templates.LongDesc(`
+		Create a new build
```
soltysh: Why build and not buildconfig? You can than create build from BC, but we don't have that option in the other direction?

pkg/cli/create/build.go
smarterclayton: Do we have any examples of this in other create commands?  Places where we've added options like volume source where if the client doesn't recognize it it either accepts or errors?

pkg/cli/create/build.go
smarterclayton: Builds are like Jobs - useful on their own.  I might do a build config later but wanted to use build first to get some experience with the args (build config is a superset of build).

soltysh: That's fair, there's `oc create job test-job --from=cronjob/a-cronjob`

pkg/cli/create/build.go
soltysh: We often default a lot in create commands, yours has more than any other create command, but that's reasonable in this particular case. Yet a proper validation here makes sense. Not sure how our apiserver will react to missing strategy type, I can't remember but I bet it will complain. That's why it's better to be more verbose sooner than later.

PR #355, opened 2020-03-23
Bug 1836833: Switch new-app to generate deployment, by default

soltysh:
/assign @adambkaplan @smarterclayton


bparees:
/hold

Pretty sure we've discussed this (or similar things) before. Doesn't this break all kinds of backwards compatibility? 

Cc @smarterclayton

soltysh:
> Pretty sure we've discussed this (or similar things) before. Doesn't this break all kinds of backwards compatibility?

For one - we need this move eventually, the sooner the better. How bad this breaks backwards compatibility - I'm not convinced. I doubt `new-app` needs any backwards compatibility, at most it will break documentation but that's a minor problem we can fix and deal with. The outcome here is higher than the docs mishaps. Eventually we can introduce this flag and set it to `true` by default with appropriate release notes and warning during this release and change it in one of the next releases. But I'm pretty sure we do want to move users towards upstream deployments by default.

bparees:
> Eventually we can introduce this flag and set it to true by default with appropriate release notes and warning during this release and change it in one of the next releases. But I'm pretty sure we do want to move users towards upstream deployments by default.

i'd certainly be more inclined toward that approach(having a flag but continuing to default to use deploymentconfigs).  Another alternative is to introduce a new command (or push people towards odo?) and deprecate new-app entirely.

Seems like this is something we should be having a broader discussion about with the right audience.

smarterclayton:
If oc create dc covers most of the use cases, I think we’d be ok with new app creating deployments.  The only issue is hooks and that’s covered by create dc and set hook

pkg/cli/newapp/newapp.go
```diff
@@ -270,6 +270,7 @@ func NewCmdNewApplication(name, baseName string, f kcmdutil.Factory, streams gen
 	o.PrintFlags.AddFlags(cmd)
 
 	cmd.Flags().BoolVar(&o.Config.AsTestDeployment, "as-test", o.Config.AsTestDeployment, "If true create this application as a test deployment, which validates that the deployment succeeds and then scales down.")
+	cmd.Flags().BoolVar(&o.Config.DeploymentConfig, "as-deployment-config", o.Config.DeploymentConfig, "If true create this application as a deployment config, which allows for hooks and custom strategies.")
```
adambkaplan: Echoing and expanding some of the discussion with @soltysh @smarterclayton and @bparees on slack:

Instead of having a boolean flag here, we should use a string. This gives us flexibility in terms of how we roll out the new behavior.

Ex - change this to `--deploy-with`, with allowed values `deployment` and `deploymentConfig`. We can default to `deploymentConfig` when we introduce the flag, then later switch the default to `deployment`.

pkg/helpers/newapp/app/pipeline.go
```diff
@@ -396,6 +423,26 @@ func AddServices(objects Objects, firstPortOnly bool) Objects {
 			if svc != nil {
 				svcs = append(svcs, svc)
 			}
+		case *kappsv1.Deployment:
+			svc := addService(t.Spec.Template.Spec.Containers, t.ObjectMeta, t.Spec.Template.Labels, firstPortOnly)
+			if svc != nil {
+				svcs = append(svcs, svc)
+			}
+		case *extensionsv1beta1.Deployment:
```
adambkaplan: Didn't we remove this in 4.4? Or are we keeping logic here for backwards compatibility?

pkg/cli/newapp/newapp.go
soltysh: Why? From what I remember our discussions we've talked about flag. Since this is meant to have only true/false values I don't see any reason why we'd need a string flag, which is also harder to understand for users.

pkg/helpers/newapp/app/pipeline.go
soltysh: Backwards compatibility, that is correct.

soltysh:
I've added necessary information in 4.5 release notes, if this gets green I'm planning to merge this right away.

pkg/cli/newapp/newapp.go
adambkaplan: My impression was that we were going to keep `DeploymentConfig` as the default initially, but switch to `Deployment` in a future release (4.6?). If we just want to make a clean break for 4.5 with the appropriate release note, I'm OK.

soltysh: Yes, the release note is here: https://github.com/openshift/openshift-docs/issues/19847#issuecomment-606540724

shvasude:
> /assign @adambkaplan @smarterclayton

Could you please estimate when this will be done.... As I am working on service binding where i need to use deployment instead of deploymentconfigs.

adambkaplan:
@soltysh looks like we have a chicken/egg problem with the build test suite - we have a new-app test which verifies a DeploymentConfig was created.

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_oc/355/pull-ci-openshift-oc-master-e2e-aws-builds/156

soltysh:
@adambkaplan this should be now ready to go, ptal

soltysh:
> @soltysh looks like we have a chicken/egg problem with the build test suite - we have a new-app test which verifies a DeploymentConfig was created.
> 
> https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_oc/355/pull-ci-openshift-oc-master-e2e-aws-builds/156

This was solved by making the tests wait for one or the other, see https://github.com/openshift/origin/pull/24760

soltysh:
/hold
for debugging flakes

soltysh:
/hold cancel
looks like quay is having issues and aws-e2e is struggling with infra problems

PR #359, opened 2020-03-26
Bug 1848408: WIP: Reenable validation just like kubectl has

soltysh:


sbose78:
Curious, why aren't we taking this in?

soltysh:
> Curious, why aren't we taking this in?

We need green test which ensure that we are ready to land this. This is changing old behaviour so we need extra scrutiny on it.

PR #360, opened 2020-03-27
images/tools: Don't change user

smarterclayton:
It's not truly necessary and complicates downstream images.  I broke origin tests because it tries to clear the root passwd user.


smarterclayton:
Forcing through to get origin back to running.

PR #362, opened 2020-03-27
image/mirror: Upconvert v1 schemas to v2 schemas by default

smarterclayton:
In general we want the world to be schema v2, so mirror should
make a best effort to upconvert as necessary. This logic will
eventually be necessary for OCI registries as well.


soltysh:
/hold
for rebase to land first

PR #368, opened 2020-04-03
debug: Correctly default the tools image now that it is on the cluster

smarterclayton:
Now debug uses openshift/tools:latest which is provided by the payload.


smarterclayton:
Trivial fix and I am eager to begin testing tools...

zhengbin78:
这 bug 存在于 4.4 和 4.3 版本，能否尽快将此合并到 4.3 和 4.4 版本？

PR #369, opened 2020-04-03
debug: The `--dry-run` flag does not work with debug

smarterclayton:
The addition of the new flag masked the existing check. Use the new
logic and also exit early (at the correct point).


sallyom:
i see this is since kube bump but also, debug dry-run was broken anyways, i confirmed this fixes it. 

/lgtm

PR #370, opened 2020-04-04
adm: Add a command that creates a token for a user

smarterclayton:
Creating a token for an admin or user is more difficult than it
has to be for scripting automation, and encourages use of service
account tokens instead. Add an administrative command that creates
a token under an oauth-client that may not be used for the standard
web flows for the current user or another.  Typical use would be:

    oc adm oauth create-token --user NAME

or if scoped tokens are desired, create a view token in the provided
namespace for the user.

    oc adm oauth create-token --user NAME --role=view -n NS

Admins can use this to get a token that is time bound but privileged.

In the future, we may wish to consider having an out of the box
client and admission control filter that allows a user to create
their own tokens with some constraints to simplify this flow.


smarterclayton:
This is almost exactly the flow e2e uses to get a token for users

pkg/cli/admin/oauth/oauth.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return err
+	}
+	o.UserInterface = userClient.UserV1().Users()
+	o.OAuthAccessTokenInterface = oauthClient.OauthV1().OAuthAccessTokens()
+	o.OAuthClientInterface = oauthClient.OauthV1().OAuthClients()
+
+	if len(o.UserUID) == 0 {
+		userName := o.UserName
+		if len(userName) == 0 {
```
stlaz, line 184: You set this to "~" above, you should add validation preventing it from being set to an empty string

pkg/cli/admin/oauth/oauth.go
```diff
@@ -0,0 +1,261 @@
...
+				return err
+			}
+		}
+	}
+
+	tokenBytes := make([]byte, 64)
+	if _, err := io.ReadAtLeast(rand.Reader, tokenBytes, 64); err != nil {
+		return err
+	}
+	accessToken := base64.RawURLEncoding.EncodeToString(tokenBytes)
```
stlaz, line 236: The code that generates tokens for the oauth-server prevents leading dashes, that might be something to consider if the main use of this is for scripts: https://github.com/openshift/oauth-server/blob/14ca5c7dcdfbbc4bb7935de071247ccc4f8499bf/pkg/osinserver/tokengen.go#L36-L36

pkg/cli/admin/oauth/oauth.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+			o.UserScopes = append(o.UserScopes, fmt.Sprintf("role:%s:%s", o.RoleScope, ns))
+		}
+	}
+	if len(o.UserScopes) == 0 {
+		o.UserScopes = append(o.UserScopes, "user:full")
+	}
+
+	// use of this redirectURI prohibits normal login
+	o.RedirectURI = "https://localhost:8443/oauth/token/implicit"
```
stlaz, line 214: What do you mean by "normal login"? You can still do `oc login --token="$(oc adm oauth create-token --user test)"`.

The token with `user:full` scopes will also work the same as any other token retrieved by the normal login flow (although the redirect URI will be different and the name is longer)

PR #379, opened 2020-04-08
Bug 1822253: Remove invalid character from manifest meta.name field

### LINKED ISSUE #378: "oc adm catalog mirror" set an invalid value for field meta.name for imagecontentsourcepolicy manifest
When you create the mirror of the artifacts of the catalog, the content of the field meta.name of the manifest file is based on the path of the source registry, including / which are invalid.

so for the command

```
oc adm catalog mirror registry:5000/bmillemathias/olm/redhat-operators:v202004071506 \
 registry:5000/bmillemathias/olm/redhat-operators-build
```

You end with 
```
metadata:
  name: olm/redhat-operators
...
```

bmillemathias-1a:
```oc adm catalog mirror``` uses the path from source registry to generate the
field "meta.name" including all characters, even ones invalid for field meta.name

This patch replaces / by -

Really starting golang and ocp. I don't know if there is some builtin function to sanitize this field.

fix #378

### DIFF
--- a/pkg/cli/admin/catalog/mirror.go
+++ b/pkg/cli/admin/catalog/mirror.go
@@ -272,6 +272,8 @@ func WriteManifests(name, dir string, mapping map[string]string) error {
 		}
 	}()
 
+	name = strings.ReplaceAll(name, "/", "-")
+
 	icsp := operatorv1alpha1.ImageContentSourcePolicy{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: operatorv1alpha1.GroupVersion.String(),

### CODE REVIEW

File: pkg/cli/admin/catalog/mirror.go [hunk 1]
```diff
@@ -272,6 +272,8 @@ func WriteManifests(name, dir string, mapping map[string]string) error {
 		}
 	}()
 
+	name = name.strings.ReplaceAll("/", "-")
```
soltysh, line 275: this should rather look like `name = strings.ReplaceAll(name, "/", "-")` but I'll also defer whether this is a good solution to the OLM folks

bmillemathias-1a: The replacement is a bit naive indeed, that why I was asking if there is some function to properly sanitize the name.

bmillemathias-1a: Fixed the syntax, sorry for not having tested on my own, I was unable to build oc.
Best.

PR #380, opened 2020-04-08
Bug 1819688: login: choose the CAs based on the remote server cert

stlaz:
The previous check to help decide which CA to use when
performing a login is performed in such a way that fails in most
cases, causing the oauth-server to log `remote error: bad certificate`.

This commit fixes that by retrieving the server cert, closing the
connection and extracting the logic for certificate verification
locally without ruining the handshake with the server.

### DIFF
--- a/pkg/helpers/tokencmd/request_token.go
+++ b/pkg/helpers/tokencmd/request_token.go
@@ -397,39 +397,54 @@ func request(rt http.RoundTripper, requestURL string, requestHeaders http.Header
 	return rt.RoundTrip(req)
 }
 
+// transportWithSystemRoots tries to retrieve the serving certificate from the
+// issuer, validates it against the system roots and if the validation passes,
+// returns transport using just system roots, otherwise it returns a transport
+// that uses the CA from kubeconfig
 func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (http.RoundTripper, error) {
-	// copy the config so we can freely mutate it
-	configWithSystemRoots := restclient.CopyConfig(clientConfig)
-
-	// explicitly unset CA cert information
-	// this will make the transport use the system roots or OS specific verification
-	// this is required to have reasonable behavior on windows (cannot get system roots)
-	// in general there is no good with to say "I want system roots plus this CA bundle"
-	// so we just try system roots first before using the kubeconfig CA bundle
-	configWithSystemRoots.CAFile = ""
-	configWithSystemRoots.CAData = nil
-
-	systemRootsRT, err := restclient.TransportFor(configWithSystemRoots)
+	issuerURL, err := url.Parse(issuer)
 	if err != nil {
 		return nil, err
 	}
 
-	// build a request to probe the OAuth server CA
-	req, err := http.NewRequest(http.MethodHead, issuer, nil)
+	// perform the retrieval with insecure transport, otherwise oauth-server
+	// logs remote tls error which is confusing during troubleshooting
+	config := tls.Config{
+		Certificates:       []tls.Certificate{},
+		InsecureSkipVerify: true,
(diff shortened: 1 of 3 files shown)


pkg/helpers/tokencmd/request_token.go
```diff
@@ -449,3 +464,27 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
...
 }
+
+// verifyCertChain uses the system trust bundle in order to perform validation
+// of a certificate chain that needs to have a leaf server cert
+func verifyServerCertChain(dnsName string, chain []*x509.Certificate) ([][]*x509.Certificate, error) {
+	intermediates := x509.NewCertPool()
+	var serverCert *x509.Certificate
+
+	for _, c := range chain {
+		if !c.IsCA {
```
vareti, line 475: i know it is weird but server cert can have `IsCA` field set to `true` . I think it is safe to follow what `tls` package is doing

https://github.com/golang/go/blob/08dd4ad7e30313d089793f19baff1855bbaa004e/src/crypto/tls/handshake_client.go#L805-L818

pkg/helpers/tokencmd/request_token.go
```diff
@@ -397,39 +397,54 @@ func request(rt http.RoundTripper, requestURL string, requestHeaders http.Header
...
 	}
+	conn.Close()
 
-	// see if get a certificate error when using the system roots
-	// we perform the check using this transport (instead of the kubeconfig based one)
-	// because it is most likely to work with a route (which is what the OAuth server uses in 4.0+)
-	// note that both transports are "safe" to use (in the sense that they have valid TLS configurations)
-	// thus the fallback case is not an "unsafe" operation
-	_, err = systemRootsRT.RoundTrip(req)
+	_, err = verifyServerCertChain(issuer, conn.ConnectionState().PeerCertificates)
```
vareti, line 428: Is this supposed to be `issuerURL.Hostname()`? or `issuer` will work with this? Not sure if there is a difference between both.

pkg/helpers/tokencmd/request_token.go
stlaz: That's for the case of self-signed certs, I'll fix that in the next revision

pkg/helpers/tokencmd/request_token.go
stlaz: it seems that both work but hostname might actually be a better choice :+1:

pkg/helpers/tokencmd/request_token.go
```diff
@@ -397,39 +397,54 @@ func request(rt http.RoundTripper, requestURL string, requestHeaders http.Header
...
+		Certificates:       []tls.Certificate{},
+		InsecureSkipVerify: true,
+		ServerName:         issuerURL.Hostname(),
+	}
+
+	port := issuerURL.Port()
+	if len(port) == 0 {
+		port = "443"
+	}
+	conn, err := tls.Dial("tcp", net.JoinHostPort(issuerURL.Hostname(), port), &config)
```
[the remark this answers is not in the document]
stlaz: Hm, it would make sense, although it adds custom verification to all calls with the client, not just the one for discovery. I'll think about it.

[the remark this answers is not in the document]
stlaz: Actually `VerifyPeerCertificate()` only happens after the normal certificate verification, so it serves for tightening purposes, not loosening.

To make the validation reach it, we would have to set `InsecureSkipVerify = true` and I'd rather not touch those parts in login.

stlaz:
/retest
wat

stlaz:
tested on Windows

stlaz:
/retest
infra

PR #384, opened 2020-04-08
oc adm catalog mirror updates: mac/windows, bundle image support, quay.io support

ecordell:
This simplifies and improves the mirroring support in `oc adm catalog mirror`.

1. operator-registry is bumped to a version compatible with kube 1.18 client-go
2. The mirror command is no longer linux-only
3. The mirror command will now mirror "bundle images" in addition to operator/related images.
4. The mirror command supports quay.io as a target
    - Images are given a tag when mirroring if they don't already have one
    - Specify `DEST` as a quay.io org, and the image will be renamed for quay (1 path component for org, the rest converted to a repo name).

Example:

```
$ oc adm catalog mirror quay.io/operator-framework/upstream-community-operators:latest quay.io/my-org
```
will generate mappings like 
`quay.io/coreos/prometheus-operator@sha256:3daa69a8c6c2f1d35dcf1fe48a7cd8b230e55f5229a1ded438f687debade5bcf` -> `quay.io/my-org/coreos-prometheus-operator:1ebe036a`


pkg/cli/admin/catalog/mirror.go
```diff
@@ -285,47 +311,34 @@ func WriteManifests(name, dir string, mapping map[string]string) error {
...
-		}
 	}
 
 	// Create an unstructured object for removing creationTimestamp
 	unstructuredObj := unstructured.Unstructured{}
+	var err error
 	unstructuredObj.Object, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&icsp)
 	if err != nil {
-		return fmt.Errorf("ToUnstructured error: %v", err)
+		return nil, fmt.Errorf("ToUnstructured error: %v", err)
```
soltysh, line 334: errors always lower case

pkg/cli/admin/catalog/mirror.go
```diff
@@ -285,47 +311,34 @@ func WriteManifests(name, dir string, mapping map[string]string) error {
...
 	}
 	delete(unstructuredObj.Object["metadata"].(map[string]interface{}), "creationTimestamp")
 
 	icspExample, err := yaml.Marshal(unstructuredObj.Object)
 	if err != nil {
-		return fmt.Errorf("Unable to marshal ImageContentSourcePolicy example yaml: %v", err)
+		return nil, fmt.Errorf("Unable to marshal ImageContentSourcePolicy yaml: %v", err)
```
soltysh, line 340: lowercase always

pkg/cli/admin/catalog/mirror.go
ecordell: The ICSP generation is verbatim from `oc adm release mirror`, I can update both

ecordell:
> move mirroring from operator-registry to oc <- will this get us closer to dropping the operator-registry dep?

Yes it will - we're actually fine to remove `oc adm catalog build` at any time, which will remove operator-registry entirely as a dependency. I can start the deprecation process by marking that command as deprecated now?

ecordell:
@soltysh updated and matches your suggested commit splitting

pkg/cli/admin/catalog/mirror.go
soltysh: Yes - please.

soltysh:
> Yes it will - we're actually fine to remove `oc adm catalog build` at any time, which will remove operator-registry entirely as a dependency. I can start the deprecation process by marking that command as deprecated now?

Yes - please.

PR #388, opened 2020-04-16
Bug 1826533: pkg/cli/admin/upgrade: Warn when --allow-explicit-upgrade let you through

wking:
To raise awareness that users are initiating an update which was not recommended by their configured Cincinnati upstream, so they are on the hook for validating its stability.


wking:
[images][1]:

```
Job cannot be started: pods "f35d62ae-8021-11ea-9883-0a58ac10fbe1" already exists
```

/retest

[1]: https://prow.svc.ci.openshift.org/prowjob?prowjob=f35d62ae-8021-11ea-9883-0a58ac10fbe1

PR #387, opened 2020-04-16
Bug 1826533: pkg/cli/admin/upgrade: Warn when --force is used

wking:
Also soften "operator blocking upgrades" to the more generic "update precondition failures", because the cluster-version operator may learn about other preconditions with time.  The warning is intended to raise awareness about `--force` being an "I've exhausted all other options and performed a bunch of undocumented manual verification" safety valve, and not something that anyone should be doing frequently.

PR #389, opened 2020-04-16
Bug 1826533: pkg/cli/admin/upgrade: Warn when --allow-upgrade-with-warnings let you through

wking:
To raise awareness that users are initiating or re-targeting an update despite their cluster being unhealthy, which may help (updating to a fixed release) or hurt (re-targeting a 4.1 -> 4.2 update to start installing 4.3 components, which will not play nicely with any remaining 4.1 components).  This is bit ugly due to the two `checkForUpgrade` callsites, although I haven't tried to DRY that up.  And I've done some minimal shuffling so `checkForUpgrade` returns error messages that [are lowercased][1] and which can still be used in both warning strings and fatal messages.

[1]: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -174,10 +174,11 @@ func (o *Options) Run() error {
 			return nil
 		}
 
-		if !o.AllowUpgradeWithWarnings {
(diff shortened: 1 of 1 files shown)


wking:
Rebased on master with 528a90f70 -> b26467a05 to avoid context overlap with #387.

PR #391, opened 2020-04-17
pkg/cli/admin/upgrade: Delegate to update-approvers

wking:
These to subcommands interact with the cluster-version operator, which is now maintained by the updates team.  I'm seeding the alias with the content from [the CVO][1].  I've left the `component` property (added to the root `OWNERS` in e83f2ac9ae, #367) off for now, because the tooling around that only consumes entries at the repo level.

[1]: https://github.com/openshift/cluster-version-operator/blob/914467c03b5bc25dbe4c9295c420fdb2598db821/OWNERS

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,12 @@
+# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#OWNERS_ALIASES
+
+aliases:
+  update-approvers:
+    - abhinavdahiya
+    - crawford
+    - jottofar
+    - LalatenduMohanty
+    - sdodson
+    - smarterclayton
+    - vrutkovs
+    - wking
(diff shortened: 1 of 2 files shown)


wking:
Dropped `release` for now, per @sdodson's preference.

wking:
@abhinavdahiya @crawford @jottofar @sdodson @smarterclayton @vrutkovs, all included because you're existing CVO approvers, but let us know if you want to opt out of being and `oc adm upgrade` approver.

OWNERS_ALIASES
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#OWNERS_ALIASES
+
+aliases:
+  update-approvers:
+    - abhinavdahiya
+    - crawford
+    - jottofar
+    - sdodson
+    - smarterclayton
```
LalatenduMohanty, line 9: Can I be included or I need to be in CVO list first.

wking: Done with 2abda3af3 -> 38ca22dac.

PR #394, opened 2020-04-21
pkg/cli/admin/upgrade: Display the upstream and channel

wking:
When users want to know what channel they're in, make life more convienent then [pointing them at a bare `oc get clusterversion`][1].

[1]: https://github.com/openshift/openshift-docs/blame/c710f7cf5fe8327d6d4275b8a894efb08cc80342/modules/update-upgrading-cli.adoc#L34-L38

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -302,6 +302,9 @@ func (o *Options) Run() error {
 		} else {
 			fmt.Fprintln(o.ErrOut, "warning: No current status info, see `oc describe clusterversion` for more details")
 		}
+		if len(cv.Spec.Upstream) > 0 && len(cv.Spec.Channel) > 0 { // if these are unset, we notify the user via RetrievedUpdates handling below
+			fmt.Fprintf(o.Out, "Cluster is watching %s for updates in channel %s\n", cv.Spec.Upstream, cv.Spec.Channel)
+		}
 		fmt.Fprintln(o.Out)
 
 		if len(cv.Status.AvailableUpdates) > 0 {


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -302,6 +302,9 @@ func (o *Options) Run() error {
 		} else {
 			fmt.Fprintln(o.ErrOut, "warning: No current status info, see `oc describe clusterversion` for more details")
 		}
+		if len(cv.Spec.Upstream) > 0 && len(cv.Spec.Channel) > 0 { // if these are unset, we notify the user via RetrievedUpdates handling below
+			fmt.Fprintf(o.Out, "Cluster is watching %s for updates in channel %s\n", cv.Spec.Upstream, cv.Spec.Channel)
+		}
```
LalatenduMohanty, line 307: So as per the PR when users runs the `oc adm upgrade` command we show the user the Cincinnati API link and the channel name, right? I do think this is a good UX as user has asked for upgrades not for the details. I wonder if this is right place for this kind of information. Ideally we should have a command which should information about the cluster to the user and this information should be part of that command output.

wking: > Ideally we should have a command which should information about the cluster to the user and this information should be part of that command output.

Knowing your upstream and channel gives context for the update source.  Otherwise we can... point them at `oc get clusterversion -o=jsonpath...`, which seems... at least hard to remember.  You don't like `oc adm upgrade` listing the source from which it pulled the available updates along with the "current" version?

wking:
Obsoleted by #576.  Hopefully that one doesn't rot out too :p.

/close

PR #395, opened 2020-04-22
Bug 1823143: Try user-given registry/repository for oc adm release extract|info|mirror  before defaulting to referenced image

sallyom:
When working with mirrored release payloads, a release from a mirrored registry,
`mylocalregistry/ocp/release:4.5.0-0.nightly-2020-04-18-093630` mirrored from
`registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-04-18-093630` -
In both, the tags reference `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2eb0a51...`.
In `oc adm release extract`, always try extracting from the user-given release first.  If this fails, then fall back to the release image-references.  In disconnected environments, currently, oc adm release extract fails when the given release is mirrored from a connected registry. 

Different from mirrored, when working with a tagged release that was uploaded to another repository, such as `quay.io/sallyom/release:new`, since that release wasn't mirrored, the underlying images still point to the original tag. In this case, `oc adm release extract --command openshift-install` will try the sallyom/release image but since that image does not exist, proceed to use the underlying tagged installer image - that will succeed as long as the user has access to the original release image. Note many devs build release w/ substitutions for one or a few component images, and this is the way they sometimes push images around. 

**To test this PR, do something like this:**
```
$ podman run -p 5000:5000 -d registry:2
$ oc adm release mirror --insecure=true -a ~/pull-secret --from=registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-04-29-144201 --to localhost:5000/ocp/release
$ oc adm release --insecure=true extract --command openshift-install localhost:5000/ocp/release:4.5.0-0.nightly-2020-04-29-144201 -v=2  
```
you'll find the failure from master, fixed w /this PR

Scenarios i tested:
* build new release w/ substituting some component images from a ci/nightly/whatever, then extract or info from that
* extract from release that was mirrored from ci or nightly and is now in disconnected registry - verify this image is extracted from disconnected registry with -v=2
* pull down a release from ci, tag it as mine, push to my quay, extract from that
* extract or info from a ci image (no openshift-release-dev/ocp-v4.0-art-dev reference) - if you can access registry.svc.ci images this succeeds, if not, you'll get unauthorized
* mirror a release from local file or from localregistry to remote registry

Bugs that are resolved with these changes (or the equivalent):
- https://bugzilla.redhat.com/show_bug.cgi?id=1823143
- https://bugzilla.redhat.com/show_bug.cgi?id=1812814
- https://bugzilla.redhat.com/show_bug.cgi?id=1806779 (& backport to 4.3)

### DIFF
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -792,7 +792,11 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 				newRef.ID = tagRef.ID
 				// If the user-given registry/repo/name:digest exists, replace with that, if not keep the
 				// is.Spec.Tag from release image-reference
-				if verifyImageExists(opts, newRef) {
+				fromContext, err := opts.SecurityOptions.Context()
+				if err != nil {
+					return true, nil
+				}
+				if verifyImageExists(fromContext, opts.FileDir, o.SecurityOptions.Insecure, opts.FilterOptions.Include, newRef) {
 					tag.From.Name = newRef.String()
 				}
 			}
@@ -895,35 +899,6 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 	sort.Strings(release.Warnings)
 
 	return release, nil
-}
-
-func verifyImageExists(opts *extract.Options, ref imagereference.DockerImageReference) bool {
-	from := imagesource.TypedImageReference{Type: imagesource.DestinationRegistry, Ref: ref}
-	ctx := context.Background()
-	fromContext, err := opts.SecurityOptions.Context()
-	if err != nil {
-		return false
-	}
-	fromOptions := &imagesource.Options{
-		FileDir:         opts.FileDir,
-		Insecure:        opts.SecurityOptions.Insecure,
-		RegistryContext: fromContext,
-	}
-
-	repo, err := fromOptions.Repository(ctx, from)
-	if err != nil {
-		klog.V(2).Infof("unable to connect to image repository %s: %v", from.String(), err)
-		return false
-	}
-	_, _, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, opts.FilterOptions.Include)
-	if err != nil {
-		if imagemanifest.IsImageNotFound(err) {
-			return false
-		}
-		klog.V(2).Infof("unable to read image %s: %v", from.String(), err)
-		return false
-	}
-	return true
 }
 
 func readComponentVersions(is *imageapi.ImageStream) (ComponentVersions, []error) {
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -775,6 +775,28 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 				errs = append(errs, err)
 				return true, nil
 			}
+			// try to verify user-passed image first, in case of mirrored images this will
+			// exist, and if so, use this instead of the release image-reference. If this doesn't
+			// exist, proceed with release image-reference from the readReleaseImageReference above
+			newRef := ref.Ref.AsRepository()
+			// Try the registry/repo passed from the user first.  If image does not exist,
+			// proceed with the release info from the release image-references, from readReleaseImageReference above
+			for _, tag := range is.Spec.Tags {
+				// tagRef is the sha of each component in the release image-reference.
+				// If can't get digest ID, skip this tag, this happens when user has built a payload by
+				// replacing component images in the release with a new image
+				tagRef, err := imagereference.Parse(tag.From.Name)
+				if err != nil {
+					continue
+				}
+				newRef.ID = tagRef.ID
+				// If the user-given registry/repo/name:digest exists, replace with that, if not keep the
+				// is.Spec.Tag from release image-reference
+				if verifyImageExists(opts, newRef) {
+					tag.From.Name = newRef.String()
+				}
+			}
+
 			release.References = is
 		case "release-metadata":
 			data, err := ioutil.ReadAll(r)
@@ -873,6 +895,35 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 	sort.Strings(release.Warnings)
 
 	return release, nil
+}
+
+func verifyImageExists(opts *extract.Options, ref imagereference.DockerImageReference) bool {
+	from := imagesource.TypedImageReference{Type: imagesource.DestinationRegistry, Ref: ref}
+	ctx := context.Background()
+	fromContext, err := opts.SecurityOptions.Context()
+	if err != nil {
+		return false
+	}
+	fromOptions := &imagesource.Options{
+		FileDir:         opts.FileDir,
+		Insecure:        opts.SecurityOptions.Insecure,
+		RegistryContext: fromContext,
+	}
+
+	repo, err := fromOptions.Repository(ctx, from)
+	if err != nil {
+		klog.V(2).Infof("unable to connect to image repository %s: %v", from.String(), err)
+		return false
+	}
+	_, _, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, opts.FilterOptions.Include)
+	if err != nil {
+		if imagemanifest.IsImageNotFound(err) {
+			return false
+		}
+		klog.V(2).Infof("unable to read image %s: %v", from.String(), err)
+		return false
+	}
+	return true
 }
 
 func readComponentVersions(is *imageapi.ImageStream) (ComponentVersions, []error) {
--- a/pkg/cli/admin/release/mirror.go
+++ b/pkg/cli/admin/release/mirror.go
@@ -377,6 +377,35 @@ func (o *MirrorOptions) handleSignatures(context context.Context, signaturesByDi
 	return nil
 }
 
+func verifySrcImageExists(o *mirror.MirrorImageOptions, ref imagereference.DockerImageReference) bool {
+	from := imagesource.TypedImageReference{Type: imagesource.DestinationRegistry, Ref: ref}
+	ctx := context.Background()
+	fromContext, err := o.SecurityOptions.Context()
+	if err != nil {
+		return false
+	}
+	fromOptions := &imagesource.Options{
+		FileDir:         o.FileDir,
+		Insecure:        o.SecurityOptions.Insecure,
+		RegistryContext: fromContext,
+	}
+
+	repo, err := fromOptions.Repository(ctx, from)
+	if err != nil {
+		klog.V(2).Infof("unable to connect to image repository %s: %v", from.String(), err)
+		return false
+	}
+	_, _, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
+	if err != nil {
+		if imagemanifest.IsImageNotFound(err) {
+			return false
+		}
+		klog.V(2).Infof("unable to read image %s: %v", from.String(), err)
+		return false
+	}
+	return true
+}
+
 func (o *MirrorOptions) Run() error {
 	var recreateRequired bool
 	var hasPrefix bool
@@ -517,10 +546,12 @@ func (o *MirrorOptions) Run() error {
 	if err := imageVerifier.Verify(ctx, releaseDigest); err != nil {
 		fmt.Fprintf(o.ErrOut, "warning: An image was retrieved that failed verification: %v\n", err)
 	}
+	var srcRef imagesource.TypedImageReference
 	var mappings []mirror.Mapping
 	if len(o.From) > 0 {
+		var err error
 		src := o.From
-		srcRef, err := imagesource.ParseReference(src)
+		srcRef, err = imagesource.ParseReference(src)
 		if err != nil {
 			return fmt.Errorf("invalid --from: %v", err)
 		}
@@ -579,6 +610,7 @@ func (o *MirrorOptions) Run() error {
 
 	repositories := make(map[string]struct{})
 
+	newRef := srcRef.Ref.AsRepository()
 	// build the mapping list for mirroring and rewrite if necessary
 	for i := range is.Spec.Tags {
 		tag := &is.Spec.Tags[i]
@@ -593,9 +625,23 @@ func (o *MirrorOptions) Run() error {
 			return fmt.Errorf("image-references should only contain pointers to images by digest: %s", tag.From.Name)
 		}
 
+		opts := mirror.NewMirrorImageOptions(genericclioptions.IOStreams{Out: o.Out, ErrOut: o.ErrOut})
+		opts.SecurityOptions = o.SecurityOptions
+		opts.ParallelOptions = o.ParallelOptions
+		opts.FileDir = o.ToDir
 		// Allow mirror refs to be sourced locally
 		srcMirrorRef := imagesource.TypedImageReference{Ref: from, Type: imagesource.DestinationRegistry}
 		srcMirrorRef = sourceFn(srcMirrorRef)
+		// Try the registry/repo passed from the user first.  If image does not exist,
+		// proceed with the release info from the release image-references
+		// ID is the sha of each component in the release image-reference.
+		newRef.ID = from.ID
+		// If the user-given registry/repo/name:digest exists, replace with that, if not keep the
+		// is.Spec.Tag from release image-reference
+		if verifySrcImageExists(opts, newRef) {
+			srcMirrorRef = imagesource.TypedImageReference{Ref: newRef, Type: imagesource.DestinationRegistry}
+			srcMirrorRef = sourceFn(srcMirrorRef)
+		}
 
 		// Create a unique map of repos as keys
 		currentRepo := from.AsRepository().String()
--- a/pkg/cli/admin/release/mirror.go
+++ b/pkg/cli/admin/release/mirror.go
@@ -377,35 +377,6 @@ func (o *MirrorOptions) handleSignatures(context context.Context, signaturesByDi
 	return nil
 }
 
-func verifySrcImageExists(o *mirror.MirrorImageOptions, ref imagereference.DockerImageReference) bool {
-	from := imagesource.TypedImageReference{Type: imagesource.DestinationRegistry, Ref: ref}
-	ctx := context.Background()
-	fromContext, err := o.SecurityOptions.Context()
-	if err != nil {
-		return false
-	}
-	fromOptions := &imagesource.Options{
-		FileDir:         o.FileDir,
-		Insecure:        o.SecurityOptions.Insecure,
-		RegistryContext: fromContext,
-	}
-
-	repo, err := fromOptions.Repository(ctx, from)
-	if err != nil {
-		klog.V(2).Infof("unable to connect to image repository %s: %v", from.String(), err)
-		return false
-	}
-	_, _, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
-	if err != nil {
-		if imagemanifest.IsImageNotFound(err) {
-			return false
-		}
-		klog.V(2).Infof("unable to read image %s: %v", from.String(), err)
-		return false
-	}
-	return true
-}
-
 func (o *MirrorOptions) Run() error {
 	var recreateRequired bool
 	var hasPrefix bool
@@ -638,7 +609,11 @@ func (o *MirrorOptions) Run() error {
 		newRef.ID = from.ID
 		// If the user-given registry/repo/name:digest exists, replace with that, if not keep the
 		// is.Spec.Tag from release image-reference
-		if verifySrcImageExists(opts, newRef) {
+		fromContext, err := opts.SecurityOptions.Context()
+		if err != nil {
+			return err
+		}
+		if verifyImageExists(fromContext, opts.FileDir, opts.SecurityOptions.Insecure, opts.FilterOptions.Include, newRef) {
 			srcMirrorRef = imagesource.TypedImageReference{Ref: newRef, Type: imagesource.DestinationRegistry}
 			srcMirrorRef = sourceFn(srcMirrorRef)
 		}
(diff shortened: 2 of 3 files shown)


pkg/cli/admin/release/extract_tools.go
```diff
@@ -320,6 +320,12 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			missing.Insert(target.Mapping.Image)
 			continue
 		}
+		// make sure registry is the one we want, if not, replace it
+		specSplit := strings.Split(spec, "@sha256")
+		desired := strings.Split(o.From, "@sha256")
+		if specSplit[0] != desired[0] {
+			spec = fmt.Sprintf("%s@sha256%s", desired[0], specSplit[1])
+		}
```
sallyom, line 328: This is what needs to happen.. now is there a less hacky way to do it?

sallyom:
/hold

this isn't right, will fix

pkg/cli/admin/release/extract_tools.go
```diff
@@ -320,6 +320,14 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			missing.Insert(target.Mapping.Image)
 			continue
 		}
+		// make sure registry is the one we want, if not, replace it
+		specSplit := strings.Split(spec, "@sha256")
+		desired := strings.Split(o.From, "@sha256")
+		if specSplit[0] != desired[0] &&
+			!strings.Contains(desired[0], "registry.svc.ci.openshift.org") &&
+			!strings.Contains(desired[0], "quay.io/openshift-release-dev") {
```
soltysh, line 328: You can't hard code any of these, you'd need to take the address from user's input and make sure that it does not appear in the payload anywhere.

sallyom: I've updated this, please give it another pass,  thanks

pkg/cli/admin/release/extract_tools.go
```diff
@@ -307,6 +309,12 @@ func (o *ExtractOptions) extractCommand(command string) error {
 	refExact.Ref.ID = release.Digest.String()
 	exactReleaseImage := refExact.String()
 
+	// will extract in parallel
+	opts := extract.NewOptions(genericclioptions.IOStreams{Out: o.Out, ErrOut: o.ErrOut})
+	opts.ParallelOptions = o.ParallelOptions
+	opts.SecurityOptions = o.SecurityOptions
+	opts.OnlyFiles = true
+
```
sallyom, line 317: the opts are moved from below, to use to verify user-given release:tag-tool image exists.

sallyom:
/hold

This PR uses a bump in library-go for a not-yet-merged PR: https://github.com/openshift/library-go/pull/785

pkg/cli/admin/release/extract_tools.go
```diff
@@ -320,20 +328,29 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
-			target.Mapping.Name = fmt.Sprintf(target.ArchiveFormat, releaseName)
-			target.Mapping.To = filepath.Join(dir, target.Mapping.Name)
-		} else {
-			target.Mapping.To = filepath.Join(dir, target.Command)
-			target.Mapping.Name = fmt.Sprintf("%s-%s", target.OS, target.Command)
+		digestTag, err := imagereference.GetSpecDigestTag(spec)
+		if err != nil {
+			return err
+		}
+		specFirst := fmt.Sprintf("%s%s", desired.AsRepository(), digestTag)
```
sallyom, line 341: @dmage this is where the "@" comes in (look at `oc adm release info anyrelease | grep baremetal-installer`  -or grep any component)

sallyom: i'm updated here to add the "@", removed it from library-go

pkg/cli/admin/release/extract_tools.go
```diff
@@ -320,20 +328,30 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
-			target.Mapping.Name = fmt.Sprintf(target.ArchiveFormat, releaseName)
-			target.Mapping.To = filepath.Join(dir, target.Mapping.Name)
-		} else {
-			target.Mapping.To = filepath.Join(dir, target.Command)
-			target.Mapping.Name = fmt.Sprintf("%s-%s", target.OS, target.Command)
+		digest, err := imagereference.GetSpecDigest(spec)
+		if err != nil {
+			return err
+		}
+		specFirst := fmt.Sprintf("%s@%s", desired.AsRepository(), digest)
```
soltysh, line 341: This is wrong and might cause problems in the future, I'd prefer to use `ref.Exact()` instead or any other for printing.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -320,20 +328,30 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
+		klog.V(2).Infof("Will extract %s from %s", target.Mapping.From, specFirst)
+		target, willArchive, err = getValidTarget(specFirst, dir, releaseName, target, willArchive)
```
soltysh, line 343: I'd prefer passing `ref` here instead of that `specFirst`.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -632,6 +644,51 @@ const (
 	releaseVersionMarker = "!_RELEASE_VERSION_LOCATION_\x00XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX\x00"
 )
 
+func getValidTarget(spec, dir, releaseName string, target extractTarget, willArchive bool) (extractTarget, bool, error) {
+	ref, err := imagereference.Parse(spec)
```
soltysh, line 648: This won't be necessary if you pass `DockerImageReference` as mentioned before.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -632,6 +644,51 @@ const (
...
+	if err != nil {
+		return fmt.Errorf("unable to connect to image repository %s: %v", from.String(), err)
+	}
+	_, _, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, opts.FilterOptions.Include)
+	if err != nil {
+		if imagemanifest.IsImageNotFound(err) {
+			return imagemanifest.NewImageNotFound("image does not exist", err)
+		}
+		return fmt.Errorf("unable to read image %s: %v", from, err)
+	}
```
soltysh, line 688: This error checking seems weird, given that you're not doing anything with that error I don't think that's needed. Although you're mentioning a few lines above you want to ignore not found and fallback to release image, but I'm not seeing this here.

go.mod
```diff
@@ -84,6 +84,8 @@ replace (
 	// d1d5f6476656c6aad457e2a91d3436e66b6f2251 is more suitable since it does not break fsouza/go-clientdocker,
 	// yet provides the same fix.
 	github.com/docker/docker => github.com/docker/docker v1.4.2-0.20191121165722-d1d5f6476656
+	// TODO: REMOVE: FAKE BUMP FOR TESTING
+	github.com/openshift/library-go => github.com/sallyom/library-go v0.0.0-20200424135032-af2fa04036d6
```
soltysh, line 88: Fix this!

sallyom: yup this is in a TOBEREMOVED commit - I need this to test the change until that library-go PR merges

pkg/cli/admin/release/extract_tools.go
sallyom: refExact is something like this: `localhost:5000/ocp/release:4.5.0-0.nightly-2020-04-18-093630` but we need just the `localhost:5000/ocp/release`  and for the tag, we need the sha of 
`localhost:5000/ocp/release:4.5.0-0.nightly-2020-04-18-093630-installer` (or whatever tool you're extracting)
that sha is always the same as it's reference image, in this case, 
`localhost:5000/ocp/release@sha256:22204fa241eaad3f04d88bfa79629d...` that is the same for the reference image (that a disconnected user will not have access to):
`quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:22204fa241eaad3f04d8...`

pkg/cli/admin/release/extract_tools.go
sallyom: this is a first pass - we're checking here that the user-given reg/repo/name@sha-from-ref-image exists, ( the splice of the user-given registry/repository/name and the @sha... of the referenced release), if it doesn't exist or there's an error returning the manifest,   then default to using the referenced image - in this case would be quay.io/openshift-release-dev.. and return to the regularly scheduled program of error-checking -  (are you authorized to access the image? does the image contain the expected binary? etc) If it _does_ exist, then return to the regularly scheduled program of error-checking, also, but with the user-given-registry/repo/name@sha rather than the reference one (that won't be accessible in a disconnected env)

soltysh: So this is still WIP?

sallyom: first pass, meaning, first check the user-given registry/repo/name@sha-for-tool-image-from-ref-image and if that image exists, carry on with extracting the binary from it, but if it does not exist, then carry on with the referenced image (that may or may not be accessible in disconnected-but we're not checking that here)

pkg/cli/admin/release/extract_tools.go
sallyom: the only way extract from user-provided image that was mirrored from a `registry.svc.ci.openshift.org/ocp/release:tag` that I've found is to 
1) substitute the registry/repo/name like so in specFirst
2) replace the `is.Tags.From.Name` for the extract target  in LoadReleaseInfo here: https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/info.go#L778

 perhaps this needs to happen during the image mirror instead.

sallyom:
To test: do something like this:
```
$ podman run -p 5000:5000 -d registry:2
$ oc adm release mirror --insecure=true -a ~/pull-secret --from=registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-04-29-144201 --to localhost:5000/ocp/release
$ oc adm release --insecure=true extract --command openshift-install localhost:5000/ocp/release:4.5.0-0.nightly-2020-04-29-144201 -v=2  
```
do that from master and from this PR

pkg/cli/admin/release/extract_tools.go
sallyom: that will give the release sha but not the release-tool

pkg/cli/admin/release/extract_tools.go
sallyom: resolving this, as i've moved all changes to pkg/cli/admin/release/info

pkg/cli/admin/release/extract_tools.go
sallyom: resolving this, since i've moved all changes to pkg/cli/admin/info

pkg/cli/admin/release/extract_tools.go
sallyom: resolving this, i've pushed a change to make this cleaner

pkg/cli/admin/release/extract_tools.go
sallyom: resolving this, as I've updated to make all changes in pkg/cli/admin/release/info

sallyom:
@soltysh I've reworked this PR, moved all changes to pkg/cli/admin/release/info.go rather than release/extract_tools.go - this is more clear.  ptal, thanks. 
If this direction looks better, ptal here, also, so I can remove the fake bump: https://github.com/openshift/library-go/pull/785

pkg/cli/admin/release/info.go
```diff
@@ -775,6 +775,33 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
...
 			}
+			// Try the registry/repo passed from the user first.  If image does not exist,
+			// proceed with the release info from the release image-references, from readReleaseImageReference above
+			for _, tag := range is.Spec.Tags {
+				// this is user-passed image name
+				userGivenImage, err := imagereference.Parse(image)
+				if err != nil {
+					errs = append(errs, err)
+					return true, nil
+				}
```
soltysh, line 786: This could be outside of the loop.

soltysh: Actually looking inside `ParseReference` invoked at the beginning of the method that one already has that invocation see https://github.com/openshift/oc/blob/29e9a330f0ab0eed17b5ba6a8cc198ccc4332567/pkg/cli/image/imagesource/reference.go#L96 Iow. you could just check `ref.Type == DestinationRegistry` and drop this above code entirely.

pkg/cli/admin/release/info.go
```diff
@@ -775,6 +775,33 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
...
+				// If can't get digest, skip this tag, this happens when user has
+				// replaced a component in the release with a new image.
+				digest, err := imagereference.GetSpecDigest(tag.From.Name)
+				if err != nil {
+					continue
+				}
+				// try to verify user-passed image tag first, in case of mirrored images this will
+				// exist, and if so, use this instead of the release image-reference. If this doesn't
+				// exist, proceed with tag from release image-reference from the readReleaseImageReference above
+				givenTag := fmt.Sprintf("%s@%s", userGivenImage.AsRepository(), digest)
```
soltysh, line 797: tag is confusing, since this is more `imageID`, `imageDigest`, right?

pkg/cli/admin/release/info.go
```diff
@@ -875,6 +902,40 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
...
+	repo, err := fromOptions.Repository(ctx, from)
+	if err != nil {
+		klog.V(2).Infof("unable to connect to image repository %s: %v", from.String(), err)
+		return false
+	}
+	_, _, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, opts.FilterOptions.Include)
+	if err != nil {
+		if imagemanifest.IsImageNotFound(err) {
+			return false
+		}
```
soltysh, line 932: This condition is not necessary, since in both cases you just return `false`.

pkg/cli/admin/release/info.go
soltysh: I'm a bit confused, you're creating a new reference from user-passed registry and release's image SHA, right? And then you parse it once again in `verifyImageExists`. If this works you replace the tag contents, correct?

soltysh: I still don't understand why you can't just use the `DockerImageReference` returned from `userGivenImage.AsRepository()` invocation and set digest there. Iow.
```go
newRef := userGivenImage.AsRepository() // this will clear both Tag and ID fields
newRef.ID = digest
if verifyImageExists(opts, newRef)...
```
This way `verifyImageExists` will work with already passed reference, so there's no need for double parsing.

pkg/cli/admin/release/info.go
sallyom: kind of, but if imageNotFound, no logging (this means it's not a mirrored image) but for any other reason should log the error, this would indicate there was an error reading the mirrored image (it exists, but there is a problem).  I'm fine not logging, too, but this was my reasoning.

pkg/cli/admin/release/info.go
sallyom: thanks, i've updated.  much cleaner. 
userGivenImage.AsRepository() == `ref.Ref.AsRepository()` :) so i can drop the userGivenImage nonsense altogether

pkg/cli/admin/release/info.go
sallyom: thanks! updated, now passing imagereference.DockerImageReference to verifyImageExists, and if found, `tag.From.Name = newRef.String()`

go.mod
sallyom: actually i don't need that change so removed the bump : )

sallyom:
/hold cancel

fake bump removed

pkg/cli/admin/release/info.go
```diff
@@ -775,6 +775,28 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 				errs = append(errs, err)
 				return true, nil
 			}
+			// try to verify user-passed image first, in case of mirrored images this will
+			// exist, and if so, use this instead of the release image-reference. If this doesn't
+			// exist, proceed with release image-reference from the readReleaseImageReference above
+			newRef := ref.Ref.AsRepository()
```
smarterclayton, line 781: Assuming that the release image and the target share the same repository is a really ugly hack.  I don't think it's appropriate without the user explicitly telling us where it is (perhaps via a flag, perhaps by providing an ICSP).

wking: So is this "pull down any ICSPs from the cluster and teach the local `oc` image resolvers how to process them"?  That would be great, but seems like a lot of work without vendoring more of `containers/image`.  Or does the native `oc` image-resolution code already support mirrors and we'd just need to feed the ICSPs in?

wking: ... and there is no guarantee that ICSPs that make sense for the cluster also make sense for wherever the user is running `oc`.  Would be nice if we had image-streams for every image referenced from the release image or some such, so we could pull from there without having to reach out and hit some external registry.

sallyom: so ICS is printed when mirroring image like so:
```
imageContentSources:
- mirrors:
  - quay.io/sallyom/release-test
  source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
- mirrors:
  - quay.io/sallyom/release-test
  source: registry.svc.ci.openshift.org/ocp/release
```
so if an image is mirrored pass a flag to provide that info? i don't have details for how to use this info atm, i'll take a look.

soltysh: OK, that's fair and w/o that information it's ok to not to have properly working disconnected mirrors working, exactly like described in the bug. The second question is I can't seem to find which piece of `containers/image` is responsible for that, can you point to it?

sallyom: @smarterclayton I'm confused by your comment, `Assuming that the release image and the target share the same repository`.  I don't see where this assumes the same repository.  You mean the shas aren't necessarily the same, if you mirror from one registry to another?

sallyom: (also, as an aside, the mirror that produced that ICS above isn't possible without this PR: https://github.com/openshift/oc/pull/408 as it requires a pull-secret for the quay.io/openshift-release-dev separate than the push-creds for quay.io/sallyom)

sallyom: I just mirrored to a docker registry and it succeeds w/ this PR, fails from master:
```console
first the mirror (requires https://github.com/openshift/oc/pull/408):
$ oc adm release mirror --pull-secret ~/installer/pull-secret registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-05-05-205255 --to docker.io/sallyom/release-test

then extract: (this PR):
$ oc adm release extract --command openshift-install docker.io/sallyom/release-test:4.5.0-0.nightly-2020-05-05-205255
$ ./openshift-install version
./openshift-install 4.5.0-0.nightly-2020-05-05-205255
built from commit abefc8a4e5cb3b8a947722384ddd3c6dfb7fb72d
release image docker.io/sallyom/release-test@sha256:a8b4208250700a544149124bfdbcdc2382490ced704f21ac8e93afd51333ba80

extract (from master)
$ oc adm release extract --command openshift-install docker.io/sallyom/release-test:4.5.0-0.nightly-2020-05-05-205255
error: unable to connect to image repository quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:a455c2ffb2e34d951fe8e12c56800a00f4c35420643237c509f1e36350502f98: Get "https://quay.io/v2/": EOF
```

pkg/cli/admin/release/mirror.go
```diff
@@ -377,6 +377,35 @@ func (o *MirrorOptions) handleSignatures(context context.Context, signaturesByDi
...
+		if imagemanifest.IsImageNotFound(err) {
+			return false
+		}
+		klog.V(2).Infof("unable to read image %s: %v", from.String(), err)
+		return false
+	}
+	return true
+}
+
 func (o *MirrorOptions) Run() error {
```
sallyom, line 409: TODO: collapse the 2 verifyImage functions into 1, pass the o.FileDir and o.SecurityDir separately

pkg/cli/admin/release/info.go
smarterclayton: It would be more appropriate right now to add a flag that is "--mirror-repository=foo/bar" (possibly that supports multiple flags via StringSlice) than to add magic lookup.  In the future, reeading from an ICSP file or ICSP on the cluster might be reasonable.  Right now, don't do magic.

smarterclayton: Adding error path when image can't be found to suggest that if this is a mirrored repository to provide that flag, and to add a paragraph to help that describes how to work with mirrored repositories, would also be required to make this clear to users.

sallyom: but how will that solve 'I can't `oc adm release extract` in disconnected environment'? 
As/is, the cmd does the right thing and extracts from the mirrored repository (you don't have to pass it), and succeeds if you are connected and have the correct pull secret for that repo but fails otherwise.

smarterclayton: That flag would add additional search locations (like ICSP) if the first fetch fails.  All locations that look at an image would support the alternates.

smarterclayton: You would be designing in (vs adding hacks) a mechanism that says "if i get an image, if i can't get it at location A, try B, C, and D by digest".  That mechanism *probably* should be in the lowest level client (the registry fetcher) which would transparently support this for all commands (i.e. putting code into mirror is probably wrong).

sallyom:
closing in favor of https://github.com/openshift/oc/pull/427 , it's less ugly, but maybe still too ugly

sallyom:
Please see https://github.com/openshift/oc/pull/439 and https://github.com/openshift/enhancements/pull/334 for progress

PR #397, opened 2020-04-23
Bug 1801292: support -f for rsh command

soltysh:
This PR  adds support -f for rsh command.

With the first of the following will work just fine:
```
oc rsh nginx cat /etc/resolv.conf 
oc rsh -f pod.yaml cat /etc/resolv.conf 
```

/assign @smarterclayton @mfojtik


smarterclayton:
```
oc rsh -f pod.yaml  -- cat /etc/resolv.conf 
```

is an api compatibility break because existing users can call that and expect `--` to be passed to the shell argument.  So no, we can't change to support it.

Behavior of `oc rsh` matches `rsh` behavior and rules (as interpreted originally), and we cannot change how arguments are interpreted because that would be break calls

/hold

Other changes are fine, just not supporting `--`.

soltysh:
> Other changes are fine, just not supporting `--`.

Support for `--` was removed, I've even added explicit fail with pointers to rsh API.
/hold cancel

soltysh:
/hold
for debugging why this is failing CI

soltysh:
/hold cancel
I fixed the problem

PR #399, opened 2020-04-24
Bug 1920024: Allow debug to start with a single `--image` argument

smarterclayton:
The command `oc debug` and `oc debug --image=foo` should be allowed to start foo exactly as if you ran `oc debug istag/foo:latest` or ran `oc debug node/bar`. This allows us to have a minimal command for scenarios where you want to test the
image in the debug setup.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -21,6 +21,7 @@ import (
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	kapierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


smarterclayton:
agnostic is broken, this is ready for review

pkg/cli/debug/debug.go
```diff
@@ -81,13 +82,17 @@ var (
...
+		resources that can be used to create pods (such as image stream tags), or simply pass
+		'--image=IMAGE' to start a simple shell session in an image with a shell program
 
 		The debug pod is deleted when the remote command completes or the user interrupts
 		the shell.
 	`)
 
 	debugExample = templates.Examples(`
+		# Start a shell session into a pod using the OpenShift tools image
+		oc debug
```
soltysh, line 94: Why not just use `oc run --image=X`, there's nothing debug specific in doing this like so.

smarterclayton: Debug is better than run because debug is solving a different problem - debug is "how do i get a shell in an environment efficiently".  Run is about "how do i actually launch a pod".  The two are not equal.  And run frankly is a ux nightmare, while debug is about making opinionated choices to get that experience cleanly.

soltysh (approved):
/lgtm

pkg/cli/debug/debug.go
soltysh: Fair, I'm not a fan of run either :sweat:

PR #404, opened 2020-05-01
Bug 1823143: Try user-given registry/repository for extract before defaulting to referenced image (take 2)

sallyom:
See https://github.com/openshift/oc/pull/395


sallyom:
folding this in with https://github.com/openshift/oc/pull/395 closing this

PR #405, opened 2020-05-04
Bug 1802719: oc adm groups sync  IsQueryOutOfBoundsError warning,not error

sallyom:
/cc @soltysh

### DIFF
--- a/pkg/helpers/groupsync/groupsyncer.go
+++ b/pkg/helpers/groupsync/groupsyncer.go
@@ -15,6 +15,7 @@ import (
 
 	userv1 "github.com/openshift/api/user/v1"
 	userv1client "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
+	"github.com/openshift/library-go/pkg/security/ldapquery"
 	"github.com/openshift/oc/pkg/helpers/groupsync/interfaces"
 )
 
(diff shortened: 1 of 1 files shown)


sallyom:
spent some time adding a unit and/or extended test for this, will sync w/ @stlaz

pkg/helpers/groupsync/groupsyncer.go
```diff
@@ -85,6 +86,10 @@ func (s *LDAPGroupSyncer) Sync() ([]*userv1.Group, []error) {
 		// update the OpenShift Group corresponding to this record
 		openshiftGroup, err := s.makeOpenShiftGroup(ldapGroupUID, usernames)
 		if err != nil {
+			if ldapquery.IsQueryOutOfBoundsError(err) {
+				klog.Warningf("%s", err.Error())
```
soltysh, line 90: We rarely use `klog` for these warnings, `fmt.Fprintf(os.Stderr` is the preferred.

stlaz:
/hold
why sync groups that are outside the intended tree?

pkg/helpers/groupsync/groupsyncer.go
sallyom: updated

stlaz:
/hold cancel
nvm, you're not actually syncing it, my bad

PR #406, opened 2020-05-05
Bug 1802580: return ProjectRequestMessage if it is set with login,project cmds

sallyom:
This PR updates return msg to users that cannot create projects for `oc login`, `oc new-project`, and `oc project`
If ProjectRequestMessage is set, and user without self-provisioner clusterrole logs in or tries to create  projects, return the ProjectRequestMessage. This is consistent with console.  

If no ProjectRequestMessage set, then return the same message for login|project|new-project: 
```console
$ oc project
You may not request a new project via this API.
```
`oc project` without this PR is misleading:
```console
$ oc project
No project has been set. Pass a project name to make that the default.
```

see comment below to set projectRequestMsg and remove self-provisioner clusterrole

### DIFF
--- a/pkg/cli/login/loginoptions.go
+++ b/pkg/cli/login/loginoptions.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"net"
 	"os"
@@ -266,6 +267,29 @@ func (o *LoginOptions) gatherProjectInfo() error {
 		return err
 	}
 
+	canRequest, err := loginutil.CanRequestProjects(o.Config, o.DefaultNamespace)
+	if err != nil {
+		return err
+	}
+	// check if ProjectRequestMessage is set
+	if !canRequest {
+		msg := ""
+		res, err := projectClient.RESTClient().Get().Resource("projectrequests").DoRaw(context.TODO())
+		if err != nil && res != nil {
+			// if err != nil, try to extract ProjectRequestMessage from status if set
+			status := metav1.Status{}
+			err := json.Unmarshal(res, &status)
+			if err != nil {
+				return err
+			}
+			if len(status.Details.Causes) != 0 && status.Details.Causes[0].Message != "" {
+				msg = status.Details.Causes[0].Message
+			}
+			// now return redirected error from !canRequest with the extracted ProjectRequestMessage if set
+			fmt.Fprintf(o.Out, errors.NoProjectsExistMessage(canRequest, msg, o.CommandName))
+			return nil
+		}
+	}
 	projectsList, err := projectClient.Projects().List(context.TODO(), metav1.ListOptions{})
 	// if we're running on kube (or likely kube), just set it to "default"
 	if kerrors.IsNotFound(err) || kerrors.IsForbidden(err) {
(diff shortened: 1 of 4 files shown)


sallyom:
to test
```console
$ oc edit project.config.openshift.io/cluster 
```
add this:
```yaml
spec:
  projectRequestMessage: Contact your system administrator admin@company.com
```
```console
$ oc login -u user1 -p pass1
Login successful.

Contact your system administrator admin@company.com
$ oc project
You do not have rights to view projects: Contact your system administrator admin@company.com
$ oc new-project testproj
Error from server (Forbidden): Contact your system administrator admin@company.com
```

pkg/cli/login/loginoptions.go
```diff
@@ -266,6 +267,26 @@ func (o *LoginOptions) gatherProjectInfo() error {
...
+		res, err := projectClient.RESTClient().Get().Resource("projectrequests").DoRaw(context.TODO())
+		if err != nil {
+			status := metav1.Status{}
+			err := json.Unmarshal(res, &status)
+			if err != nil {
+				return err
+			}
+			if len(status.Details.Causes) != 0 && status.Details.Causes[0].Message != "" {
+				msg = status.Details.Causes[0].Message
+			}
```
sallyom, line 285: This is where the `project.config.openshift.io/cluster Spec.ProjectRequestMessage` is stored.  If it exists, return it to the NoProjectsExistMessage, if not, 
the default `You don't have any projects. Contact your system administrator to request a project.` will display

pkg/cli/login/loginoptions.go
```diff
@@ -266,6 +267,27 @@ func (o *LoginOptions) gatherProjectInfo() error {
...
 	}
 
+	canRequest, err := loginutil.CanRequestProjects(o.Config, o.DefaultNamespace)
+	if err != nil {
+		return err
+	}
+	// check if ProjectRequestMessage is set
+	if !canRequest {
+		msg := ""
+		res, err := projectClient.RESTClient().Get().Resource("projectrequests").DoRaw(context.TODO())
```
soltysh, line 277:
```
status := metav1.Status{}
projectClient.RESTClient().Get().Resource("projectrequests").Do(context.TODO()).Into(&status)
```
will drop the entire encoding code from here.

pkg/cli/project/project.go
```diff
@@ -158,14 +158,17 @@ func (o ProjectOptions) Run() error {
 		currentProject = currentContext.Namespace
 	}
 
+	client, kubeclient, err := o.ClientFn()
+	if err != nil {
+		return err
+	}
+	if err := client.RESTClient().Get().Resource("projectrequests").Do(context.TODO()).Into(&metav1.Status{}); err != nil {
+		fmt.Fprintf(o.Out, "%v\n", err)
+		return nil
```
soltysh, line 167: If this is error, just `return err` is the best option.

pkg/cli/login/loginoptions.go
sallyom: The issue with that is that here:
https://github.com/openshift/oc/blob/master/vendor/k8s.io/client-go/rest/request.go#L920-#L936
The resp.body is transformed into the error, so you lose the status with the projectRequestMessage.  Then, if you instead try to use DoRaw you'll get the response body, but as []byte which cannot be used in `(r Result) Into(obj runtime.Object)` :( I tried, but the `type Result` has only unexported fields.

pkg/cli/project/project.go
sallyom: updated, fixed

sallyom: but this  changes behavior of `oc project` to return an error.  
As/is, if user w/out self-provisioner role runs `oc project` they'll see this:
```console
$ oc project
No project has been set. Pass a project name to make that the default.
```
and now, with this PR, it will be this:
```console
$ oc project
Error from server (Forbidden): This is a friendly message...
```
or, if no projectRequestMessageSet, this:
```console
$ oc project
Error from server (Forbidden): You may not request a new project via this API.
```
(this message without the set projRequestMessage is now consistent with the error a user gets with `oc new-project foo` now, which makes more sense to me)

PR #408, opened 2020-05-07
Bug 1799020: Add a pull-secret flag to 'oc adm release new' 

sallyom:
Bug 1799020: Add a pull-secret flag for `oc adm release` commands that both push and pull images: `oc adm release new`  and `oc adm release mirror`  
    
It is not possible to configure 2 auths for a single image registry in a single authfile.
This is consistent with behavior of docker, podman, and oc.
Allow 'oc adm release new|mirror' to succeed when a user pulls from one (Red Hat owned) quay.io repository that requires one auth and then push to another (personal) quay.io repository that requires a different auth.  The assumption here is that a user keeps his personal quay credentials in the default location (~/.docker/config.json for now but we need to update that) and a separate pull-secret for pulling the Red Hat release payload images. 

This PR enables the following:
```console
$ oc adm release new --pull-secret ~/installer/pull-secret --from-release registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-05-06-003431 cli=quay.io/sallyom/testcli:test --to-image quay.io/sallyom/release-test:testcli
``` 
And for users who don't use Docker, the default podman authfile is in `/run/user/1000/containers/auth.json` so:
 ```console
$ oc adm release new --pull-secret ~/installer/pull-secret --registry-config /run/user/1000/containers/auth.json --from-release registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-05-06-003431 cli=quay.io/sallyom/testcli:test --to-image quay.io/sallyom/release-test:testcli
```

### CODE REVIEW

File: pkg/cli/image/manifest/manifest.go [hunk 1]
```diff
@@ -40,14 +40,16 @@ func (o *ParallelOptions) Bind(flags *pflag.FlagSet) {
...
 	Insecure         bool
 	SkipVerification bool
 
-	CachedContext *registryclient.Context
+	CachedContext []*registryclient.Context
 }
 
 func (o *SecurityOptions) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&o.RegistryConfig, "registry-config", "a", o.RegistryConfig, "Path to your registry credentials (defaults to ~/.docker/config.json)")
+	flags.StringVar(&o.PushConfig, "push-config", o.PushConfig, "Path to your registry credentials for pushing images to (defaults to ~/.docker/config.json)")
```
wking, line 52: nit: Seems like `release new` is more focused on pushing than pulling, so I'd have expected pushing to use `--registry-config` with a new `--pull-config` to override in the pull case (instead of having a `--push-config` to override in the push case).  But obviously is could work either way.

Also, I'd expect (and for backwards compat, we probably need to require) the alternative config to fall back to `--registry-config`, not to fall back straight to `~/.docker/config.json`.  So:

* User sets nothing, we use `~/.docker/config.json` for both.
* User sets `--registry-config`, we use that for both.
* User sets `--push-config`, we use that for push and `~/.docker/config.json` for pull (or vice versa if we go with `--pull-config).
* User sets both, we use the explicitly-provided paths for their respective activities.

wking:
Might want to add the same option to the vanilla `oc image mirror` as well, since you could have the same issue there.

pkg/cli/image/manifest/manifest.go
```diff
@@ -109,9 +113,18 @@ func (o *SecurityOptions) NewContext() (*registryclient.Context, error) {
 			return nil, fmt.Errorf("unable to load --registry-config: %v", err)
 		}
 	}
-	context := registryclient.NewContext(rt, insecureRT).WithCredentials(creds)
-	context.DisableDigestVerification = o.SkipVerification
-	return context, nil
+	pushCreds := dockercredentials.NewLocal()
+	if len(o.PushConfig) > 0 {
+		pushCreds, err = dockercredentials.NewFromFile(o.PushConfig)
```
p0lyn0mial, line 118: haven't looked in great detail but it seems to return a custom implementation of a store that relies on a `credentialprovider.DockerKeyring` with the following interface:

```
// DockerKeyring tracks a set of docker registry credentials, maintaining a
// reverse index across the registry endpoints. A registry endpoint is made
// up of a host (e.g. registry.example.com), but it may also contain a path
// (e.g. registry.example.com/foo) This index is important for two reasons:
// - registry endpoints may overlap, and when this happens we must find the
//   most specific match for a given image
// - iterating a map does not yield predictable results
type DockerKeyring interface {
	Lookup(image string) ([]AuthConfig, bool)
}
```

it looks like we could actually find the most specific match for a given image, wdyt?

pkg/cli/image/manifest/manifest.go
sallyom: agreed, i'll update, thanks - also still looking into whether it's possible to use the single auth file _somehow_

pkg/cli/image/manifest/manifest.go
sallyom: i will try : ) will report back

sallyom:
> Might want to add the same option to the vanilla `oc image mirror` as well, since you could have the same issue there.

yes, was planning on that! I was waiting to see if we all agree on the general direction and if so, add it to mirror but if not, less to undo : )

pkg/cli/image/manifest/manifest.go
sallyom: I've modified this PR: Now, the new flag is `--pull-secret` and if not specified, defaults to `--registry-config` and that defaults to `~/.docker/config.json` 
For users, they should have their Red Hat pull-secret separate from their containers authfile.

sallyom: for example, for a command that only pulls, like `oc adm release extract` you can either
`oc adm release extract -a ~/authfile` or `oc adm release extract --pull-secret ~/authfile`  
(but if you are in habit of pushing releases to a personal quay repo via `release mirror` or `release new`, you _should_ have separate auth files for these as the mirror or release new will require separate auths)

sallyom:
instead of adding a new flag to handle multiple auth files, we are documenting the workaround for this, that is:
`$ oc adm release mirror registry/repo/release:tag --to-dir=some-local-dir`
`$ oc image mirror --from-dir=some-local-dir 'file://openshift/release:tag*' registry/another-repo`
and where the `oc image mirror` command is provided as output of the first `oc adm release mirror` cmd

closing this PR in favor of docs update

PR #413, opened 2020-05-11
bug 1826676: oc status: check if the current project exists before printing status

ingvagabund:
It can happen a project is no longer available when running the status.
Right now, oc status considers "namespaces XXX not found" error message
as valid and keeps querying for services, deployment, etc. over
on-existing namespace.

New output for non-existing namespaces:
```
$ oc status
error: the project "jenkins-csb-skynet" specified in your config does not exist.
```

### DIFF
--- a/pkg/cli/status/status.go
+++ b/pkg/cli/status/status.go
@@ -1,12 +1,14 @@
 package status
 
 import (
+	"context"
 	"errors"
 	"fmt"
 
 	"github.com/gonum/graph/encoding/dot"
 	"github.com/spf13/cobra"
 
+	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


pkg/cli/status/status.go
```diff
@@ -148,6 +150,15 @@ func (o *StatusOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, baseCLI
 		if err != nil {
 			return err
 		}
+		_, err = projectClient.Projects().Get(context.TODO(), namespace, metav1.GetOptions{})
+		switch {
+		case kapierrors.IsForbidden(err):
+			return fmt.Errorf("you do not have rights to view project %q.", namespace)
```
soltysh, line 156: Drop the dot.

pkg/cli/status/status.go
```diff
@@ -148,6 +150,15 @@ func (o *StatusOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, baseCLI
...
+		case kapierrors.IsNotFound(err):
+			return fmt.Errorf("the project %q specified in your config does not exist.", namespace)
```
soltysh, line 158: Drop the dot.

PR #414, opened 2020-05-11
Bug 1827073: Support mirroring catalogs using default-tagged source images.

benluddy:
Catalogs may reference image pull specs without a tag or digest, with
the expectation that the tag "latest" will be assumed.

### DIFF
--- a/pkg/cli/admin/catalog/mirrorer_test.go
+++ b/pkg/cli/admin/catalog/mirrorer_test.go
@@ -1,10 +1,10 @@
 package catalog
 
 import (
-	"github.com/openshift/library-go/pkg/image/reference"
 	"reflect"
 	"testing"
 
+	"github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/oc/pkg/cli/image/imagesource"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/catalog/mirrorer_test.go
```diff
@@ -169,11 +169,11 @@ func TestMirror(t *testing.T) {
 					WithTag:    "quay.io/org/coreos-prometheus-operator:76771fef",
 					WithDigest: "quay.io/org/coreos-prometheus-operator@sha256:5037b4e90dbb03ebdefaa547ddf6a1f748c8eeebeedf6b9d9f0913ad662b5731",
 				},
-				"quay.io/test/etcd.0.9.2": {
-					WithTag: "quay.io/org/test-etcd.0.9.2:f0e557b2",
```
awgreene, line 173: Was there a reason you opted to update existing tests rather than create new tests?

benluddy: This changes the existing behavior -- mappings from name-only sources get an explicit "latest" tag -- so the affected test cases needed to have their expectations updated.

PR #417, opened 2020-05-12
Bug 1835011: 'oc project' improve error when project has been deleted

sallyom:
/cc @ingvagabund 
/cc @soltysh 

Fixes the misleading error that a user gets with `oc project` when current-context is a deleted project 
now will see 
```console
error: You are not a member of any projects. You can request a project to be created with the 'new-project' command.
```
instead of
```console
error: you do not have rights to view project "test".
```
NOTE: 
User sees the correct info when run `oc project non-existing-proj`, that they are not a member.  Run from test2 project:
```console
$ oc project test
error: You are not a member of project "test".
Your projects are:
* test2
* test4
```

### DIFF
--- a/pkg/cli/project/project.go
+++ b/pkg/cli/project/project.go
@@ -219,35 +219,22 @@ func (o ProjectOptions) Run() error {
 			}
 
 			if err := ConfirmProjectAccess(argument, client, kubeclient); err != nil {
+				msg := ""
 				if isNotFound, isForbidden := kapierrors.IsNotFound(err), kapierrors.IsForbidden(err); isNotFound || isForbidden {
-					var msg string
 					if isForbidden {
 						msg = fmt.Sprintf("You are not a member of project %q.", argument)
 					} else {
 						msg = fmt.Sprintf("A project named %q does not exist on %q.", argument, clientCfg.Host)
 					}
 
-					projects, err := GetProjects(client, kubeclient)
-					if err == nil {
-						switch len(projects) {
-						case 0:
-							msg += "\nYou are not a member of any projects. You can request a project to be created with the 'new-project' command."
-						case 1:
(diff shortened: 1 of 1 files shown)


pkg/cli/project/project.go
```diff
@@ -320,13 +320,22 @@ func (o *ProjectOptions) GetContextFromName(contextName string) (*clientcmdapi.C
 }
 
 func ConfirmProjectAccess(currentProject string, projectClient projectv1client.ProjectV1Interface, kClient corev1client.CoreV1Interface) error {
+	projects, err := GetProjects(projectClient, kClient)
```
ingvagabund, line 323: `GetProjects` calls Projects.List which you might not have a permission to do while you could still have one for Projects.Get. Though, not sure if it's a real scenario since usually when you are given Get perms, you get List as well.

pkg/cli/project/project.go
```diff
@@ -320,18 +307,43 @@ func (o *ProjectOptions) GetContextFromName(contextName string) (*clientcmdapi.C
...
+			matched = true
+			break
+		}
+	}
+	if matched {
 		return nil
 	}
-
+	var msg string
+	switch len(projects) {
```
ingvagabund, line 333: Original code iterates through projects only when `GetProjects`'s err is nil. This changes adds another case where err  is either IsNotFound or IsForbidden but the namespace does not exist. In which case `projects` is nil. Which is still aligned with the comment and the error message for case 0.

pkg/cli/project/project.go
soltysh: I don't like we list all projects and if that fails we get the actual projects, this seems backwards. `ConfirmProjectAccess` should try a get on a project and alternatively on a namespace, that's it. I wouldn't change any code in it. It seems overly complicated. It's the caller business to handle the error accordingly to what I said in the comment yesterday.

sallyom:
closing, in favor of https://github.com/openshift/oc/pull/431

PR #419, opened 2020-05-14
[WIP] Introduce oc create project to replace oc new-project

soltysh:
This came out earlier today during discussion with Michal, he proposed to be more k8s like and introduce `oc create project` instead of `oc new-project`. I started wondering about `oc projects` vs `oc get projects` as well. Let's use this PR to discuss these options.

/assign @mfojtik @smarterclayton

### DIFF
--- a/contrib/completions/bash/oc
+++ b/contrib/completions/bash/oc
@@ -10214,6 +10214,72 @@ _oc_create_priorityclass()
     noun_aliases=()
 }
 
+_oc_create_project()
+{
+    last_command="oc_create_project"
+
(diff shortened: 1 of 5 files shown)


mfojtik:
+1

I like to be more kubernetes-alike if possible... I had a brain fart in morning, low on caffeine, when I wanted to create project and the natural thing for me was to type `kubectl create project foo`. I would keep `new-project` as alias maybe and set it deprecated, but keep it for long time as this command is probably used in endless tutorials.

soltysh:
This is not happening.
/close

PR #423, opened 2020-05-15
Bug 1827544: provide better defaults for oc adm catalog build

ecordell:
- will try to find the base image to use from an existing ocp connection or release image
- will throw an error if it can't find a base image from a release or explicitly with `--from`
- will default to linux/amd64 for the base image arch when building


pkg/cli/admin/catalog/build.go
```diff
@@ -48,21 +50,30 @@ type BuildImageOptions struct {
...
+	// default the base image os to linux/amd64 (the most common case)
+	pattern := o.FilterOptions.FilterByOS
+	if len(pattern) == 0 {
+		o.FilterOptions.FilterByOS = regexp.QuoteMeta(fmt.Sprintf("%s/%s", "linux", "amd64"))
+	}
+
+	// default the base image tag to match the version of oc being used
+	_, reportedVersion, _ := version.ExtractVersion()
+	if len(reportedVersion) > 0 {
+		o.From = defaultFromImage + ":v" + reportedVersion
```
ecordell, line 73: Experimentally, this seems to do what we want:

```
$ oc version --client
Client Version: 4.3.5
```
corresponds to:
https://catalog.redhat.com/software/containers/detail/5cddd0bed70cc57c44b2e1f3?tag=v4.3.5&architecture=amd64

ecordell:
/hold

I'm going to change this to grab the base image from a target ocp release

pkg/cli/admin/catalog/build.go
ecordell: Changed to pull from a release instead.

pkg/cli/admin/catalog/build.go
```diff
@@ -26,13 +27,29 @@ var (
...
+		The base image used for the catalog should match the target version of ocp. This can be set manually with 
+		the '--from' flag, or it can be inferred from an ocp release image via '--release-image', or directly from
+		the target cluster if logged in via 'oc login'. 
+
+		The base image will often be a multi-arch image. By default, the linux/amd64 variant is chosen. This can be 
+		overridden with '--filter-by-os'.
 		`)
 	buildExample = templates.Examples(`
-# Build an operator catalog from an appregistry repo and store in a file 
+# Build an operator catalog from an appregistry repo and store in a file. Assumes you are logged in via 'oc login'.
```
soltysh, line 39: Every command in oc assumes you're logged in, so I don't think you need to state that here explicitly.

pkg/cli/admin/catalog/build.go
```diff
@@ -44,25 +61,50 @@ type BuildImageOptions struct {
...
 		ParallelOptions:         imagemanifest.ParallelOptions{MaxPerRegistry: 4},
 	}
 }
 
-func (o *BuildImageOptions) Complete(cmd *cobra.Command, args []string) error {
+func (o *BuildImageOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
+	// expect to extract the base image from an active ocp connection or a release image
+	if len(o.From) == 0 {
+		// get an iostream buffer for extracting the image name from the release image
+		streams, _, out, errOut := genericclioptions.NewTestIOStreams()
```
soltysh, line 84: This is definitely wrong usage of that method, it's meant only for testing. I'd prefer you manually create that structure, if you need one.

pkg/cli/admin/catalog/build.go
```diff
@@ -44,25 +61,50 @@ type BuildImageOptions struct {
...
+		infoOpts := release.NewInfoOptions(streams)
+		infoOpts.ImageFor = "operator-registry"
+		images := []string{}
+		if len(o.ReleaseImage) > 0 {
+			images = append(images, o.ReleaseImage)
+		}
+		if err := infoOpts.Complete(f, nil, images); err != nil {
```
soltysh, line 91: Seems like you could just use `LoadReleaseInfo` from `InfoOptions` which would be much nicer than what you're doing here.

pkg/cli/admin/catalog/build.go
```diff
@@ -44,25 +61,50 @@ type BuildImageOptions struct {
...
+		}
+		if err := infoOpts.Run(); err != nil {
+			return fmt.Errorf("error extracting operator-registry image from release: %s", errOut.String())
+		}
+		o.From = strings.TrimSpace(out.String())
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "using %s as a base image for building", o.From)
+
+	// default the base image os to linux/amd64 (the most common case)
```
soltysh, line 102: Shouldn't this option be exposed to users?

pkg/cli/admin/catalog/build.go
ecordell: Do you have a preference between renaming this `NewBufferIOStream` vs. making a new structure? Do you anticipate testing needs to diverge from simple buffer storage?

pkg/cli/admin/catalog/build.go
ecordell: it is, via filter-by-os

pkg/cli/admin/catalog/build.go
soltysh: I wouldn't guarantee stability of that test method vs stability of the struct itself.

ecordell:
/retest

gcp error: lookupDiskImageSources

pkg/cli/admin/catalog/build.go
```diff
@@ -44,25 +67,69 @@ type BuildImageOptions struct {
...
 		IOStreams:               streams,
 		ParallelOptions:         imagemanifest.ParallelOptions{MaxPerRegistry: 4},
 	}
 }
 
-func (o *BuildImageOptions) Complete(cmd *cobra.Command, args []string) error {
+func (o *BuildImageOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
+	// expect to extract the base image from an active ocp connection or a release image
+	if len(o.From) == 0 {
+		fmt.Fprintf(o.IOStreams.Out, "--from not specified, attempting to get base image from release image\n")
```
soltysh, line 89: These look more like debugging messages, rather than normal output. If you want to, I'd put one statement at the end where you get the image from.

pkg/cli/admin/catalog/build.go
```diff
@@ -44,25 +67,69 @@ type BuildImageOptions struct {
...
+					o.From = tag.From.Name
+				}
+			}
+		}
+		if len(o.From) == 0 {
+			return fmt.Errorf("unable to find release tag %s in release image %s", releaseImageStreamTag, info.Image)
+		}
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "using %s as a base image for building", o.From)
```
soltysh, line 125: Yeah, leave only this and the rest of prints turn into `klog.V(2).Infof`

ecordell:
> Which means you should be able to pass whatever you wanted through --from, no?

Yes, you can pass whatever you want as a base image. 

Typically you'll want to pick the base image from a release, but there are cases where that's not what you want (you might want to pick a newer image to fix a bug, or you might want be targeting an upstream cluster, or you might be QE testing different variants).

So the order is: if you've told us the base image (`--from`) we'll use that, or if you've told us a specific release (`--release-image`) we'll use the image from that, or we'll try to figure out a good base image from the cluster you're logged into.

soltysh:
> So the order is: if you've told us the base image (`--from`) we'll use that, or if you've told us a specific release (`--release-image`) we'll use the image from that, or we'll try to figure out a good base image from the cluster you're logged into.

If that's the case, I still think all of that should be baked in one flag, having more is just confusing for user as in which one should I use, what happens if both are set, etc. Having a single gives simple information - that's the one we use, or we default to a cluster.

pkg/cli/admin/catalog/build.go
```diff
@@ -26,16 +29,36 @@ var (
 
 		Extracts the contents of a collection of operator manifests to disk, and builds them into
 		an operator registry catalog image.
+
+		The base image used for the catalog should match the target version of ocp. This can be set manually with 
+		the '--from' flag, or it can be inferred from an ocp release image via '--release-image', or directly from
```
soltysh, line 34: You left `--release-image` mentioned here.

pkg/cli/admin/catalog/build.go
```diff
@@ -113,7 +184,7 @@ func NewBuildImage(streams genericclioptions.IOStreams) *cobra.Command {
 	o.FilterOptions.Bind(flags)
 	o.ParallelOptions.Bind(flags)
 
-	flags.StringVar(&o.From, "from", o.From, "The image to use as a base.")
+	flags.StringVar(&o.From, "from", o.From, "The image to use as a base. This can be omitted if `--release-image` is specified or if oc is already logged into the target cluster.")
```
soltysh, line 187: Again `--release-image` being mentioned.

PR #425, opened 2020-05-18
TEST UPSTREAM CHANGE: DO NOT MERGE

sallyom:
@soltysh this resolves https://bugzilla.redhat.com/show_bug.cgi?id=1806785  
I'm not so familiar w/ describe code, and/or where the rbacv1.PolicyRule is getting the duplicate verbs for these: 
```
namespaces                                       []         []        [get get list watch]
packagemanifests.packages.operators.coreos.com   []         []        [get list watch get list 
```
The change here 'fixes' the duplicates but is there something I'm missing b4 I open an upstream PR? thanks

### DIFF
--- a/vendor/k8s.io/kubectl/pkg/util/rbac/rbac.go
+++ b/vendor/k8s.io/kubectl/pkg/util/rbac/rbac.go
@@ -42,7 +42,11 @@ func CompactRules(rules []rbacv1.PolicyRule) ([]rbacv1.PolicyRule, error) {
 				if existingRule.Verbs == nil {
 					existingRule.Verbs = []string{}
 				}
-				existingRule.Verbs = append(existingRule.Verbs, rule.Verbs...)
+				for _, v := range rule.Verbs {
+					if !isDuplicateVerb(existingRule.Verbs, v) {
+						existingRule.Verbs = append(existingRule.Verbs, v)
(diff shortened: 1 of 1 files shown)


sallyom:
/hold

opened only for testing, not for merge

vendor/k8s.io/kubectl/pkg/util/rbac/rbac.go
```diff
@@ -42,7 +42,11 @@ func CompactRules(rules []rbacv1.PolicyRule) ([]rbacv1.PolicyRule, error) {
 				if existingRule.Verbs == nil {
 					existingRule.Verbs = []string{}
 				}
-				existingRule.Verbs = append(existingRule.Verbs, rule.Verbs...)
+				for _, v := range rule.Verbs {
```
soltysh, line 45: I'd suggest using `k8s.io/apimachinery/pkg/util/sets#String` here, so you don't have to deal with duplicates manually.
Just create a `sets.String` and add verbs to it, after loop you'll assign it back to `existingRule.Verbs`

sallyom: thanks!

sallyom:
opened upstream

PR #427, opened 2020-05-19
Bug 1823143: oc adm release mirror|extract allow user to force from image prefix

sallyom:
@smarterclayton while thinking about the flag suggestion from https://github.com/openshift/oc/pull/395 I quickly modified that PR to use a flag instead.  I'm still looking at how this can happen at a lower level, but this is less ugly of a hack - still too ugly, though? 

**To test this PR, do something like this:**
```
$ podman run -p 5000:5000 -d registry:2
$ oc adm release mirror --insecure=true -a ~/pull-secret --from=registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-05-18-131223 --to localhost:5000/ocp/release
$ oc adm release --insecure=true extract --force-prefix=true --command openshift-install localhost:5000/ocp/release:4.5.0-0.nightly-2020-05-18-131223 -v=2  
```
you'll find the failure from master, fixed w /this PR


wking:
> To test this PR, do something like this...

But then your suggestion doesn't mention `--force-prefix`?

pkg/cli/admin/release/info.go
```diff
@@ -775,6 +775,20 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 				errs = append(errs, err)
 				return true, nil
 			}
+			for _, tag := range is.Spec.Tags {
+				// If ForcePrefix true, use user-provided image rather than it's mirrored source
```
wking, line 779: nit: "it's" -> "its"

pkg/cli/admin/release/info.go
```diff
@@ -775,6 +775,20 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
...
+				// imagereference.Parse returns the digest ID of each component in the release image-reference.
+				// If can't get digest ID, skip this tag, this happens when user has built a payload by
+				// replacing component images in the release with a new image
```
wking, line 782: Is this "user builds a release replacing a component image with a by-tag pullspec"?  Can we forbid that?  I would hope users replacing component images would use by-digest pullspecs, and expect the new release image to include digests for those images too.

pkg/cli/admin/release/mirror.go
```diff
@@ -593,6 +595,12 @@ func (o *MirrorOptions) Run() error {
 			return fmt.Errorf("image-references should only contain pointers to images by digest: %s", tag.From.Name)
 		}
 
+		// if ForcePrefix set, set from to be user-provided image rather than it's mirrored source
```
wking, line 598: nit: "it's" -> "its"

pkg/cli/admin/release/info.go
sallyom: > > To test this PR, do something like this...
> 
> But then your suggestion doesn't mention `--force-prefix`?

oops, fixed.. plus, I am going to rework this to use ICSP, but wanted to push this in the meantime bc it's simpler than the other PR.

sallyom: > nit: "it's" -> "its"
thanks, that one always gets me

pkg/cli/image/manifest/manifest.go
```diff
@@ -42,12 +42,14 @@ type SecurityOptions struct {
...
 	Insecure         bool
 	SkipVerification bool
+	ForcePrefix      bool
 
 	CachedContext *registryclient.Context
 }
 
 func (o *SecurityOptions) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&o.RegistryConfig, "registry-config", "a", o.RegistryConfig, "Path to your registry credentials (defaults to ~/.docker/config.json)")
+	flags.BoolVar(&o.ForcePrefix, "force-prefix", o.ForcePrefix, "Force lookup of named prefix (registry/repository/name) for an image source")
```
wking, line 52: It feels strange to me to have this flag down on `oc image info ...` and such but have handling only up under `oc adm release ...`.

pkg/cli/admin/release/info.go
sallyom: When I create a release payload w/ a substituted image I pass something like `cli=quay.io/sallyom/cli:test` I think that's what most devs would do? That's the case I added that skip for.  :shrug:

sallyom:
> > To test this PR, do something like this...
> 
> But then your suggestion doesn't mention `--force-prefix`?

fixed it, added to the description, thanks, and oops

pkg/cli/image/manifest/manifest.go
sallyom: yes, wasn't sure where to place the flag.  I was going to take suggestions.  I'm reworking this to use the ImageContentSourcePolicy information, though, and possibly remove the need for a user to pass a flag at all - working on that now

soltysh:
/hold
this is being discussed in https://github.com/openshift/enhancements/pull/334

sallyom:
Direction of this PR has changed, closing this in favor of https://github.com/openshift/oc/pull/439 and https://github.com/openshift/enhancements/pull/334

PR #430, opened 2020-05-20
Bug 1837375: use http.Client.Head instead of tls.Dial to be able to consume proxy settings

soltysh:
/assign @stlaz 
In https://github.com/openshift/oc/commit/eabf7a06457d4c645167cb150167fb3076bb2899 you moved to `tls.Dial` the problem is that it does not use proxy, when one is specified. It needs to wrapped in a full http client to consume one, when needed.

### DIFF
--- a/pkg/helpers/tokencmd/request_token.go
+++ b/pkg/helpers/tokencmd/request_token.go
@@ -407,25 +407,29 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 		return nil, err
 	}
 
-	// perform the retrieval with insecure transport, otherwise oauth-server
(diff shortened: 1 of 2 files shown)


stlaz:
/hold
this would just bring the old problem back and the BZ has nothing to do with this as it's the DNS lookup that fails

soltysh:
> this would just bring the old problem back and the BZ has nothing to do with this as it's the DNS lookup that fails

Verified just now and only when I'm using 4.4 binary I'm getting:

stlaz:
/hold cancel
/lgtm
I was wrong about the old problem reappearing, I still don't understand the DNS issue reported at the BZ, but that does not matter, this is a good change

PR #433, opened 2020-05-21
Bug 1774616: Update BuildLogOptions

coreydaley:
Adds InsecureSkipTLSVerifyBackend to BuildLogOptions


adambkaplan:
@coreydaley please clean up this PR such that vendor changes only appear in a `bump(xx)` commit.

coreydaley:
> @coreydaley please clean up this PR such that vendor changes only appear in a `bump(xx)` commit.

I think they already are cleaned up?

bparees:
vendor update commits should be titled `bump(*)` (or whatever example you can find in the commit history)

coreydaley:
@bparees can you re-lgtm me?  I updated the commit message

bparees:
it's bump(*).

you may have been only wanting to pick up openshift/api, but as you can see you also picked up a k8s api bump here.

coreydaley:
@bparees fixed.

bparees:
> /hold for @bparees to decide if he needs it for 4.5. We did this for pod logs to solve a rebootstrapping debugging scenario when kubelets running static pods cannot start. Builds don't require that.

definitely not a blocker for 4.5, so i'm going to defer to @soltysh as the defacto owner of the binary as to whether he's ok taking on the (low, imho) risk of including this change in 4.5.

coreydaley:
@soltysh I think we should be able to merge this since 4.6 is open now.

PR #435, opened 2020-05-22
Bug 1827486: clarify usage of 'oc adm create-kubeconfig'

sallyom:
As discussed on slack, we will deprecate 'oc adm create-kubeconfig' in its current form. Will replace with an equivalent 'oc config set ...' in the future.  

/assign @soltysh


pkg/cli/admin/createkubeconfig/create_kubeconfig.go
```diff
@@ -93,6 +93,8 @@ func NewCommandCreateKubeConfig(commandName string, fullName string, streams gen
 				kcmdutil.CheckErr(err)
 			}
 		},
+		Deprecated: "and will be removed in the future version",
+		Hidden:     true,
```
soltysh, line 97: Nit: this is not needed, each deprecated command is automatically hidden.

sallyom: oops and updated

PR #438, opened 2020-05-26
Bug 1829832: support mirroring for index images

ecordell:
`oc adm catalog mirror` initially targeted "registry images" built via `oc adm catalog build`.

Newer `opm` tooling creates index images with slight differences. This PR accounts for them in two ways:

- The database folder has always been configurable, but the database name was hardcoded to bundles.db. The first commit removes the hardcoding, so index images could be supported via the `--path` flag.
- `mirror` now detects the "index label", if present, and uses that to find the database in the image. `--path` takes precedence, if present.

### DIFF
--- a/pkg/cli/admin/catalog/mirror.go
+++ b/pkg/cli/admin/catalog/mirror.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/alicebob/sqlittle"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/catalog/mirror.go
```diff
@@ -226,8 +227,30 @@ func (o *MirrorCatalogOptions) Complete(cmd *cobra.Command, args []string) error
...
+		var dbPath string
+		errFound := fmt.Errorf("found valid db file")
+		err := filepath.Walk(e.Mappings[0].To, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return nil
+			}
+			if info.IsDir() {
+				return nil
+			}
+			if _, err := sqlittle.Open(path); err == nil {
```
soltysh, line 301: Is it possible that more than one valid sqlite database file will be present there?

ecordell: Possible but unlikely (no tooling for building these images will do it) - but if that's the case, the output will indicate which one was picked, and if it's not the one you wanted you can override with `--path`

PR #439, opened 2020-05-26
Bug 1823143: Implement enhancement for "Add ImageContentSource awareness to oc"

sallyom:
This PR implements https://github.com/openshift/enhancements/pull/334:
* writes an ICSP file as part of `oc adm release mirror` to either:
    * ./config 
    * --release-image-icsp-to-dir config subdirectory  or --to-dir config subdirectory
* brings in library-go change to allow lookup of alternative image sources from mirrored registries

**To Test** - See these test commands: https://github.com/openshift/oc/pull/439#issuecomment-668228752
 [set up a secure local registry with this script.](https://gist.github.com/sallyom/dc168cfecfd478547cbf5e0a85fb9334)

For previous discussions and closed/related PRs:
https://github.com/openshift/oc/pull/427
https://github.com/openshift/oc/pull/395

Includes cherry-picks of commits in #745
Includes fake bump with library-go change: https://github.com/openshift/library-go/pull/939


pkg/cli/image/manifest/manifest.go
```diff
@@ -157,10 +161,44 @@ func (o *ImageSourceOptions) Complete(flags *pflag.FlagSet, args []string) error
...
+		for _, icsp := range icsps.Items {
+			repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
+			var sourceMatches bool
+			for _, rdm := range repoDigestMirrors {
+				if strings.Contains(args[0], rdm.Source) {
+					sourceMatches = true
+				}
+				for _, m := range rdm.Mirrors {
+					if sourceMatches {
+						o.ImageSourcePrefix = m
```
sallyom, line 190: TODO: how to tell which rdm.Mirror if there is more than 1

sallyom:
/hold

until enhancement https://github.com/openshift/enhancements/pull/334 merges

pkg/cli/image/manifest/manifest.go
```diff
@@ -114,6 +122,113 @@ func (o *SecurityOptions) NewContext() (*registryclient.Context, error) {
...
+			var icsp *operatorv1alpha1.ImageContentSourcePolicy
+			var ok bool
+			if icsp, ok = icspObj.(*operatorv1alpha1.ImageContentSourcePolicy); !ok {
+				klog.Infof("HERE")
+				return fmt.Errorf("could not decode ImageContentSourcePolicy from %s", o.ImageContentSourcePolicyFile)
+			}
+			repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
+			var sourceMatches bool
+			for _, rdm := range repoDigestMirrors {
+				if strings.Contains(images[0], rdm.Source) {
```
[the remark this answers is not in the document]
sallyom: yes, this part was a poc, should've put a note there

pkg/cli/image/manifest/manifest.go
```diff
@@ -114,6 +122,113 @@ func (o *SecurityOptions) NewContext() (*registryclient.Context, error) {
...
+	ImageContentSourcePolicyFile string
+}
+
+// Bind adds the options to the flag set.
+func (o *ImageSourceOptions) Bind(flags *pflag.FlagSet) {
+	flags.StringVar(&o.ImageContentSourcePolicyFile, "icsp-file", o.ImageContentSourcePolicyFile, "Path to an ImageContentSourcePolicy file.  If set, data from this file will be used to set source release image.")
+}
+
+// Complete completes ImageSourceOptions flags.
+func (o *ImageSourceOptions) Complete(f kcmdutil.Factory, images []string) error {
```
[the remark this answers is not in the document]
sallyom: ok i've split that function and moved it to manifest/sources.go

pkg/cli/image/manifest/manifest.go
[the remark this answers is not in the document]
sallyom: updated : )

pkg/cli/admin/release/mirror.go
```diff
@@ -885,15 +938,30 @@ func printImageContentInstructions(out io.Writer, from, to string, signatureToDi
 	}
 	delete(unstructuredObj.Object["metadata"].(map[string]interface{}), "creationTimestamp")
 
-	icspExample, err := yaml.Marshal(unstructuredObj.Object)
+	icspDataBytes, err := yaml.Marshal(unstructuredObj.Object)
 	if err != nil {
-		return fmt.Errorf("Unable to marshal ImageContentSourcePolicy example yaml: %v", err)
+		return fmt.Errorf("Unable to marshal ImageContentSourcePolicy yaml: %v", err)
+	}
+	icspFile := o.ICSPTo
```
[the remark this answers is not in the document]
sallyom: grrr ok

pkg/cli/admin/release/info.go
```diff
@@ -735,16 +740,47 @@ func (i *ReleaseInfo) Platform() string {
 }
 
 func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*ReleaseInfo, error) {
+	opts := extract.NewOptions(genericclioptions.IOStreams{Out: o.Out, ErrOut: o.ErrOut})
+	opts.SecurityOptions = o.SecurityOptions
+	opts.ImageSourceOptions = o.ImageSourceOptions
+	opts.FileDir = o.FileDir
+
+	setPrefix, imageSources, err := imagemanifest.GetImageSourcePrefixes(opts.ImageSourceOptions.ImageContentSourcePolicyList, image)
```
soltysh, line 748: I'd suggest either calling this `useMirrors` or check `imageSources` length and act accordingly in the condition below.

pkg/cli/admin/release/mirror.go
```diff
@@ -140,13 +142,15 @@ func NewMirror(f kcmdutil.Factory, parentName string, streams genericclioptions.
...
 	o.SecurityOptions.Bind(flags)
+	o.ImageSourceOptions.Bind(flags)
 	o.ParallelOptions.Bind(flags)
 
 	flags.StringVar(&o.From, "from", o.From, "Image containing the release payload.")
 	flags.StringVar(&o.To, "to", o.To, "An image repository to push to.")
 	flags.StringVar(&o.ToImageStream, "to-image-stream", o.ToImageStream, "An image stream to tag images into.")
 	flags.StringVar(&o.FromDir, "from-dir", o.FromDir, "A directory to import images from.")
 	flags.StringVar(&o.ToDir, "to-dir", o.ToDir, "A directory to export images to.")
+	flags.StringVar(&o.ICSPTo, "icsp-to", o.ICSPTo, "Path to write ImageContentSourcePolicy file to.  If not set, will write to current directory.")
```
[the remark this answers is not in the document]
soltysh: I think it's about using the same mechanics for saving ISCP as in your proposal for saving release signatures.

pkg/cli/image/manifest/source.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		// may or may not be connected to a cluster
+		// don't error if can't connect
+		klog.V(4).Infof("did not connect to an OpenShift 4.x server")
+		return nil
+	}
+	icsps, err := icspClient.ImageContentSourcePolicies().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		// may or may not have access to ICSPs in cluster
+		// don't error if can't access ICSPs
```
soltysh, line 72: Hmm... I would want to know that I don't have access and I maybe should or that my mirror/copy/else will be missing something.

pkg/cli/admin/release/mirror.go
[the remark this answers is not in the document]
sallyom: oh, the ICSP file can be saved to the "config" directory, in absence of a --icsp-to-dir flag - same as what we do with the signature file.  i'll make that change.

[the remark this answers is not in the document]
sallyom: I was going to do that initially, but the flag is `--release-image-signature-to-dir`  - should that be more generic for both files - the signature + icsp files? @wking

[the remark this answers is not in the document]
sallyom: Pushed a commit w/ a single flag `--release-config-dir` that sets the directory for both release-image-signature-dir and icsp-to dir.

pkg/cli/image/manifest/source.go
sallyom: updated the message there.

pkg/cli/admin/release/info.go
sallyom: moved `setPrefix` to `useImageContentSources`

pkg/cli/admin/release/mirror.go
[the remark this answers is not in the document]
sallyom: The release-signature configmap change that writes the configmap to disk added a flag, 
`release-image-signature-to-dir` that I modified in this PR to `release-image-config-dir` so it can be dual purpose.  This requires a change to current users, though, so we might want to keep the release-image-signature-to-dir and add a new flag for the icsp-to-dir.

pkg/cli/admin/release/mirror.go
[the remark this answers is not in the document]
wking: Has @smarterclayton signed off on the generic approach whose enhancement he rejected earlier (linked above)?  Because if that's back on the table, the generic option that it died with was [`--config-to-dir`][1].

[1]: https://github.com/openshift/enhancements/pull/188/files#diff-13e449411659f84a0f8827c8a909c727R131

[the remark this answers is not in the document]
smarterclayton: I rejected recursive handling, the idea of the command outputting multiple config to the same dir is fine.  If we already pick a namespace for ICSP, and we expect that you have to apply ICSP, why do we need more than one dir?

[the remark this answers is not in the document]
sallyom: I'd like the one directory, `--release-image-config-dir` that a user can provide for path to write either icsp-file and/or release-signature file.  So the current `-release-image-signature-to-dir` will go away in favor of `--release-image-config-dir`.    
(the --icsp-to flag above this comment is outdated, i've removed that and this PR has the 1 dir --release-image-config-dir)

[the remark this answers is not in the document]
sallyom: I've updated so that both flags will be accepted - deprecated message with `release-image-signature-to-dir` in favor of `release-image-config-dir`

sallyom:
sample output (logging currently at v(2) but will increase):
```console
$ oc adm release extract --command openshift-install -a ~/installer/pull-secret-local localhost:5000/ocp/release:4.6.0-0.nightly-2020-08-02-091622 -v=2
I0803 16:26:00.685200  238616 extract_tools.go:448] Skipping openshift-install-mac-%s.tar.gz, does not match current OS darwin
I0803 16:26:00.685266  238616 extract_tools.go:456] Will extract usr/bin/openshift-install from quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d
I0803 16:26:00.867524  238616 extract_tools.go:329] Failed lookup of release quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d, will try other sources.
I0803 16:26:00.867579  238616 extract_tools.go:335] Could not load release info from image reference, will try ICSP and mirrored sources
I0803 16:26:01.765293  238616 extract_tools.go:448] Skipping openshift-install-mac-%s.tar.gz, does not match current OS darwin
I0803 16:26:01.765314  238616 extract_tools.go:456] Will extract usr/bin/openshift-install from localhost:5000/ocp/release@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d
I0803 16:26:04.023551  238616 extract_tools.go:738] Found match for release version at 3194 (len=4096, offset=327, n=3769)
I0803 16:26:04.031978  238616 extract_tools.go:738] Found match for release image at 1822 (len=4096, offset=327, n=3769)

$ oc adm release extract --command openshift-install -a ~/installer/pull-secret localhost:5000/ocp/release:4.6.0-0.nightly-2020-08-02-091622 -v=2
I0803 16:26:13.026902  238653 extract_tools.go:448] Skipping openshift-install-mac-%s.tar.gz, does not match current OS darwin
I0803 16:26:13.026960  238653 extract_tools.go:456] Will extract usr/bin/openshift-install from quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d
I0803 16:26:15.551124  238653 extract_tools.go:738] Found match for release version at 3194 (len=4096, offset=327, n=3769)
I0803 16:26:15.559095  238653 extract_tools.go:738] Found match for release image at 1822 (len=4096, offset=327, n=3769)

$ oc adm release extract --command openshift-install -a ~/installer/pull-secret --icsp-file /tmp/config/icsp-sha256-a0cd5e461757e8c0.yaml localhost:5000/ocp/release:4.6.0-0.nightly-2020-08-02-091622 -v=2
I0803 16:26:58.180344  238689 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:26:58.180399  238689 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:26:58.180404  238689 source.go:130] Will try these images: [localhost:5000/ocp/release]
I0803 16:26:58.768833  238689 extract_tools.go:448] Skipping openshift-install-mac-%s.tar.gz, does not match current OS darwin
I0803 16:26:58.768849  238689 extract_tools.go:456] Will extract usr/bin/openshift-install from localhost:5000/ocp/release@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d
I0803 16:27:01.057220  238689 extract_tools.go:738] Found match for release version at 3194 (len=4096, offset=327, n=3769)
I0803 16:27:01.065715  238689 extract_tools.go:738] Found match for release image at 1822 (len=4096, offset=327, n=3769)

$ oc apply -f /tmp/config/icsp-sha256-a0cd5e461757e8c0.yaml 
imagecontentsourcepolicy.operator.openshift.io/release created

$ oc adm release extract --command openshift-install -a ~/installer/pull-secret localhost:5000/ocp/release:4.6.0-0.nightly-2020-08-02-091622 -v=2
I0803 16:27:32.024492  238741 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:27:32.024598  238741 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:27:32.024618  238741 source.go:130] Will try these images: [localhost:5000/ocp/release]
I0803 16:27:32.095067  238741 extract_tools.go:448] Skipping openshift-install-mac-%s.tar.gz, does not match current OS darwin
I0803 16:27:32.095084  238741 extract_tools.go:456] Will extract usr/bin/openshift-install from quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d
I0803 16:27:34.765492  238741 extract_tools.go:738] Found match for release version at 3194 (len=4096, offset=327, n=3769)
I0803 16:27:34.773889  238741 extract_tools.go:738] Found match for release image at 1822 (len=4096, offset=327, n=3769)

$ oc adm release extract --command openshift-install -a ~/installer/pull-secret-local localhost:5000/ocp/release:4.6.0-0.nightly-2020-08-02-091622 -v=2
I0803 16:27:54.859013  238771 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:27:54.859094  238771 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:27:54.859101  238771 source.go:130] Will try these images: [localhost:5000/ocp/release]
I0803 16:27:54.938538  238771 extract_tools.go:448] Skipping openshift-install-mac-%s.tar.gz, does not match current OS darwin
I0803 16:27:54.938558  238771 extract_tools.go:456] Will extract usr/bin/openshift-install from quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d
I0803 16:27:55.098415  238771 extract_tools.go:329] Failed lookup of release quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d, will try other sources.
I0803 16:27:55.098468  238771 extract_tools.go:335] Could not load release info from image reference, will try ICSP and mirrored sources
I0803 16:27:55.098557  238771 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:27:55.098600  238771 source.go:115] localhost:5000/ocp/release RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy
I0803 16:27:55.098617  238771 source.go:130] Will try these images: [localhost:5000/ocp/release]
I0803 16:27:56.091969  238771 extract_tools.go:448] Skipping openshift-install-mac-%s.tar.gz, does not match current OS darwin
I0803 16:27:56.091991  238771 extract_tools.go:456] Will extract usr/bin/openshift-install from localhost:5000/ocp/release@sha256:ebb435d6f8ab070c6be2ba7dedc54413272f6d294897be71e3473b1d20e36c3d
I0803 16:27:58.403849  238771 extract_tools.go:738] Found match for release version at 3194 (len=4096, offset=327, n=3769)
I0803 16:27:58.412707  238771 extract_tools.go:738] Found match for release image at 1822 (len=4096, offset=327, n=3769)
```

pkg/cli/admin/release/mirror.go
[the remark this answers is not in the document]
sallyom: updated to add a parallel flag --release-image-icsp-to-dir  instead, after our discussion.

sallyom:
@smarterclayton I've updated this PR based on discussions.
* un-deprecated, un-joined --release-image-signature-to-dir and --release-image-icsp-to-dir.  both default to 'config' subdir of --to-dir, or just 'config' dir
* modified order in which sources will be tried.
    *  try image-ref from current image (same as now)
    * look for ICSP, try that if found
    * try user-given image

In case of disconnected, either 2 or 3 will succeed, if not return original error from 1.

pkg/cli/admin/release/mirror.go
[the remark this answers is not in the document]
sallyom: updated again - 2 flags - --release-image-signature-to-dir, --release-image-icsp-to-dir.  Both default to config or --to-dir dir/config

pkg/cli/image/manifest/source.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		var icsp *operatorv1alpha1.ImageContentSourcePolicy
+		var ok bool
+		if icsp, ok = icspObj.(*operatorv1alpha1.ImageContentSourcePolicy); !ok {
+			return fmt.Errorf("could not decode ImageContentSourcePolicy from %s", o.ImageContentSourcePolicyFile)
+		}
+		o.ImageContentSourcePolicyList = append(o.ImageContentSourcePolicyList, *icsp)
+		return nil
+	}
+	restConfig, err := f.ToRESTConfig()
```
smarterclayton, line 55: Unfortunately i think this structure probably won't work right - we don't want to make these calls here (which could be fairly expensive or slow), we want to only do them if we need them.  The check above is fine (since a user asked).

I think this method needs to provide a function that when asked can return alternates, that hides all of these details.  That function would be provided to the repository lookup client that could delegate to it late (and return an error or not).

I don't think doing this up front is going to scale, especially since this forces us to talk to a client in every image code path, even in places where we didn't do so before.

sallyom: got it, working on it, ok

sallyom: updated, now ImageSourceOptions has fn parameter, to pass an ICSPClient - and this client is _only_ used if the image lookup fails (via the function `GetICSPs(icspClient)`) (when using mirrored registries)

sallyom:
/hold cancel 
enhancement merged

sallyom:
@dmage If you can ptal at this PR it would be very much appreciated, thanks!

pkg/cli/image/manifest/source.go
```diff
@@ -0,0 +1,141 @@
...
+
+// GetImageSourcePrefixes returns list of 'registry/repository/name' options to try
+// 'registry/repository/name' determined from examining ImageContentSourcePolicy objects.
+// If a valid prefix is obtained from ICSP, return true
+func GetImageSourcePrefixes(icsps []operatorv1alpha1.ImageContentSourcePolicy, image string) (bool, []string, error) {
+	var imageSources []string
+	for _, icsp := range icsps {
+		repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
+		var sourceMatches bool
+		imageRef, err := reference.Parse(image)
```
dmage, line 99: You don't need to re-parse it on every iteration.

pkg/cli/image/manifest/source.go
```diff
@@ -0,0 +1,141 @@
...
+			for _, m := range rdm.Mirrors {
+				if sourceMatches {
+					klog.V(2).Infof("%v RepositoryDigestMirrors mirror added to potential ImageSourcePrefixes from ImageContentSourcePolicy", m)
+					imageSources = append(imageSources, m)
+				}
+				mRef, err := reference.Parse(m)
+				if err != nil {
+					return false, imageSources, err
+				}
+				if imageRef.AsRepository() == mRef.AsRepository() {
```
dmage, line 121: Why do you need this?

pkg/cli/image/manifest/source.go
```diff
@@ -0,0 +1,141 @@
...
+	if len(icsps.Items) == 0 {
+		klog.V(4).Info("no ImageContentSourcePolicies found in cluster")
+	}
+	o.ImageContentSourcePolicyList = append(o.ImageContentSourcePolicyList, icsps.Items...)
+}
+
+// GetImageSourcePrefixes returns list of 'registry/repository/name' options to try
+// 'registry/repository/name' determined from examining ImageContentSourcePolicy objects.
+// If a valid prefix is obtained from ICSP, return true
+func GetImageSourcePrefixes(icsps []operatorv1alpha1.ImageContentSourcePolicy, image string) (bool, []string, error) {
```
dmage, line 94: Note that ICSP should work only for images that are referenced by digest. We don't know if the mirrored images are fresh enough and we can trust its tags.

pkg/cli/image/manifest/source.go
```diff
@@ -0,0 +1,141 @@
...
+		imageRef, err := reference.Parse(image)
+		if err != nil {
+			return false, imageSources, err
+		}
+		for _, rdm := range repoDigestMirrors {
+			rdmRef, err := reference.Parse(rdm.Source)
+			if err != nil {
+				return false, imageSources, err
+			}
+			if imageRef.AsRepository() == rdmRef.AsRepository() {
```
dmage, line 108: ICSP is used in CRI-O and builds via `registries.conf`, so see their example: https://github.com/containers/image/blob/f1a6f9b50d0286bf208935d4b9ab48117294ac31/registries.conf

If you have the policy
```yaml
source: example.com/foo
mirrors:
- internal-registry-for-example.net/bar
```
then the image `example.com/foo/myimage:latest` should be pulled from `internal-registry-for-example.net/bar/myimage:latest`.

dmage (changes requested):
I reviewed only GetImageSourcePrefixes, it doesn't look like a correct implementation. ICSPs define rules for prefixes, not full repository names.

See also https://github.com/openshift/runtime-utils/blob/c45d4ff3f912/pkg/registries/registries.go#L104, it merges multiple ICSPs together and deals with some edge-cases using topological sorting.

pkg/cli/image/manifest/source.go
sallyom: fixed, updated

pkg/cli/image/manifest/source.go
sallyom: this is what allows
`oc adm release extract -a ~/pull-secret-local --icsp-file icsp.yaml --command openshift-install 4.6.0` to succeed, with this icsp.yaml, generated from `oc adm release mirror 4.6.0 --to localhost:5000/openshift/release`:
```
metadata:
  name: example
spec:
  repositoryDigestMirrors:
  - mirrors:
    - localhost:5000/openshift/release
    source: quay.io/openshift-release-dev/ocp-release
  - mirrors:
    - localhost:5000/openshift/release
    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
```

pkg/cli/image/manifest/source.go
sallyom: Yes, ICSP uses digests only. This change enables extracting from mirrored images and a disconnected registry. The
`registry/repository/name` is substituted.  When mirroring, `registry/repo/image@sha256:123456` becomes `mirroredregistry/repo2/image2@sha256:123456` The sha is the same.  This change enables that substitution during the extract, so you no longer require the auth for the original registry/repo to extract from a mirrored release.

pkg/cli/image/manifest/source.go
sallyom: I've added comment explaining that GetImageSourcePrefixes function.

pkg/cli/image/manifest/source.go
sallyom: > I reviewed only GetImageSourcePrefixes, it doesn't look like a correct implementation. ICSPs define rules for prefixes, not full repository names.
> 
> See also https://github.com/openshift/runtime-utils/blob/c45d4ff3f912/pkg/registries/registries.go#L104, it merges multiple ICSPs together and deals with some edge-cases using topological sorting.

I've added explanation to GetImageSourcePrefixes - I believe the implementation is what is required. From the client side, when extracting from mirrored images, currently this requires access & authorization to the original registry/repository defeating the purpose of mirroring.  This change adds logic to `oc` to try various imageSources (all verified through digests) since `oc` currently _only_ tries the underlying image reference of a mirrored image - for example, if `4.6.0` is mirrored to `localhost:5000/ocp/release`, 
it's reference remains `quay.io/openshift-release-dev/ocp-release@sha256:579cd61047067d8d642dc0bf8b3...` and currently, that is the only location `oc` knows to extract from.  This change allows `oc` to substitute the `registry/repository/name` so the release will be `localhost:5000/openshift/release@sha256:579cd61047067d8d6` and then the extract will then happen from `localhost:5000/openshift/release@sha256:sha-of-some-tool-or-component` that was originally `quay.io/openshift-release-dev/ocp-release@sha256:sha-of-some-tool-or-component`  This substitution/lookup happens in the following order:
*   if --icsp-file is set, then decode/use that - don't look further
*   image reference from user-passed release image (the original un-mirrored source)
*   if ICSP found in cluster, add that to list of image source options, those will be:
      * icsp from cluster
      * user-passed release image (set registry/repo/name from user-given, and use dame digest)

pkg/cli/image/manifest/source.go
sallyom: With extract from release there is a call to this fn to find the correct release image, then to find imageSources for the image from which you'll extract (say, the `openshift-install` binary from the installer image) the first, for mirrored, you'd need to replace `quay.io/openshift-release-dev/ocp-release@sha256:something` with `localhost:5000/openshift/release@sha256:something`  _then_  to extract, have to match again for @sha256:component but this time you'll match the passed image with the mirrors instead of the source - since now the passed image will be the substituted release.

pkg/cli/image/manifest/source.go
dmage: I don't doubt it works as you expect it. I'm trying to understand its purpose and assumptions.

It seems you call prefixes strings like `quay.io/openshift-release-dev/ocp-release`. In Docker (and CRI-O/podman/buildah) terminology they are called repositories. CRI-O calls prefixes strings like `quay.io`, `quay.io/openshift-release-dev`, or `quay.io/openshift-release-dev/ocp-release`. So it returns a list of repositories that may contain the image, right?

Do you assume that ICSP is generated by `oc adm release mirror`? If so, it's a strong assumption and it should be clearly documented.

pkg/cli/image/manifest/source.go
dmage: Right, but where is the check that the image is referenced by sha?

pkg/cli/image/manifest/source.go
sallyom: right, it returns `repositories` i call them prefixes because repositories is confusing/misleading. I can update to note that `prefix == repositories == registry/repository/name` that makes sense, and yes, this code assumes ICSP is generated by `oc adm release mirror`  but it should not, you are right.  let me look to see if i can remove that assumption

pkg/cli/image/manifest/source.go
sallyom: whatever image is user-given (it might be `registry/repo/name:tag` or `registry/repo/name@sha256:digest`) is parsed to use digest, explained here: https://github.com/openshift/oc/blob/d30b2d0a4f1bab7ec8bd51d13c50371c730eee84/pkg/cli/admin/release/info.go#L853-#L896  and verified if the substituted `registry/rep/name@sha256:digest` exists or if it does not exist, the substitution is not made.

pkg/cli/image/manifest/source.go
sallyom: I got rid of this, by adding the image-passed to the list of image sources

pkg/cli/image/manifest/source.go
sallyom: I've moved the logic for getting distribution.Repository to library-go see here: https://github.com/openshift/library-go/pull/939

sallyom:
To test this PR set up a local registry and run the following:
```console
$ oc adm release mirror registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-11-10-053716 --to localhost:5000/local/test
$ oc adm release extract -a ~/ps-local --command openshift-install localhost:5000/local/test:4.7.0-0.ci-2020-11-10-053716
$ oc adm release extract -a ~/ps-local --command openshift-install --icsp-file /tmp/config/icsp-sha256-f11....yaml registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-11-10-053716
$ oc apply -f config/icsp-sha256-f1....yaml  (if connected to cluster, and this file is generated from mirror cmd above)
$ oc adm release extract -a ~/ps-local --command openshift-install registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-11-10-053716  (this will succeed, with -v=2 you'll see it finds ICSP from cluster)
```
and to set up a localhost:5000 test registry:
https://gist.github.com/sallyom/dc168cfecfd478547cbf5e0a85fb9334#file-local-registry-10-yr-cert-with-san-sh

sallyom:
@smarterclayton this PR has grown exponentially - from our discussion the other day I've added logic to allow for alternative image source awareness in `oc image extract, oc image mirror` - I have not added to append or info - there is quite a bit of repeated code so I'm working on adding a fn for that and also adding the alternative-source-awareness in info, append.  
* The `oc image *` commands require either `--icsp-file` or `lookup-cluster-icsp` to try alternative sources. 
* The `oc adm release` commands try alternative sources implicitly, and preference is given if `--icsp-file` or `--lookup-cluster-icsp` is passed. 

This PR is huge now, I'll spend tomorrow DRY/rewriting unit tests and then hoping to have a list of things to accomplish in a follow-up PR.

sallyom:
`oc adm release` test commands:
NOTE: `oc adm release` implicitly looks for icsp in cluster if other paths fail, you do not have to pass the --lookup-cluster-icsp flag.
```console
$ oc adm release mirror registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-12-02-015504 --to localhost:5000/test/release
then
$ oc adm release -a ~/pull-secret-local-only extract --command openshift-install localhost:5000/test/release:4.7.0-0.ci-2020-12-02-015504 
or
$ oc adm release -a ~/pull-secret-local-only extract --command openshift-install  registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-12-02-015504 --icsp-file config/icsp..yaml
or
$ oc apply -f config/icsp...yaml
then
$ oc adm release -a ~/pull-secret-local-only extract --command openshift-install  registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-12-02-015504
```

`oc image`  test commands:
NOTE: `oc image` will not implicitly look for icsp in cluster, you have to pass the --lookup-cluster-icsp flag.
```console
$ oc adm release info registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-12-02-015504 --image-for cli-artifacts
registry.svc.ci.openshift.org/ocp/4.7-2020-12-02-015504@sha256:b2a9b9cbf7d36cea817739677c4123ebb3e3e0263ae7d48723097be0b6547096
$ oc image mirror registry.svc.ci.openshift.org/ocp/4.7-2020-12-02-015504@sha256:b2a9b9cbf7d36cea817739677c4123ebb3e3e0263ae7d48723097be0b6547096=localhost:5000/cli-artifacts:4.7.0-0.ci-2020-12-02-015504
$ oc image -a ~/ps-local info registry.svc.ci.openshift.org/ocp/4.7-2020-12-02-015504@sha256:b2a9b9cbf7d36cea817739677c4123ebb3e3e0263ae7d48723097be0b6547096
error: unable to read image registry.svc.ci.openshift.org/ocp/4.7-2020-12-02-015504@sha256:b2a9b9cbf7d36cea817739677c4123ebb3e3e0263ae7d48723097be0b6547096: Get "https://registry.svc.ci.openshift.org/v2/ocp/4.7-2020-12-02-015504/manifests/sha256:b2a9b9cbf7d36cea817739677c4123ebb3e3e0263ae7d48723097be0b6547096": unauthorized: authentication required

failed because not authorized - now run the same but pass the icsp file:
$ oc image -a ~/ps-local info registry.svc.ci.openshift.org/ocp/4.7-2020-12-02-015504@sha256:b2a9b9cbf7d36cea817739677c4123ebb3e3e0263ae7d48723097be0b6547096 --icsp-file config/icsp-sha256-2c0f33960669ce07.yaml
or, after 'oc apply -f config/icsp...yaml':
$ oc image -a ~/ps-local info registry.svc.ci.openshift.org/ocp/4.7-2020-12-02-015504@sha256:b2a9b9cbf7d36cea817739677c4123ebb3e3e0263ae7d48723097be0b6547096 --lookup-cluster-icsp=true

**Comment by sallyom** [comment]
..fixing bug in setting TryAlternativeSources bool, this is broken until fixed

File: contrib/completions/bash/oc [hunk 13]
```diff
@@ -576,8 +576,13 @@ _oc_adm_catalog_build()
     flags+=("--from-dir=")
     two_word_flags+=("--from-dir")
     local_nonpersistent_flags+=("--from-dir=")
+    flags+=("--icsp-file=")
```
soltysh, line 579: Pls split the generated bits like this to a separate commit.

pkg/cli/admin/release/extract.go
```diff
@@ -174,6 +174,12 @@ func (o *ExtractOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args [
 	}
 	o.From = args[0]
 
+	if err := o.SecurityOptions.Complete(f); err != nil {
+		return err
+	}
+	if !o.SecurityOptions.LookupClusterICSP && len(o.SecurityOptions.ICSPFile) == 0 {
+		o.SecurityOptions.TryAlternativeSources = true
```
soltysh, line 181: I don't understand, I thought that we want to check alternative sources when either cluster-icsp or icsp-file is passed?

pkg/cli/admin/release/extract_tools.go
```diff
@@ -297,16 +297,159 @@ func (o *ExtractOptions) extractCommand(command string) error {
 	infoOptions := NewInfoOptions(o.IOStreams)
 	infoOptions.SecurityOptions = o.SecurityOptions
 	infoOptions.FileDir = o.FileDir
-	release, err := infoOptions.LoadReleaseInfo(o.From, false)
-	if err != nil {
-		return err
+
+	var releaseFromRefErr error
+	targetRelease := targetReleaseInfo{}
+	if len(o.SecurityOptions.ICSPFile) > 0 {
```
soltysh, line 303: I thought this entire logic of picking file first and than using server or nothing should be part of the library-go code, why repeating similar logic here? As a consumer of the library-go code I'll pass both icspfile and flag saying to read it from cluster and library-go will know the priority and which to pick.

soltysh: It's entirely transparent for consumer.

pkg/cli/image/manifest/security.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	creds := dockercredentials.NewLocal()
+	if len(o.RegistryConfig) > 0 {
+		creds, err = dockercredentials.NewFromFile(o.RegistryConfig)
+		if err != nil {
+			return nil, fmt.Errorf("unable to load --registry-config: %v", err)
+		}
+	}
+
+	imageSources := registryclient.AlternativeImageSources{
```
soltysh, line 120: This should be transparent for users as well. You shouldn't be setting `AlternativeImageSources` by yourself.

pkg/cli/image/manifest/security.go
```diff
@@ -0,0 +1,238 @@
...
+		ICSPFile:     o.ICSPFile,
+		ICSPList:     o.ICSPList,
+		ICSPClientFn: o.ICSPClientFn,
+	}
+
+	context := registryclient.NewContext(rt, insecureRT, imageSources).WithCredentials(creds)
```
soltysh, line 126: I'd imagine we could do `WithICSPFile` and `WithClusterICSP` methods similar to `WithCredentials` above and those would be set based on `--icsp-file` and/or `--lookup-cluster-icsp` flags. And the `Context` method will know both the order (iow. what is more important file vs cluster ICSP) and if those are set and should be used. I think that's what @smarterclayton had in mind and how I understood him.

soltysh (changes requested):

contrib/completions/bash/oc
sallyom: done

pkg/cli/admin/release/extract.go
sallyom: I'll rename that variable, that should be something like `TryAlternativeSourcesImplicitly` this is because for the `oc adm release` commands, the expectation is that oc should look in cluster without being askes.. but actually, I think this variable might not be necessary, I think I can do without it :eyes: will report back

sallyom: renamed to `ImplicitTryAlternativeSources` and set to true with `oc adm release extract` and `oc adm release info`

pkg/cli/image/manifest/security.go
sallyom: here i'm taking the info from the user (icsp flags + the icspclient) and passing it to library-go.

pkg/cli/image/manifest/security.go
sallyom: i see, ok, lemme update that

sallyom: updated, and see above comment

pkg/cli/admin/release/extract_tools.go
sallyom: Yes, thanks, I missed this - this was leftover from previous implementation - I've cleaned this up and pushed the update :bowing_woman:  All the icsp gathering happens in lib-go.

sallyom:
@soltysh I've addressed your comments, thanks for your review again, it's very much appreciated : )

pkg/cli/admin/release/extract.go
soltysh: Probably `AlwaysTryAlternativeSources` would be better and more explicit.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -297,16 +297,115 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
+	typedReleaseRef, err := infoOptions.LoadReleaseInfo(o.From, false)
 	if err != nil {
-		return err
+		return fmt.Errorf("could not load release %s info: %v", o.From, err)
+	}
+	targetRelease, err = o.setReleaseLookup(typedReleaseRef, targets, currentOS, willArchive)
+	if err != nil {
+		return fmt.Errorf("failed lookup of release %s: %v", o.From, err)
+	}
+	if targetRelease.willArchive {
```
soltysh, line 310: I'd suggest the change which re-shuffles this code from one place to be re-usable to be in a separate commit, it'll be easier to grasp and understand.

pkg/cli/admin/release/info.go
```diff
@@ -735,15 +741,28 @@ func (i *ReleaseInfo) Platform() string {
...
+	opts.FileDir = o.FileDir
 	ref, err := imagesource.ParseReference(image)
 	if err != nil {
 		return nil, err
 	}
+	fromContext, err := opts.SecurityOptions.Context()
+	if err != nil {
+		return nil, err
+	}
+	fromContext.AlternativeSources = &opts.SecurityOptions
```
soltysh, line 756: This looks off, why you're assigning `opts.SecurityOptions` to a context which was just retrieved from `opt.SecurityOptions` a few lines above?

soltysh: I was expecting that all of alternative sources is done through `WithICSP` you've introduced in library-go.

vendor/github.com/openshift/library-go/pkg/image/registryclient/alternative-sources.go
```diff
@@ -0,0 +1,135 @@
...
+	ICSPFile     string
+	ICSPList     []operatorv1alpha1.ImageContentSourcePolicy
+	ICSPClientFn func() (operatorv1alpha1client.ImageContentSourcePolicyInterface, error)
+}
+
+type AlternativeSources interface {
+	PreferredImageSource(imageRef imagereference.DockerImageReference) (imagereference.DockerImageReference, distribution.Repository, distribution.ManifestService, error)
+}
+
+func (c *Context) WithICSP(icspFile string, icspList []operatorv1alpha1.ImageContentSourcePolicy, icspClientFn func() (operatorv1alpha1client.ImageContentSourcePolicyInterface, error)) *Context {
```
soltysh, line 30: Why you need both the list and the client? I'd expect either not both.

pkg/cli/image/manifest/security.go
```diff
@@ -0,0 +1,206 @@
...
+			return nil, fmt.Errorf("unable to load --registry-config: %v", err)
+		}
+	}
+
+	context := registryclient.NewContext(rt, insecureRT).WithCredentials(creds).WithICSP(o.ICSPFile, o.ICSPList, o.ICSPClientFn)
+	context.DisableDigestVerification = o.SkipVerification
+	return context, nil
+}
+
+func (s *SecurityOptions) PreferredImageSource(image imagereference.DockerImageReference) (imagereference.DockerImageReference, distribution.Repository, distribution.ManifestService, error) {
```
soltysh, line 125: The entire preferred image sources should be inside library-go context, that's what agreed to, no?

soltysh (changes requested):

pkg/cli/image/manifest/security.go
sallyom: moved, updated

pkg/cli/admin/release/info.go
sallyom: moved that all to library-go, updated

pkg/cli/admin/release/extract_tools.go
sallyom: updated, split this change into its own commit

pkg/cli/admin/release/extract.go
sallyom: updated to `AlwaysTryAlternativeSources`

pkg/cli/admin/release/info.go
```diff
@@ -734,16 +740,33 @@ func (i *ReleaseInfo) Platform() string {
 	return fmt.Sprintf("%s/%s", os, arch)
 }
 
-func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*ReleaseInfo, error) {
+// LoadReleaseInfo takes an image and returns release payload info or an error. The release image will be looked up using
+// alternative image sources from image content source policy objects and/or from the user-given string, if setRef is true
+// If retrieveImages is true, release payload component images will also be fetched.
+func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool, setRef bool) (*ReleaseInfo, error) {
```
soltysh, line 746: Nit: maybe `icspReference` or `alternateReference` instead of `setRef`?

pkg/cli/admin/release/info.go
```diff
@@ -734,16 +740,33 @@ func (i *ReleaseInfo) Platform() string {
...
+	replacedRef := false
+	opts := extract.NewExtractOptions(genericclioptions.IOStreams{Out: o.Out, ErrOut: o.ErrOut})
```
soltysh, line 748: `opts := extract.NewExtractOptions(o.IOStreams)` will be sufficient

pkg/cli/admin/release/info.go
```diff
@@ -734,16 +740,33 @@ func (i *ReleaseInfo) Platform() string {
...
+	if err != nil {
+		return nil, err
+	}
+	newRef, _, _, err := fromContext.PreferredImageSource(ref.Ref, o.SecurityOptions.AlwaysTryAlternativeSources, o.SecurityOptions.Insecure)
+	if err != nil {
+		return nil, err
+	}
+	if ref.Ref != newRef {
+		replacedRef = true
+	}
```
soltysh, line 765:
```go
replacedRef = ref.Ref != newRef
```
You could actually drop the above initialization of `replacedRef` and do that in just this single line.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -297,16 +297,123 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
+}
+
+func (o *ExtractOptions) setReleaseLookup(release *ReleaseInfo, targets []extractTarget, currentOS string, toArchive bool) (targetReleaseInfo, error) {
 	releaseName := release.PreferredName()
 	refExact := release.ImageRef
 	refExact.Ref.Tag = ""
 	refExact.Ref.ID = release.Digest.String()
 	exactReleaseImage := refExact.String()
 
+	tr := targetReleaseInfo{
```
soltysh, line 410: More descriptive names, pls, `targetRelease`, for example.

vendor/github.com/openshift/library-go/pkg/image/registryclient/alternative-sources.go
soltysh: This still holds, I don't think you need both the actual list of ICSP and client, either of the two should be sufficient, no?

soltysh: From looking at the code it seems that `icspList` is not being used here nor in `pkg/cli/image/manifest/security.go` I'd just drop it. Also you can probably just pass file and client. The presence of either is a sign that a user requested ICSP. `pkg/cli/image/manifest/security.go` will use the flag only to read user intent, iow. user will pass `--lookup-cluster-icsp` or `--icsp-file` and you pass both to this method.

soltysh (changes requested):
One nit and a big one about `WithICSP`

pkg/cli/admin/release/info.go
sallyom: updated, to `alternateImageReference`

pkg/cli/admin/release/info.go
sallyom: updated, thanks ;)

pkg/cli/admin/release/info.go
sallyom: thanks, updated

pkg/cli/admin/release/extract_tools.go
sallyom: updated

vendor/github.com/openshift/library-go/pkg/image/registryclient/alternative-sources.go
sallyom: don't need to pass the icspList,but the LookupClusterICSP is needed afaict. It has to be carried thru to differentiate from implicit icsp lookup from explicit lookup. I added comments to the library-go file to help clarify - I'll keep testing though to be sure.

sallyom:
> One nit and a big one about `WithICSP`

I removed the ICSPList from being passed, but the LookupClusterICSP I _think_ is required, bc that fn can't otherwise know if should explicitly look for alternative from cluster (with oc adm release) or not (oc image) I'll keep testing though to be sure. I added a few comments to the library-go file to clarify

vendor/github.com/openshift/library-go/pkg/image/registryclient/alternative-sources.go
sallyom: nm, updated & merged LookupClusterICSP w/ TryAlternativeSources.

pkg/cli/admin/release/info.go
```diff
@@ -734,16 +742,34 @@ func (i *ReleaseInfo) Platform() string {
...
+	opts.FileDir = o.FileDir
 	ref, err := imagesource.ParseReference(image)
 	if err != nil {
 		return nil, err
 	}
+	fromContext, err := opts.SecurityOptions.Context()
+	if err != nil {
+		return nil, err
+	}
+	newRef, repo, _, err := fromContext.PreferredImageSource(ref.Ref, o.SecurityOptions.AlwaysTryAlternativeSources, o.SecurityOptions.Insecure)
```
smarterclayton, line 761: This goes against the whole point of hiding this inside the registryclient interfaces (hiding the detection of failover).  This is exactly the sort of change I didn't want to make in oc - all we should have to do is identify the input ICSP and provide that to the context, and then everything else should work automatically with no change.  Doing it here defaults the point.

sallyom: I've updated both PRs, ptal

pkg/cli/image/manifest/security.go
```diff
@@ -112,7 +115,60 @@ func (o *SecurityOptions) NewContext() (*registryclient.Context, error) {
...
 
-	context := registryclient.NewContext(rt, insecureRT).WithCredentials(creds).WithICSP(o.ICSPFile, o.LookupClusterICSP, o.ICSPClientFn)
-	context.DisableDigestVerification = o.SkipVerification
-	return context, nil
+	var regContext *registryclient.Context
+	imageSources := imagecontentsources.ICSPSources{
+		ICSPFile:     o.ICSPFile,
+		ICSPClientFn: o.ICSPClientFn,
+	}
+	regContext = registryclient.NewContext(rt, insecureRT).WithCredentials(creds)
```
soltysh, line 123: yeah, so with my proposed changes from https://github.com/openshift/library-go/pull/939 this would be as simple as:
```golang
regContext = registryclient.NewContext(rt, insecureRT).WithCredentials(creds)
if len(o.ICSPFile) > 0 || o.LookupClusterICSP {
    regContext.WithImageResolver(imagecontentsources.NewICSPSources(o.ICSPFile, o.ICSPClientFn)
}
```

sallyom: updated, got it

pkg/cli/admin/release/info.go
```diff
@@ -783,6 +818,21 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 				errs = append(errs, err)
 				return true, nil
 			}
+			// If using an alternative image source, have to update all tags
+			if replacedRef || alternativeSource {
+				for _, tag := range is.Spec.Tags {
+					// If can't get digest ID, skip this tag, this happens when user has built a payload by
+					// replacing component images in the release with a new image (unsupported configuration)
+					tagRef, err := imagereference.Parse(tag.From.Name)
+					// if err != nil, skip this tag
```
soltysh, line 821: I think I missed it yesterday, fail early and try to avoid unnecessary blocks:
```golang
if err != nil {
    continue
}
ref.Ref.ID = tagRef.ID
...
```

soltysh (commented):
One nit, I must have missed before.

pkg/cli/admin/release/info.go
sallyom: fixed

pkg/cli/image/imagesource/options.go
```diff
@@ -35,8 +45,48 @@ func (o *Options) Repository(ctx context.Context, ref TypedImageReference) (dist
 		url := ref.Ref.DockerClientDefaults().RegistryURL()
 		return driver.Repository(ctx, url, ref.Ref.RepositoryName(), o.Insecure)
 	default:
-		return nil, fmt.Errorf("unrecognized image reference type %s", ref.Type)
+		return nil, nil, fmt.Errorf("unrecognized image reference type %s", ref.Type)
+	}
+}
+
+// PreferredImageSource returns an appropriate image reference, repository, and manifest service  for a given Context, or an error. Alternative image sources
```
sallyom, line 52: Since we already run Context.Repository from imagesource.Options, I added the new PreferredImageSource method here, as well.

pkg/cli/admin/release/info.go
```diff
@@ -587,7 +596,17 @@ func findImageSpec(image *imageapi.ImageStream, tagName, imageName string) (stri
...
+				if err != nil {
+					return "", err
+				}
+				preferred, err := imagereference.Parse(imageName)
+				if err != nil {
+					return "", err
+				}
+				preferred.ID = ref.ID
+				preferred.Tag = ref.Tag
+				return preferred.String(), nil
```
sallyom, line 609: there is a way this won't be necessary.. looking for it

sallyom: kind of removed it - moved to the `readReleaseImageReferences` in adm/release/image_mapper.go

pkg/cli/image/manifest/manifest.go
```diff
@@ -114,6 +122,113 @@ func (o *SecurityOptions) NewContext() (*registryclient.Context, error) {
...
+// ImageSourceOptions sets correct image source for oc adm commands
+type ImageSourceOptions struct {
+	ImageSourcePrefix            []string
+	SetPrefix                    bool
+	ImageContentSourcePolicyFile string
+}
+
+// Bind adds the options to the flag set.
+func (o *ImageSourceOptions) Bind(flags *pflag.FlagSet) {
+	flags.StringVar(&o.ImageContentSourcePolicyFile, "icsp-file", o.ImageContentSourcePolicyFile, "Path to an ImageContentSourcePolicy file.  If set, data from this file will be used to set source release image.")
```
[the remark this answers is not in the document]
sallyom: no longer needed

sallyom:
@smarterclayton most of the changes from the old approach https://github.com/openshift/oc/pull/790 are not required, without the requirement of pulling the original imageReference from a mirrored location using ICSP lookup/replace (with tags or digests). Therefor, the changes here are minimal (still have to do the replace when extracting mirrored releases, since there's no other way for oc to know that the internal referenced images (like tools, installer, oc) should be pulled from the same registry as the release imageRef (fetched from the alternative repo w/ lib-go blobMirrored client) **see the readReleaseImageReference fn** in release/image_mapper.go, the parsed user-passed imageRef (the o.From usually) must be known to apply the imageRef.ID of a component image to the desired registry.

Note that `oc` is no longer using any of the ICSP added features to registryclient, only added feature to `oc` is the switching of lookup from original ref/registry to mirrored ref/registry so `image-registry` will definitely need to verify that it's all working (however, I did verify it worked w/ the old approach, so I'm pretty confident w/ an ICSP client it will all be working well for them).

romfreiman:
@smarterclayton @sallyom how do we make a progress on this one? We're blocked with the central host management for 4.8 without it

sallyom:
@smarterclayton another issue I see w/ hiding the replacement imageRef from oc is you cannot do something like 
```
$ oc adm release info localhost:5000/openshift/release:4.7.5-x86_64  --image-for tools
quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:e59e69fa1539693867c2e4ec9c02d0d6091fac35fb6417a8b1e5abc87d326cae
```
(you can run it, but user will be confused from the output)

sallyom:
I've pushed an older working version of the branch since it came to my attn that some were using the branch until it merges. I probably broke it while incorporating the latest version of the registry client. It should be working now

smarterclayton:
> $ oc adm release info localhost:5000/openshift/release:4.7.5-x86_64  --image-for tools
> quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:e59e69fa1539693867c2e4ec9c02d0d6091fac35fb6417a8b1e5abc87d326cae

I think this is the correct behavior.  If someone wants to view that image (with `oc image info`) they have to pass an ICSP.

sallyom:
> > $ oc adm release info localhost:5000/openshift/release:4.7.5-x86_64  --image-for tools
> > quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:e59e69fa1539693867c2e4ec9c02d0d6091fac35fb6417a8b1e5abc87d326cae
> 
> I think this is the correct behavior. If someone wants to view that image (with `oc image info`) they have to pass an ICSP.

With this PR, oc can now access the mirrored location rather than only the original. No ICSP is necessary for this, as the user-provided location is simply added to the alternates list (ICSP would be necessary if you needed to `oc image info quay.io/openshift...@sha...` and expect to behind-the-scenes map that to and access `mirrored/location..@sha...`.That is why oc is no longer using the ICSP, only part oc needs is the simple addition of the user-given location to list of alternates (in the simple strategy lookup from registryclient). If the ICSP feature is desired, I can add that back, but that can be in a follow-up, since the source of the bugs are all that a user simply wants this to work:
`oc adm release extract mirrored/location...@sha.... or :tag`

soltysh (approved):

sallyom:
@smarterclayton @soltysh with the added *WithLocation fns in registryclient.client_mirrored, I added the ICSP function back, it is working. See the test cmds for examples: https://github.com/openshift/oc/pull/439#issuecomment-668228752

eranco74:
I've tried using the oc from this PR to extract the baremetal-installer from a nightly image (edge-01.edge.lab.eng.rdu2.redhat.com:5000/ocp:nightly-2021-03-16-221720)

Seems that without the changes from this PR, `oc` knows where to extract from:
``` Will extract usr/bin/openshift-install from quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:13714f2d3e2db23d44158bad03dcd017971c9674a9c5904b4700f31ff24b67e9```
According to the content of `release-manifests/image-references` in the release image.

But when I use the `oc` from this PR it tries to extract from the same registry as the release image:
``` Will extract usr/bin/openshift-install from registry.ci.openshift.org/ocp/release@sha256:13714f2d3e2db23d44158bad03dcd017971c9674a9c5904b4700f31ff24b67e9```
 and fail with this error:
```error: image "registry.ci.openshift.org/ocp/release@sha256:13714f2d3e2db23d44158bad03dcd017971c9674a9c5904b4700f31ff24b67e9" not found: manifest unknown: manifest unknown ```

romfreiman:
cc: @soltysh

soltysh:
This will be split into several PRs, first is https://github.com/openshift/oc/pull/829
/close

PR #442, opened 2020-05-28
adm/release/info: add `--list-bugs` flag

AlexNPavel:
Add a new `--list-bugs` flag for `oc adm release info` that returns a newline separated list of bug IDs based on git commit info. It does not verify that it can access the bugs, as opposed to `--bugs=... -o name`, which checks that it can access the bugs before returning. This prevents issues when the bugs list contains private bugs.

/cc @stevekuznetsov


pkg/cli/admin/release/info.go
```diff
@@ -397,12 +400,12 @@ func (o *InfoOptions) Validate() error {
...
-	if len(o.ChangelogDir) > 0 || len(o.BugsDir) > 0 {
+	if len(o.ChangelogDir) > 0 || len(o.BugsDir) > 0 || len(o.ListBugsDir) > 0 {
 		if len(o.From) == 0 {
-			return fmt.Errorf("--changelog/--bugs require --from")
+			return fmt.Errorf("--changelog/--bugs/--list-bugs require --from")
 		}
 	}
-	if len(o.ChangelogDir) > 0 && len(o.BugsDir) > 0 {
+	if (len(o.ChangelogDir) > 0 && len(o.BugsDir) > 0) || (len(o.ChangelogDir) > 0 && len(o.ListBugsDir) > 0) || (len(o.BugsDir) > 0 && len(o.ListBugsDir) > 0) {
 		return fmt.Errorf("--changelog and --bugs may not both be specified")
```
stevekuznetsov, line 409: If someone does `--changelog --list-bugs` this error message will be confusing

pkg/cli/admin/release/info.go
```diff
@@ -1563,6 +1567,35 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
 			bugIDs.Insert(commit.Bugs...)
 		}
 	}
+	return bugIDs, utilerrors.NewAggregate(errs)
+}
+
+func listBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string) error {
+	if diff.To.Digest == diff.From.Digest {
```
stevekuznetsov, line 1574: Since this is always done can we DRY in in `getBugSet()`?

pkg/cli/admin/release/info.go
```diff
@@ -1563,6 +1567,35 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+		return fmt.Errorf("releases are identical")
+	}
+	bugs, err := getBugSet(out, errOut, diff, dir)
+	if err != nil {
+		fmt.Fprintf(errOut, "error: %v\n", err)
+		return kcmdutil.ErrExit
+	}
+	for id := range bugs {
+		fmt.Fprintln(out, id)
```
stevekuznetsov, line 1583: This means on the consumer's end we will need to read a file line-by-line and parse integers, which seems less straightforward than if this tool output JSON or something. WDYT?

stevekuznetsov:
No, that is literally what we are trying not to do. We just want to list the bugs, we do NOT want any validation, and that `--output=name` explicitly does not do what we need.

soltysh:
OK, I spoke with Steve on slack, this is about skipping that verification of BZ. @AlexNPavel my proposal is to add `--skip-bug-check`, `--offline` or `--skip-check`. I think I'm leaning towards the first.

pkg/cli/admin/release/info.go
```diff
@@ -1564,6 +1572,13 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
 		}
 	}
 
+	if skipBugCheck {
```
soltysh, line 1575: I'd rather have the check inside `if !skipBugCheck` and leave the printing as is defined below. This introduces unnecessary code duplication .

AlexNPavel: Updated

PR #446, opened 2020-06-01
Bug 1806780: release: Mirroring from custom directory via oc adm release mirror fails

wking:
The command fails when specifying a non-default from directory because the from directory was not passed to either extract or mirror. Add that transformation.

Also include the Parallel options as good practice even if they are not heavily used in these flows.

Backporting e1567dd1f291bc2264c6423afa17a4084ee900ee (#424) and resolving a context conflict because 4.4 does not include 2725cfa7ce (#358).


wking:
I'll try again after #456 lands...

wking:
Manually backported to 4.3 with #458.

PR #449, opened 2020-06-02
make oc adm inspect run in parallel and remove rate limits

deads2k:
This allows `oc adm inspect co` to run 90 seconds for me locally against my aws cluster.

/assign @soltysh


deads2k:
just an import conflict.  re-lgtming

PR #452, opened 2020-06-04
bug 1843853: [release-4.2] Introduce MOTD for login command (backport from 4.3)

fmartingr:
Backport of #67 for 4.2
Bugzilla for reference [#1843853](https://bugzilla.redhat.com/show_bug.cgi?id=1843853)
---
Due to legal requirements, some deployments need to display a legal
notice for users "entering" a system.

This legal notice can range from telling the user that they are entering
a government system and are being monitored, to just displaying the
rules of usage in the system. We'll implement this requirements in the
form of a "message of the day" (MOTD).

This is required by law by several countries, with the intent that in
case the user attacks the system and needs to be prosecuted, the message
can be used as proof that the attacker knew against whom the attack was being
executed.

In this PR, the "oc login" command will search for the motd
ConfigMap, which would ideally be configured in the "openshift"
namespace.

If the notice is not present, or the user has no rights to access that
ConfigMap, the command will not fail.

Note that the legal notice will be displayed for the login command
regardless if the user is authenticated or not.

As a separate PR, the (empty) ConfigMap and the needed ClusterRole and
ClusterRoleBinding will be populated by default in the deployment.

### DIFF
--- a/pkg/cli/login/loginoptions.go
+++ b/pkg/cli/login/loginoptions.go
@@ -14,6 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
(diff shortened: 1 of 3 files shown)


fmartingr:
Since this is not going to be backported to 4.2 I'm closing the PR.

Users on 4.2 can use the 4.3 client to get the MOTD feature.

PR #459, opened 2020-06-08
bug 1844354: oc set probe: add support for startup probe

ingvagabund:
Startup probes were introduced with Kubernetes 1.16


ingvagabund:
/hold
It's actually a valid fail since origin tests for oc set probe expects a different error message when no probe flag is provided :-|

PR #460, opened 2020-06-08
Bug 1844998: Fix --save-config for kubectl create commands, where it was missing

soltysh:
/assign @sallyom

PR #461, opened 2020-06-08
Get rid of passing oc name to every command

soltysh:
This builds on top of https://github.com/openshift/oc/pull/460 so that other goes first. 
This is cleaning that I've been putting aside for quite some time, hopefully I didn't break anything :wink: 

/assign @sallyom


soltysh:
Rebased, tagging back manually to move forward.

petr-muller:
/hold

If the test is broken then there's no point in retesting. Either button merge it or `/override` after a failure is reported and then unhold (but Tide may decide to re-run the tests before merge so button merge is probably better).

PR #467, opened 2020-06-16
oc adm release extract - add option to extract credential requests only

csrwng:
The purpose of this feature is to enable the manual creation of credentials required by a particular Openshift release.

https://issues.redhat.com/browse/CO-972


pkg/cli/admin/release/extract.go
```diff
@@ -64,6 +79,10 @@ func NewExtract(f kcmdutil.Factory, parentName string, streams genericclioptions
 			signed by the key. For more advanced signing use the generated sha256sum.txt and an
 			external tool like gpg.
 
+			The --credentials-requests flag filters extracted manifests to only cloud credential
+			requests. The --cloud flag further filters credential requests to a specific cloud.
+			Valid values for --cloud include aws, gcp, azure, vsphere, openstack, ovirt, and vsphere.
```
dgoodwin, line 84: vsphere listed twice.

csrwng:
> I feel like an option to direct to stdout instead of files might be useful, wdyt?

@dgoodwin Actually thinking some about this a bit... maybe the default should be to output to stdout. Unlike a full extract, there's really not much of a benefit in having the original file names created. Especially if we're filtering stuff out of those files.

dgoodwin:
Personally stdout is what I would be looking for as well.

csrwng:
Changed to output to stdout, fixed description

PR #470, opened 2020-06-18
Add force-stdin flag to oc debug

juanluisvaladas:
oc debug discards the stdin when a command is present on the command
args. This behavior is usually good, however it prevents using oc debug
in a pipeline.

Example usage:
cat README.md | oc debug node/my-node \
                   --force-stdin \
                   -- bash -c 'cat > /host/tmp/README.md'


juanluisvaladas:
@mfojtik @deads2k could you PTAL?

PR #475, opened 2020-06-23
Replace custom printing with built-in printing in obseve commad

soltysh:
/assing @sallyom


soltysh:
This is waiting for https://github.com/openshift/origin/pull/25278
/assign @sallyom

PR #482, opened 2020-07-02
Bug 1853116: write to files instead of stdout

soltysh:
/assign @csrwng

### DIFF
--- a/pkg/cli/admin/release/extract.go
+++ b/pkg/cli/admin/release/extract.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -116,7 +117,7 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be copied under.")
 
 	flags.BoolVar(&o.CredentialsRequests, "credentials-requests", o.CredentialsRequests, "Extract credential request manifests only")
-	flags.StringVar(&o.Cloud, "cloud", o.Cloud, "Specify the cloud for which credential request manifests should be extracted.")
+	flags.StringVar(&o.Cloud, "cloud", o.Cloud, "Specify the cloud for which credential request manifests should be extracted. Works only in combination with --credentials-requests.")
 
 	flags.StringVarP(&o.Output, "output", "o", o.Output, "Output format. Supports 'commit' when used with '--git'.")
 	return cmd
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/release/extract.go
```diff
@@ -118,7 +118,7 @@ func NewExtract(f kcmdutil.Factory, parentName string, streams genericclioptions
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be copied under.")
 
 	flags.BoolVar(&o.CredentialsRequests, "credentials-requests", o.CredentialsRequests, "Extract credential request manifests only")
-	flags.StringVar(&o.Cloud, "cloud", o.Cloud, "Specify the cloud for which credential request manifests should be extracted.")
+	flags.StringVar(&o.Cloud, "cloud", o.Cloud, "Specify the cloud for which credential request manifests should be extracted. Works only in combination with --credentials-requests.")
```
csrwng, line 121: :+1:

pkg/cli/admin/release/extract.go
```diff
@@ -377,18 +377,25 @@ func (o *ExtractOptions) Run() error {
...
+			for _, m := range o.Manifests {
 				yamlBytes, err := yaml.JSONToYAML(m.Raw)
 				if err != nil {
 					return false, errors.Wrapf(err, "error serializing manifest in %s", hdr.Name)
 				}
-				fmt.Fprintf(o.Out, "---\n")
-				o.Out.Write(yamlBytes)
+				out := o.Out
+				if len(o.Directory) > 0 {
+					out, err = os.Create(path.Join(o.Directory, hdr.Name))
```
csrwng, line 389: The creation of the file should happen outside the loop of manifests. Otherwise for files that contain multiple manifests, you'll create the file multiple times, leaving it only with the last manifest.
Also s/`path.Join`/`filepath.Join`/ so windows clients work.
After creating a file outside of the loop, you also need to add  a `defer out.Close()`

soltysh:
@csrwng comment addressed, ptal

pkg/cli/admin/release/extract.go
```diff
@@ -377,18 +377,24 @@ func (o *ExtractOptions) Run() error {
...
+				o.Manifests = append(o.Manifests, m)
 			}
-			if len(credRequestManifests) == 0 {
-				return true, nil
+			out := o.Out
+			if len(o.Directory) > 0 {
+				out, err = os.Create(path.Join(o.Directory, hdr.Name))
+				if err != nil {
+					return false, errors.Wrapf(err, "error creating manifest in %s", hdr.Name)
+				}
```
csrwng, line 387: Hey @soltysh , looks great... just one more suggestion (use filepath instead of path, include original error in return error message, and defer closing of the file so it doesn't stay open throughout the run):
```suggestion
				file, err := os.Create(filepath.Join(o.Directory, hdr.Name))
				if err != nil {
					return false, errors.Wrapf(err, "error creating manifest %s: %v", hdr.Name, err)
				}
				defer file.Close()
				out = file
```

soltysh: @csrwng yeah you're right, I always forget about it. Should be all addressed. thx!

soltysh:
@csrwng ptal, this got rebased against master and ready for merge

wking:
/cherrypick release-4.6

I want this for openshift/release#24166

PR #487, opened 2020-07-08
inspect: improve the event filter page

mfojtik:
Improves the event filter HTML page (use bootstrap, bootstrap-table) and allow to pass a single `events.json` file to `oc adm inspect` and get the `events-filter.html` out, which is useful when inspect local files (collected either via CI artifacts or `oc get events -o json &> events.json`.

Sample page: https://mfojtik.keybase.pub/events.html


mfojtik:
missed updating completetion, fixed and lgtm reapplied

PR #493, opened 2020-07-13
 Bug 1850931: 'oc image mirror' Add warning when mirroring SchemaVersion:1 images

sallyom:
Add warning when mirroring SchemaVersion:1 images that digests are not preserved.

Also, update TODOs with functions handling schema conversions to drop support for Schema v1 in future release.
/assign @soltysh

### DIFF
--- a/pkg/cli/image/mirror/mirror.go
+++ b/pkg/cli/image/mirror/mirror.go
@@ -339,7 +339,7 @@ func (o *MirrorImageOptions) Run() error {
 								srcDigest := digest
 								tags := op.digestsToTags[srcDigest].List()
 								w.Parallel(func() {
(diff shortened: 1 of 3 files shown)


pkg/cli/image/mirror/mirror.go
```diff
@@ -735,6 +735,9 @@ func copyManifestToTags(
 		for _, desc := range srcManifest.References() {
 			plan.parent.parent.AssociateBlob(desc.Digest, plan.parent.name)
 		}
+		if srcDigest != toDigest {
+			fmt.Fprintf(out, "warning: Digests are not preserved with schema version 1 images. Support for schema version 1 images will be removed in a future release.\n")
```
soltysh, line 739: Wire `errout` and use that, from what I'm grepping through our commands we tend to warn through stderr, not stdout.

sallyom: fixed, done

PR #500, opened 2020-07-22
Prepare for RHEL 8

yselkowitz:


deads2k:
@crawford says that a rhel7 based image will work on a rhel8 host and that a rhel8 based image will work on a rhel7 host, so this is fine.

/approve

PR #509, opened 2020-07-29
Bug 1861559: Only schedule must-gather pod on Linux nodes

sebsoto:
oc adm must-gather is creating a pod which is getting scheduled on
Windows nodes. This commit changes the podspec to include a linux
node selector.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -520,6 +520,7 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 					},
 				},
 			},
+			NodeSelector:                  map[string]string{corev1.LabelOSStable: "linux"},
 			TerminationGracePeriodSeconds: &zero,
 			Tolerations: []corev1.Toleration{
 				{


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -520,6 +520,7 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 					},
 				},
 			},
+			NodeSelector:                  map[string]string{"kubernetes.io/os": "linux"},
```
tnozicka, line 523: Is there a preexisting place already using this approach?
I don't see this used in deployer controller - are DeploymentConfigs working with windows nodes?
@damemi @ingvagabund could scheduler know better and place pods referencing linux image only on linux nodes?

ingvagabund: Scheduler does not know anything about node architectures. The closest thing is to use node selector or taints to repel the pod from unwanted nodes. Taints are not suitable here since every pod would need to tolerate it.

Though, not sure if relying on presence of `kubernetes.io/os=linux` label is bullet proof (it always has to be set). Also, not every step in must-gather sequence is for Linux only. There are other parts as collecting namespaces, pods, etc. which are architecture independent.

sebsoto: OpenShift does not support running linux pods on Windows. We stop Linux pods from running on Windows nodes by adding a specific taint to the node which needs to be tolerated. Because this podspec tolerates all taints, this nodeSelector should be used.

> Though, not sure if relying on presence of kubernetes.io/os=linux label is bullet proof

You can see the label documented in [external docs](kubernetes.io/os), and where it is applied within the [kubelet](https://github.com/kubernetes/kubernetes/blob/c53ce48d48372c30053a9b67c6d1dee237b5cd69/pkg/kubelet/kubelet_node_status.go#L201).

> I don't see this used in deployer controller - are DeploymentConfigs working with windows nodes?

DeploymentConfigs are [not yet supported with Windows nodes](https://github.com/openshift/enhancements/blob/master/enhancements/windows-containers/windows-machine-config-operator.md#non-goals)

> Also, not every step in must-gather sequence is for Linux only.
Are you sure that the must-gather pod must run on a node to collect information from that specific node? It seems like node specific information is gathered via `oc adm node-logs` such as [here](https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_service_logs), and therefore if run on a linux node, it can gather information from any other node, as long as the node supports `oc adm node-logs`.

sebsoto: To be clear we have no intention for must-gather pods to run on Windows nodes for now. Adding the node-selector flag would not be useful for the purpose of fixing this bug

aravindhp: The reason the @openshift/openshift-team-windows-containers  is not going down the direction of running `must-gather` pods on Windows nodes is that it would require building Windows images. All Windows images contains a kernel and OpenShift / Red Hat has a policy of not publishing/shipping any software that has a 3rd party closed source kernel. Plus this would also require our CI / ART infrastructure to build Windows images.

sdodson:
/approve
This is in line with the design constraints that Windows nodes operate under today.

pkg/cli/admin/mustgather/mustgather.go
sdodson: I agree, this is in line with the design constraints that Windows nodes operate under today.

tnozicka: nit: 
```suggestion
			NodeSelector:                  map[string]string{corev1.LabelOSStable: "linux"},
```

tnozicka:
nit, otherwise lgtm. thanks

pkg/cli/admin/mustgather/mustgather.go
sebsoto: @tnozicka Good suggestion, made the change

PR #512, opened 2020-08-06
use filepath not path to manipulate filepaths

sallyom:
from go docs:
```
from "path" overview:
Package path implements utility routines for manipulating slash-separated paths.
The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the path/filepath package.
```
```
"path/filepath" overview:
Package filepath implements utility routines for manipulating filename paths in a way compatible with the target operating system-defined file paths.
The filepath package uses either forward slashes or backslashes, depending on the operating system. To process paths such as URLs that always use forward slashes regardless of the operating system, see the path package.
```

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -7,7 +7,7 @@ import (
 	"io"
 	"math/rand"
 	"os"
-	"path"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


sallyom:
@soltysh see this: https://bugzilla.redhat.com/show_bug.cgi?id=1807714#c10   🤷

sallyom:
verified in Windows env, oc sets mountPath correctly so closing this.  path instead of path/filepath is necessary to set forward slashes instead of backward slashes in container mountPath paths.

PR #513, opened 2020-08-08
Bug 1788156: With 'oc adm release new' allow a component image substitution w/ a multi-arch component image

sallyom:
This PR will allow:
* `oc adm release new --from-release anything --to anythingelse component=multi-arch-image` 
    * this currently fails bc `oc image extract` can't handle DeserializedManifestList with 
       `oc adm release new` flow

Note: `oc image extract|info|append` can handle multi-arch images, it defaults to extract system os/arch unless `--filter-by-os` is passed.  It doesn't, however, work with `--filter-by-os=.*` or with any expandable regex, as it can only handle a literal single filter (linux/amd64 not linux/* ). This PR updates logic around choosing a single manifest from a manifest list w/ `oc image append|info|extract`. 

See sample commands with improved output[ in comments below.](https://github.com/openshift/oc/pull/513#issuecomment-725061916)

Also resolves https://bugzilla.redhat.com/show_bug.cgi?id=1906276

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -993,6 +994,10 @@ func (o *NewOptions) extractManifests(is *imageapi.ImageStream, name string, met
 				var labels map[string]string
 				if imageConfig.Config != nil {
 					labels = imageConfig.Config.Labels
+					// when user provides an override, the platform spec of the release payload will be passed to extract and
+					// used to set the os/arch of the override image if it is a manifestList.  This is an unsupported release configuration.
+					opts.FilterOptions.FilterByOS = fmt.Sprintf("^%s/%s$", imageConfig.OS, imageConfig.Architecture)
+					opts.FilterOptions.OSFilter = regexp.MustCompile(opts.FilterOptions.FilterByOS)
 				}
 				if tag.Annotations == nil {
 					tag.Annotations = make(map[string]string)
--- a/pkg/cli/image/append/append.go
+++ b/pkg/cli/image/append/append.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/docker/distribution"
+	"github.com/docker/distribution/manifest/manifestlist"
 	"github.com/docker/distribution/manifest/schema2"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/distribution/registry/client"
@@ -30,6 +31,7 @@ import (
 	"github.com/openshift/library-go/pkg/image/dockerv1client"
 	"github.com/openshift/library-go/pkg/image/registryclient"
 	"github.com/openshift/oc/pkg/cli/image/imagesource"
+	imageinfo "github.com/openshift/oc/pkg/cli/image/info"
 	imagemanifest "github.com/openshift/oc/pkg/cli/image/manifest"
 	"github.com/openshift/oc/pkg/cli/image/workqueue"
 	"github.com/openshift/oc/pkg/helpers/image/dockerlayer"
@@ -276,6 +278,13 @@ func (o *AppendImageOptions) Run() error {
 		srcManifest, manifestLocation, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
 		if err != nil {
 			return fmt.Errorf("unable to read image %s: %v", from, err)
+		}
+		if _, ok := srcManifest.(*manifestlist.DeserializedManifestList); ok {
+			// If type is a ManifestList here, did not filter to single manifest from manifest list
+			// print os/arch options for this image as a convenience and error
+			if _, err = imageinfo.FilterManifestList(from.String(), srcManifest.(*manifestlist.DeserializedManifestList), nil, o.FilterOptions); err != nil {
+				return err
+			}
 		}
 		base, layers, err = imagemanifest.ManifestToImageConfig(ctx, srcManifest, repo.Blobs(ctx), manifestLocation)
 		if err != nil {
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -16,6 +16,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/docker/distribution"
+	"github.com/docker/distribution/manifest/manifestlist"
 	dockerarchive "github.com/docker/docker/pkg/archive"
 	digest "github.com/opencontainers/go-digest"
 
@@ -28,6 +29,7 @@ import (
 	"github.com/openshift/library-go/pkg/image/registryclient"
 	"github.com/openshift/oc/pkg/cli/image/archive"
 	"github.com/openshift/oc/pkg/cli/image/imagesource"
+	imageinfo "github.com/openshift/oc/pkg/cli/image/info"
 	imagemanifest "github.com/openshift/oc/pkg/cli/image/manifest"
 	"github.com/openshift/oc/pkg/cli/image/workqueue"
 )
@@ -313,7 +315,7 @@ func (o *ExtractOptions) Complete(cmd *cobra.Command, args []string) error {
 	}
 
 	if len(args) == 0 {
-		return fmt.Errorf("you must specify at least one image to extract as an argument")
+		return fmt.Errorf("you must specify an image to extract as an argument")
 	}
 
 	if len(o.Paths) == 0 && len(o.Files) == 0 {
@@ -325,6 +327,7 @@ func (o *ExtractOptions) Complete(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
 	return nil
 }
 
@@ -372,7 +375,13 @@ func (o *ExtractOptions) Run() error {
 					}
 					return fmt.Errorf("unable to read image %s: %v", from, err)
 				}
-
+				if _, ok := srcManifest.(*manifestlist.DeserializedManifestList); ok {
+					// If type is a ManifestList here, did not filter to single manifest from manifest list
+					// print os/arch options for this image as a convenience and error
+					if _, err = imageinfo.FilterManifestList(from.String(), srcManifest.(*manifestlist.DeserializedManifestList), nil, o.FilterOptions); err != nil {
+						return err
+					}
+				}
 				contentDigest, err := registryclient.ContentDigestForManifest(srcManifest, location.Manifest.Algorithm())
 				if err != nil {
 					return err
--- a/pkg/cli/image/info/info.go
+++ b/pkg/cli/image/info/info.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -99,6 +100,9 @@ func (o *InfoOptions) Complete(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("info expects at least one argument, an image pull spec")
 	}
 	o.Images = args
+	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -138,26 +142,7 @@ func (o *InfoOptions) Run() error {
 				FileDir:         o.FileDir,
 				SecurityOptions: o.SecurityOptions,
 				ManifestListCallback: func(from string, list *manifestlist.DeserializedManifestList, all map[digest.Digest]distribution.Manifest) (map[digest.Digest]distribution.Manifest, error) {
-					filtered := make(map[digest.Digest]distribution.Manifest)
-					for _, manifest := range list.Manifests {
-						if !o.FilterOptions.Include(&manifest, len(list.Manifests) > 1) {
-							klog.V(5).Infof("Skipping image for %#v from %s", manifest.Platform, from)
-							continue
-						}
-						filtered[manifest.Digest] = all[manifest.Digest]
-					}
-					if len(filtered) == 1 {
-						return filtered, nil
-					}
-
-					buf := &bytes.Buffer{}
-					w := tabwriter.NewWriter(buf, 0, 0, 1, ' ', 0)
-					fmt.Fprintf(w, "  OS\tDIGEST\n")
-					for _, manifest := range list.Manifests {
-						fmt.Fprintf(w, "  %s\t%s\n", imagemanifest.PlatformSpecString(manifest.Platform), manifest.Digest)
-					}
-					w.Flush()
-					return nil, fmt.Errorf("the image is a manifest list and contains multiple images - use --filter-by-os to select from:\n\n%s\n", buf.String())
+					return FilterManifestList(from, list, all, o.FilterOptions)
 				},
 
 				ImageMetadataCallback: func(from string, i *Image, err error) error {
@@ -458,4 +443,33 @@ func (o *ImageRetriever) Images(ctx context.Context, refs map[string]imagesource
 			})
 		}
 	})
+}
+
+// FilterManifestList returns map of digest to manifest, or a list of available platforms upon error,
+// to distinguish errors between multiple manifests found and no manifests found
+func FilterManifestList(from string, list *manifestlist.DeserializedManifestList, all map[digest.Digest]distribution.Manifest, filterOptions imagemanifest.FilterOptions) (map[digest.Digest]distribution.Manifest, error) {
+	filtered := make(map[digest.Digest]distribution.Manifest)
+	for _, manifest := range list.Manifests {
+		if !filterOptions.Include(&manifest, len(list.Manifests) > 1) {
+			klog.V(5).Infof("Skipping image for %#v from %s", manifest.Platform, from)
+			continue
+		}
+		filtered[manifest.Digest] = all[manifest.Digest]
+	}
+	if len(filtered) == 1 {
+		return filtered, nil
+	}
+	msg := fmt.Sprintf("no manifests retrieved that match filter %s", filterOptions.FilterByOS)
+	if len(filtered) > 1 {
+		msg = fmt.Sprintf("multiple manifests that match filter %s", filterOptions.FilterByOS)
+	}
+
+	buf := &bytes.Buffer{}
+	w := tabwriter.NewWriter(buf, 0, 0, 1, ' ', 0)
+	fmt.Fprintf(w, "  OS\tDIGEST\n")
+	for _, manifest := range list.Manifests {
+		fmt.Fprintf(w, "  %s\t%s\n", imagemanifest.PlatformSpecString(manifest.Platform), manifest.Digest)
+	}
+	w.Flush()
+	return nil, fmt.Errorf("%s\nThe image is a manifest list - use --filter-by-os to select only a single option from below\nIf the --filter-by-os option you passed is listed below, try again with '^os/arch$' to avoid multiple manifests matching with regex expansion:\n\n%s\nor, to default to system '%s/%s' don't pass --filter-by-os.\n\n", msg, buf.String(), runtime.GOOS, runtime.GOARCH)
 }
--- a/pkg/cli/image/manifest/manifest.go
+++ b/pkg/cli/image/manifest/manifest.go
@@ -282,23 +282,21 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
 	if err != nil {
 		return nil, ManifestLocation{}, err
 	}
-
 	originalSrcDigest := srcDigest
 	srcManifests, srcManifest, srcDigest, err := ProcessManifestList(ctx, srcDigest, srcManifest, manifests, from, filterFn, false)
 	if err != nil {
 		return nil, ManifestLocation{}, err
 	}
 	if len(srcManifests) == 0 {
-		return nil, ManifestLocation{}, fmt.Errorf("filtered all images from manifest list")
+		return nil, ManifestLocation{}, fmt.Errorf("filtered all images from list and no matching manifests found")
 	}
-
 	if srcDigest != originalSrcDigest {
 		return srcManifest, ManifestLocation{Manifest: srcDigest, ManifestList: originalSrcDigest}, nil
 	}
 	return srcManifest, ManifestLocation{Manifest: srcDigest}, nil
 }
 
-// ManifestToImageConfig takes an image manifest and converts it into a structured object.
+// ManifestToImageConfig takes an image manifest and converts it into a DockerImageConfig and digest Descriptors.
 func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifest, blobs distribution.BlobService, location ManifestLocation) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, error) {
 	switch t := srcManifest.(type) {
 	case *schema2.DeserializedManifest:
@@ -316,13 +314,12 @@ func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifes
 		}
 
 		base := config
-		layers := t.Layers
 		base.Size = 0
 		for _, layer := range t.Layers {
 			base.Size += layer.Size
 		}
 
-		return base, layers, nil
+		return base, t.Layers, nil
 
 	case *ocischema.DeserializedManifest:
 		if t.Config.MediaType != imagespecv1.MediaTypeImageConfig {
@@ -439,7 +436,6 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
 			if err != nil {
 				return nil, nil, "", err
 			}
-			klog.Warningf("Chose %s/%s manifest from the manifest list.", t.Manifests[0].Platform.OS, t.Manifests[0].Platform.Architecture)
 			return srcManifests, srcManifests[0], manifestDigest, nil
 		default:
 			return append(srcManifests, manifestList), manifestList, manifestDigest, nil
--- a/pkg/cli/image/mirror/mirror.go
+++ b/pkg/cli/image/mirror/mirror.go
@@ -514,7 +514,6 @@ func (o *MirrorImageOptions) plan() (*plan, error) {
 			})
 
 			canonicalFrom := srcRepo.Named()
-
 			rq.Queue(func(w workqueue.Work) {
 				for key := range src.digests {
 					srcDigestString, pushTargets := key, src.digests[key]


pkg/cli/image/extract/extract.go
```diff
@@ -357,7 +360,13 @@ func (o *ExtractOptions) Run() error {
...
 				}
 
-				imageConfig, layers, err := imagemanifest.ManifestToImageConfig(ctx, srcManifest, repo.Blobs(ctx), location)
+				manifests, err := repo.Manifests(ctx)
+				if err != nil {
+					return err
+				}
+
+				// If empty, set platformSpec to match other mappings
+				imageConfig, layers, platformSpec, err = imagemanifest.ManifestToImageConfig(ctx, srcManifest, repo.Blobs(ctx), manifests, location, true, platformSpec)
```
sallyom, line 369: TODO: have to wire in FilterByOS if `oc image extract` should work w/ --filter-by-os.  Set platformSpec to that, and then this PR will fix both cmds.

pkg/cli/image/manifest/manifest.go
```diff
@@ -276,47 +276,84 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
...
+func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifest, blobs distribution.BlobService, manifests distribution.ManifestService, location ManifestLocation, setPlatform bool, platformSpec string) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, string, error) {
+	switch t := srcManifest.(type) {
+	case *manifestlist.DeserializedManifestList:
+		// Only way here is with 'oc adm release new' flow
+		// In case of 'oc adm release new' with a component image substituted and that component image is multi-arch,
+		// the platformSpec will be populated from the base release image passed, since release payload is a separate image for each os/arch.
+		// If not in extract flow, this will be false
+		if !setPlatform {
+			klog.V(2).Infof("Should not have gotten here. Should only be here if substituting component image with multi-arch image in 'oc adm release new'.")
+			return nil, nil, "", fmt.Errorf("unable to use image manifest of type %T from %s", srcManifest, location)
```
sallyom, line 288: This will go away, still working on it

pkg/cli/image/manifest/manifest.go
```diff
@@ -327,15 +357,16 @@ func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifes
 				Digest:    t.FSLayers[i].BlobSum,
 				// size must be reconstructed from the blobs
 			}
+			platformSpec = fmt.Sprintf("linux/%s", t.Architecture)
```
sallyom, line 360: not sure about this

pkg/cli/image/extract/extract.go
```diff
@@ -357,7 +370,21 @@ func (o *ExtractOptions) Run() error {
...
+				manifests, err := repo.Manifests(ctx)
+				if err != nil {
+					return err
+				}
+
+				if len(o.FilterOptions.FilterByOS) > 0 {
+					// platformSpec will only be used to set ImageConfig if srcManifest is DeserializedManifestList
+					// If set here and is a ManifestList, user passed a bad os/arch - pass it along so user sees the FilterByOS in the error.
+					// Won't get here with 'oc adm release new' flow because the FilterByOS flag not exposed.
+					platformSpec = o.FilterOptions.FilterByOS
```
sallyom, line 382: instead, could error here if FilterByOS && is manifest list

pkg/cli/image/manifest/manifest.go
```diff
@@ -327,15 +358,16 @@ func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifes
...
```
sallyom, line 361: this i don't think is necessary but have to check

pkg/cli/image/manifest/manifest.go
```diff
@@ -276,47 +276,78 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
 }
 
 // ManifestToImageConfig takes an image manifest and converts it into a structured object.
-func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifest, blobs distribution.BlobService, location ManifestLocation) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, error) {
+func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifest, blobs distribution.BlobService, manifests distribution.ManifestService, location ManifestLocation, platformSpec string) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, string, error) {
```
smarterclayton, line 279: You need to document using variable names in the signature what the new return value actually is.  Also improve the godoc.

smarterclayton: The godoc for instance needs to describe what this method does with its inputs.

pkg/cli/image/manifest/manifest.go
```diff
@@ -276,47 +276,78 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
...
+	switch t := srcManifest.(type) {
+	case *manifestlist.DeserializedManifestList:
+		// In case of 'oc adm release new' with a component image substituted and that component image is multi-arch,
```
smarterclayton, line 282: This comment is wierdly specific to a oc, you really need to describe in the godoc how ManifestToImageConfig handles manifest lists.  And passing platformSpec in here is kind of wierd - platform spec is a high level concern of the `oc` command, not of low level code like this utility method.  You probably need to provide something like a handler function that, if specified, is passed the manifest list and tells you which image to select (something like `listSelectionFn func(DeserializedManifestList) (Manifest, error)` which returns a manifest, nil, or an error).  

Also, you should only fetch the child manifest AFTER you filter, not before (this will load O(N) manifests, instead of O(1) which it should be).

pkg/cli/image/manifest/manifest.go
sallyom: updated

pkg/cli/image/manifest/manifest.go
sallyom: restructured, updated the godoc, no need to return the platformSpec there as it is the DockerImageConfig.OS/Architecture already returned.  

I'm not sure what you mean w/ fetching the child manifest after filtering. The filtering [here](https://github.com/openshift/oc/blob/master/pkg/cli/image/extract/extract.go#L366) [to filter operators](https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/new.go#L978)?

pkg/cli/image/extract/extract.go
```diff
@@ -311,6 +316,9 @@ func (o *ExtractOptions) Validate() error {
 	if len(o.Mappings) == 0 {
 		return fmt.Errorf("you must specify one or more paths or files")
 	}
+	if o.FilterOptions.IsWildcardFilter() {
+		return fmt.Errorf("you must specify a single os/arch with --filter-by-os=os/arch, wildcards cannot be used with extract")
```
smarterclayton, line 320: That's not really true (if wildcards filter to a single image it's fine), and we don't do this in other contexts like image info.  I don't think this error should be here.

pkg/cli/image/extract/extract.go
```diff
@@ -357,7 +371,29 @@ func (o *ExtractOptions) Run() error {
...
 				}
 
-				imageConfig, layers, err := imagemanifest.ManifestToImageConfig(ctx, srcManifest, repo.Blobs(ctx), location)
+				manifests, err := repo.Manifests(ctx)
+				if err != nil {
+					return err
+				}
+
+				// If empty, platformSpec will be set to match other mappings
+				// platformSpec will be used to determine which manifest from a DeserializedManifestList should be chosen, when appropriate.
```
smarterclayton, line 380: I'm confused - why is this being defaulted in extract?  If this is about oc adm release new, you need to calculate what platform you are looking for in oc adm release new, then pass down the more specific filter to all the sub commands (not interpret the logic here in extract). 

Only oc adm release new knows about releases, so that's the only code that can translate "release image + initial filterByOS" into "release image + fully qualified filterByOS"

pkg/cli/image/extract/extract.go
sallyom: with `quay.io/prometheus/prometheus:v2.18.1` as an example:
`oc image info` gives a nice error msg with the list of os/arch choices available for the image.  `oc image extract` gives the confusing: `error: unable to parse image quay.io/prometheus/prometheus:v2.18.1: unknown image manifest of type *manifestlist.DeserializedManifestList`

sallyom: er, that ugly error msg is only produced if a user passes the wildcard filter-by-os, see here:
```console
$ oc image extract quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=foo
error: unable to read image quay.io/prometheus/prometheus:v2.18.1: filtered all images from manifest list
$ oc image extract quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=./*
error: unable to parse image quay.io/prometheus/prometheus:v2.18.1: unknown image manifest of type *manifestlist.DeserializedManifestList from manifest sha256:5880ec936055fad18ccee798d2a63f64ed85bd28e8e0af17c6923a090b686c3d
```
and also, if no filter-by-os passed then ProcessManifestList does its thing.  There's a slight room for improvement here, but yea, my original plan of erroring w/ wildcard filter is wrong, i'll update that.

pkg/cli/image/manifest/manifest.go
sallyom: updated

pkg/cli/image/extract/extract.go
sallyom: updated. PlatformSpec is set in release/new and passed to extract, to be used to inform user w/ better error msg when passing an unavailable os/arch like --filter-by-os=foo/foo  or if --filter-by-os=.* but manifests in manifestLIst >1, then print the ImageRetriever error msg that tells user to choose an os/arch.

pkg/cli/image/extract/extract.go
sallyom: the platformSpec is found in `oc adm release new` but then has to be passed to extract->manifest.go bc that's where all the filtering/choosing happens.  Also, if _not_ in `oc adm release new`  it's still useful to pass the 'filterByOS' from a user to get better error msg if the given os/arch can't be matched to an available manifest in a manifestLIst.

pkg/cli/image/manifest/manifest.go
```diff
@@ -284,7 +284,7 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
 		return nil, ManifestLocation{}, err
 	}
 	if len(srcManifests) == 0 {
-		return nil, ManifestLocation{}, fmt.Errorf("filtered all images from manifest list")
+		return nil, ManifestLocation{}, fmt.Errorf("filtered all images from manifest list and did not retrieve a valid source")
```
dmage, line 287: `manifest list does not have images that match filtering criteria`?

pkg/cli/image/manifest/manifest.go
```diff
@@ -293,8 +293,34 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
...
-// ManifestToImageConfig takes an image manifest and converts it into a structured object.
-func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifest, blobs distribution.BlobService, location ManifestLocation) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, error) {
+// ManifestToImageConfig takes an image manifest and converts it into a DockerImageConfig and digest Descriptors.
+// If srcManifest is a DeserializedManifestList, the platformSpec determines which manifest (os/arch) will be used.
+func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifest, blobs distribution.BlobService, manifests distribution.ManifestService, location ManifestLocation, platformSpec string) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, error) {
+	switch t := srcManifest.(type) {
+	case *manifestlist.DeserializedManifestList:
+		matched := false
+		for _, manifest := range t.Manifests {
+			childManifest, err := manifests.Get(ctx, manifest.Digest, distribution.WithManifestMediaTypes([]string{manifestlist.MediaTypeManifestList, schema2.MediaTypeManifest}))
```
dmage, line 303: `manifests.Get` is a relatively expensive operation. The result will be thrown away if `manifest.Platform` doesn't match platformSpec.

dmage (commented):
@sallyom I'm out of context, can you briefly explain the difference between PlatfromSpec and FilterOptions.FilterByOS?

sallyom:
sample `oc image extract` output:
```console
$ oc image extract quay.io/prometheus/prometheus:v2.14.0 --filter-by-os=.*
error: the image is a manifest list and contains multiple images - use --filter-by-os to select from:

  OS          DIGEST
  linux/amd64 sha256:3d53ce329b25cc0c1bfc4c03be0496022d81335942e9e0518ded6d50a5e6c638
  linux/arm64 sha256:ed001f5f670df3eb750f482db6dbd3a4d0561644c1c943375c381271e9d41b71
  linux/arm   sha256:888bcfa3886da8290f79e8ba3bcc1cb3a7b20d50a3363117899f9ade66593592

or, to default to system 'linux/amd64' don't pass --filter-by-os.

$ oc image extract quay.io/prometheus/prometheus:v2.14.0 --filter-by-os=foo/foo
error: unable to read image quay.io/prometheus/prometheus:v2.14.0: for platform foo/foo: filtered all images from manifest list and did not retrieve image that matches filtering criteria
```

pkg/cli/image/extract/extract.go
```diff
@@ -357,7 +369,51 @@ func (o *ExtractOptions) Run() error {
...
+									return err
+								}
+							}
+							// Setting PlatformSpec here enables better error message if given --filter-by-os os/arch does not match an
+							// available os/arch when srcManifest is a manifest list and imagemanifest.ManifestToImageConfig errors.
+							o.PlatformSpec = o.FilterOptions.FilterByOS
+						}
+					}
+				}
+				imageConfig, layers, err := imagemanifest.ManifestToImageConfig(ctx, srcManifest, repo.Blobs(ctx), manifests, location, o.PlatformSpec)
```
dmage, line 416: @sallyom so what I see there:
* imagemanifest.FirstManifest filters manifest lists and returns `srcManifest`: either a manifest for the selected platform or a manifest list
* if PlatformSpec is not set and FilterOptions is a wildcard filter and `srcManifest` is a manifest list:
    * if `srcManifest` has more than one manifest, print an error and exit
    * otherwise set PlatformSpec equal to FilterByOS
* if we still have a manifest list, ManifestToImageConfig is going to filter it again, but using PlatformSpec

Can you incorporate filtering by PlatformSpec into FirstManifest? I think you can update FirstManifest and rename it into, let's say, PlatformManifest. It's called from two places and both places don't want it to return a manifest list.

sallyom: I will try, thanks!

pkg/cli/image/manifest/manifest.go
sallyom: updated, moved this all to FirstManifest

pkg/cli/image/extract/extract.go
sallyom: ptal, updated, thanks

sallyom:
sample commands to test updated `oc image extract` with a multi-arch image:
```console
$ oc image extract --filter-by-os=linux/arm quay.io/prometheus/prometheus:v2.18.1

$ oc image extract --filter-by-os=.* quay.io/prometheus/prometheus:v2.18.1
```
sample cmd to test updated `oc amd release new` with multi-arch image substitution:
```console
$ oc adm release new --from-release registry.svc.ci.openshift.org/ocp/release:4.7.0-0.ci-2020-11-10-053716 --to-image localhost:5000/substitute/release prometheus=quay.io/prometheus/prometheus:v2.18.1
```

sallyom:
@smarterclayton @dmage ptal, I've addressed feedback, anything else? thanks

pkg/cli/image/manifest/manifest.go
dmage: The manifest `childManifest` is still downloaded before checking that it is needed.

sallyom: oh right, thanks, updated, ptal

pkg/cli/image/manifest/manifest.go
sallyom: updated the msg

pkg/cli/admin/release/new.go
```diff
@@ -979,6 +979,9 @@ func (o *NewOptions) extractManifests(is *imageapi.ImageStream, name string, met
 				var labels map[string]string
 				if imageConfig.Config != nil {
 					labels = imageConfig.Config.Labels
+					// when user provides an override, the platform spec of the release payload will be passed to extract and
+					// used to set the os/arch of the override image if it is a manifestList.  This is an unsupported release configuration.
+					opts.PlatformSpec = fmt.Sprintf("%s/%s", imageConfig.OS, imageConfig.Architecture)
```
smarterclayton, line 984: Why isn't this just setting FilterOptions on the extract command?

sallyom: right, i've updated, ptal thanks

pkg/cli/image/extract/extract.go
```diff
@@ -339,7 +347,7 @@ func (o *ExtractOptions) Run() error {
 					return fmt.Errorf("unable to connect to image repository %s: %v", from.String(), err)
 				}
 
-				srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
+				srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include, o.FilterOptions.FilterByOS, o.FilterOptions.IsWildcardFilter())
```
soltysh, line 346: how about passing entire `o.FilterOptions`?

sallyom: updated, FirstManifest is now a FilterOptions method

pkg/cli/image/manifest/manifest.go
```diff
@@ -372,6 +381,13 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
...
 			klog.V(5).Infof("Including image %s for %#v from %s", manifest.Digest, manifest.Platform, ref)
 			filtered = append(filtered, manifest)
+			if PlatformSpecString(manifest.Platform) == platformSpec {
+				// If exact match, there's no regex expansion (no '*' in FilterByOS), so clear any filtered that may have
+				// been added already, since in this case we only want the exact match.
+				klog.V(5).Infof("Exact match found for %s, only including image %s from %s", platformSpec, manifest.Digest, ref)
+				filtered = []manifestlist.ManifestDescriptor{manifest}
+				break
+			}
 		}
```
sallyom, line 391: Looking at ~1st 10 lines of output with this command:
`oc image extract --filter-by-os=linux/arm quay.io/prometheus/prometheus:v2.18.1 -v=5`  
the linux/arm64, then linux/arm - that's good for cmds (like mirror) that allow ' * ' in --filter-by-os=linux/* but always want to only return the single one if an exact match was found, regardless of what's allowed via filterFn

pkg/cli/image/manifest/manifest.go
```diff
@@ -279,21 +280,30 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
...
 	if err != nil {
 		return nil, ManifestLocation{}, err
 	}
-	if len(srcManifests) == 0 {
-		return nil, ManifestLocation{}, fmt.Errorf("filtered all images from manifest list")
+	// Should only ever return 1 manifest from here
+	m := len(srcManifests)
+	switch {
+	case m > 1:
+		return nil, ManifestLocation{}, fmt.Errorf("multiple manifests match platform %s, failed to choose single manifest", o.FilterByOS)
```
sallyom, line 296: with the fix below of removing dupes from filtered if/only if filterByOS == manifest.Platform, i'm not sure the returning multiple manifests is possible anymore but in case, better to differentiate. It would be if an image had duplicate digests for same platform os/arch (don't know if there's anything preventing that w/ image builds)

soltysh:
@sallyom this needs a rebase

soltysh (approved):
/approve
leaving final tag to @smarterclayton

pkg/cli/image/manifest/manifest.go
smarterclayton: I don't think this method should at all be aware of filterSpec.  If you only want to match exact, the filterFn should implement exact matching.  Are there examples of manifest lists that have 2 linux/amd64 images?  In those cases the user needs to tell us which one to use, you can't just exit early.

pkg/cli/image/manifest/manifest.go
```diff
@@ -259,7 +259,8 @@ func (m ManifestLocation) String() string {
 }
 
 // FirstManifest returns the first manifest at the request location that matches the filter function.
-func FirstManifest(ctx context.Context, from imagereference.DockerImageReference, repo distribution.Repository, filterFn FilterFunc) (distribution.Manifest, ManifestLocation, error) {
+// If srcManifest is a DeserializedManifestList, the os/arch filter determines which manifest will be returned.
+func (o *FilterOptions) FirstManifest(ctx context.Context, from imagereference.DockerImageReference, repo distribution.Repository) (distribution.Manifest, ManifestLocation, error) {
```
smarterclayton, line 263: See my other comment, I don't think this is a valid change because filterFn can't just arbitrarily hide duplicates with early exit.

pkg/cli/image/manifest/manifest.go
sallyom: without this change of insisting on only a single manifest returned, when a user runs this:
`oc image extract --filter-by-os=linux/arm quay.io/prometheus/prometheus:v2.18.1 -v=5`  extract will error because it can only extract a single manifest and 2 manifests pass filterFn (arm, arm64 are available w/ that image) _note: that fails today because extract can't take a ManifestList at all, so you won't see this issue unless you're testing w/ this PR_ - you want multiple returned w/ mirror and regex expansion , but if a user passed `linux/arm` to extract & append you _only_ want that exact match - since the command will error if you get `linux/arm` + `linux/arm64` (such as w/ this prometheus image: `quay.io/prometheus/prometheus:v2.18.1`)  Here's the output of the cmd:
```console
$ mkdir ext; cd ext; oc image extract --filter-by-os=linux/arm quay.io/prometheus/prometheus:v2.18.1 -v=5
I0120 16:38:59.320392 4083752 config.go:128] looking for config.json at /home/somalley/.docker/config.json
I0120 16:38:59.320540 4083752 config.go:136] found valid config.json at /home/somalley/.docker/config.json
I0120 16:38:59.470831 4083752 credentials.go:108] Found secret to match https://quay.io/v2/auth (quay.io/auth): 
I0120 16:38:59.938937 4083752 manifest.go:405] Skipping image sha256:b4e6cd0275a26750505e539f8528e891053434ebd3972be02645bed5f02f0795 for manifestlist.PlatformSpec{Architecture:"amd64", OS:"linux", OSVersion:"", OSFeatures:[]string(nil), Variant:"", Features:[]string(nil)} from quay.io/prometheus/prometheus:v2.18.1
I0120 16:38:59.938975 4083752 manifest.go:408] Including image sha256:33cd2bbeea5c52add05be111fdccd54d289aa609f9059b22e9476865fd22b5b5 for manifestlist.PlatformSpec{Architecture:"arm64", OS:"linux", OSVersion:"", OSFeatures:[]string(nil), Variant:"", Features:[]string(nil)} from quay.io/prometheus/prometheus:v2.18.1
I0120 16:38:59.938993 4083752 manifest.go:408] Including image sha256:19bce37348f8e9fc652a0a562cad12a85abd3bde2c5ed905b36df24efec30b48 for manifestlist.PlatformSpec{Architecture:"arm", OS:"linux", OSVersion:"", OSFeatures:[]string(nil), Variant:"", Features:[]string(nil)} from quay.io/prometheus/prometheus:v2.18.1
I0120 16:38:59.939008 4083752 manifest.go:413] Exact match found for linux/arm, only including image sha256:19bce37348f8e9fc652a0a562cad12a85abd3bde2c5ed905b36df24efec30b48 from quay.io/prometheus/prometheus:v2.18.1
I0120 16:38:59.939069 4083752 manifest.go:439] Filtered manifest list to new digest sha256:f2e73eb62077979c29bc47242defbfa6875fdeee249a49bcf6067fd0ed09dead:
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2824,
         "digest": "sha256:19bce37348f8e9fc652a0a562cad12a85abd3bde2c5ed905b36df24efec30b48",
         "platform": {
            "architecture": "arm",
            "os": "linux"
         }
      }
   ]
}
W0120 16:39:00.000542 4083752 manifest.go:456] Chose linux/arm manifest from the manifest list.
```
You can see that 2 possilble manifests were included, and we want that if the user passed --filter-by-os=linux/ar* to a mirror cmd, but if user passed `--filter-by-os=linux/arm` only the exact match should be chosen (otherwise extract &append fail, bc they only can handle a single manifest)

sallyom: > the user needs to tell us which one to use, you can't just exit early.

Right, so if the user _does_ tell us which one to use (an exact match is found) then the command should not fail bc it found both `linux/arm` and `linux/arm64`

soltysh: @sallyom we clearly state that `--filter-by-os` is a regexp, if we pass `linux/arm` and we match arm and arm64 then the bug is in the filter. If you're not passing `linux/arm.*` it should be equivalent with `^linux/arm$`, is that right @smarterclayton

dmage: I'd like to note that it's perfectly valid to have two images for `linux/arm`:

```console
$ boater get-manifest -a nginx | jq .manifests[].platform -c
{"architecture":"amd64","os":"linux"}
{"architecture":"arm","os":"linux","variant":"v5"}
{"architecture":"arm","os":"linux","variant":"v7"}
{"architecture":"arm64","os":"linux","variant":"v8"}
{"architecture":"386","os":"linux"}
{"architecture":"mips64le","os":"linux"}
{"architecture":"ppc64le","os":"linux"}
{"architecture":"s390x","os":"linux"}
```

or for `windows/amd64`:

```console
$ boater get-manifest -a mongo | jq .manifests[].platform -c
{"architecture":"amd64","os":"linux"}
{"architecture":"arm64","os":"linux","variant":"v8"}
{"architecture":"s390x","os":"linux"}
{"architecture":"amd64","os":"windows","os.version":"10.0.17763.1697"}
{"architecture":"amd64","os":"windows","os.version":"10.0.14393.4169"}
```

sallyom: yes, I meant 2 digests for same platform, the duplicate linux/arm have different variants

sallyom: Here's correct output for extracting multi-arch w/ same os/arch, different variants:
```console
$ oc image extract --filter-by-os=linux/arm docker.io/library/nginx:latest --confirm
error: unable to read image docker.io/library/nginx:latest: filtered all images from manifest list and did not retrieve image that matches platform linux/arm
$ oc image extract --filter-by-os=linux/arm/v5 docker.io/library/nginx:latest --confirm
W0121 13:23:25.441694 4152421 manifest.go:465] Chose linux/arm/v5 manifest from the manifest list.
```

sallyom: I've pushed a new filter function, `IncludeSingle` that filters for exact match only, and passed to FirstManifest for extract & append @smarterclayton @soltysh ptal

pkg/cli/image/manifest/manifest.go
sallyom: still not sure an image can't have multiple platforms per digest, so leaving this check here since otherwise the error will be misleading.

pkg/cli/image/manifest/manifest.go
sallyom: I've pushed a new filter func, "IncludeSingle" that returns true only w/ exact match. ptal

pkg/cli/image/manifest/manifest.go
sallyom: ^^ oops forgot to remove that code that is no longer needed... removing/pushing now...

smarterclayton: > @sallyom we clearly state that --filter-by-os is a regexp, if we pass linux/arm and we match arm and arm64 then the bug is in the filter. If you're not passing linux/arm.* it should be equivalent with ^linux/arm$, is that right @smarterclayton

This is what I meant.  If you want a single image, pass a --filter-by-os that is exact.

Then oleg's point is about indicating that the point of the processing is to get a single element out, and it depends on whether the use case wants "all images matching the filter" or "single image matching the filter", but that's a multiple choice option on process that I would expect to get passed in to `ProcessManifestList`.  I.e. I want to process this manifest and either get all that match, get the first that matches, or fail if there are multiple matches and return an error so that the appropriate command can do the right thing. I would expect `FirstManifest` on `--filter-by-os=^linux/arm$` to already do the right thing (if we didn't need to fail in multiple matches, which would depend on the current behavior of the command).  In this case, in this command, `FirstManifest` with a correctly specified regex would return the first of duplicates and that's ok.

`release new --filter-by-os=X` should assume that X is an exact match, and so it should already be converting the value passed as X to `^X$` (i guess for backcompat you should probably see whether it's a regex or a constant, and if it's a constant you should be doing the transform, and if it's not a constant you should have the whole thing fail).  If we do that, do we need these other changes?

smarterclayton:
/hold

for the long comment asking why the scope of change is necessary if we just interpret --filter-by-os is necessary

pkg/cli/image/manifest/manifest.go
sallyom: `oc adm release new` does not expose `--filter-by-os` nor should it. The only time it needs this is when a user passes a multi-arch image to substitute (devs only do this) 
like w/ this: 
`oc adm release new --from-release registry.ci.openshift.org/ocp/release:4.7.0-0.nightly-2021-01-18-180737 --to-image localhost:5000/sub-release prometheus=quay.io/prometheus/prometheus:v2.18.1` that fails w/  `unable to parse manifestList` error

sallyom: and yes, with `oc image extract` `oc image append` you have to pass the exact match, or you'll get an error - but if you pass a regex that matches more than a single, you'll get a confusing error, see here:
**output from master**
```console
$ mkdir ext; cd ext; oc image extract quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=linux/arm --confirm 
error: unable to parse image quay.io/prometheus/prometheus:v2.18.1: unknown image manifest of type *manifestlist.DeserializedManifestList from manifest sha256:5880ec936055fad18ccee798d2a63f64ed85bd28e8e0af17c6923a090b686c3d
```
the above error is wrong, it's erroring bc it found multiple that match, not because it's not handling the manifestList (linux/arm and linux/arm64)
**and output from this PR**:
```console
$ mkdir ext; cd ext; oc image extract quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=linux/arm --confirm 
W0121 16:55:15.901894 4171937 manifest.go:457] Chose linux/arm manifest from the manifest list.
```

sallyom: In these instances, filter-by-os has to be considered a string, not a regex because the cmds can only handle a single and it's not enough to pass exactly what you want `linux/arm` bc that will not always yield a single w/ regex.MatchString.

soltysh: > get the first that matches, or fail if there are multiple matches and return an error
@smarterclayton I'd personally be against match first, b/c it will lead to different results each time, no matter where. I'd rather have always an explicit choice, either through expanding the filter with digest or something else.

pkg/cli/image/manifest/manifest.go
```diff
@@ -191,6 +191,14 @@ func (o *FilterOptions) Include(d *manifestlist.ManifestDescriptor, hasMultiple
 	return o.OSFilter.MatchString(s)
 }
 
+// IncludeSingle returns true if the provided manifest platform spec is an exact match to FilterByOS
+func (o *FilterOptions) IncludeSingle(d *manifestlist.ManifestDescriptor, hasMultiple bool) bool {
+	if o.FilterByOS == PlatformSpecString(d.Platform) {
+		return true
+	}
+	return false
```
soltysh, line 199: Nit:
```
return o.FilterByOS == PlatformSpecString(d.Platform)
```

soltysh: What's the difference between `IncludeSingle` and `Include`?

soltysh (commented):
I was rather expecting that `FilterOptions#Complete` method would handle wrapping argument of Filter with `^$`. As it was discussed, this is a regexp filter, and it should be treated as such explicitly. If you pass `linux/ar` that's the string we will look for and it should not match `linux/arm` unless the regexp is `linux/ar.*`. So instead of adding additional method which complicates API I'd just transform non-nil filteros flag.

sallyom:
@soltysh @smarterclayton last commit is what has to happen to keep regex a regex but bookend with ^,$ if/only if info, append, and extract - not mirror. 
Also, I found a bug with 'oc image info' that I added a fix for in that last commit - bug is that if you mirror only a single os/arch of an image then run `oc image info thatimage --filter-by-os=os/arch` oc image info will happily give you info for unmatching filter - if that's considered a bug an ugly fix is in info.go. cmds to see the issue:
```console
$ oc image mirror quay.io/prometheus/prometheus:v2.18.1=localhost:5000/prom-amd64:local (this will only mirror linux/amd64)
$ oc image info localhost:5000/prom-amd64:local --filter-by-os=linux/arm
```
info is printed for the linux/amd64 manifest, but user explicitly asked for linux/arm (that does not exist bc it wasn't mirrored, so should be an error)

pkg/cli/image/manifest/manifest.go
sallyom: IncludeSingle was taking the regex and pretending it was an exact string - however I removed that and added `^, $` to bookend the filter, but _only_ with `oc image append, extract, and info`, _not_ for `oc imag mirror`

pkg/cli/image/manifest/manifest.go
sallyom: ok removed the IncludeSingle function, added a bit of logic to handle regex as regex by adding `^, $` bookends

pkg/cli/image/info/info.go
```diff
@@ -139,6 +143,14 @@ func (o *InfoOptions) Run() error {
 			if err != nil {
 				return err
 			}
+			strippedFilter := strings.TrimSuffix(o.FilterOptions.FilterByOS, "$")
```
smarterclayton, line 146: This doesn't seem appropriate.  This should be a method on FilterOptions that gives you singleton platform.

pkg/cli/image/info/info.go
```diff
@@ -99,6 +99,9 @@ func (o *InfoOptions) Complete(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("info expects at least one argument, an image pull spec")
 	}
 	o.Images = args
+	if err := o.FilterOptions.BookendRegex(); err != nil {
```
smarterclayton, line 102: I would expect this to be handled inside Complete() by a boolean on FilterOptions set in NewInfo/NewAppend (`filter options ExpectSingle`).

pkg/cli/image/info/info.go
```diff
@@ -444,3 +424,53 @@ func (o *ImageRetriever) Run() error {
 		}
 	})
 }
+
+// ImageRetrieverErr differentiates extract errors due to manifestLists, to enable better error message
+func (o *InfoOptions) ImageRetrieverManifestList(src string) (*Image, error) {
```
smarterclayton, line 429: Don't invoke run in this method.  Return ImageRetriever

pkg/cli/image/info/info.go
sallyom: right, this was the ugly fix to the `oc image info` i mentioned. I'll improve it

pkg/cli/image/info/info.go
sallyom: ok, will do that instead

pkg/cli/image/info/info.go
sallyom: updated, done.

pkg/cli/image/info/info.go
sallyom: done

pkg/cli/image/info/info.go
sallyom: updated

pkg/cli/image/manifest/manifest.go
```diff
@@ -282,21 +304,30 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
 	}
 
 	originalSrcDigest := srcDigest
-	srcManifests, srcManifest, srcDigest, err := ProcessManifestList(ctx, srcDigest, srcManifest, manifests, from, filterFn, false)
+	// Do not return a manifest list, filter here to return appropriate os/arch
+	if o.IsWildcardFilter() {
```
smarterclayton, line 308: I don't understand why this is here?  This seems unnecessary and is better handled outside this method.

pkg/cli/image/manifest/manifest.go
```diff
@@ -282,21 +304,30 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
...
+	}
+	srcManifests, srcManifest, srcDigest, err := ProcessManifestList(ctx, srcDigest, srcManifest, manifests, from, o.FilterByOS, o.Include, false)
 	if err != nil {
 		return nil, ManifestLocation{}, err
 	}
-	if len(srcManifests) == 0 {
-		return nil, ManifestLocation{}, fmt.Errorf("filtered all images from manifest list")
+	// Should only ever return 1 manifest from here
+	m := len(srcManifests)
+	switch {
```
smarterclayton, line 318: Thinking about this more, I don't see a reason for this just not to truncate the manifest at the first item.  I don't think we care if there are m > 1 manifests, we should just select `srcManifests[0]` if `len(srcManifests) > 0`, return the generic error (I don't think its appropriate to print o.FilterByOS at this layer, but you can return a typed error if you want `ErrNoMatchingImage` that uses the same generic message and then transform it at a higher method level).

I would prefer this not become a method but instead stay a function given this comment and the one above.

pkg/cli/image/manifest/manifest.go
```diff
@@ -437,7 +467,7 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
 			if err != nil {
 				return nil, nil, "", err
 			}
-			klog.Warningf("Chose %s/%s manifest from the manifest list.", t.Manifests[0].Platform.OS, t.Manifests[0].Platform.Architecture)
+			klog.Warningf("Chose %s manifest from the manifest list.", PlatformSpecString(t.Manifests[0].Platform))
```
smarterclayton, line 470: I don't think this is appropriate at all - we sholud be returning a typed error, NOT printing inline warnings from library methods.  No CLI code should ever call into stuff using klog.

sallyom: It slid by before, in a previous PR (that was mine) - I've removed this, makes sense, thanks for clarifying that.

pkg/cli/image/manifest/manifest.go
sallyom: I see, makes sense. I've updated, FirstManifest back to a function - doing all the error handling higher in append, extract

pkg/cli/image/manifest/manifest.go
sallyom: I removed the need for this error, removed this code

pkg/cli/image/manifest/manifest.go
sallyom: also removed the WildcardFilterErr, found it wasn't necessary  - added a ErrNoMatchingManifestsErr when handling manifest list but no matches found to pass back to append,extract

sallyom:
new output when pass a filter that matches more than a single manifest from a manifest list:
```console
$ oc image info quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=.*
error: filter .* failed to find a single manifest. Use --filter-by-os again to select only a single option from:

  OS          DIGEST
  linux/amd64 sha256:b4e6cd0275a26750505e539f8528e891053434ebd3972be02645bed5f02f0795
  linux/arm64 sha256:33cd2bbeea5c52add05be111fdccd54d289aa609f9059b22e9476865fd22b5b5
  linux/arm   sha256:19bce37348f8e9fc652a0a562cad12a85abd3bde2c5ed905b36df24efec30b48
  linux/s390x sha256:575eb2a6d4c2fb7b15a55a2788294925ab148da77b21dc1c92d0557693fdf749

or, to default to system 'linux/amd64' don't pass --filter-by-os.

$ mkdir ext; cd ext
$ oc image extract quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=.*
error: filter .* failed to find a single manifest. Use --filter-by-os again to select only a single option from:

  OS          DIGEST
  linux/amd64 sha256:b4e6cd0275a26750505e539f8528e891053434ebd3972be02645bed5f02f0795
  linux/arm64 sha256:33cd2bbeea5c52add05be111fdccd54d289aa609f9059b22e9476865fd22b5b5
  linux/arm   sha256:19bce37348f8e9fc652a0a562cad12a85abd3bde2c5ed905b36df24efec30b48
  linux/s390x sha256:575eb2a6d4c2fb7b15a55a2788294925ab148da77b21dc1c92d0557693fdf749

or, to default to system 'linux/amd64' don't pass --filter-by-os.

$ oc image append --from docker.io/library/busybox:latest --to localhost:5000/busybox:local file.tar.gz --filter-by-os=.*
error: filter .* failed to find a single manifest. Use --filter-by-os again to select only a single option from:

  OS             DIGEST
  linux/amd64    sha256:0415f56ccc05526f2af5a7ae8654baec97d4a614f24736e8eef41a4591f08019
  linux/arm/v5   sha256:19ea09c75445a0949e1a39fb100233eded8d3a0b9f6bc0f3e6f237b6c74f0930
  linux/arm/v6   sha256:43c54eaff5af90941cc9895821dfa09448e1c8889de6815aadd4cb9d7e1fb5d7
  linux/arm/v7   sha256:de2179c139422b80990c400ef4021aadb61f1fa6113c8ed0a214f981059ebaa8
  linux/arm64/v8 sha256:bf920ca7f146b802e1c9a8aab1fba3a3fe601c56b075ecef90834c13b90bb5bb
  linux/386      sha256:9bae0ce4b604680683a3f8d538c8f86ffab8b982577eee240e0accf369b3ac78
  linux/mips64le sha256:2c3ded2309184bedf27a03dd99e8f6b1bd28febd7fc9b64dc76a862e3a221214
  linux/ppc64le  sha256:a1dc9b439291e609fc80847b5c7954e117af4b74b7b60506a07b200e9c08d272
  linux/s390x    sha256:16a47375542a993a78368ac80f729382277853523c2baf237c0ae27e41e40fd3

or, to default to system 'linux/amd64' don't pass --filter-by-os.

**Comment by sallyom** [comment]
@smarterclayton while going thru your review/implementing and testing, I found I do not at all like the behavior of setting `^, $` in Complete, because `we want the user to know if they passed "linux/*" that multiple manifests were found, give them the list of options to choose the single, and error`. Adding the bookends results in the above returning the crappy & confusing error of `did not find manifest for linux/*` The _only_ issue with not adding the `^,$` is that you run into the confusing-yet-not-wrong error of getting `multiple manifests found` when a user passes, for example,  `linux/arm` and there are `linux/arm, linux/arm64` available. In this case there are 2 options of what `oc` should do. 
1) Add to the ImageRetriever error message a suggestion that if the filter you passed (linux/arm) is listed below, try again but with `^linux/arm$` to avoid finding multiple with regex expansion.  
2) on the spot add the ^,$ only _after_ multiple were found, as a convenience to the user. This would look like
```
        srcManifest, manifestLocation, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
        if err == nil {
            if ml, ok := srcManifest.(*manifestlist.DeserializedManifestList); ok {
                if len(ml.Manifests) > 1 { 
                    // if multiple matches found, try once more with an exact filter match
                    // have to do this here, need to differentiate between a filter such as 'linux/*',
                    // that may have more than one match, from 'linux/arm' that should only have a single
                    // match and should not also include linux/arm64.
                    multipleMatch = true
                    filterFn := func(d *manifestlist.ManifestDescriptor, hasMultiple bool) bool {
                        return imagemanifest.PlatformSpecString(d.Platform) == o.FilterOptions.FilterByOS
                    }
                    srcManifest, manifestLocation, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, filterFn)
                    ml, ok := srcManifest.(*manifestlist.DeserializedManifestList)
                    if (ok && len(ml.Manifests) == 1) || !ok {
                        multipleMatch = false
                    }
                }
            }
        }
        if err != nil || multipleMatch {
```
and that would be run only in extract, append, info. I prefer 2), but I also tie my son's shoes still, and he's 10.  If I had to guess, you are going to choose 1). Wdyt? Thanks!

File: pkg/cli/image/append/append.go [hunk 25]
```diff
@@ -258,8 +259,50 @@ func (o *AppendImageOptions) Run() error {
...
+						return imagemanifest.PlatformSpecString(d.Platform) == o.FilterOptions.FilterByOS
+					}
+					srcManifest, manifestLocation, err = imagemanifest.FirstManifest(ctx, from.Ref, repo, filterFn)
+					ml, ok := srcManifest.(*manifestlist.DeserializedManifestList)
+					if (ok && len(ml.Manifests) == 1) || !ok {
+						multipleMatch = false
+					}
+				}
+			}
+		}
```
sallyom, line 284: other option for the above code is to add a suggestion with the ImageRetriever error that if the filter you passed is listed below in the available os/arch options for the image, then try again but with `^,$` to avoid multiple matches.

sallyom: I'm guessing this is the option that will be preferred, so I'll push that now.

soltysh:
> @smarterclayton while going thru your review/implementing and testing, I found I do not at all like the behavior of setting `^, $` in Complete, because `we want the user to know if they passed "linux/*" that multiple manifests were found, give them the list of options to choose the single, and error`. Adding the bookends results in the above returning the crappy & confusing error of `did not find manifest for linux/*` The _only_ issue with not adding the `^,$` is that you run into the confusing-yet-not-wrong error of getting `multiple manifests found` when a user passes, for example, `linux/arm` and there are `linux/arm, linux/arm64` available. In this case there are 2 options of what `oc` should do.

But `linux/*` is not a valid regular expression so the error that nothing was matched is perfectly reasonable. `linux/.*` should give you the list of matched manifests, though. Notice the dot `.*` before the asterisk in that expression.

sallyom:
> > @smarterclayton while going thru your review/implementing and testing, I found I do not at all like the behavior of setting `^, $` in Complete, because `we want the user to know if they passed "linux/*" that multiple manifests were found, give them the list of options to choose the single, and error`. Adding the bookends results in the above returning the crappy & confusing error of `did not find manifest for linux/*` The _only_ issue with not adding the `^,$` is that you run into the confusing-yet-not-wrong error of getting `multiple manifests found` when a user passes, for example, `linux/arm` and there are `linux/arm, linux/arm64` available. In this case there are 2 options of what `oc` should do.
> 
> But `linux/*` is not a valid regular expression so the error that nothing was matched is perfectly reasonable. `linux/.*` should give you the list of matched manifests, though. Notice the dot `.*` before the asterisk in that expression.

Ok, I've pushed what I believe is the best option, here's output:
```console
$ oc image info quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=linux/arm
error: multiple manifests that match filter linux/.*
The image is a manifest list - use --filter-by-os to select only a single option from below
If the --filter-by-os option you passed is listed below, try again with '^os/arch$' to avoid multiple manifests matching with regex expansion:

  OS          DIGEST
  linux/amd64 sha256:b4e6cd0275a26750505e539f8528e891053434ebd3972be02645bed5f02f0795
  linux/arm64 sha256:33cd2bbeea5c52add05be111fdccd54d289aa609f9059b22e9476865fd22b5b5
  linux/arm   sha256:19bce37348f8e9fc652a0a562cad12a85abd3bde2c5ed905b36df24efec30b48
  linux/s390x sha256:575eb2a6d4c2fb7b15a55a2788294925ab148da77b21dc1c92d0557693fdf749

or, to default to system 'linux/amd64' don't pass --filter-by-os.
```
and you'll get that same output if you pass `linux/*`, `linux/.*`, `.*` etc
and if none found you'll get slightly different msg:
```console
$ oc image info quay.io/prometheus/prometheus:v2.18.1 --filter-by-os=linux/foo
error: no manifests retrieved that match filter linux/foo
The image is a manifest list - use --filter-by-os to select only a single option from below
If the --filter-by-os option you passed is listed below, try again with '^os/arch$' to avoid multiple manifests matching with regex expansion:

  OS          DIGEST
  linux/amd64 sha256:b4e6cd0275a26750505e539f8528e891053434ebd3972be02645bed5f02f0795
  linux/arm64 sha256:33cd2bbeea5c52add05be111fdccd54d289aa609f9059b22e9476865fd22b5b5
  linux/arm   sha256:19bce37348f8e9fc652a0a562cad12a85abd3bde2c5ed905b36df24efec30b48
  linux/s390x sha256:575eb2a6d4c2fb7b15a55a2788294925ab148da77b21dc1c92d0557693fdf749

or, to default to system 'linux/amd64' don't pass --filter-by-os.
```

pkg/cli/image/info/info.go
```diff
@@ -435,12 +404,66 @@ func (o *ImageRetriever) Run() error {
...
 			})
 		}
 	})
+	if err != nil {
+		return nil, err
+	}
+	return imageRetrieved, nil
+}
+
+// ImageRetrieverOrList returns ImageRetriever or lists os/arch options for manifestList source
```
sallyom, line 422: I can update this comment to 
// ImageRetrieverOrList returns ImageRetriever and adds to ManifestListCallback  a msg to distinguish errors from  multiple manifests found from no manifests found, and lists os/arch options for manifestList source if either is the case

sallyom:
> /hold
> 
> for the long comment asking why the scope of change is necessary if we just interpret --filter-by-os is necessary

@smarterclayton @soltysh  ptal, I've improved output with errors due to multiple manifests or no manifests filtered from manifest list, I've removed the logic of adding the ^, $ to filter in favor of suggesting to the user they do that themselves, if the filter they want is provided in the list of available os/arch options for that image.

pkg/cli/image/extract/extract.go
```diff
@@ -343,8 +346,14 @@ func (o *ExtractOptions) Run() error {
 					return fmt.Errorf("unable to connect to image repository %s: %v", from.String(), err)
 				}
 
+				isList := false
 				srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
-				if err != nil {
+				if err == nil {
```
soltysh, line 351: Here similarly to above, `isList` is escaping its necessary scope by being defined outside the error check condition's body. You don't need that outside it. 
So I'd re-write it as:
```golang
srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
if err != nil {
    .. // the 2 conditions ISImageForbidden & IsImageNotFound
    if _, ok := srcManifest.(*manifestlist.DeserializedManifestList); ok {
    .. // the body of if isList condition from below
```

soltysh: Also since you're using that two times already, I'd consider creating a function to encapsulate that logic.

soltysh (commented):
I've noticed you're setting `^..$` only in a single plase in `pkg/cli/admin/release/new.go` but from the discussions earlier that was meant to happen in `FilterOptions`'s `Complete` method. I remember you had some doubts, but I thought that one of my previous explanation made it clear.

sallyom:
> I've noticed you're setting `^..$` only in a single plase in `pkg/cli/admin/release/new.go` but from the discussions earlier that was meant to happen in `FilterOptions`'s `Complete` method. I remember you had some doubts, but I thought that one of my previous explanation made it clear.

The `oc adm release new` does not expose the filter-by-os flags, so no Complete - this is solely for a user creating a new release while providing a multi-arch image for a component substitution ([see the referenced bz](https://bugzilla.redhat.com/show_bug.cgi?id=1859452) for this PR) - this is an unsupported release configuration but one that oc should be able to handle, for dev/testing/non-prod purposes. For this purpose only, have to manually add the os/arch to the `adm new` cmd here.  The `^,$` probably aren't necessary, but might-as-well since it's most correct. 

In all other cases, I've removed the need to add the ^,$, since all was working exactly as it should - I found that when a user passes a wildcard, the cmds that don't handle multiple manifests (info, append, extract) _should_ at least report to the user that either _multiple_ manifests were retrieved or _no_ manifests retrieved for the given filter. If we add the ^,$, user will get a misleading error, IMO, that `no manifest found to match given regex` when in many cases that's not true, the issue is that _multiple_ were found so have to pass a more exact filter.

pkg/cli/image/append/append.go
```diff
@@ -258,10 +259,33 @@ func (o *AppendImageOptions) Run() error {
 		}
 		fromRepo = repo
 
+		isList := false
 		srcManifest, manifestLocation, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
+		if err == nil {
```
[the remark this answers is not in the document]
sallyom: updated, fixed

pkg/cli/image/extract/extract.go
sallyom: updated, fixed, didn't add the fn bc append doesn't include the 2 extra error checks, i'll add a fn if we should add the ImageForbidden/ImageNotFound checks though.

soltysh:
Fine by me, we need final justification from @smarterclayton

pkg/cli/image/append/append.go
```diff
@@ -262,6 +263,23 @@ func (o *AppendImageOptions) Run() error {
...
+		if _, ok := srcManifest.(*manifestlist.DeserializedManifestList); ok {
+			// If type is a ManifestList here, did not filter to single manifest from manifest list
+			// print os/arch options for this image as a convenience and error
+			// ImageRetrieverOrList prints os/arch options for an image upon error
+			infoOpts := &imageinfo.InfoOptions{
+				IOStreams:       o.IOStreams,
+				SecurityOptions: o.SecurityOptions,
+				FilterOptions:   o.FilterOptions,
+			}
+			retriever, err := infoOpts.ImageRetrieverOrList(from.String(), o.FilterOptions.FilterByOS)
```
smarterclayton, line 275: Reretrieving the manifest seems like not the right thing to do - we already have it loaded.

pkg/cli/image/info/info.go
```diff
@@ -435,12 +404,67 @@ func (o *ImageRetriever) Run() error {
...
+		return &ImageRetriever{}, err
+	}
+	retriever := &ImageRetriever{
+		Image: map[string]imagesource.TypedImageReference{
+			src: srcRef,
+		},
+		FileDir:         o.FileDir,
+		SecurityOptions: o.SecurityOptions,
+		ManifestListCallback: func(from string, list *manifestlist.DeserializedManifestList, all map[digest.Digest]distribution.Manifest) (map[digest.Digest]distribution.Manifest, error) {
+			filtered := make(map[digest.Digest]distribution.Manifest)
```
smarterclayton, line 436: It looks like you want to split this out into a function `filterManifestList(manifestList, filterOptions, w io.Writer) (map[digest]manifest, error)` and then call this from multiple places.

pkg/cli/image/info/info.go
```diff
@@ -435,12 +404,67 @@ func (o *ImageRetriever) Run() error {
...
 	})
+	if err != nil {
+		return nil, err
+	}
+	return imageRetrieved, nil
+}
+
+// ImageRetrieverOrList returns ImageRetriever and with the ManifestListCallbackFn provides a message
+// to distinguish errors between multiple manifests found and no manifests found, and lists os/arch options for manifestList upon error
+func (o *InfoOptions) ImageRetrieverOrList(src, filterByOS string) (*ImageRetriever, error) {
```
smarterclayton, line 424: If you extract the lower method this larger function should go back to being inlined (you're just passing a standard function to ManifestListCallback at that point)

smarterclayton: I'm not completely opposed to this method being extracted, it's just going to have a lot of arguments and that's a bit of a code smell.

pkg/cli/image/info/info.go
```diff
@@ -435,12 +404,67 @@ func (o *ImageRetriever) Run() error {
...
+				if !o.FilterOptions.Include(&manifest, len(list.Manifests) > 1) {
+					klog.V(5).Infof("Skipping image for %#v from %s", manifest.Platform, from)
+					continue
+				}
+				filtered[manifest.Digest] = all[manifest.Digest]
+			}
+			if len(filtered) == 1 {
+				return filtered, nil
+			}
+			msg := fmt.Sprintf("no manifests retrieved that match filter %s", filterByOS)
```
smarterclayton, line 447: You should be getting filterByOS from o.FilterOptions here, not an argument

sallyom: duh, right, fixed.

pkg/cli/image/info/info.go
sallyom: no reason to extract this, un-extracted it

pkg/cli/image/append/append.go
sallyom: right, all i need is the callback error, i've extracted that (as you suggested below)

pkg/cli/image/info/info.go
sallyom: done

sallyom:
@smarterclayton please review, i've addressed all feedback, thanks 

commands to see fix & improved output: https://github.com/openshift/oc/pull/513#issuecomment-725061916

soltysh:
@sallyom this needs a rebase
@smarterclayton bump for review

sallyom:
> @sallyom this needs a rebase
> @smarterclayton bump for review

rebased and included the [revert pr](https://github.com/openshift/oc/pull/780), also, since they conflict

soltysh:
@sallyom verify failure is real, you need to generate completions.

sallyom:
> > @sallyom this needs a rebase
> > @smarterclayton bump for review
> 
> rebased and included the [revert pr](https://github.com/openshift/oc/pull/780), also, since they conflict

This us updated, ready for final review. Please see sample commands with improved output and the fix:
https://github.com/openshift/oc/pull/513#issuecomment-725061916

sallyom:
> > > @sallyom this needs a rebase
> > > @smarterclayton bump for review
> > 
> 
> This us updated, ready for final review. Please see sample commands with improved output and the fix:
> [#513 (comment)](https://github.com/openshift/oc/pull/513#issuecomment-725061916)

@smarterclayton bump for review

sallyom:
@smarterclayton afaict this doesn't need a /hold, i've updated and this is awaiting review
-  improved behavior here: https://github.com/openshift/oc/pull/513#issuecomment-725061916 
- as you noted previously, this log should be removed - if this PR will not be merged then I'll open a PR w/ only this change: https://github.com/openshift/oc/pull/513/files#diff-88a19aa652d0dad2300f5b2080f14494546c5036b5b357e07ae150e24d67c072L442

Issue #496, opened 2020-07-17
oc login error: "net/http: TLS handshake timeout"


fckbo:
Hello,

I've got an oc login error: "net/http: TLS handshake timeout" on my mac running mojave, even though connecting with the same user using the ui or using oc on a linux system works fine

command used and output:
```
 oc login -u kubeadmin -p <mypwd> api.myocp.mydomain.local:6443 --insecure-skip-tls-verify=true
error: net/http: TLS handshake timeout
```

with log
```
oc login -u kubeadmin -p <mypwd> api.myocp.mydomain.local:6443 --insecure-skip-tls-verify=true --loglevel=9
I0717 19:04:01.721157    8593 loader.go:375] Config loaded from file:  /Users/fb/.kube/config
I0717 19:04:01.722654    8593 round_trippers.go:423] curl -k -v -XHEAD  'https://myocp.mydomain.local:6443/'
I0717 19:04:11.763330    8593 round_trippers.go:443] HEAD https://api.myocp.mydomain.local:6443/  in 10040 milliseconds
I0717 19:04:11.763367    8593 round_trippers.go:449] Response Headers:
F0717 19:04:11.763447    8593 helpers.go:115] error: net/http: TLS handshake timeout
```

Same problem without specifying the option `--insecure-skip-tls-verify=true`

oc version:
Client Version: 4.5.2 on my mac mojave (Version 10.14.6)
Openshift Server Version: 4.2.21 (same problem with another server in 4.3)

The funny thing is that if I do a sudo it works, here is the beginning of the log with same loglevel
```
sudo oc login -u kubeadmin -p <mypwd> api.myocp.mydomain.local:6443 --insecure-skip-tls-verify=true --loglevel=9
Password:
I0717 19:09:31.987104    8867 loader.go:375] Config loaded from file:  /Users/fb/.kube/config
I0717 19:09:31.988190    8867 round_trippers.go:423] curl -k -v -XHEAD  'https://api.myocp.mydomain.local:6443/'
I0717 19:09:32.226527    8867 round_trippers.go:443] HEAD https://api.myocp.mydomain.local:6443/  in 238 milliseconds
I0717 19:09:32.226548    8867 round_trippers.go:449] Response Headers:
I0717 19:09:32.226598    8867 round_trippers.go:423] curl -k -v -XHEAD  'https://api.myocp.mydomain.local:6443/'
[...]

PR #514, opened 2020-08-10
Bug 1880865: fix part 1 of issue https://github.com/openshift/oc/issues/496

fckbo:
Issue:

- Observed behaviour: Cannot login to an openshift cluster due to TLS Handshake timeout even with option --insecure-skip-tls-verify=true (See description of the root cause here: https://github.com/openshift/oc/issues/496)

- Expected behaviour: Connection should succeed

### DIFF
--- a/pkg/cli/login/loginoptions.go
+++ b/pkg/cli/login/loginoptions.go
@@ -133,6 +133,9 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 		clientConfig.CAData = caData
 	}
 
+	// if user has selected option --insecure-skip-tls-verify=true, TLS server certificate verification should not be attempted
+	clientConfig.Insecure = o.InsecureTLS
+
 	// try to TCP connect to the server to make sure it's reachable, and discover
 	// about the need of certificates or insecure TLS
 	if err := dialToServer(*clientConfig); err != nil {


stlaz:
This won't fix whatever issue you're having. If you're getting `TLS handshake timeout`, TLS handshake is still attempted even with `--insecure-skip-tls-verify=true`, that option only skips verifying the server certificate.

stlaz:
/ok-to-test
Should have read the issue prior to commenting, the client validating is taking too long is... Well, I've never seen anything like that before.

pkg/cli/login/loginoptions.go
```diff
@@ -133,6 +133,12 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 		clientConfig.CAData = caData
 	}
 
+	// if user has selected option --insecure-skip-tls-verify=true on purpose, the login should not fail due to TLS verification
+	// errors (like TLS Handshake Timeout for example see issue: https://github.com/openshift/oc/issues/496)
+	if o.InsecureTLS {
+		clientConfig.Insecure = true
+	}
+
```
stlaz, line 141: I don't think the comment is necessary (and I don't really like issue/BZ/JIRA links in comments).

Also
```golang
  clientConfig.Insecure = o.InsecureTLS
```

stlaz (changes requested):
squash the changes into a single commit please

stlaz:
This is not really a good solution to your problem though. The proper one would be to find out why retrieving the certs takes so long - slow disk/not enough memory/hole in space-time continuum? I am not sure 300 is a high enough number for the cert verification to take over 10s.

fckbo:
@stlaz,
sorry was out in vac for a while... to your question : "find out why it takes so long to retrieve certificate?"  I actually re-compile the go 'net' & 'crypto' packages putting my own traces to analyse the issue and found that the parseCertificate function in crypto/x509/x509.go (https://github.com/golang/go/blob/master/src/crypto/x509/x509.go) is taking quite some time and that around 250 certificates can be parse in 10 sec. 
As mentioned the problem only happens on macOS.

pkg/cli/login/loginoptions.go
fckbo: Sure

fckbo:
I'll investigate why these 2 tests fail in the coming days

fckbo:
@stlaz,

I found the time to do add more traces in the go crypto functions being used and asked some colleagues to test on other macs having same & different os flavor: Mojave (like mine) & Catalina and obviously different content in their keychain.

I can now confirm that the problem is not due to performance problems on my mac, I actually found similar timing (and worst ones) on my colleague's mac and with the traces could pin-point that the duration of the checking is mostly coming from certificate with trust policies that lead to the invocation of the function 'verifyCertWithSystem' mentioned in a post above.

This function actually invokes the following command: 
```/usr/bin/security verify-cert -p ssl -c  <tempCertFileName> -l -L
```

pkg/cli/login/loginoptions.go
fckbo: Hello,

I've reworked the code & comments, squashed the changes in a single commits and made additional investigations to clarify the root cause of the issue (see 2 comments below) and re-ran the tests.

Shall I open a bug in bugzilla to move this forward ?
Thx

fckbo:
@soltysh ,
hello, I see the msg above: 'expected the bug to target the "4.6.0" release, but it targets "4.7.0" instead'. May I ask you what the best way to move this forward ? 
Thanks.

soltysh:
Both @stlaz and myself are not comfortable to make this particular change, so I'd prefer leave this code as is.

PR #516, opened 2020-08-11
Bug 1867792: Allow prune to move on with Removed registry

ricardomaraschini:
Pruner attempts to derive internal image registry host from the last
unmanaged image or from the last image stream created. In some cases
OpenShift may not contain any of those (for instance in a fresh install
with Removed registry), this makes pruner to fail even with the flag
--prune-registry set to false.

This commit allows the pruner job to move on with the --prune-registry
flag set to false even when we can't derive the internal registry host.

### DIFF
--- a/pkg/cli/admin/prune/imageprune/helper.go
+++ b/pkg/cli/admin/prune/imageprune/helper.go
@@ -40,6 +40,10 @@ func (ba isByAge) Less(i, j int) bool {
 	return ba[i].CreationTimestamp.After(ba[j].CreationTimestamp.Time)
 }
 
+// ErrRegistryHostNotFound is an error returned when we could not find an image from where we can
+// derive the image registry host.
+var ErrRegistryHostNotFound = fmt.Errorf("no image found from where we can derive registry host")
+
 // DetermineRegistryHost returns registry host embedded in a pull-spec of the latest unmanaged image or the
 // latest imagestream from the provided lists. If no such pull-spec is found, error is returned.
 func DetermineRegistryHost(images *imagev1.ImageList, imageStreams *imagev1.ImageStreamList) (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/prune/imageprune/helper.go
```diff
@@ -40,6 +40,10 @@ func (ba isByAge) Less(i, j int) bool {
 	return ba[i].CreationTimestamp.After(ba[j].CreationTimestamp.Time)
 }
 
+// ErrNoImageFound is an error returned when we could not find an image from where we can
+// derive the image registry host.
+var ErrNoImageFound = fmt.Errorf("no image found from where we can derive registry host")
```
adambkaplan, line 45: nit - perhaps rename to `ErrRegistryHostNotFound`:

```suggestion
// ErrRegistryHostNotFound is an error returned when we could not find an image from where we can
// derive the image registry host.
var ErrRegistryHostNotFound = fmt.Errorf("no image found from where we can derive registry host")
```

pkg/cli/admin/prune/images/images.go
```diff
@@ -436,7 +436,14 @@ func (o PruneImagesOptions) Run() error {
...
 			if err != nil {
-				return fmt.Errorf("unable to determine registry: %v", err)
+				if err != imageprune.ErrNoImageFound {
+					return fmt.Errorf("unable to determine registry: %v", err)
+				}
+				// if we can't determine the registry host (ErrNoImageFound) and
+				// we need to prune it as well, bail out immediately.
+				if o.PruneRegistry != nil && *o.PruneRegistry {
+					return fmt.Errorf("unable to find the remote registry host: %v", err)
+				}
```
adambkaplan, line 446: @ricardomaraschini how will this prevent the cluster from being degraded? Will `o.PruneRegistry` be `false` and/or nil if the image registry is `Removed`?

ricardomaraschini: The problem here is that even with `--prune-registry=false` it was still attempting and failing to parse the registry URL.

ricardomaraschini:
After @adambkaplan  comment I realized there is no point in having an specific error to report that we can't find the remote registry when we are not going to access it in the first place. I am closing this PR and opening a new one (https://github.com/openshift/oc/pull/520)

PR #517, opened 2020-08-11
fix get printers when output format specified

sallyom:
TEST: DO NOT MERGE: OPEN UPSTREAM, WILL CLOSE THIS


vendor/k8s.io/kubectl/pkg/cmd/get/get.go
```diff
@@ -263,7 +263,10 @@ func (o *GetOptions) Complete(f cmdutil.Factory, cmd *cobra.Command, args []stri
 		if outputObjects != nil {
 			printer = &skipPrinter{delegate: printer, output: outputObjects}
 		}
-		if o.ServerPrint {
+
+		outputFormatSpecified := o.PrintFlags.OutputFormat != nil && len(*o.PrintFlags.OutputFormat) > 0
```
soltysh, line 267: Close but not quite right, we have several output formats:
```
  -o, --output='': Output format. One of:
json|yaml|wide|name|custom-columns=...|custom-columns-file=...|go-template=...|go-template-file=...|jsonpath=...|jsonpath-file=...
```
only some should not use server printing, but `name`, `wide` will certainly do. I'd check what has changed in get and printers between 1.18 and 1.19 that might have caused this regression.

sallyom: ah right, ok :eyes:

sallyom:
closing, opened upstream

PR #520, opened 2020-08-12
Bug 1867792: Allow prune to move on with Removed registry

ricardomaraschini:
Pruner attempts to derive internal image registry host from the last
unmanaged image or from the last image stream created. In some cases
OpenShift may not contain any of those (for instance in a fresh install
with Removed registry), this makes pruner to fail even with the flag
--prune-registry set to false.

This commit allows the pruner job to move on with the --prune-registry
flag set to false even when we can't derive the internal registry host.

### DIFF
--- a/pkg/cli/admin/prune/images/images_test.go
+++ b/pkg/cli/admin/prune/images/images_test.go
@@ -39,6 +39,42 @@ import (
 
 var logLevel = flag.Int("loglevel", 0, "")
 
+func TestPruneWithoutImagesWithRegistryAccess(t *testing.T) {
+	pruneRegistry := true
+	opts := &PruneImagesOptions{
+		Namespace:     "foo",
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/prune/images/images_test.go
```diff
@@ -39,6 +39,52 @@ import (
 
 var logLevel = flag.Int("loglevel", 0, "")
 
+func TestPruneWithoutImagesWithRegistryAccess(t *testing.T) {
+	kFake := fakekubernetes.NewSimpleClientset()
+	imageFake := &fakeimagev1client.FakeImageV1{
```
dmage, line 44: Why does image client have its own variable, while apps client and build client are created inline?

ricardomaraschini: Because `copy&paste` reasons. Fixed, thanks.

soltysh (approved):
/approve

PR #521, opened 2020-08-16
Bug 1868324: oc logout should make the token invalid

barleyer:
Since openshift/enhancements#323, oc logout should delete the object name of oauthaccesstoken instead of token value.

And I didnt **import** [openshift/oauth-server](https://github.com/openshift/oauth-server),because this fix  just needs few funcs from that repo.

### DIFF
--- a/pkg/cli/logout/logout.go
+++ b/pkg/cli/logout/logout.go
@@ -2,8 +2,11 @@ package logout
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


pkg/cli/logout/logout.go
```diff
@@ -149,3 +154,25 @@ func deleteTokenFromConfig(config kclientcmdapi.Config, pathOptions *kclientcmd.
 	}
 	return kclientcmd.ModifyConfig(pathOptions, config, true)
 }
+
+// TokenToObjectName returns the oauthaccesstokens object name for the given raw token,
+// i.e. the sha256 hash prefixed with "sha256~".
+func TokenToObjectName(code string) string {
```
sttts, line 160: nit: lower-case these new funcs as they are not used externally. Otherwise :+1:

sttts:
Only a nit about the func names. Otherwise lgtm.

pkg/cli/logout/logout.go
soltysh: Similarly the other two functions lower-case as well.

barleyer:
@sttts @soltysh Thanks both.I have updated those func names and tested locally via `make oc`.

PR #523, opened 2020-08-19
must-gather: use exponential backoff for polling

yuvalturg:
When running must-gather on a bare metal cluster, the command will fail
to find the pod after a 1 minute duration, and exit with an error
message in the form of `Get: $pod-url: unexpected EOF`.

This patch replaces the normal way of polling (every 1 minute), with an
exponential backoff starting with 1 second, and a factor of 2.  The
timeout value is unchanged.

Bug 1859972

Signed-off-by: Yuval Turgeman <yturgema@redhat.com>

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -397,9 +397,18 @@ func newPrefixWriter(out io.Writer, prefix string) io.Writer {
 	return writer
 }
 
+func (o *MustGatherOptions) getBackoff() wait.Backoff {
+	return wait.Backoff{
+		Duration: time.Second,
+		Factor:   2,
+		Steps:    10,
+		Cap:      time.Duration(o.Timeout) * time.Second,
+	}
+}
+
 func (o *MustGatherOptions) waitForPodRunning(pod *corev1.Pod) error {
 	phase := pod.Status.Phase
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -397,9 +397,18 @@ func newPrefixWriter(out io.Writer, prefix string) io.Writer {
...
+		Factor:   2,
+		Steps:    10,
+		Cap:      time.Duration(o.Timeout) * time.Second,
+	}
+}
+
 func (o *MustGatherOptions) waitForPodRunning(pod *corev1.Pod) error {
 	phase := pod.Status.Phase
-	err := wait.PollImmediate(time.Minute, time.Duration(o.Timeout)*time.Second, func() (bool, error) {
+	err := wait.ExponentialBackoff(o.getBackoff(), func() (bool, error) {
```
soltysh, line 411: Does shortening this initial timeout like here https://github.com/openshift/oc/pull/527 does the trick for you? Why you're proposing the exponential backoff?

yuvalturg: I thought you needed a longer interval, so I figure this was a decent compromise.  If #527 works for you, then yes, it will solve our issue as well and we can close this.

yuvalturg:
Closing this in favor of #527

PR #526, opened 2020-08-20
Bug 1866876: Signature verify error msgs to klog.V(4) rather than klog.Infof

sallyom:
I am not able to reproduce the info messages reported in the bug.  I do see the messages are double printed, though.  There might be a good reason why these are both printing klog.Infof and klog.V(4)? 

/assign @soltysh 
/cc @smarterclayton


soltysh (approved):
/lgtm
/approve

soltysh:
I'm manually applying valid bug label, since the bot can't access private bugs.

PR #530, opened 2020-08-25
Bug 1872709: Copy foreign layers from external sources during mirroring

smarterclayton:
A foreign image layer has a URL that points outside the current registry, such that the registry returns not found if the blob
is requested. Since mirror attempts to ensure the destination image can be pulled, even in the presence of disjoint networks,
and given that the Docker daemon will by default attempt the blob lookup first before checking URLs, always copy the blob
by checking the source registry, then trying to pull the layer URLs.

There may be scenarios where this is not desirable, and a future change may allow skipping mirror of those blobs.

### DIFF
--- a/pkg/cli/image/manifest/manifest.go
+++ b/pkg/cli/image/manifest/manifest.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"regexp"
 	"runtime"
 	"sync"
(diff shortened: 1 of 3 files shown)


soltysh (approved):
/lgtm
/retest

smarterclayton:
/hold

Testing out whether i can do a silghtly better fix

smarterclayton:
/hold cancel

This is updated to actually mirror URLs after reading the spec and discussing with oleg.  Verified this works between gcr -> internal, internal -> internal, and gcr -> quay.

soltysh (approved):
Got :+1: from Oleg on slack
/lgtm
/retest

dmage:
title needs to be updated, otherwise LGTM

PR #537, opened 2020-08-28
Bug 1868384: Login: Avoid overwriting same user from different cluster

alvaroaleman:


alvaroaleman:
that e2e test is just 99% percent red, so I don't think its the fault of this PR: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-oc-master-e2e-cmd
/retest

pkg/helpers/originkubeconfignames/smart_merge.go
```diff
@@ -35,8 +35,7 @@ func GetUserNicknameFromCert(clusterNick string, chain ...*x509.Certificate) (st
 }
 
 func GetContextNickname(namespace, clusterNick, userNick string) string {
-	tokens := strings.SplitN(userNick, "/", 2)
-	return namespace + "/" + clusterNick + "/" + tokens[0]
+	return namespace + "/" + clusterNick + "/" + userNick
```
alvaroaleman, line 38: This is needed to not make `test/cmd/projects.sh` fail due to the output being
```
Now using project \"test4\" from context named \"test4/api-ci-op-rv3cdw69-d5f1e-origin-ci-int-gce-dev-openshift-com:6443/system:admin/api-ci-op-rv3cdw69-d5f1e-origin-ci-int-gce-dev-openshift-com:6443\" on server \"https://api.ci-op-rv3cdw69-d5f1e.origin-ci-int-gce.dev.openshift.com:6443\"
```

instead of the expected

```
Now using project \"test4\" on server 
```
because the `contextName` is not the `defaultContextName` anymore here: so we fallthrough and print it here:
https://github.com/openshift/oc/blob/89c013fb9dd83b2e7e47cfcfc97b01b4b4b0d1d9/pkg/cli/project/project.go#L317

When we generate the `contextNick` this splitting code doesn't exist so not sure why it is here: https://github.com/openshift/oc/blob/89c013fb9dd83b2e7e47cfcfc97b01b4b4b0d1d9/pkg/helpers/kubeconfig/smart_merge.go#L52

pkg/helpers/kubeconfig/smart_merge.go
```diff
@@ -47,6 +47,7 @@ func CreateConfig(namespace, userName string, clientCfg *restclient.Config) (*cl
 	if err != nil {
 		return nil, err
 	}
+	userName = userName + "/" + clusterNick
```
soltysh, line 50: You'd need to move this after `getContextNicknameFromConfig` otherwise you'll end up with context having `clusterNick` twice.
I also wonder if not reversing the order, even though this seems intuitive, contextNick has them reversed so it'll be more reasonable to keep that same approach here as well.

pkg/helpers/originkubeconfignames/smart_merge.go
soltysh: Right this is failing because you're modifying the `contextNick`, see my previous comment.

soltysh (changes requested):

soltysh (approved):
/lgtm
/approve

soltysh:
Failures unrelated to this PR

PR #541, opened 2020-08-31
Bug 1874293: pkg/cli/admin/upgrade/: Mention Upgradeable=False

wking:
When `Upgradeable=False`, the cluster-version operator will block (usually) minor version bumps.  [Sometimes][1] it will block all version bumps.  Users running `oc adm upgrade` should be notified about these conditions, so they are not surprised if they choose a target and the CVO subsequently blocks the chosen update.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1822844


soltysh (approved):
/lgtm
/retest

PR #542, opened 2020-08-31
Bug 1874290: Do not generate REST config when only basic auth credentials are specified

bostrt:


soltysh:
Since this command is targeted at oc users, I'm not sure this approach is a good one. 
/hold

PR #545, opened 2020-09-01
Bug 1866399: lessen retries in must-gather

sallyom:
Increase retry intervals in must-gather. 3s proved too short for some,
resulted in too many retries.  10s is a fine compromise.


soltysh (approved):
/lgtm
/approve

PR #546, opened 2020-09-01
Bug 1812813: oc adm must-gather: add empty node-selector annotation to namespace

sallyom:
Without this, 'oc adm must-gather' fails when defaultNodeSelector is set and the
must-gather namespace is created on a node without the matching labels.
This PR adds the annotation 'openshift.io/node-selector: ""' to the
must-gather namespace to override a configured defaultNodeSelector.

/assign @soltysh 
/cc @deads2k

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -236,7 +236,8 @@ func (o *MustGatherOptions) Run() error {
 				"openshift.io/run-level": "0",
 			},
 			Annotations: map[string]string{
-				"oc.openshift.io/command": "oc adm must-gather",
+				"oc.openshift.io/command":    "oc adm must-gather",
+				"openshift.io/node-selector": "",
 			},
 		},
 	}, metav1.CreateOptions{})


soltysh (approved):
/lgtm
/approve
@sallyom you need to fix go-fmt failure, it's legit

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -237,6 +237,7 @@ func (o *MustGatherOptions) Run() error {
 			},
 			Annotations: map[string]string{
 				"oc.openshift.io/command": "oc adm must-gather",
+				"openshift.io/node-selector": "",
```
deads2k, line 240: yeah, this seems fine.  YOu want it for oc debug node, too, right?

sallyom:
fixed verify :facepalm:

pkg/cli/admin/mustgather/mustgather.go
sallyom: yes, i added that in a [separate PR](https://github.com/openshift/oc/pull/550), bc wasn't sure on direction of that.

soltysh (approved):
/hold cancel
/lgtm

PR #518, opened 2020-08-11
Bug 1866380: must-gather ImagePullPolicy should be 'Always'

sallyom:
/assign @soltysh 

`oc adm must-gather` should use `ImagePullPolicy: "Always"`? 
 issue this PR resolves: 
```
Run oc adm must-gather --image docker.io/sallyom/must-gather:local then look at pod yaml of must-gather pod:
(pod yaml snippet)
start with:
      image: docker.io/sallyom/must-gather:local
      imageID: docker.io/sallyom/must-gather@sha256:b20516b99edb0dd40010f4084b16117e936863dd25ea24b363489

committed and pushed new image id to docker.io/sallyom/must-gather:local

    with "IfNotPresent" imagePullPolicy:
      image: docker.io/sallyom/must-gather:local
      imageID: docker.io/sallyom/must-gather@sha256:b20516b99edb0dd40010f4084b16117e936863dd25ea24b36348961cb4a
    with "Always" imagePullPolicy:
      image: docker.io/sallyom/must-gather:local
      imageID: docker.io/sallyom/must-gather@sha256:21aed5526cbc8db38c350cc74c57510f09b8b37425326c20ce2f9e7931b
```

PR #549, opened 2020-09-02
bug 1874856: use local must-gather image if it is present

deads2k:


soltysh (approved):
/lgtm

sallyom:
@soltysh @deads2k  this will re-open this bz then:  https://bugzilla.redhat.com/show_bug.cgi?id=1866380 
(https://github.com/openshift/oc/pull/518)  The issue is a must-gather:tag with an update but still pushed as must-gather:tag the latest must-gather push will not be fetched.  Is this a problem or should we follow the best practice of not pushing image updates to the same tag (if we're depending on always getting those updates).

soltysh:
> @soltysh @deads2k this will re-open this bz then: https://bugzilla.redhat.com/show_bug.cgi?id=1866380

yeah, that's painful only for testing, but not for production and that we care more about

PR #550, opened 2020-09-02
Bug 1812813: oc debug node: create debug namespace with empty node-selector annotation

sallyom:
When a defaultNodeSelector is configured in the scheduler, `oc debug node` fails
if the node to debug does not have the matching labels.  To fix this, with
`oc debug node` create a temporary namespace for the debug pod that has an empty
`openshift.io/node-selector: ""` annotation.  Only create the debug ns if
 the user hasn't explicitly set the debug-pod namespace with
`--to-namespace` or `--namespace|-n`.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
@@ -375,11 +376,49 @@ func (o *DebugOptions) RunDebug() error {
 	if err != nil {
 		klog.V(4).Infof("Unable to get exact template, but continuing with fallback: %v", err)
 	}
+	var debugNodeNS *corev1.Namespace
+	generateName := names.SimpleNameGenerator.GenerateName("openshift-debug-node-")
+	// create namespace only if debugging a node
+	if o.IsNode && len(o.ToNamespace) == 0 && !o.ExplicitNamespace {
+		if !o.Attach.Quiet {
+			fmt.Fprintf(o.ErrOut, "Creating debug namespace/%s ...\n", generateName)
+		}
+		debugNodeNS, err = o.CoreClient.Namespaces().Create(context.TODO(), &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: generateName,
+				Labels: map[string]string{
+					"openshift.io/run-level": "0",
+				},
+				Annotations: map[string]string{
+					"openshift.io/node-selector": "",
+				},
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	defer func() {
+		if debugNodeNS != nil {
+			if !o.Attach.Quiet {
+				fmt.Fprintf(o.ErrOut, "Removing debug namespace/%s ...\n", generateName)
+			}
+
+			if err := o.CoreClient.Namespaces().Delete(context.TODO(), debugNodeNS.Name, metav1.DeleteOptions{}); err != nil {
+				fmt.Fprintf(o.ErrOut, "%v\n", err)
+				return
+			}
+		}
+	}()
+
 	pod := &corev1.Pod{
 		ObjectMeta: template.ObjectMeta,
 		Spec:       template.Spec,
 	}
 	ns := infos[0].Namespace
+	if debugNodeNS != nil {
+		ns = debugNodeNS.Name
+	}
 	if len(ns) == 0 {
 		ns = o.Namespace
 	}


pkg/cli/debug/debug.go
```diff
@@ -338,6 +341,11 @@ func (o *DebugOptions) Complete(cmd *cobra.Command, f kcmdutil.Factory, args []s
 		return err
 	}
 
+	o.RBACClient, err = rbacv1client.NewForConfig(config)
```
deads2k, line 344: if you can make a namespace, you don't need this binding.

pkg/cli/debug/debug.go
```diff
@@ -375,11 +383,67 @@ func (o *DebugOptions) RunDebug() error {
...
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+		// TODO fix this
+		//if debugNodeNS != nil {
+		//  o.Printer.PrintObj(debugNodeNS, o.Out)
+		//}
+		clusterRoleBinding, err = o.RBACClient.ClusterRoleBindings().Create(context.TODO(), o.newClusterRoleBinding(debugNodeNS.Name), metav1.CreateOptions{})
```
deads2k, line 408: wrong scope

pkg/cli/debug/debug.go
```diff
@@ -375,11 +383,67 @@ func (o *DebugOptions) RunDebug() error {
 	if err != nil {
 		klog.V(4).Infof("Unable to get exact template, but continuing with fallback: %v", err)
 	}
+	var debugNodeNS *corev1.Namespace
+	var clusterRoleBinding *rbacv1.ClusterRoleBinding
+	// create namespace only if debugging a node
+	if o.IsNode {
+		debugNodeNS, err = o.CoreClient.Namespaces().Create(context.TODO(), &corev1.Namespace{
```
deads2k, line 390: only do this if the ns isn't explicitly set

pkg/cli/debug/debug.go
sallyom: right, thanks - removed it!

pkg/cli/debug/debug.go
sallyom: removed this

pkg/cli/debug/debug.go
sallyom: right, fixed that, thanks again

pkg/cli/debug/debug.go
```diff
@@ -375,11 +376,52 @@ func (o *DebugOptions) RunDebug() error {
 	if err != nil {
 		klog.V(4).Infof("Unable to get exact template, but continuing with fallback: %v", err)
 	}
+	var debugNodeNS *corev1.Namespace
+	generateName := names.SimpleNameGenerator.GenerateName("openshift-debug-node-")
+	// create namespace only if debugging a node
+	if o.IsNode && len(o.ToNamespace) == 0 && len(o.Namespace) == 0 {
+		if !o.Attach.Quiet {
+			fmt.Fprintf(o.Attach.ErrOut, "\nCreating debug namespace/%s ...\n", generateName)
```
soltysh, line 384: `o.ErrOut`

soltysh: Drop the newline at the beginning too

pkg/cli/debug/debug.go
```diff
@@ -375,11 +376,52 @@ func (o *DebugOptions) RunDebug() error {
...
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	defer func() {
+		if debugNodeNS != nil {
+			if !o.Attach.Quiet {
+				fmt.Fprintf(o.Attach.ErrOut, "Removing debug namespace/%s ...\n", generateName)
```
soltysh, line 404: `o.ErrOut`

pkg/cli/debug/debug.go
```diff
@@ -375,11 +376,52 @@ func (o *DebugOptions) RunDebug() error {
...
+			}
+
+			if err := o.CoreClient.Namespaces().Delete(context.TODO(), debugNodeNS.Name, metav1.DeleteOptions{}); err != nil {
+				fmt.Printf("%v", err)
```
soltysh, line 408: I'd print that to `o.ErrOut` just like above.

pkg/cli/debug/debug.go
```diff
@@ -375,11 +376,52 @@ func (o *DebugOptions) RunDebug() error {
...
 		Spec:       template.Spec,
 	}
-	ns := infos[0].Namespace
+	ns := ""
+	if debugNodeNS != nil {
+		ns = debugNodeNS.Name
+	}
+	if len(ns) == 0 {
+		ns = infos[0].Namespace
+	}
```
soltysh, line 424: The above will be simpler if you write it like this:
```go
ns := infos[0].Namespace
if debugNodeNS != nil {
    ns = debugNodeNS.Name
}
```

soltysh (approved):
A few nits
/approve

pkg/cli/debug/debug.go
sallyom: updated, and also I changed the above`if len(o.Namespace) == 0`  to `if !o.ExplicitNamespace`  bc that was broken

pkg/cli/debug/debug.go
sallyom: updated

pkg/cli/debug/debug.go
sallyom: updated

pkg/cli/debug/debug.go
sallyom: updated

soltysh (approved):
/lgtm
/retest

smarterclayton:
This behavior is really concerning.  Creating a new namespace is a big operation, and assumes a lot of privileges.  I don't think this is the appropriate fix for this problem, but instead, we should tell the user that they may not be able to schedule because they don't have permission to select.  A user can easily work around this problem by switching to another namespace, and creating a new namespace is a very, very dramatic operation.  It also violates a core assumption of debug - that the context you are running in doesn't change.

I would like to revert this and do a much simpler operation like warning the user when we don't schedule, not creating a namespace for them.

sallyom:
> This behavior is really concerning. Creating a new namespace is a big operation, and assumes a lot of privileges. I don't think this is the appropriate fix for this problem, but instead, we should tell the user that they may not be able to schedule because they don't have permission to select. A user can easily work around this problem by switching to another namespace, and creating a new namespace is a very, very dramatic operation. It also violates a core assumption of debug - that the context you are running in doesn't change.
> 
> I would like to revert this and do a much simpler operation like warning the user when we don't schedule, not creating a namespace for them.

There was discussion in the bug report and on slack, and my initial feeling was the same :( 
https://bugzilla.redhat.com/show_bug.cgi?id=1812813#c35

soltysh:
I agree with Clayton, that was my view when we initially put this together.

sallyom:
I'll open a revert PR

PR #553, opened 2020-09-03
oc login: Show tokenURL message if only IDP is basic and user has not provided username

sallyom:
* If a user provides `oc login -u anything`, try for basic auth prompt. 
* If only password IDP available, and no username was provided, show the tokenURL message instead of basic auth prompt.

### DIFF
--- a/pkg/cli/login/login.go
+++ b/pkg/cli/login/login.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/openshift/oc/pkg/helpers/flagtypes"
 	kubeconfiglib "github.com/openshift/oc/pkg/helpers/kubeconfig"
+	"github.com/openshift/oc/pkg/helpers/tokencmd"
 )
 
 var (
@@ -59,10 +60,11 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra
 			kcmdutil.CheckErr(o.Validate(cmd, kcmdutil.GetFlagString(cmd, "server"), args))
 
 			if err := o.Run(); kapierrors.IsUnauthorized(err) {
-				fmt.Fprintln(streams.Out, "Login failed (401 Unauthorized)")
-				fmt.Fprintln(streams.Out, "Verify you have provided correct credentials.")
-
 				if err, isStatusErr := err.(*kapierrors.StatusError); isStatusErr {
+					if err.Status().Message != tokencmd.BasicAuthNoUsernameMessage {
+						fmt.Fprintln(streams.Out, "Login failed (401 Unauthorized)")
+						fmt.Fprintln(streams.Out, "Verify you have provided correct credentials.")
+					}
 					if details := err.Status().Details; details != nil {
 						for _, cause := range details.Causes {
 							fmt.Fprintln(streams.Out, cause.Message)
--- a/pkg/helpers/tokencmd/basicauth.go
+++ b/pkg/helpers/tokencmd/basicauth.go
@@ -14,6 +14,19 @@ import (
 	"github.com/openshift/oc/pkg/helpers/term"
 )
 
+// BasicAuthNoUsernameError is an error that means that basic authentication challenge handling was attempted
+// but the required username was not provided from the command line options
+type BasicAuthNoUsernameError struct{}
+
+func (e *BasicAuthNoUsernameError) Error() string {
+	return "did not receive username - print token url instead of basic prompt"
+}
+
+// NewBasicAuthNoUsernameError returns an error for a basic challenge without a username
+func NewBasicAuthNoUsernameError() error {
+	return &BasicAuthNoUsernameError{}
+}
+
 type BasicChallengeHandler struct {
 	// Host is the server being authenticated to. Used only for displaying messages when prompting for username/password
 	Host string
@@ -54,7 +67,15 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 	missingUsername := len(username) == 0
 	missingPassword := len(password) == 0
 
-	if (missingUsername || missingPassword) && c.Reader != nil {
+	if missingUsername {
+		return nil, false, NewBasicAuthNoUsernameError()
+	}
+	// Basic auth does not support usernames containing colons
+	// http://tools.ietf.org/html/rfc2617#section-2
+	if strings.Contains(username, ":") {
+		return nil, false, fmt.Errorf("username %s is invalid for basic auth", username)
+	}
+	if missingPassword && c.Reader != nil {
 		w := c.Writer
 		if w == nil {
 			w = os.Stdout
@@ -65,11 +86,7 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 		} else {
 			fmt.Fprintf(w, "Authentication required for %s\n", c.Host)
 		}
-		if missingUsername {
-			username = term.PromptForString(c.Reader, w, "Username: ")
-		} else {
-			fmt.Fprintf(w, "Username: %s\n", username)
-		}
+		fmt.Fprintf(w, "Username: %s\n", username)
 		if missingPassword {
 			password = term.PromptForPasswordString(c.Reader, w, "Password: ")
 		}
@@ -77,21 +94,11 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 		c.prompted = true
 	}
 
-	if len(username) > 0 || len(password) > 0 {
-		// Basic auth does not support usernames containing colons
-		// http://tools.ietf.org/html/rfc2617#section-2
-		if strings.Contains(username, ":") {
-			return nil, false, fmt.Errorf("username %s is invalid for basic auth", username)
-		}
-		responseHeaders := http.Header{}
-		responseHeaders.Set("Authorization", getBasicHeader(username, password))
-		// remember so we don't re-handle non-interactively
-		c.handled = true
-		return responseHeaders, true, nil
-	}
-
-	klog.V(2).Info("no username or password available")
-	return nil, false, nil
+	responseHeaders := http.Header{}
+	responseHeaders.Set("Authorization", getBasicHeader(username, password))
+	// remember so we don't re-handle non-interactively
+	c.handled = true
+	return responseHeaders, true, nil
 }
 func (c *BasicChallengeHandler) CompleteChallenge(requestURL string, headers http.Header) error {
 	return nil
@@ -112,6 +119,8 @@ var basicRegexes = []*regexp.Regexp{
 	regexp.MustCompile(`(?i)^\s*basic(?:\s+|$)`),
 }
 
+// basicRealm returns true if a header indicates a basic auth challenge,
+// and the realm if one exists.
 func basicRealm(headers http.Header) (bool, string) {
 	for _, challengeHeader := range headers[http.CanonicalHeaderKey("WWW-Authenticate")] {
 		for _, r := range basicRegexes {
--- a/pkg/helpers/tokencmd/basicauth_test.go
+++ b/pkg/helpers/tokencmd/basicauth_test.go
@@ -44,7 +44,7 @@ func TestHandleChallenge(t *testing.T) {
 					ExpectedCanHandle: true,
 					ExpectedHeaders:   nil,
 					ExpectedHandled:   false,
-					ExpectedErr:       nil,
+					ExpectedErr:       NewBasicAuthNoUsernameError(),
 					ExpectedPrompt:    "",
 				},
 			},
@@ -94,34 +94,6 @@ func TestHandleChallenge(t *testing.T) {
 					ExpectedPrompt: `Authentication required for myhost (myrealm)
 Username: myuser
 Password: `,
-				},
-			},
-		},
-
-		"interactive challenge": {
-			Handler: &BasicChallengeHandler{
-				Host:     "myhost",
-				Reader:   bytes.NewBufferString("myuser\nmypassword\n"),
-				Username: "",
-				Password: "",
-			},
-			Challenges: []Challenge{
-				{
-					Headers:           basicChallenge,
-					ExpectedCanHandle: true,
-					ExpectedHeaders:   http.Header{AUTHORIZATION: []string{getBasicHeader("myuser", "mypassword")}},
-					ExpectedHandled:   true,
-					ExpectedErr:       nil,
-					ExpectedPrompt: `Authentication required for myhost (myrealm)
-Username: Password: `,
-				},
-				{
-					Headers:           basicChallenge,
-					ExpectedCanHandle: true,
-					ExpectedHeaders:   nil,
-					ExpectedHandled:   false,
-					ExpectedErr:       nil,
-					ExpectedPrompt:    ``,
 				},
 			},
 		},
@@ -185,9 +157,7 @@ Username: Password: `,
 					ExpectedHeaders:   nil,
 					ExpectedHandled:   false,
 					ExpectedErr:       errors.New("username system:admin is invalid for basic auth"),
-					ExpectedPrompt: `Authentication required for myhost (myrealm)
-Username: system:admin
-Password: `,
+					ExpectedPrompt:    "",
 				},
 			},
 		},
--- a/pkg/helpers/tokencmd/request_token.go
+++ b/pkg/helpers/tokencmd/request_token.go
@@ -43,6 +43,9 @@ const (
 
 	// token fakes the missing osin.TOKEN const
 	token osincli.AuthorizeRequestType = "token"
+
+	// BasicAuthNoUsernameMessage will differentiate unauthorized errors from basic login with no username
+	BasicAuthNoUsernameMessage = "BasicChallengeNoUsername"
 )
 
 // ChallengeHandler handles responses to WWW-Authenticate challenges.
@@ -225,13 +228,23 @@ func (o *RequestTokenOptions) RequestToken() (string, error) {
 		defer resp.Body.Close()
 
 		if resp.StatusCode == http.StatusUnauthorized {
-			if resp.Header.Get("WWW-Authenticate") != "" {
+			if len(resp.Header.Get("WWW-Authenticate")) > 0 {
 				if !o.Handler.CanHandle(resp.Header) {
 					return "", apierrs.NewUnauthorized("unhandled challenge")
 				}
 				// Handle the challenge
 				newRequestHeaders, shouldRetry, err := o.Handler.HandleChallenge(requestURL, resp.Header)
 				if err != nil {
+					if _, ok := err.(*BasicAuthNoUsernameError); ok {
+						tokenPromptErr := apierrs.NewUnauthorized(BasicAuthNoUsernameMessage)
+						klog.V(4).Infof("%v", err)
+						tokenPromptErr.ErrStatus.Details = &metav1.StatusDetails{
+							Causes: []metav1.StatusCause{
+								{Message: fmt.Sprintf("You must obtain an API token by visiting %s/request", o.OsinConfig.TokenUrl)},
+							},
+						}
+						return "", tokenPromptErr
+					}
 					return "", err
 				}
 				if !shouldRetry {
--- a/pkg/helpers/tokencmd/request_token_test.go
+++ b/pkg/helpers/tokencmd/request_token_test.go
@@ -149,6 +149,7 @@ func TestRequestToken(t *testing.T) {
 
 	basicChallenge1 := resp{401, "", []string{"Basic realm=foo"}}
 	basicRequest1 := req{"Basic bXl1c2VyOm15cGFzc3dvcmQ=", "", ""} // base64("myuser:mypassword")
+	basicRequestOnlyUsername := req{"Basic bXl1c2VyOg==", "", ""}  // base64("myuser:")
 	basicChallenge2 := resp{401, "", []string{"Basic realm=seriously...foo"}}
 
 	negotiateChallenge1 := resp{401, "", []string{"Negotiate"}}
@@ -213,18 +214,27 @@ func TestRequestToken(t *testing.T) {
 			},
 			ExpectedError: "unhandled challenge",
 		},
-		"failing basic handler, basic challenge, failure": {
+		"no username, basic challenge, failure": {
 			Handler: &BasicChallengeHandler{},
 			Requests: []requestResponse{
 				{initialHead, initialHeadResp},
 				{initialRequest, basicChallenge1},
+			},
+			ExpectedError: BasicAuthNoUsernameMessage,
+		},
+		"failing basic handler, basic challenge, failure": {
+			Handler: &BasicChallengeHandler{Username: "myuser"},
+			Requests: []requestResponse{
+				{initialHead, initialHeadResp},
+				{initialRequest, basicChallenge1},
+				{basicRequestOnlyUsername, basicChallenge1},
 			},
 			ExpectedError: "challenger chose not to retry the request",
 		},
 
 		// Prompting basic handler
 		"prompting basic handler, no challenge, success": {
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+			Handler: &BasicChallengeHandler{Username: "myuser", Reader: bytes.NewBufferString("mypassword\n")},
 			Requests: []requestResponse{
 				{initialHead, initialHeadResp},
 				{initialRequest, success},
@@ -232,7 +242,7 @@ func TestRequestToken(t *testing.T) {
 			ExpectedToken: successfulToken,
 		},
 		"prompting basic handler, basic challenge, success": {
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+			Handler: &BasicChallengeHandler{Username: "myuser", Reader: bytes.NewBufferString("mypassword\n")},
 			Requests: []requestResponse{
 				{initialHead, initialHeadResp},
 				{initialRequest, basicChallenge1},
@@ -241,7 +251,7 @@ func TestRequestToken(t *testing.T) {
 			ExpectedToken: successfulToken,
 		},
 		"prompting basic handler, basic+negotiate challenge, success": {
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+			Handler: &BasicChallengeHandler{Username: "myuser", Reader: bytes.NewBufferString("mypassword\n")},
 			Requests: []requestResponse{
 				{initialHead, initialHeadResp},
 				{initialRequest, doubleChallenge},
@@ -250,11 +260,11 @@ func TestRequestToken(t *testing.T) {
 			ExpectedToken: successfulToken,
 		},
 		"prompting basic handler, basic challenge, failure": {
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+			Handler: &BasicChallengeHandler{Username: "myuser"},
 			Requests: []requestResponse{
 				{initialHead, initialHeadResp},
 				{initialRequest, basicChallenge1},
-				{basicRequest1, basicChallenge2},
+				{basicRequestOnlyUsername, basicChallenge2},
 			},
 			ExpectedError: "challenger chose not to retry the request",
 		},
@@ -381,7 +391,7 @@ func TestRequestToken(t *testing.T) {
 		"failing negotiate+prompting basic handler, negotiate+basic challenge, success": {
 			Handler: NewMultiHandler(
 				&NegotiateChallengeHandler{negotiator: &failingNegotiator{}},
-				&BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+				&BasicChallengeHandler{Username: "myuser", Reader: bytes.NewBufferString("mypassword\n")},
 			),
 			Requests: []requestResponse{
 				{initialHead, initialHeadResp},


sallyom:
/assign @soltysh 
/cc @stlaz 

this is a guess as to what should happen

soltysh:
I don't think this is a good approach, b/c you're basically forcing username to be passed to pick basic auth. I was rather imagining that when you get information from the server about supported auth methods, and if only basic you'd go with that always. But when more than basic is present always prefer the other one. @stlaz might help you with that and I think that's what @smarterclayton had in mind

soltysh:
The actual response you're faking here is coming from oauth-server, from here: https://github.com/openshift/oauth-server/blob/dcbeb48c9cedcf0827629556f1e6cdb8538d4ebe/pkg/authenticator/challenger/placeholderchallenger/placeholder_challenger.go#L24

sallyom:
> The actual response you're faking here is coming from oauth-server, from here: https://github.com/openshift/oauth-server/blob/dcbeb48c9cedcf0827629556f1e6cdb8538d4ebe/pkg/authenticator/challenger/placeholderchallenger/placeholder_challenger.go#L2

I saw that, ok  - i'll rework this.

sallyom:
> I don't think this is a good approach, b/c you're basically forcing username to be passed to pick basic auth. I was rather imagining that when you get information from the server about supported auth methods, and if only basic you'd go with that always. But when more than basic is present always prefer the other one. @stlaz might help you with that and I think that's what @smarterclayton had in mind

I've reworked this:
[Change in openshift/oauth-server ](https://github.com/openshift/oauth-server/pull/55) to mark the bootstrap basic auth with realm "bootstrap" (as @stlaz and I discussed via slack).  Here in `oc` if bootstrap basic challenge and no other password IDP & no username passed, then show the tokenURL.

stlaz (changes requested):
you're also going to have to improve
https://github.com/openshift/oc/blob/ea0d54068621ec0f95973068729f739f3dacfef7/pkg/helpers/tokencmd/basicauth.go#L115
so that it prefers non-kubeadmin realms

pkg/helpers/tokencmd/request_token.go
```diff
@@ -224,11 +226,29 @@ func (o *RequestTokenOptions) RequestToken() (string, error) {
...
 
+		unauthorizedError := apierrs.NewUnauthorized("")
 		if resp.StatusCode == http.StatusUnauthorized {
-			if resp.Header.Get("WWW-Authenticate") != "" {
+			header := resp.Header.Get("WWW-Authenticate")
+			if len(header) > 0 {
 				if !o.Handler.CanHandle(resp.Header) {
 					return "", apierrs.NewUnauthorized("unhandled challenge")
 				}
+				if strings.Contains(header, "bootstrap") && !o.BootstrapBasic {
```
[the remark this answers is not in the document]
sallyom: updated the realm in the oauth-server PR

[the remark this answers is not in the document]
sallyom: updated the handling, also

sallyom:
@stlaz pt-another-l, thanks

stlaz (changes requested):
is there any specific code that will handle the warning received from the oauth-server? I couldn't find where the message about the token URL is actually printed in the case of the placeholderchallenger on the oauth-server side

pkg/helpers/tokencmd/request_token.go
```diff
@@ -105,6 +105,8 @@ func NewRequestTokenOptions(clientCfg *restclient.Config, reader io.Reader, defa
 		ClientConfig: clientCfg,
 		Handler:      handler,
 		TokenFlow:    tokenFlow,
+		// if bootstrap user, display tokenURL if haven't passed username
+		BootstrapBasic: defaultUsername == bootstrap_user,
```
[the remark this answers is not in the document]
sallyom: > is there any specific code that will handle the warning received from the oauth-server? I couldn't find where the message about the token URL is actually printed in the case of the placeholderchallenger on the oauth-server side

it's here: https://github.com/openshift/oauth-server/blob/master/pkg/authenticator/challenger/placeholderchallenger/placeholder_challenger.go#L24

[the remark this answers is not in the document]
sallyom: so yea, we're faking that msg by adding it to oc.

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -14,6 +15,20 @@ import (
...
+
+	// bootstrap_user is the username for basic auth bootstrap user
+	bootstrap_user = "kubeadmin"
+
+	tokenURLPromptErrReason = "token URL prompt"
+)
+
+func IsTokenURLPromptErr(err error) bool {
+	return err.Error() == tokenURLPromptErrReason
+}
```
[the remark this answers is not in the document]
sallyom: yup i wasn't fully committed to the custom error so took a shortcut

pkg/helpers/tokencmd/request_token.go
```diff
@@ -225,13 +227,27 @@ func (o *RequestTokenOptions) RequestToken() (string, error) {
 		defer resp.Body.Close()
 
 		if resp.StatusCode == http.StatusUnauthorized {
-			if resp.Header.Get("WWW-Authenticate") != "" {
+			header := resp.Header.Get("WWW-Authenticate")
+			if len(header) > 0 {
```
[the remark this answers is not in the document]
sallyom: updated

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -116,8 +139,16 @@ func basicRealm(headers http.Header) (bool, string) {
...
 				if len(matches) > 1 {
-					// We got a realm as well
+					if !strings.Contains(challengeHeader, bootstrap_realm) {
+						// We got non-bootstrap realm
+						return true, matches[1]
+					}
+					matched_bootstrap = true
+				}
+				// We got bootstrap realm
+				if matched_bootstrap {
```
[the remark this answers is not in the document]
sallyom: of course, duh, thanks

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -116,8 +139,16 @@ func basicRealm(headers http.Header) (bool, string) {
 	for _, challengeHeader := range headers[http.CanonicalHeaderKey("WWW-Authenticate")] {
 		for _, r := range basicRegexes {
 			if matches := r.FindStringSubmatch(challengeHeader); matches != nil {
+				matched_bootstrap := false
 				if len(matches) > 1 {
-					// We got a realm as well
+					if !strings.Contains(challengeHeader, bootstrap_realm) {
```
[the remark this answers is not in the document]
sallyom: this causes an issue, as matches[1] `"openshift:kubeadmin"` never matches bootstrapRealm `openshift:kubeadmin`  so strip the `""`?  I'm retesting/will report back.. this doesn't quite add up, maybe I was doing something wrong.

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -116,8 +139,16 @@ func basicRealm(headers http.Header) (bool, string) {
 	for _, challengeHeader := range headers[http.CanonicalHeaderKey("WWW-Authenticate")] {
 		for _, r := range basicRegexes {
 			if matches := r.FindStringSubmatch(challengeHeader); matches != nil {
+				matched_bootstrap := false
```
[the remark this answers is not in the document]
sallyom: moved/fixed

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -113,18 +146,26 @@ var basicRegexes = []*regexp.Regexp{
 }
 
 func basicRealm(headers http.Header) (bool, string) {
+	matchedBootstrap := false
```
stlaz, line 149: nit: I think we should now document what `basicRealm` does in a godoc + please write a unit test :)

pkg/helpers/tokencmd/request_token.go
```diff
@@ -225,13 +224,27 @@ func (o *RequestTokenOptions) RequestToken() (string, error) {
...
 				}
 				// Handle the challenge
 				newRequestHeaders, shouldRetry, err := o.Handler.HandleChallenge(requestURL, resp.Header)
 				if err != nil {
+					if IsBootstrapPromptErr(err) {
+						tokenPromptErr := apierrs.NewUnauthorized("")
+						klog.V(4).Infof("%v", err)
+						msg := fmt.Sprintf("You must obtain an API token by visting %s/request\n", o.OsinConfig.TokenUrl)
+						details, err := ioutil.ReadAll(resp.Body)
+						if err == nil && len(details) == 0 {
```
stlaz, line 239: details are unused

pkg/helpers/tokencmd/request_token.go
```diff
@@ -225,13 +224,27 @@ func (o *RequestTokenOptions) RequestToken() (string, error) {
...
+							tokenPromptErr.ErrStatus.Details = &metav1.StatusDetails{
+								Causes: []metav1.StatusCause{
+									{Message: msg},
+								},
+							}
+						}
+						return "", tokenPromptErr
```
stlaz, line 246: should this be just
```golang
unauthorizedErr := apierrs.NewUnauthorized("")
unauthorizedErr.ErrStatus.Details = &metav1.StatusDetails{
    Causes: []metav1.StatusCause{
        {Message: fmt.Sprintf("You must obtain an API token by visting %s/request\n", o.OsinConfig.TokenUrl)},
    },
}
return "", unauthorizedErr
```

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -85,6 +115,9 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 		}
 		responseHeaders := http.Header{}
 		responseHeaders.Set("Authorization", getBasicHeader(username, password))
+		if realm == bootstrapRealm && username != bootstrapUser {
+			return nil, false, BootstrapPromptError()
+		}
```
stlaz, line 120: leave this to the challenging server

stlaz (changes requested):

pkg/helpers/tokencmd/basicauth.go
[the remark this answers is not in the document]
sallyom: updated this

pkg/helpers/tokencmd/basicauth.go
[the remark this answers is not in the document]
sallyom: oh hm there's `strconv.Quote(str)`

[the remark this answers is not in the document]
sallyom: updated

pkg/helpers/tokencmd/request_token.go
sallyom: Don't need to read Body at all, all that's needed is whether err is bootstrapPromptErr.
Updated.

pkg/helpers/tokencmd/request_token.go
sallyom: updated

pkg/helpers/tokencmd/basicauth.go
sallyom: removed

pkg/helpers/tokencmd/basicauth.go
sallyom: godoc added, unit test cases added

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -54,13 +69,17 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 	missingUsername := len(username) == 0
 	missingPassword := len(password) == 0
 
+	_, realm := basicRealm(headers)
```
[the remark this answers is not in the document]
sallyom: updated

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -112,19 +141,29 @@ var basicRegexes = []*regexp.Regexp{
...
 			if matches := r.FindStringSubmatch(challengeHeader); matches != nil {
-				if len(matches) > 1 {
-					// We got a realm as well
-					return true, matches[1]
+				if !strings.Contains(challengeHeader, bootstrapRealm) {
+					if len(matches) > 1 {
+						return true, matches[1]
+					}
+					// No realm, but still basic
+					return true, ""
```
sallyom, line 157: @stlaz, should `basic no realm` be favored over `basic openshift:kubeadmin`?

sallyom:
@stlaz, updated this PR, I fixed all the issues and added unit tests. I have a question about the `basic no realm` see above.  thanks again  for your review.

pkg/helpers/tokencmd/basicauth.go
stlaz: yes, it should

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -112,19 +141,29 @@ var basicRegexes = []*regexp.Regexp{
...
+// basic challenge is favored.
 func basicRealm(headers http.Header) (bool, string) {
+	matchedBootstrap := false
 	for _, challengeHeader := range headers[http.CanonicalHeaderKey("WWW-Authenticate")] {
 		for _, r := range basicRegexes {
 			if matches := r.FindStringSubmatch(challengeHeader); matches != nil {
-				if len(matches) > 1 {
-					// We got a realm as well
-					return true, matches[1]
+				if !strings.Contains(challengeHeader, bootstrapRealm) {
```
smarterclayton, line 152: Knowing about the bootstrap realm in the CLI is super sketchy.  Bootstrap isn't special - it could just as easily be an htpasswd IDP.  Why would we treat kubeadmin specially?

smarterclayton: I.e., if I have the following IDPs:

```
htpasswd - one admin user
oidc - my org
```

that is no different than 

```
kubeadmin
oidc - my org
```

I don't think bootstrap should be what we determine this based on.  The client should probably always have the option of going to the token endpoint, no matter what the IDP configured is.  If I have:

```
htpasswd - one user
```

i still may prefer to go to the token endpoint than to use basic auth, since i don't have to type my password on the machine to do it.

sallyom: so.. any time no username passed, show the token URL?

sallyom: kubeadmin is different bc the user didn't specifically configure it, it comes out of the box. With htpasswd, user chose that, so thinking is that they'd want the prompt there.

stlaz: > Bootstrap isn't special

it *is* special, it's is the only opt-out IdP

> I don't think bootstrap should be what we determine this based on. The client should probably always have the option of going to the token endpoint, no matter what the IDP configured is.

add a flag `--show-token-url` then or make even the basic auth handling always print:
```
You can also get your token at %(url)

Username:
Password:
```

> since i don't have to type my password on the machine to do it.

You do, the browser is usually on your machine just as well. If you're using a password manager like LastPass, you can still do `oc login -u user -p - < $(lpass show openshift_user --password)`

> so.. any time no username passed, show the token URL?

no

sallyom: We can add a line of output to the prompt then, showing the 'you can also get your token tokenURL' and still keep the kubeadmin idp/no username special realm to _not_ show prompt in only that case.

stlaz: I slept on it, it's probably not a good idea to add the additional realm after all. It adds a side-channel which reveals existence of a super-user to an anonymous user. That means they would know there's most probably a file somewhere with a password to it, and it's highly possible that the password is going to live somewhere in mails/chats (although passwords shouldn't be shared but the real world teaches us otherwise time and time again) and/or social engineering it is a possibility.

If always adding a line of text would help the situation, I think it's the simplest thing we can do. I don't think we should be touching how challenges are handled, though.

soltysh: @smarterclayton are you basically proposing we pick basic auth as the last resort, not looking at what challenges the server presents, except for when you explicitly ask for basic-auth by passing `--username` flag?

sallyom: I've updated to that effect - if only idp is a password idp and a user does not pass --username they'll get the tokenURL message.

pkg/helpers/tokencmd/request_token_test.go
```diff
@@ -378,18 +357,6 @@ func TestRequestToken(t *testing.T) {
 			},
 			ExpectedToken: successfulToken,
 		},
-		"failing negotiate+prompting basic handler, negotiate+basic challenge, success": {
```
smarterclayton, line 381: Why is this test removed, vs changed to expected error

sallyom: updated, added that case back in w/ best guess, ptal 
was going to ask @stlaz about that as i'm a bit confused about negotiate challenge flow/forgot to ping him b4 he left today

pkg/helpers/tokencmd/request_token_test.go
```diff
@@ -219,44 +219,23 @@ func TestRequestToken(t *testing.T) {
 				{initialHead, initialHeadResp},
 				{initialRequest, basicChallenge1},
 			},
-			ExpectedError: "challenger chose not to retry the request",
-		},
-
-		// Prompting basic handler
-		"prompting basic handler, no challenge, success": {
```
smarterclayton, line 226: I expected to keep the test and change the outcome

pkg/helpers/tokencmd/request_token.go
```diff
@@ -225,13 +225,23 @@ func (o *RequestTokenOptions) RequestToken() (string, error) {
...
 				}
 				// Handle the challenge
 				newRequestHeaders, shouldRetry, err := o.Handler.HandleChallenge(requestURL, resp.Header)
 				if err != nil {
+					if IsBasicNoUsernameError(err) {
+						tokenPromptErr := apierrs.NewUnauthorized("")
+						klog.V(4).Infof("%v", err)
+						tokenPromptErr.ErrStatus.Details = &metav1.StatusDetails{
+							Causes: []metav1.StatusCause{
+								{Message: fmt.Sprintf("You must obtain an API token by visting %s/request\n", o.OsinConfig.TokenUrl)},
```
smarterclayton, line 240: Remove trailing newline, fix to `visiting`.  Is this consistent with the server message?

sallyom: yes, but i'll double check

sallyom: oops fixed spelling
removed newline, but checking w/ the 3.11 api.ci cluster server msg has a newline

sallyom: https://github.com/openshift/oauth-server/blob/master/pkg/authenticator/challenger/placeholderchallenger/placeholder_challenger.go#L23-#L30

pkg/helpers/tokencmd/request_token_test.go
sallyom: this case will never happen, AFAICT, so took it out but maybe i'm missing something - i've added it back, with a failure instead.

sallyom:
@stlaz ptal at the unit test changes, head is spinning w/ those, thank you : )

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -14,6 +14,23 @@ import (
...
+
+func BasicNoUsernameError() error {
+	return &basicNoUsername{}
+}
+
+func IsBasicNoUsernameError(err error) bool {
+	if _, ok := err.(*basicNoUsername); ok {
+		return true
+	}
+	return false
```
stlaz, line 31:
```
_, ok := err.(*basicNoUsername)
return ok
```
or export the error an inline it?

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -54,13 +71,17 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 	missingUsername := len(username) == 0
 	missingPassword := len(password) == 0
 
+	isBasic, realm := basicRealm(headers)
+	if missingUsername && isBasic {
```
stlaz, line 75: you already know basic auth is possible at this point so you can move getting the realm to the point of use and fail if username is missing

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -54,13 +71,17 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
...
+		return nil, false, BasicNoUsernameError()
+	}
 	if (missingUsername || missingPassword) && c.Reader != nil {
```
stlaz, line 78: ...and may omit `missingUsername` here

pkg/helpers/tokencmd/basicauth_test.go
```diff
@@ -101,27 +101,17 @@ Password: `,
...
 					ExpectedCanHandle: true,
 					ExpectedHeaders:   nil,
 					ExpectedHandled:   false,
-					ExpectedErr:       nil,
-					ExpectedPrompt:    ``,
+					ExpectedErr:       BasicNoUsernameError(),
+					ExpectedPrompt:    "",
 				},
 			},
 		},
```
stlaz, line 117: this now equals `"non-interactive with no defaults"` and you're testing the only other possible scenario with `"interactive challenge with default user"` so you can remove this test case

soltysh (approved):
/approve
there are some nits from Standa left, so I'm leaving him final tag

stlaz (changes requested):
Needs further improvements and refactoring due to the changes of basic challenge handler.

The test case changes seem wrong -> you can add more but keep the behavior of the old tests. Unless I missed anything, the old tests weren't prompting for password only so just do that instead of turning them into negative tests.

stlaz:
/hold
to prevent random lgtms :)

pkg/helpers/tokencmd/request_token_test.go
```diff
@@ -378,18 +365,18 @@ func TestRequestToken(t *testing.T) {
 			},
 			ExpectedToken: successfulToken,
 		},
-		"failing negotiate+prompting basic handler, negotiate+basic challenge, success": {
+		"failing negotiate+prompting basic handler, negotiate+basic challenge, failure": {
```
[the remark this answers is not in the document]
sallyom: still confused, will ask tomorrow

pkg/helpers/tokencmd/request_token_test.go
```diff
@@ -219,44 +219,31 @@ func TestRequestToken(t *testing.T) {
...
 				{initialHead, initialHeadResp},
 				{initialRequest, doubleChallenge},
-				{basicRequest1, success},
 			},
-			ExpectedToken: successfulToken,
+			ExpectedError: BasicNoUsernameMessage,
 		},
 		"prompting basic handler, basic challenge, failure": {
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+			Handler: &BasicChallengeHandler{},
```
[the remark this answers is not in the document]
sallyom: still confused by this! i'll reach out tomorrow

pkg/helpers/tokencmd/request_token_test.go
```diff
@@ -219,44 +219,31 @@ func TestRequestToken(t *testing.T) {
...
 				{initialHead, initialHeadResp},
 				{initialRequest, basicChallenge1},
-				{basicRequest1, success},
 			},
-			ExpectedToken: successfulToken,
+			ExpectedError: BasicNoUsernameMessage,
 		},
-		"prompting basic handler, basic+negotiate challenge, success": {
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+		"prompting basic handler, basic+negotiate challenge, failure": {
```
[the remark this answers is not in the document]
sallyom: still confused by this! i'll reach out tomorrow

pkg/helpers/tokencmd/request_token_test.go
```diff
@@ -219,44 +219,31 @@ func TestRequestToken(t *testing.T) {
...
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
-			Requests: []requestResponse{
-				{initialHead, initialHeadResp},
-				{initialRequest, success},
-			},
-			ExpectedToken: successfulToken,
-		},
-		"prompting basic handler, basic challenge, success": {
-			Handler: &BasicChallengeHandler{Reader: bytes.NewBufferString("myuser\nmypassword\n")},
+		"prompting basic handler, no challenge, failure": {
```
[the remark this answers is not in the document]
sallyom: right,i think this can be omitted and the TC you suggested above added.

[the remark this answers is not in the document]
sallyom: updated

pkg/helpers/tokencmd/request_token_test.go
```diff
@@ -219,44 +219,31 @@ func TestRequestToken(t *testing.T) {
 				{initialHead, initialHeadResp},
 				{initialRequest, basicChallenge1},
 			},
-			ExpectedError: "challenger chose not to retry the request",
```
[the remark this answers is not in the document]
sallyom: added : )

pkg/helpers/tokencmd/basicauth_test.go
sallyom: removed "interactive challenge" case,

pkg/helpers/tokencmd/basicauth.go
sallyom: ah, right, thanks - got it, updated

pkg/helpers/tokencmd/basicauth.go
sallyom: right, updated.

pkg/helpers/tokencmd/basicauth.go
sallyom: done, exported/inline, thanks!

pkg/helpers/tokencmd/request_token_test.go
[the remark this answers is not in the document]
sallyom: changed this to `Handler: &BasicChallengeHandler{Username: "myuser"},` then
`{basicRequestOnlyUsername, basicChallenge2},` and keeping same error

pkg/helpers/tokencmd/request_token_test.go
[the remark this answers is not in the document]
sallyom: for this i added `&BasicChallengeHandler{Username: "myuser", Reader: bytes.NewBufferString("mypassword\n")},` thanks for clarifying :blush:

pkg/helpers/tokencmd/request_token_test.go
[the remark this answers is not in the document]
sallyom: and for this, again adding `BasicChallengeHandler{Username: "myuser", Reader: bytes.NewBufferString("mypassword\n")}` and keeping success 
Thanks, again, @stlaz

pkg/helpers/tokencmd/request_token_test.go
sallyom: updated

pkg/helpers/tokencmd/request_token_test.go
sallyom: oops added this back, with `Handler: &BasicChallengeHandler{Username: "myuser", Reader: bytes.NewBufferString("mypassword\n")},` and successfulToken

pkg/helpers/tokencmd/basicauth_test.go
```diff
@@ -98,34 +98,6 @@ Password: `,
...
 
-		"interactive challenge": {
```
soltysh, line 101: @sallyom can this be changed to half-interactive when you pass username it should still ask you for a password, or we have one covering that case?

sallyom: this can be omitted, as the one above covers that, it's `"interactive challenge with default user"`

sallyom: (afaict :) )

soltysh: :+1:

sallyom:
@stlaz I've updated all test cases, i think you can remove the hold now, and thanks

stlaz (commented):
nits only \o/

The PR looks good to me now

pkg/cli/login/login.go
```diff
@@ -59,10 +60,12 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra
 			kcmdutil.CheckErr(o.Validate(cmd, kcmdutil.GetFlagString(cmd, "server"), args))
 
 			if err := o.Run(); kapierrors.IsUnauthorized(err) {
-				fmt.Fprintln(streams.Out, "Login failed (401 Unauthorized)")
-				fmt.Fprintln(streams.Out, "Verify you have provided correct credentials.")
-
-				if err, isStatusErr := err.(*kapierrors.StatusError); isStatusErr {
+				err, isStatusErr := err.(*kapierrors.StatusError)
+				if isStatusErr {
```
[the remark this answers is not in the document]
sallyom: updated, thanks

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -14,6 +14,16 @@ import (
 	"github.com/openshift/oc/pkg/helpers/term"
 )
 
+type BasicNoUsername struct{}
```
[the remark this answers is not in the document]
sallyom: added!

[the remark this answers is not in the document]
stlaz:
```
// BasicAuthNoUsernameError is an error that means that basic authentication challenge handling was attempted but the required username was not provided from the command line options
```

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -14,6 +14,18 @@ import (
...
 
+// BasicNoUsername enables BasicNoUsernameError implementation
+type BasicNoUsername struct{}
+
+func (e *BasicNoUsername) Error() string {
+	return "did not receive username - print token url instead of basic prompt"
+}
+
+// BasicNoUsernameError returns an error for a basic challenge without a username
+func BasicNoUsernameError() error {
```
stlaz, line 25: `NewBasicAuthNoUsernameError()`

stlaz (commented):
last nits, I promise

pkg/helpers/tokencmd/basicauth.go
sallyom: yes, much better, updated : )

Also, in request_token.go, updated `BasicNoUsernameMessage` to `BasicAuthNoUsernameMessage`

pkg/helpers/tokencmd/basicauth.go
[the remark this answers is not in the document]
sallyom: updated, much better, thanks

PR #540, opened 2020-08-31
bug 1805891: must-gather: move gather init container under containers

ingvagabund:
Move `gather` container from init containers into containers so the pod does not stay in `Pending` state until gather container is finished. Otherwise, `KubePodNotReady` alert gets triggered.

Once the gathering phase is finished, must-gather pod drops to NotReady state since `gather` container terminates:
```
NAME                READY   STATUS     RESTARTS   AGE
must-gather-sr295   1/2     NotReady   0          4m18s
```

Having a pod in `NotReady` status might still be better than having it in `Pending`. The copy phase should take less than the gathering one (depending on the network bandwidth).

PR #554, opened 2020-09-03
bug 1875551: must-gather: move gather init container under containers

ingvagabund:
Move `gather` container from init containers into containers so the pod does not stay in `Pending` state until gather container is finished. Otherwise, `KubePodNotReady` alert gets triggered.

Once the gathering phase is finished, must-gather pod drops to NotReady state since `gather` container terminates:
```
NAME                READY   STATUS     RESTARTS   AGE
must-gather-sr295   1/2     NotReady   0          4m18s
```

Having a pod in `NotReady` status might still be better than having it in `Pending`. The copy phase should take less than the gathering one (depending on the network bandwidth).

Backporting https://github.com/openshift/oc/pull/540

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -300,7 +300,7 @@ func (o *MustGatherOptions) Run() error {
 
 			// wait for pod to be running (gather has completed)
 			log("waiting for gather to complete")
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/approve

ingvagabund:
Bot seems to be stucked for 7 hours in "tide Pending — Not mergeable. Needs cherry-pick-approved label"

/approve

PR #558, opened 2020-09-06
cli/admin/release/new: typo fix

miabbott:
Looks like the correct word in the help should be `precede`


soltysh (approved):
/approve
/retest

PR #566, opened 2020-09-14
Bug 1878925: pkg/cli/admin/upgrade: Remove help text around history lookups

wking:
Copying from [the CVO][1], and adjusting to return a pointer instead of a (non-pointer,found) tuple.  Future work can shift this function into library-go.  This avoids:

```console
$ oc adm upgrade --to 4.6.0-fc.5
error: Can't look up image for version 4.6.0-fc.5. The update channel has not been configured.
```

When that lookup would succeed when investigating ClusterVersion's `status.history`:

```console
$ oc patch clusterversion version --type json -p '[{"op": "add", "path": "/spec/desiredUpdate", "value": {"version": "4.6.0-fc.5"}}]'
$ oc adm upgrade
info: An upgrade is in progress. Working towards 4.6.0-fc.5: downloading update

warning: Cannot display available updates:
  Reason: NoChannel
  Message: The update channel has not been configured.
```

[1]: https://github.com/openshift/cluster-version-operator/blob/6d56c655ea16f6faee4b65ffef43dcd912657bc6/pkg/cvo/updatepayload.go#L305-L325

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -78,7 +78,7 @@ func New(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command
 		},
 	}
 	flags := cmd.Flags()
-	flags.StringVar(&o.To, "to", o.To, "Specify the version to upgrade to. The version must be on the list of previous or available updates.")
+	flags.StringVar(&o.To, "to", o.To, "Specify the version to upgrade to. The version must be on the list of available updates.")
 	flags.StringVar(&o.ToImage, "to-image", o.ToImage, "Provide a release image to upgrade to. WARNING: This option does not check for upgrade compatibility and may break your cluster.")
 	flags.BoolVar(&o.ToLatestAvailable, "to-latest", o.ToLatestAvailable, "Use the next available version")
 	flags.BoolVar(&o.Clear, "clear", o.Clear, "If an upgrade has been requested but not yet downloaded, cancel the update. This has no effect once the update has started.")


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -221,23 +221,15 @@ func (o *Options) Run() error {
...
+			update = findUpdateFromConfigVersion(cv, o.To)
 			if update == nil {
 				if len(cv.Status.AvailableUpdates) == 0 {
 					if c := findCondition(cv.Status.Conditions, configv1.RetrievedUpdates); c != nil && c.Status == configv1.ConditionFalse {
 						return fmt.Errorf("Can't look up image for version %s. %v", o.To, c.Message)
 					}
 					return fmt.Errorf("No available updates, specify --to-image or wait for new updates to be available")
 				}
-				return fmt.Errorf("The update %s is not one of the available updates: %s", o.To, strings.Join(versionStrings(cv.Status.AvailableUpdates), ", "))
+				return fmt.Errorf("The update %s is neither one of the available updates (%s) nor a history entry", o.To, strings.Join(versionStrings(cv.Status.AvailableUpdates), ", "))
```
LalatenduMohanty, line 232: Why can not we just say `The %s is not available to update`? Do we really need to print the available update versions?

LalatenduMohanty: I not comfortable in printing about `nor a history entry` when we do not encourage customers to downgrade.

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/upgrade.go
wking: Should we put this behind `--allow-versions-from-history` or some such with a warning when it's needed, like our other `--allow-...` handling?

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -460,3 +452,23 @@ func checkForUpgrade(cv *configv1.ClusterVersion) error {
...
+func findUpdateFromConfigVersion(config *configv1.ClusterVersion, version string) *configv1.Update {
+	for _, update := range config.Status.AvailableUpdates {
+		if update.Version == version && len(update.Image) > 0 {
+			return &configv1.Update{
+				Version: version,
+				Image:   update.Image,
+			}
+		}
+	}
+	for _, history := range config.Status.History {
```
jottofar, line 465: Is it okay to bypass the channel lookup for "available updates", i.e. is it okay to assume that just because a version was in the channel at some point in "history" that it still is? For example, the cluster may have been updated to 4.6.0-fc.5 in the past but we have since pulled that version for some reason.

wking: Let's collapse warning about reaching into history into [this thread][1].

[1]: https://github.com/openshift/oc/pull/566#discussion_r488575538

pkg/cli/admin/upgrade/upgrade.go
jottofar: But is a version from history still necessarily available in channel for an update?

wking: Good point from both of you about having a guard around downgrades.  I've pushed 2b9210665 -> adb05a76b to wire this up to the existing `--allow-explicit-upgrade`.  If `--to ...` picks a value from `availableUpdates`, it doesn't matter whether you've set `--allow-explicit-upgrade` or not.  But with adb05a76b, when `--to ...` picks a value from the history, you will either fail to update (when `--allow-explicit-upgrade` was not set) or successfully trigger the update but with a stderr complaint (when `--allow-explicit-upgrade` was set).  I think that addresses the safety concerns.

> Do we really need to print the available update versions?

No, but mentioning them means folks can skip a separate `oc adm upgrade` step that may return a different list of available updates than the set we just used to reject their choice.

> I not comfortable in printing about `nor a history entry`...

Does the new `--allow-explicit-upgrade` integration address your concerns there?

> But is a version from history still necessarily available in channel for an update?

Nope.  And the point of this PR is allowing these updates to previous versions without requiring either `oc patch ...` or folks looking up the by-digest pullspec themselves.  But yes, there is a safety issue with rolling the cluster back.  Does the new `--allow-explicit-upgrade` integration address your concerns there?

LalatenduMohanty: @wking `nor a history entry` , will the customer understand what this means? If yes then I am ok with this one.

wking: > @wking nor a history entry , will the customer understand what this means?

Are you asking for a rephrase that says "you can draw from either set" without using "neither ... nor ..."?  Or something that mentions the fact this history entries may not be recommended updates?  Or...?

LalatenduMohanty: Here is my suggestion `The update %s is neither one of the available updates (%s) nor present in the update history`

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -221,23 +221,21 @@ func (o *Options) Run() error {
...
 					}
 					return fmt.Errorf("No available updates, specify --to-image or wait for new updates to be available")
 				}
-				return fmt.Errorf("The update %s is not one of the available updates: %s", o.To, strings.Join(versionStrings(cv.Status.AvailableUpdates), ", "))
+				return fmt.Errorf("The update %s is neither one of the available updates (%s) nor a history entry", o.To, strings.Join(versionStrings(cv.Status.AvailableUpdates), ", "))
+			} else if !fromAvailableUpdates {
+				if !o.AllowExplicitUpgrade {
+					return fmt.Errorf("The requested upgrade image is not one of the available updates, you must pass --allow-explicit-upgrade to continue")
+				}
+				fmt.Fprintln(o.ErrOut, "warning: The requested upgrade version is not one of the available updates.  You have used --allow-explicit-upgrade to the update to proceed anyway")
```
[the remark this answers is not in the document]
wking: It matches [the existing pattern][1] I used in previous work which has already landed.  If you would rather not point out why particular guard failures are being waved, can you submit a master PR to remove the existing instances?

[1]: https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/cli/admin/upgrade/upgrade.go#L265

pkg/cli/admin/upgrade/upgrade.go
wking: > Here is my suggestion...

Rebased onto master and updated to use your text with adb05a76b -> 754657b01.

wking:
I've pushed 754657b01 -> fcc58db01 to pivot to dropping `previous` from the `--help` text, per @sdodson's request.

PR #569, opened 2020-09-16
Bug 1861746: run inspect in parallel

soltysh:
This brings back commit 7ec02e2eacdf66c5a6b77e44ee779398d5aee80b which was reverted as a culprit to https://bugzilla.redhat.com/show_bug.cgi?id=1861746 which it actually was not.

### DIFF
--- a/pkg/cli/admin/inspect/resource.go
+++ b/pkg/cli/admin/inspect/resource.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/inspect/util.go
```diff
@@ -30,15 +31,35 @@ import (
 
 // resourceContext is used to keep track of previously seen objects
 type resourceContext struct {
-	visited sets.String
+	lock sync.Mutex
```
mfojtik, line 34: is it possible to use RWMutex ?

pkg/cli/admin/inspect/resource.go
```diff
@@ -137,15 +165,11 @@ func gatherRelatedObjects(context *resourceContext, unstr *unstructured.Unstruct
...
-			if err := InspectResource(relatedInfo, context, o); err != nil {
-				errs = append(errs, fmt.Errorf("skipping gathering %s due to error: %v", objectReferenceToString(relatedRef), err))
-				continue
-			}
-		}
+		allRelatedInfos = append(allRelatedInfos, relatedInfos...)
 	}
 
+	gatherErrs := ParallelInspectResource(allRelatedInfos, context, o)
+	errs = append(errs, gatherErrs...)
```
mfojtik, line 172: nit: `return errors.NewAggregate(append(errs, gatherErrs...))`

mfojtik: or more fancy:

```
return errors.NewAggregate(append(errs, ParallelInspectResource(allRelatedInfos, context, o)...))
```

pkg/cli/admin/inspect/resource.go
```diff
@@ -24,14 +25,42 @@ const (
 	operatorResourceDataKey = "/cluster-scoped-resources/operator.openshift.io"
 )
 
+func ParallelInspectResource(infos []*resource.Info, context *resourceContext, o *InspectOptions) []error {
```
mfojtik, line 28: nit: godoc

pkg/cli/admin/inspect/resource.go
```diff
@@ -24,14 +25,42 @@ const (
...
+	if len(infos) == 0 {
```
mfojtik, line 29: nit: no need for this, it will die fast

pkg/cli/admin/inspect/util.go
soltysh: It's possible, but I don't think we'd benefit that much from it.

pkg/cli/admin/inspect/resource.go
soltysh: I hate overly complicated statements, this reads cleanly and probably gets optimized by the compiler, so why not making our life's easier :wink:

PR #570, opened 2020-09-16
Bug 1874106: Split work of oc image mirror into chunks

sallyom:
While mirroring multiple images, quay.io sends "http2: server sent GOAWAY and closed the connection; LastStreamID=1, ErrCode=ENHANCE_YOUR_CALM, debug=""""
To avoid this, split the mappings into small chunks that the registry server can easily handle.

This PR sets a hard-coded chunk size of 10 for []Mapping from which the imageTrees and plan are built.  I've seen from local testing that `oc image mirror` fails when given > 20 SRC=DST mappings.

(currently invalid bz, so linking here: https://bugzilla.redhat.com/show_bug.cgi?id=1874106)

/cc @smarterclayton 
/assign @soltysh

### DIFF
--- a/pkg/cli/image/mirror/mappings.go
+++ b/pkg/cli/image/mirror/mappings.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/docker/distribution/registry/client/auth"
 	digest "github.com/opencontainers/go-digest"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/oc/pkg/cli/image/imagesource"
 )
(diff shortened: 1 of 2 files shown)


soltysh:
@sallyom have you considered approach (as suggested in BZ also) to start with as and only backoff when we get 429s? It looks like the problem is only with quay.io so it might seem artificial to always batch those, wdyt?

sallyom:
getting back to this, yes i did try that approach it did nothing :( retries do nothing - however, since I don't exactly know what limitation of quay we're hitting, I will reach out to try and pin that down.

pkg/cli/image/mirror/mappings.go
```diff
@@ -169,47 +170,70 @@ func (d *destinations) mergeIntoDigests(srcDigest digest.Digest, target pushTarg
...
-			dst.tags = append(dst.tags, m.Destination.Ref.Tag)
-		}
-		current[dstKey] = dst
+func chunk(mappings []Mapping, size int) [][]Mapping {
+	var chunkedMappings [][]Mapping
+	for i := 0; i < len(mappings); i += size {
+		m := mappings[i:min(i+size, len(mappings))]
+		chunkedMappings = append(chunkedMappings, m)
+	}
+	klog.Infof("LEN CHUNKED MAPPING: %d", len(chunkedMappings))
```
sallyom, line 228: TODO: remove

pkg/cli/image/mirror/mirror.go
```diff
@@ -226,6 +226,7 @@ func (o *MirrorImageOptions) Complete(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("SRC and DST may not be the same")
 		}
 	}
+	klog.Infof("LENGTH OF MAPPINGS: %v", len(o.Mappings))
```
sallyom, line 229: TODO: remove

sallyom:
/hold

Moving this bz to quay, will re-open if we decide oc should provide a workaround for any quay limitation

sallyom:
closing this, hoping to resolve this with quay.io. if not, will re-open/re-work a fix in oc.

sallyom:
@soltysh the issue is when sending more that ~20 mappings with `oc image mirror -f file-w-more-than-20` and/or w/ `oc adm catalog mirror`  the `Request Header Or Cookie Too Large</center>\r\n<hr><center>nginx/1.14.1` or the `http2 ErrCode=ENHANCE_YOUR_CALM` error is encountered. This PR splits large groups of mappings into smaller groups - since this PR is fixing the issue and we've determined this is not a quay limitation, I think we should reopen for discussion.

sallyom:
reopened here: https://github.com/openshift/oc/pull/761

PR #578, opened 2020-09-21
Bug 1872520: Replace bitbucket.org/ww/goautoneg with its mirror

dmage:
This PR bumps library-go to get changes from openshift/library-go#898: bitbucket.org/ww/goautoneg is no longer available, and it is replaced by github.com/munnerz/goautoneg (which is already vendored and has the same content).


soltysh (approved):
/lgtm
/approve

markmc:
/retitle Bug 1872520: Replace bitbucket.org/ww/goautoneg with its mirror

PR #595, opened 2020-09-30
bug 1882304: oc adm must-gather: have must-gather pods run on master nodes if --node-name is not specified

ingvagabund:


soltysh (approved):
/lgtm
/approve

PR #603, opened 2020-10-05
Bug 1885170: fix nil pointer in release code 

soltysh:
This currently is waiting for https://github.com/openshift/library-go/pull/910
/hold

/assign @jottofar @wking


soltysh:
With the real bump
/hold cancel

wking (approved):

PR #604, opened 2020-10-06
Bug 1885179: Avoid deadlock by closing pipe reader end

ricardomaraschini:
If we don't close this pipe reader with an error we ended up having a
deadlock as we keep trying to read indefinitely from it. This deadlock
causes the error message to be misguiding when attempting to upload a
layer with invalid format (gzip: invalid header).

Without this patch oc image append using an invalid layer format fails
with either "timeout" (docker.io) or "protocol error" (quay.io) instead
of reporting appropriate error ("invalid gzip").

### DIFF
--- a/pkg/helpers/image/dockerlayer/add/add.go
+++ b/pkg/helpers/image/dockerlayer/add/add.go
@@ -47,12 +47,14 @@ func DigestCopy(dst io.ReaderFrom, src io.Reader) (layerDigest, blobDigest diges
 	// calculate the diffID as the sha256 sum of the layer contents
 	pr, pw := io.Pipe()
 	layerhash := algo.Hash()
-	ch := make(chan error)
+	ch := make(chan error, 1)
 	go func() {
 		defer close(ch)
 		gr, err := gzip.NewReader(pr)
 		if err != nil {
-			ch <- fmt.Errorf("unable to create gzip reader layer upload: %v", err)
+			err := fmt.Errorf("unable to create gzip reader layer upload: %v", err)
+			pr.CloseWithError(err)
+			ch <- err
 			return
 		}
 		if !gr.Header.ModTime.IsZero() {


pkg/helpers/image/dockerlayer/add/add.go
```diff
@@ -52,7 +52,7 @@ func DigestCopy(dst io.ReaderFrom, src io.Reader) (layerDigest, blobDigest diges
 		defer close(ch)
 		gr, err := gzip.NewReader(pr)
 		if err != nil {
-			ch <- fmt.Errorf("unable to create gzip reader layer upload: %v", err)
+			pr.CloseWithError(fmt.Errorf("unable to create gzip reader layer upload: %v", err))
```
dmage, line 55: My only concern is that gzip.NewReader may have already read all data from `pr` at this point (for example, when there were less than 10 bytes to read). In this case this error won't be detected as nothing will write into the writer.

ricardomaraschini: Yes, you are absolutely right. I think it makes sense to still send the error down through that error channel. I have updated this to just close the pipe reader end with an error.

sallyom:
thanks @ricardomaraschini and @dmage

/lgtm

sallyom:
and I've confirmed the desired behavior:
with this change:
```console
$ oc image append --from=quay.io/sallyom/testcli:test --to=quay.io/sallyom/testcli:append new-layer.tar
Uploading ... error: unable to upload new layer (0): Patch "https://quay.io/v2/sallyom/testcli/blobs/uploads/ab100e66-f2c6-4c8a-b4dc-b326d6c563fb": unable to create gzip reader layer upload: gzip: invalid header
```
and from master (and after a minute of waiting...):
```console
$ oc image append --from=quay.io/sallyom/testcli:test --to=quay.io/sallyom/testcli:append new-layer.tar
Uploading ... error: unable to upload new layer (0): Patch "https://quay.io/v2/sallyom/testcli/blobs/uploads/b438f781-3070-45b2-8164-8dc14eba9c19": stream error: stream ID 15; PROTOCOL_ERROR
```

soltysh (approved):
/approve

PR #611, opened 2020-10-07
Bug 1841885: Support airgapped mirroring with `oc adm catalog mirror`

ecordell:
When mirroring to a file, `oc adm catalog mirror` will not rename any images. This means that image names can be inferred from the index contents when mirroring back out of a file store later.

Airgap mirroring then becomes a two step:

1. mirror from registry to file:///
2. mirror from file:// to the airgapped registry

This required a few changes:

1. the index image itself needed to be included in the mirror, so that it is available on the other side of the airgap
2. the catalog mirroring is updated to use typed image refs (i.e. file / registry discriminator)
3. bonus: parallelized the mirroring process


pkg/cli/admin/catalog/mirror.go
```diff
@@ -375,29 +375,76 @@ func WriteManifests(out io.Writer, name, dir, icspScope string, mapping map[stri
...
 	}
-	fmt.Fprintf(out, "wrote mirroring manifests to %s\n", dir)
+
+	catalogSource, err := generateCatalogSource(source, mapping)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(filepath.Join(dir, "catalogSource.yaml"), catalogSource, os.ModePerm); err != nil {
+		return fmt.Errorf("error writing CatalogSource")
+	}
```
sallyom, line 402: Should these files be uniquely named - icsp.yaml, catalogSource.yaml, mapping.txt? (I see "imageContentSourcePolicy.yaml" is not part of this PR)  See how the signature file is named here: https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/mirror.go#L302-#L313  and a PR to add the writing of icsp file w/ `oc adm release mirror` will also use this pattern to name the icsp file.  Won't these files be overwritten when mirror cmd is run more than once, and is this a potential problem?

sallyom (changes requested):
This PR should add examples to the catalog mirror command help:
`oc adm catalog mirror quay.io/my/image:latest file://local-catalog`
and then
`oc adm catalog mirror file://local-catalog some-registry:5000/local-catalog` 
I'm not seeing any examples of this in the documentation, either. (I'm not sure the above is the correct use but the correct way to mirror to disk then from disk to registry should be documented) 

I'm wondering why `oc adm catalog mirror` varies from `oc adm release mirror` in ways that should be similar, for example, why is there no `--to-dir` or `--to` flag? To mirror to directory w/ `oc adm release mirror` you can either:
`oc adm mirror 4.5.0 --to-dir=release` and the output shows the user this:
```
To upload local images to a registry, run:
    oc image mirror --from-dir=release 'file://openshift/release:4.5.0*' REGISTRY/REPOSITORY
```
_or_ `oc adm mirror 4.5.0 --to file://release` and output is shown:
```
To upload local images to a registry, run:
    oc image mirror 'file://release:4.5.0*' REGISTRY/REPOSITORY
```
Without the added examples or documentation on how to mirror to local disk and from local disk, it would be very difficult for users to use the `oc adm mirror catalog` command in this way.

pkg/cli/admin/catalog/mirror.go
ecordell: the manifests all go into a directory, so I copied what must-gather does and added a timestamp to the containing folder

ecordell:
> I'm wondering why oc adm catalog mirror varies from oc adm release mirror in ways that should be similar, for example, why is there no --to-dir or --to flag? 

`--to-dir` is available here as `--dir` - I didn't add those flags in either place so I can't answer why they're different (see [commit](https://github.com/openshift/oc/commit/92718e52707d562101e2bf38610974076f67a115#diff-11350ff26f6d5d859b8f6d7cbf7de11edbd6a9306b198b1753e5b1bbd37af0ffR89), but it's good to point out that they are.

For now I have added an output like you suggested indicating what the next step should be to mirror back to a registry from files. It's probably worth looking at differences between oc adm release mirror and oc adm catalog mirror and attempting to align them more in the future, but I'd like to keep this PR scoped to addressing the BZ.

pkg/cli/admin/catalog/mirror.go
sallyom: sg, and like you said below, we can align `oc adm catalog mirror` with `oc adm release mirror` in a follow-up PR if necessary.

sallyom:
@ecordell as noted in slack, this change in library-go is breaking this command (the failing e2e-cmd):
`oc create imagestreamtag tag:6 --from=wildfly:15.0`  with this output `error: Invalid --from, registry may not be specified`

ecordell:
library-go pr to fix failing test here: https://github.com/openshift/library-go/pull/928

soltysh (approved):
/lgtm
/approve

ecordell:
/hold

Tests which seem unrelated are flaking consistently

soltysh:
The failures are not related to this PR, from what I've manually checked:
/override ci/prow/e2e-cmd 
/hold cancel

PR #613, opened 2020-10-12
Bug 1920024: debug: Allow pods to keep labels during debugging

smarterclayton:
Labels allow pods to join services, and it should be possible to add
a debug pod for a replica set that joins the service in order to target
traffic to it. We keep the default of "false", merely add `--keep-labels`.


sallyom:
There is `kubectl alpha debug` also, with less options. We'll want to add this (and others, like --keep-annotations) upstream, eventually and eventually adopt kubectl debug (i think). 

Trying this out, the debug pod with `--keep-labels=true` is failing - tried with the rails-postgresql-example app suggested with
`oc new-project` 
```
$ oc debug dc/postgresql
Starting pod/postgresql-debug, command was: container-entrypoint run-postgresql
Pod IP: 10.131.0.30
sh-4.4$ exit
exit

$ oc debug dc/postgresql --keep-labels=true
Starting pod/postgresql-debug, command was: container-entrypoint run-postgresql

Removing debug pod ...
error: unable to create the debug pod "postgresql-debug"
```
I'm not sure why, but I can look into it.

pkg/cli/debug/debug.go
```diff
@@ -214,6 +214,7 @@ func NewCmdDebug(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra
 	cmd.Flags().BoolVarP(&o.DisableTTY, "no-tty", "T", o.DisableTTY, "Disable pseudo-terminal allocation")
 	cmd.Flags().StringVarP(&o.Attach.ContainerName, "container", "c", o.Attach.ContainerName, "Container name; defaults to first container")
 	cmd.Flags().BoolVar(&o.KeepAnnotations, "keep-annotations", o.KeepAnnotations, "If true, keep the original pod annotations")
+	cmd.Flags().BoolVar(&o.KeepLabels, "keep-labels", o.KeepLabels, "If true, keep the original pod labels")
```
soltysh, line 217: You've exposed this flag, but we don't copy those labels anywhere in the code, so that's why it's failing for @sallyom

soltysh (changes requested):

pkg/cli/debug/debug.go
smarterclayton: We do copy them:

https://github.com/openshift/oc/blob/c923633547ce429cab2903f063b4e8d423227894/pkg/cli/debug/debug.go#L795

smarterclayton:
/retest

not good, a bunch of packages disappeared

soltysh (approved):
/lgtm

PR #623, opened 2020-10-27
Bug 1891555: Create Windows version information

yselkowitz:
This was previously enabled in 4.1 but was lost in 4.2 when oc was split
out from the origin repository.


soltysh (approved):
/lgtm
/approve

PR #625, opened 2020-10-27
show proper wildcard for 'oc image mirror --help example'

sallyom:
Somebody (me) wrote this `--filter-by-os=/*` as example of how to list/mirror all digests of a multi-arch image with the wildcard filter-by-os.  This PR fixes the help to show `--filter-by-os=.*` instead.


soltysh (approved):
/lgtm
/approve

PR #628, opened 2020-10-29
Add `oc ignition convert3`

cgwalters:
The transition to Ignition Spec 3 with 4.6 creates a
discontinuity. Some users want to update their bootimages,
e.g. for a cluster originally provisioned as 4.4 but upgraded
in place to 4.6, it should be possible to directly use RHCOS 4.6
bootimages for new workers.

In some cases in fact, this could be required for things like
adding a node with newer hardware.

The main stumbling block here is the pointer ignition config
which is generated by `openshift-install`.  Since the idea is
`openshift-install` should in theory be disposable after a cluster
is provisioned, let's add this to `oc` which admins will need anyways.
Vendor and use
https://github.com/coreos/ign-converter
the same as the MCO uses.

xref openshift/enhancements#492 (comment)
xref https://bugzilla.redhat.com/show_bug.cgi?id=1884750


kikisdeliveryservice:
This seems like a good idea to me!! LGTM

pkg/cli/ignition/convert3.go
```diff
@@ -0,0 +1,116 @@
...
+			return err
+		}
+		data = d
+	}
+
+	cfg, rpt, err := ignv2.Parse(data)
+	fmt.Fprintf(os.Stderr, "%s", rpt.String())
+	if err != nil || rpt.IsFatal() {
+		return errors.Errorf("Error parsing spec v2 config: %v\n%v", err, rpt)
+	}
```
bgilbert, line 95: If the report contains warnings, we may as well print them, I think?

pkg/cli/ignition/convert3.go
```diff
@@ -0,0 +1,116 @@
...
+const convert3CommandName = "convert3"
+
+type convert3Options struct {
+	// Source file
+	Source string
+	// Destination file
+	Destination string
+
+	SourceReader io.Reader
+	DestWriter   io.Writer
```
abhinavdahiya, line 29: if the options struct becomes public, all the fields of the struct will also become public API, maybe we can make all of these private for start with or only leave certain fields public that we think will be public API when this struct becomes public...

cgwalters: Yeah, will fix this though I'll note that this pattern seems fairly old and pervasive in `oc`; I think I used [extract.go](https://github.com/openshift/oc/blob/8150994e0392adb4aeade70d62974ca77b3740a4/pkg/cli/extract/extract.go) as a template.

pkg/cli/ignition/convert3.go
cgwalters: Wait actually aren't we doing this with the `fmt.Fprintf(os.Stderr, "%s", rpt.String())` above?

Just tested this, yes we are.

cgwalters:
OK resolved outstanding comments so far.

pkg/cli/ignition/convert3.go
bgilbert: Whoops, no idea how I missed that.

k-keiichi-rh:
I tested if we can add new nodes to the upgraded cluster using the converted Ignition Config by this tool
and then found an issue of failing to add new nodes to a cluster in vSphere.

[ Test conditions ]
  - Upgraded the cluster from OCP4.5 to OCP4.6 in vSphere
  - Converted the Ignition Config which is generated by openshift-install when deploying OCP4.5.
  - Added new nodes to the upgraded cluster by using the converted Ignition Config and the OCP4.6 base image.

It failed to add new nodes and the following error is repeated:

GET https://api-int.share.test:22623/config/worker: attempt GET error: GET "https://api-int.share.test:22623/config/worker": x509: certificate relies on legacy Common Name filed, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0

This issue is caused by missing "X509v3 Subject Alternative Name"(SAN) field in the OpenSSL x509 certificate of https://api-int.share.test:22623/config/worker.
After OCP4.6(go1.15 and later), the CommonName field on X.509 certificates is deprecated and the SAN field should be set in the certificates. However, OCP4.5 cluster on vSphere platform doesn't set the SAN field.

The following patch is related to this issue:

https://github.com/openshift/installer/commit/b2d424b7b44a0db5173fca37018b7af3f164d677
=> It's changed to set the SAN field when deploying OCP4.6 clusters on vSphere platform.

So this converter tool won't work in Sphere platform.

To avoid this issue, the following steps are required:

1. Retrieve Ignition Config from the upgrade cluster(OCP4.6)

```
 # curl -H "Accept: application/vnd.coreos.ignition+json; version=3.1.0" -k https://<api-server-url>:22623/config/worker
```

2. Add new nodes using the retrieved Ignition Config and OCP4.6 base image.

Or

The X.509 certificate for the api-int endpoint should be re-created with the SNA field.
I removed machine-config-server-tls in openshift-machine-config-operator.
And then it was re-created automatically, but it didn't have the SAN field.

If this pr is not the right place to discuss, please correct me.
I will move to the right place.

cgwalters:
Hi right, you hit an unrelated issue in 4.5 :arrow_right: 4.6 upgrades.  This is https://bugzilla.redhat.com/show_bug.cgi?id=1886134 for Ignition, and probably warrants an entry in the knowledge base and explicitly tracking a fix.

Perhaps that fix could go alongside here in `oc` indeed...something like `oc ignition get-pointer-config` that would just regenerate it with an updated cert?

Alternatively we could fix the above bug for Ignition and work around the behavior for a while longer.

k-keiichi-rh:
> Hi right, you hit an unrelated issue in 4.5 ➡️ 4.6 upgrades. This is https://bugzilla.redhat.com/show_bug.cgi?id=1886134 for Ignition, and probably warrants an entry in the knowledge base and explicitly tracking a fix.

Thank you for informing me. I will follow the bugzilla.

> The transition to Ignition Spec 3 with 4.6 creates a
> discontinuity. Some users want to update their bootimages,
> e.g. for a cluster originally provisioned as 4.4 but upgraded
> in place to 4.6, it should be possible to directly use RHCOS 4.6
> bootimages for new workers.

I have a question regarding adding new nodes to the upgraded OCP4.6 cluster by using RHCOS 4.6 boot images directly.
In this case, we have to convert Ignition Config Spec 2 that was used to install the OCP4.5 cluster.

There are two ways to achieve the above:
1. Convert Ignition Config from Ignition Spec 2 to Ignition Spec 3 by "oc ignition convert3"
2. Get Ignition Spec 3 from MCO
    ex) "curl -H "Accept: application/vnd.coreos.ignition+json; version=3.1.0" -k https://<api-server-url>:22623/config/worker"

Is there any difference between "Option 1" and "Option 2"?
As for Option 1, we have to keep the Ignition Config that was used to install the OCP4.5 cluster.
But, as for Option 2, we can dispose the Ignition Config and can get the Ignition Config to add new nodes from the existing cluster each time. It looks like Option 2 is easier and more maintainable way that Option 1.

miabbott:
Looks like the new subcommand is missing a help string:

```
$ ./oc --help | grep -C 3 ignition
  image           Useful commands for managing images
  registry        Commands for working with the registry
  idle            Idle scalable resources
  ignition        
  api-versions    Print the supported API versions on the server, in the form of "group/version"
  api-resources   Print the supported API resources on the server
  cluster-info    Display cluster info
```

pkg/cli/ignition/convert3.go
```diff
@@ -0,0 +1,116 @@
...
+		if err != nil {
+			return err
+		}
+		data = d
+	}
+
+	cfg, rpt, err := ignv2.Parse(data)
+	fmt.Fprintf(os.Stderr, "%s", rpt.String())
+	if err != nil || rpt.IsFatal() {
+		return errors.Errorf("Error parsing spec v2 config: %v\n%v", err, rpt)
```
miabbott, line 94: If a user feeds a v3 config to the command, it returns this same error.

```
$ jq .ignition.version < ignitionv3.json 
"3.1.0"
$ ./oc ignition convert3 --in ignitionv3.json --out ignitionv3a.json
error: Error parsing spec v2 config: unsupported config version
```

Could we check the incoming config version before trying to parse?

cgwalters:
> Get Ignition Spec 3 from MCO
ex) "curl -H "Accept: application/vnd.coreos.ignition+json; version=3.1.0" -k https://:22623/config/worker"

No, that's going to cause problems because you're getting the full config and not just the pointer.  We don't expose an explicit API to retrieve the pointer config today, see discussions in https://github.com/openshift/enhancements/pull/540

pkg/cli/ignition/convert3.go
cgwalters: > Could we check the incoming config version before trying to parse?

Not really, to get the version we need to parse...  I think you're just asking for a better error message?  Makes total sense but I think the fix would probably be in ign-converter.

cgwalters:
> Looks like the new subcommand is missing a help string:

Fixed.

soltysh:
/hold
Adding new commands post-freeze is not acceptable, so at soonest this can be a 4.8 item. Secondly I haven't singed off the addition of a new command, yet.

pkg/cli/ignition/convert3.go
bgilbert: Ignition now has `config.util.GetConfigVersion()`, which should work with 2.x configs also.

PR #631, opened 2020-10-30
Bug 1888192: add rsync & log re-tries

soltysh:
There are 2 main issues with `oc adm must-gather` that this PR addresses:
1. On the env, the logs streaming from gather pod ends after ~10min, when the gather runs longer and on that cluster I was given access it was taking ~30mins, you'll only get info about the initial 10mins. The time-out is coming from server, so we just need to ensure we re-try streaming logs after that time-out if the pod is still running.
2. Current copy logic is missing compression.
3. Current sync logic is copying the artifacts only once, if it fails it doesn't re-try. 

/assign @sallyom

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -431,8 +431,9 @@ func (o *MustGatherOptions) getGatherContainerLogs(pod *corev1.Pod) error {
 		Namespace:   pod.Namespace,
 		ResourceArg: pod.Name,
 		Options: &corev1.PodLogOptions{
-			Follow:    true,
-			Container: pod.Spec.Containers[0].Name,
+			Follow:     true,
+			Container:  pod.Spec.Containers[0].Name,
+			Timestamps: true,
 		},
 		RESTClientGetter: o.RESTClientGetter,
 		Object:           pod,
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -414,7 +421,25 @@ func (o *MustGatherOptions) getGatherContainerLogs(pod *corev1.Pod) error {
...
+		if err := opts.RunLogs(); err != nil {
+			return err
+		}
+
+		klog.V(4).Infof("lost logs, re-trying...")
+
+		// to ensure we don't print all of history set since to past 2 seconds
+		opts.Options.(*corev1.PodLogOptions).SinceSeconds = &since2s
+
+		if done, _ := o.isGatherDone(pod); done {
```
sallyom, line 439: move this above the klog "lost logs, re-trying", so that only prints if !isGatherDone -
(see easily w/ `oc adm must-gather --dest-dir=/tmp/must-gather -v=4 -- date`)

sallyom (changes requested):
suggestion to move the isGatherDone b4 the lost logs message, other than that lgtm.

pkg/cli/admin/mustgather/mustgather.go
soltysh: Good point, updated.

soltysh:
/hold
I want to add timestamps to log lines

PR #634, opened 2020-11-03
image extract: handle files at the root of a filesystem

stevekuznetsov:
Previously, files like `/app` were incorrectly rejected by faulty
validation logic.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

### DIFF
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -201,7 +201,7 @@ func parseMappings(images, paths, files []string, requireEmpty bool) ([]Mapping,
 			}
 			mappings = append(mappings, Mapping{
 				Image: image,
-				From:  strings.TrimPrefix(arg, "/"),
+				From:  path.Clean(arg),
 				To:    ".",
 			})
 		}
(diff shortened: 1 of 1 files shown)


smarterclayton:
/hold

Will review but there are a bunch of ugly edge cases in some images that may be impacted with path.clean.

pkg/cli/image/extract/extract.go
```diff
@@ -201,7 +201,7 @@ func parseMappings(images, paths, files []string, requireEmpty bool) ([]Mapping,
 			}
 			mappings = append(mappings, Mapping{
 				Image: image,
-				From:  strings.TrimPrefix(arg, "/"),
+				From:  path.Clean(arg),
```
soltysh, line 204: I'd at least start with filepath, which works fine on any OS. path is unix only.

stevekuznetsov: Isn't this the path _in the container_ - so it is going to be a UNIX filesystem? Nobody is doing e.g. `C://` to refer to something in the container FS, are they?

soltysh: What when the container is windows? Obviously it's not here, but I'd prefer to fix it now, rather then later. It's not that hard :wink:

stevekuznetsov: Ah, gotcha

stevekuznetsov:
@smarterclayton can you review this please?

PR #635, opened 2020-11-04
Bug 1891555: Accept OS_GIT_VERSION to set version information

yselkowitz:
When building in OCP, this variable is set with the intended version
information.

For CI versions, annotated tags (releases) need to be created.  This 
used to occur in openshift/origin but hasn't regularly since this was 
split out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,10 @@
 all: build
 .PHONY: all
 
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/approve

PR #637, opened 2020-11-09
Bug 1890671: use proper username for image verification

soltysh:
/assign @sallyom 

There are two possible approaches to this problem:
1. the one from this PR
2. allowing injecting `--registry-config` similarly to how we do that in `oc image mirror`

what do you think which one would be better?

### DIFF
--- a/pkg/cli/admin/verifyimagesignature/verify-signature.go
+++ b/pkg/cli/admin/verifyimagesignature/verify-signature.go
@@ -261,7 +261,9 @@ func (o *VerifyImageSignatureOptions) getImageManifest(img *imagev1.Image) ([]by
 			registryURL.Scheme = ""
 		}
 	}
-	return getImageManifestByIDFromRegistry(registryURL, parsed.RepositoryName(), img.Name, o.CurrentUser, o.CurrentUserToken, o.Insecure)
+	// when using in-cluster auth, the username is just user + token, compare
+	// https://github.com/openshift/oc/blob/9f54c1d4f68c8530ac9466c655a4e55eb04a1459/pkg/cli/registry/login/login.go#L208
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/verifyimagesignature/verify-signature.go
```diff
@@ -261,7 +261,9 @@ func (o *VerifyImageSignatureOptions) getImageManifest(img *imagev1.Image) ([]by
 			registryURL.Scheme = ""
 		}
 	}
-	return getImageManifestByIDFromRegistry(registryURL, parsed.RepositoryName(), img.Name, o.CurrentUser, o.CurrentUserToken, o.Insecure)
+	// when using in-cluster auth, the username is just user + token, compare
+	// https://github.com/openshift/oc/blob/9f54c1d4f68c8530ac9466c655a4e55eb04a1459/pkg/cli/registry/login/login.go#L208
```
soltysh, line 265: Check this URL, this answers the question why `user` is being used as a username.

sallyom: so, as long as the token is in use, then "user" is appropriate?  that's what it looks like, but I've been having a hard time verifying/reproducing this issue :(

sallyom: I see the `getImageManifestByIDFromRegistry` method is always passed the `o.CurrentUserToken` so based on that, and looking at how we force the "user" in that case with registry login code, this makes sense, but still, I haven't been able to reproduce or confirm this.

sallyom:
This is tricky to test/confirm. I did verify that with this PR I do not see an unauthorized error, while with master branch I see the unauthorized error. 

/lgtm

PR #640, opened 2020-11-12
Bug 1878823: 'oc adm release mirror' add --to-release mirrors to generated ICSP

sallyom:
When oc adm release mirror is passed both --to and --to-release flags, only the --to value is used to generate ICSP. This PR adds the --to-release mirrors to the generated ICSP.


soltysh (approved):
/lgtm
/approve
/retest

PR #627, opened 2020-10-29
Bug 1888738: fall-back must-gather to official RH supported image

soltysh:
This is the only reasonable image we can probably fall-back to, although I'm a bit hesitant on using it here. 

/assign @wking

PR #642, opened 2020-11-13
Bug 1890951: Improve output of 'oc image mirror' with multi-arch images

sallyom:
With this PR, the `--keep-manifest-list=true` command will fail if `--filter-by-os` is passed with any value other than `--filter-by-os=.*`. This is because it is not possible to preserve the manifest list digest while also altering the manifest list to only include some of its manifests (you could, but it would be rejected by registries as it will no longer be valid). Before this PR, `--keep-manifest-list=true` did not actually keep the manifest list digest, in cases where the `--filter-by-os=not/wildcard`, a new manifest list with a new digest was created. 

This PR also addresses https://bugzilla.redhat.com/show_bug.cgi?id=1882689
/assign @smarterclayton

### DIFF
--- a/pkg/cli/image/manifest/manifest.go
+++ b/pkg/cli/image/manifest/manifest.go
@@ -411,7 +411,7 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
 			if err != nil {
 				return nil, nil, "", err
 			}
-			klog.V(5).Infof("Used only one manifest from the list %s", srcDigest)
+			klog.Warningf("Chose %s/%s manifest from the list.\nTo include the manifest list digest: %s, use --keep-manifest-list=true", t.Manifests[0].Platform.OS, t.Manifests[0].Platform.Architecture, srcDigest)
 			return srcManifests, srcManifests[0], manifestDigest, nil
 		default:
 			return append(srcManifests, manifestList), manifestList, manifestDigest, nil
--- a/pkg/cli/image/mirror/mirror.go
+++ b/pkg/cli/image/mirror/mirror.go
@@ -95,11 +95,17 @@ var (
 
 		# Copy specific os/arch manifest of a multi-architecture image
 		# Run 'oc image info myregistry.com/myimage:latest' to see available os/arch for multi-arch images
+		# Note that with multi-arch images, this results in a new manifest list digest that includes only
+		# the filtered manifests.
 		oc image mirror myregistry.com/myimage:latest=myregistry.com/other:test \
 			--filter-by-os=os/arch
 
 		# Copy all os/arch manifests of a multi-architecture image
 		# Run 'oc image info myregistry.com/myimage:latest' to see list of os/arch manifests that will be mirrored.
+		oc image mirror myregistry.com/myimage:latest=myregistry.com/other:test \
+			--keep-manifest-list=true
+
+		# Note the above command is equivalent to
 		oc image mirror myregistry.com/myimage:latest=myregistry.com/other:test \
 			--filter-by-os=.*
 	`)
@@ -168,7 +174,7 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
 	flag.BoolVar(&o.SkipMount, "skip-mount", o.SkipMount, "Always push layers instead of cross-mounting them")
 	flag.BoolVar(&o.SkipMultipleScopes, "skip-multiple-scopes", o.SkipMultipleScopes, "Some registries do not support multiple scopes passed to the registry login.")
 	flag.BoolVar(&o.Force, "force", o.Force, "Attempt to write all layers and manifests even if they exist in the remote repository.")
-	flag.BoolVar(&o.KeepManifestList, "keep-manifest-list", o.KeepManifestList, "If an image is part of a manifest list, always mirror the list even if only one image is found. The default is to mirror the specific image unless unless --filter-by-os is '.*'.")
+	flag.BoolVar(&o.KeepManifestList, "keep-manifest-list", o.KeepManifestList, "If an image is part of a manifest list, always mirror the list even if only one image is found. The default is to mirror the specific image unless unless --filter-by-os is passed. This flag is equivalent to setting --filter-by-os to '.*' since you cannot preserve the manifest list digest while filtering out any of the manifests included in the list.")
 	flag.IntVar(&o.MaxRegistry, "max-registry", o.MaxRegistry, "Number of concurrent registries to connect to at any one time.")
 	flag.StringSliceVar(&o.AttemptS3BucketCopy, "s3-source-bucket", o.AttemptS3BucketCopy, "A list of bucket/path locations on S3 that may contain already uploaded blobs. Add [store] to the end to use the container image registry path convention.")
 	flag.StringSliceVarP(&o.Filenames, "filename", "f", o.Filenames, "One or more files to read SRC=DST or SRC DST [DST ...] mappings from.")
@@ -179,6 +185,10 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
 }
 
 func (o *MirrorImageOptions) Complete(cmd *cobra.Command, args []string) error {
+	if o.KeepManifestList && len(o.FilterOptions.FilterByOS) == 0 {
+		o.FilterOptions.FilterByOS = ".*"
+	}
+
 	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
 		return err
 	}
@@ -246,6 +256,9 @@ func (o *MirrorImageOptions) Repository(ctx context.Context, context *registrycl
 }
 
 func (o *MirrorImageOptions) Validate() error {
+	if o.KeepManifestList && len(o.FilterOptions.FilterByOS) > 0 && !o.FilterOptions.IsWildcardFilter() {
+		return fmt.Errorf("--keep-manifest-list=true cannot be passed with --filter-by-os, unless --filter-by-os=.*")
+	}
 	return o.FilterOptions.Validate()
 }
 


sallyom:
This PR also addresses https://bugzilla.redhat.com/show_bug.cgi?id=1882689

pkg/cli/image/mirror/mirror.go
```diff
@@ -168,7 +168,7 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
 	flag.BoolVar(&o.SkipMount, "skip-mount", o.SkipMount, "Always push layers instead of cross-mounting them")
 	flag.BoolVar(&o.SkipMultipleScopes, "skip-multiple-scopes", o.SkipMultipleScopes, "Some registries do not support multiple scopes passed to the registry login.")
 	flag.BoolVar(&o.Force, "force", o.Force, "Attempt to write all layers and manifests even if they exist in the remote repository.")
-	flag.BoolVar(&o.KeepManifestList, "keep-manifest-list", o.KeepManifestList, "If an image is part of a manifest list, always mirror the list even if only one image is found. The default is to mirror the specific image unless unless --filter-by-os is '.*'.")
+	flag.BoolVar(&o.KeepManifestList, "keep-manifest-list", o.KeepManifestList, "If an image is part of a manifest list, always mirror the list even if only one image is found. The default is to mirror the specific image unless unless --filter-by-os is passed.")
```
wking, line 171: What happens if `--filter-by-os` is set to something that matches multiple targets within the manifest list?  I dunno if anyone has a spec around manifest-list severability, but it would be nice if you could have the list and then *n* of *m* referenced children, still be able to push the manifest list up into the target registry, and only get pull errors if someone tried to use the list to get at a referenced image that had not been mirrored.

sallyom: You're stretching the limits of what I know about manifestLists, only way i'll truly be able to answer this is if you have an example of such a manifestList that I can try :)

wking: `registry.redhat.io/openshift4/ose-must-gather:latest` is an example of a multi-arch image (see inspection [here][1]), if you want one to play around with.

[1]: https://github.com/openshift/oc/pull/627#issuecomment-718924035

sallyom: I meant, a manifestList with multiple targets of same os/arch. I've been using `quay.io/prometheus/prometheus:v2.18.1` multi-arch image in my testing.  I see the must-gather image does not have multiple targets for same os/arch? 
```
  OS            DIGEST
  linux/amd64   sha256:47c2f751ab0d5ee88e2826749f1372e6a24db3d0c0c942136ae84db17cb7f086
  linux/ppc64le sha256:70ebe4f2c01e17b84617c58dd5d64b00fe91bf6a3bc216b6d9f4690552f48265
  linux/s390x   sha256:aca00f9e47286c74b5b5d9ac7b3b130b9ce3ee087cb940269a0130aa7d569d18
```

sallyom: `filter-by-os` matches `os/arch` not just `os` - is that what you mean?

wking: So I mean if you ran `oc image mirror --filter-by-os 'amd64|ppc64le quay.io/prometheus/prometheus:v2.18.1 ...` or whatever, to match both `linux/amd64` and `linux/ppc64le`.  Ideally that would bring over the manifest list, the `linux/amd64` image, and the `linux/ppc64le` image, but not the `linux/390x` image.  It's not clear to me if registries and other tooling would support the original manifest list in that situation (because it would also reference the `linux/s390x` image we were not mirroring), or if you'd need to create a new manifest-list image (breaking signatures on the manifest-list image) that only referenced the images you were mirroring.

sallyom: This change is to always mirror the manifest list along with any filtered os/arch manifests. The only way to _not_ mirror the manifest list  is to run something like:
```console
$ oc image mirror registry.redhat.io/openshift4/ose-must-gather:latest=localhost:5000/mustgather
```
and you'll get the warning letting you know you did not mirror the manifest list, only the manifest for your system os/arch:
```console
W1130 22:57:41.622731   49734 manifest.go:414] Chose linux/amd64 manifest from the list sha256:d8e0b62c75bff609394bfaa43c923480d2d0bf2f7a3f36fd28ec601f99f3d25c
To include the manifest list digest: sha256:d8e0b62c75bff609394bfaa43c923480d2d0bf2f7a3f36fd28ec601f99f3d25c, use --keep-manifest-list=true
```

sallyom: and these will mirror the manifest list along w/ your filtered os/arch manifests:
```console
$ oc image mirror registry.redhat.io/openshift4/ose-must-gather:latest=localhost:5000/mustgather --keep-manifest-list=true
$ oc image mirror registry.redhat.io/openshift4/ose-must-gather:latest=localhost:5000/mustgather --filter-by-os=any/valid
```

ecordell: `--filter-by-os` will construct a manifest list with only the filtered manifests listed, therefore changing the digest of the mirrored manifestlist (or if only one matches it will only mirror the one and not mirror the manifestlist at all)

`--keep-manifest-list` was intended to preserve the entire manifestlist, regardless of what manifests have been filtered via `--filter-by-os`.

Without doing the forensics to confirm, I suspect that keep-manifest-list was implemented without the knowledge that that `--filter-by-os` could rewrite the manifestlist. This probably went unnoticed because most of the images mirrored with the `--keep-manifest-list` flag only contained a single entry, so the digest would remain the same.

The least surprising solution would be (IMO) to not touch `--filter-by-os` behavior at all unless used in conjunction with `--keep-manifest-list`, i.e.:

`oc image mirror --filter-by-os=<filter>`: same behavior as today
`oc image mirror --filter-by-os=<filter> --keep-manifest-list=true`: manifest list is not modified and mirrored, and any manifests matching the filter are mirrored.

pkg/cli/image/mirror/mirror.go
```diff
@@ -179,14 +179,16 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
 }
 
 func (o *MirrorImageOptions) Complete(cmd *cobra.Command, args []string) error {
+	// Check before completing filter options, to see if filter-by-os flag is set
+	// Return manifest list if a user is mirroring a particular os/arch (or wildcard os/arches)
+	// The default is to mirror the single manifest that matches the user's system os/arch
+	if len(o.FilterOptions.FilterByOS) > 0 {
```
ecordell, line 185: see my other comment - I think omitting this makes the change less disruptive to existing oc image mirror users.

sallyom: I think maybe all that's needed for this PR is a validation that one cannot pass both `--filter-by-os=os/arch` (except wildcard) _and_ `--keep-manifest-list=true`  since they contradict.  wdyt?

sallyom: @ecordell ptal, I've updated.  If a user passes both `--filter-by-os=not/wildcard` _and_ `--keep-manifest-list=true`, they'll see a warning during the flag completion, then, the cmd will fail later on when it is determined that len(filtered) != len(list) - bc at that point the manifestList digest will be changed to a new one that includes only the filtered os/arch manifests.  Result will be like so:
```console
$ oc image  mirror quay.io/prometheus/prometheus:v2.18.1=localhost:5000/prom-part --keep-manifest-list=true --filter-by-os=linux/arm*
W1201 13:43:53.332411   93410 mirror.go:186] This command will fail if the filter ( linux/arm* ) does not include all available os/arch options.

error: failing because manifest list cannot be preserved with the given filters. Try running the command again without the --filter-by-os or to filter the list, omit the '--keep-manifest-list=true'. A filtered manifest list will have a new digest.
error: an error occurred during planning
```

ecordell: shouldn't that be allowed? 

--filter-by-os=os/arch and --keep-manifest-list: mirror only the manifests that match the filter in the whole manifestlist, mirror the manifestlist unmodified

--filter-by-os=os/arch: mirror only the manifests that match the filter in the whole manifestlist, mirror a modified manifestlist that only references the mirrored manifests

sallyom: i discussed this w/ @nalind and @mtrmac on slack (pinged you) - it's not possible to preserve the manifest list digest if its constituents are changed, so unless the filter includes all available os/arch, the manifest list digest needs to be updated, otherwise as i understand it, registries will refuse it.

ecordell: Ah you updated while I was typing the last one 😄 

> bc at that point the manifestList digest will be changed to a new one that includes only the filtered os/arch manifests.

The issue is that we have disconnected users that want to only mirror a subset of the manifests (i.e. the s390x ones) but need the entire manifestlist preserved (because they're referenced by digest in the cluster). So the `--keep-manifest-list` was supposed to preserve the original manifestlist even if only a subset of manifests contained within were actually mirrored.

ecordell:
Sorry, just caught up on slack. It looks like most registries (docker/distribution and quay.io at least) require all referenced manifests to exist in the registry in order for it to be in a registry. 

That's unfortunate and degrades the disconnected process for multi-arch operators, but it doesn't seem like something that can be fixed in `oc` alone right now.

pkg/cli/image/mirror/mirror.go
sallyom: Right, hm.. I see this is currently possible by passing no --filter-by-os and with --keep-manifest-list=true. It seems we are already bending the rule there (a user _can_ today successfully mirror only their system os/arch with the reference to the manifest list, and then pull their system os/arch manifest while also being able to  pull the manifest list digest.)  The issue (the bz) is that a user cannot extend that to pass --filter-by-os=anything/other than the system os/arch - let me look at this again and see if there's a solution, since we already accept that we'll mirror the manifest list with only a single constituent.  Looking...

sallyom: er, no, I was wrong above, the manifest list digest is not preserved now unless one passes the wildcard filter-by-os.  - I've been working with this image: 
`quay.io/prometheus/prometheus@sha256:5880ec936055fad18ccee798d2a63f64ed85bd28e8e0af17c6923a090b686c3d` that is a manifest list (quay.io/prometheus/prometheus:v2.18.1)
If I run oc image mirror cmds I see the following:
```console
$ oc image  mirror quay.io/prometheus/prometheus:v2.18.1=localhost:5000/prom:v2.18.1 --keep-manifest-list=true
I1202 14:47:28.599652  226774 manifest.go:404] Filtered manifest list to new digest sha256:debb32e52e5b585d84f35e7867f680c11871f9ad31aba422c8bb4d44c23aea32
(that is a new manifest list digest since it filtered my system os/arch only)
but...
$ oc image  mirror quay.io/prometheus/prometheus:v2.18.1=localhost:5000/prom:v2.18.1 --filter-by-os=.*
sha256:5880ec936055fad18ccee798d2a63f64ed85bd28e8e0af17c6923a090b686c3d localhost:5000/prom:v2.18.1  the digest is preserved  (adding --keep-manifest-list=true is unnecessary, does nothing, it's the fact that all components of the list are filtered, or len(filtered) == len(list), that result in preserving the manifest list.  
```
So, what I've uncovered is that `--keep-manifest-list=true` is broken :shrug:  and has never worked, AFAICT.

sallyom: and to "fix" this means introducing a hack, to force preserve the manifest list digest - even if we change the contents of the manifest list.

sallyom: After further consideration I think this PR makes sense, unfortunately - when a user passes `--keep-manifest-list=true` the cmd should error if that can't happen (it does not, now, see the first of the 2 example cmds, above).
I think, however, we need to fix that first command above, to _not_ default to the system os/arch if `--keep-manifest-list=true` and instead mirror all os/arches since user has signalled they _want_ to preserve the manifest list - and only way to do that is to mirror all.  
I don't believe there is a fix to `--keep-manifest-list=true` and also `--filter-by-os=anything/other-than-wildcard` and the fact that our code is currently advertising otherwise is a bug (that this PR fixes)

ecordell: I agree - not sure there's a better solution if registries reject manifestlists that reference manifests not available in the registry

sallyom:
> Sorry, just caught up on slack. It looks like most registries (docker/distribution and quay.io at least) require all referenced manifests to exist in the registry in order for it to be in a registry.
> 
> That's unfortunate and degrades the disconnected process for multi-arch operators, but it doesn't seem like something that can be fixed in `oc` alone right now.

Yup, that's what I've determined (lol was experimenting all day to confirm that oc has always been broken, and am just now seeing this comment).  Thanks. Unfortunate, indeed.

pkg/cli/image/mirror/mirror.go
smarterclayton: You can't change the behavior of the flags.  Keep-manifest-list works in some scenarios, and we can't change the flags because those flags may be used in mirroring automation.  That's why the wildcard default behavior was introduced.

Can you summarize what problem you are trying to solve right now? :)

smarterclayton:
> Then, the command will fail later when/if it is determined that the len(filtered) != len(list) - bc it is not possible to preserve the manifest list digest at this point (you could, but it would be rejected by registries).

I don't understand this comment

sallyom:
> > Then, the command will fail later when/if it is determined that the len(filtered) != len(list) - bc it is not possible to preserve the manifest list digest at this point (you could, but it would be rejected by registries).
> 
> I don't understand this comment

it's in reference to this line:
https://github.com/openshift/oc/blob/master/pkg/cli/image/manifest/manifest.go#L382
and modified in this PR by erroring if a manifest is skipped with this:
https://github.com/openshift/oc/blob/6146efd88166335b56879b487051f272b50ed9c8/pkg/cli/image/manifest/manifest.go#L371-#L386

pkg/cli/image/mirror/mirror.go
sallyom: I don't want to change the behavior of the flags, but as/is, --keep-manifest-list=true is broken, so to fix it, whenever a user passes `--keep-manifest-list=true` the mirror command must mirror _all_ manifests in the list, otherwise the manifest list digest is not preserved.

sallyom: I've updated this PR, so now, the change is that whenever `--keep-manifest-list=true` is passed, `--filter-by-os=.*`  since this is the only way to include the same manifest list digest.  The help menu explains that `--keep-manifest-list=true` will override any value passed for `--filter-by-os=not/wildcard`

pkg/cli/image/manifest/manifest.go
```diff
@@ -404,14 +405,13 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
...
 		}
-
 		switch {
 		case len(srcManifests) == 1 && !keepManifestList:
 			manifestDigest, err := registryclient.ContentDigestForManifest(srcManifests[0], srcDigest.Algorithm())
 			if err != nil {
 				return nil, nil, "", err
 			}
-			klog.V(5).Infof("Used only one manifest from the list %s", srcDigest)
+			klog.Warningf("Chose %s/%s manifest from the list %s\nTo include the manifest list digest: %s, use --keep-manifest-list=true", t.Manifests[0].Platform.OS, t.Manifests[0].Platform.Architecture, srcDigest, srcDigest)
```
soltysh, line 414: No need for printing `srcDigest` twice, just leave the first instance.

pkg/cli/image/mirror/mirror.go
```diff
@@ -183,6 +189,14 @@ func (o *MirrorImageOptions) Complete(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Regardless of what FilterByOS was passed or set, if KeepManifestList=true,
+	// have to set wildcard filter, otherwise the manifest list digest will be altered.
+	// This means that user will mirror all manifests in the list, as there is no way to
+	// preserve the manifest list digest while also modifying its constituents.
+	if o.KeepManifestList {
+		o.FilterOptions.FilterByOS = ".*"
```
soltysh, line 197: Maybe `klog.V(1)` here stating this fact.

soltysh: Alternatively, fail validation if these two conflict?

sallyom: yup, that was my first instinct - I changed to this because in discussions it was noted that it _is_ possible to have a valid combo of `--filter-by-os` with `--keep-manifest-list=true` , if and only if the filter value includes all manifest list os/arch options.  I will go back to the validation error, though, it's clearer. We can always update later if we find people complaining (pretty unlikely, i think).

pkg/cli/image/manifest/manifest.go
sallyom: updated

pkg/cli/image/mirror/mirror.go
sallyom: Updated with new Validation to error if/when `--keep-manifest-list=true` && `--filter-by-os=not/wildcard` with a clear error message.

pkg/cli/image/mirror/mirror.go
```diff
@@ -178,8 +184,23 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
...
 func (o *MirrorImageOptions) Complete(cmd *cobra.Command, args []string) error {
-	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
+	if o.KeepManifestList {
+		o.FilterOptions.FilterByOS = ".*"
+	}
+
+	o.FilterOptions.Complete(cmd.Flags())
+
+	if err := o.FilterOptions.Validate(); err != nil {
 		return err
```
sallyom, line 204: This has to run _after_ complete, otherwise the FilterOptions.OSFilter is not set correctly (due to the defaulting that happens in FilterOptions.Complete)

pkg/cli/admin/catalog/mirror.go
```diff
@@ -144,9 +144,8 @@ func (o *MirrorCatalogOptions) Complete(cmd *cobra.Command, args []string) error
 	src := args[0]
 	dest := args[1]
 
-	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
-		return err
-	}
+	o.FilterOptions.Complete(cmd.Flags())
```
soltysh, line 147: Any particular reason for swallowing the error?

pkg/cli/image/append/append.go
```diff
@@ -146,9 +146,7 @@ func NewCmdAppendImage(streams genericclioptions.IOStreams) *cobra.Command {
 }
 
 func (o *AppendImageOptions) Complete(cmd *cobra.Command, args []string) error {
-	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
-		return err
-	}
+	o.FilterOptions.Complete(cmd.Flags())
```
soltysh, line 149: Same here

pkg/cli/image/extract/extract.go
```diff
@@ -287,9 +287,7 @@ func parseMappings(images, paths, files []string, requireEmpty bool) ([]Mapping,
 }
 
 func (o *ExtractOptions) Complete(cmd *cobra.Command, args []string) error {
-	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
-		return err
-	}
+	o.FilterOptions.Complete(cmd.Flags())
```
soltysh, line 290: Ditto

pkg/cli/image/manifest/manifest.go
```diff
@@ -158,13 +158,12 @@ func (o *FilterOptions) Validate() error {
 }
 
 // Complete performs defaulting by OS.
-func (o *FilterOptions) Complete(flags *pflag.FlagSet) error {
+func (o *FilterOptions) Complete(flags *pflag.FlagSet) {
```
soltysh, line 161: The rule of thumb is that each of the 3 methods (`Complete`, `Validate` and `Run`) will return error. Please don't change that contract.

pkg/cli/image/mirror/mirror.go
```diff
@@ -151,8 +157,8 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
 		Long:    mirrorDesc,
 		Example: mirrorExample,
 		Run: func(c *cobra.Command, args []string) {
-			kcmdutil.CheckErr(o.Complete(c, args))
 			kcmdutil.CheckErr(o.Validate())
+			kcmdutil.CheckErr(o.Complete(c, args))
```
soltysh, line 161: This order doesn't look right? Why would `Validate` be before `Complete`?

pkg/cli/image/mirror/mirror.go
```diff
@@ -178,8 +184,23 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
...
+	return nil
+}
+
 func (o *MirrorImageOptions) Complete(cmd *cobra.Command, args []string) error {
-	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
+	if o.KeepManifestList {
+		o.FilterOptions.FilterByOS = ".*"
+	}
+
+	o.FilterOptions.Complete(cmd.Flags())
```
soltysh, line 201: The error from the other comment applies here too.

pkg/cli/image/mirror/mirror.go
```diff
@@ -178,8 +184,23 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
 	return cmd
 }
 
+// This needs to run before FilterOptions.Complete because filter is set to system os/arch if/when --filter-by-os flag is not set
+// and also before MirrorImageOptions.Complete since --filter-by-os is set to wildcard if/when --keep-manifest-list=true
+func (o *MirrorImageOptions) Validate() error {
+	if o.KeepManifestList && len(o.FilterOptions.FilterByOS) > 0 && !o.FilterOptions.IsWildcardFilter() {
+		return fmt.Errorf("cannot preserve the manifest list digest while also filtering its constituents. --keep-manifest-list=true cannot be passed with --filter-by-os, unless --filter-by-os=.* \nAlso, --keep-manifest-list=true is equivalent to --filter-by-os=.*, so no need to pass both flags.")
```
soltysh, line 191: Too long message, and you're repeating yourself, I'd go with just: `--keep-manifest-list=true cannot be passed with --filter-by-os, unless --filter-by-os=.*` the full explanation is in the help

pkg/cli/admin/catalog/mirror.go
sallyom: there was no error creation in that fn, so I removed it. it only returned nil,there was no path in which there would be an error

soltysh: Yeah, I imagine, but if this changes in the future, you'll have to change all the invocations and you're breaking the API this way, it's better to be able to return error, even though you're not returning one, but you're at least setting clear expectations.

pkg/cli/image/mirror/mirror.go
sallyom: it's because in FilterOptions.Validate, the option OSFilter is set, and that has to be set only after Complete (unless we fix that and rewrite both Validate and Complete, but going down that path I saw some snags so opted for this, for now, instead)

pkg/cli/image/mirror/mirror.go
sallyom: updated

pkg/cli/image/mirror/mirror.go
sallyom: updated, got it

pkg/cli/image/manifest/manifest.go
sallyom: ok, updated

pkg/cli/image/extract/extract.go
sallyom: updated

pkg/cli/image/append/append.go
sallyom: updated

pkg/cli/image/mirror/mirror.go
sallyom: although, I think _Validate_ shouldn't _set_ any option, that should be done in _Complete_ - but the order in this particular case is tricky, will take some time to detangle that (I was stuck on it for a bit so left it as/is for now)

sallyom: The only thing Validate is there to catch is the conflicting `--keep-manifest-list=true` with `--filter-by-os=not/wildcard`

pkg/cli/image/mirror/mirror.go
```diff
@@ -179,6 +185,14 @@ func NewCmdMirrorImage(streams genericclioptions.IOStreams) *cobra.Command {
 }
 
 func (o *MirrorImageOptions) Complete(cmd *cobra.Command, args []string) error {
+	if o.KeepManifestList && len(o.FilterOptions.FilterByOS) > 0 && !o.FilterOptions.IsWildcardFilter() {
```
sallyom, line 188: @soltysh wdyt about this?  add this line to Complete like so, rather than in a new Validate fn, then keep the current Validate fn as/is (wrapper for FilterOptions.Validate), then no other change necessary, and Complete then Validate order is restored.

sallyom: ok, I added this line to the Validate Fn, as it should be, and also _only_ default filter to `.*` if user has not explicitly set that flag. I think this is good now, thanks for your suggestions/review

soltysh (approved):
/lgtm
/approve

PR #644, opened 2020-11-17
Bug 1838512: Set CGO_ENABLED=1 for cross compile on Mac

damemi:
With the changes in https://github.com/openshift/ocp-build-data/pull/627, C cross-compilers should be available in our build images. Enabling this for darwin builds of `oc` would fix the referenced bug, in which cgo is required for proper dns resolution on mac (and was fixed in upstream kubectl long ago: https://github.com/kubernetes/release/issues/469)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ generate-versioninfo:
 .PHONY: generate-versioninfo
 
 cross-build-darwin-amd64:
-	+@GOOS=darwin GOARCH=amd64 $(MAKE) --no-print-directory build GO_BUILD_PACKAGES:=./cmd/oc GO_BUILD_FLAGS:="$(GO_BUILD_FLAGS_DARWIN)" GO_BUILD_BINDIR:=$(CROSS_BUILD_BINDIR)/darwin_amd64
+	+@GOOS=darwin CGO_ENABLED=1 GOARCH=amd64 $(MAKE) --no-print-directory build GO_BUILD_PACKAGES:=./cmd/oc GO_BUILD_FLAGS:="$(GO_BUILD_FLAGS_DARWIN)" GO_BUILD_BINDIR:=$(CROSS_BUILD_BINDIR)/darwin_amd64
 .PHONY: cross-build-darwin-amd64
 
 cross-build-windows-amd64: generate-versioninfo
--- a/Makefile
+++ b/Makefile
@@ -92,8 +92,8 @@ generate-versioninfo:
 	SOURCE_GIT_TAG=$(SOURCE_GIT_TAG) hack/generate-versioninfo.sh
 .PHONY: generate-versioninfo
 
-cross-build-darwin-amd64:
-	+@GOOS=darwin CGO_ENABLED=1 GOARCH=amd64 $(MAKE) --no-print-directory build GO_BUILD_PACKAGES:=./cmd/oc GO_BUILD_FLAGS:="$(GO_BUILD_FLAGS_DARWIN)" GO_BUILD_BINDIR:=$(CROSS_BUILD_BINDIR)/darwin_amd64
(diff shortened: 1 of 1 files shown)


damemi:
/cc @soltysh 
@ecordell and @sosiouxme, could you also please take a look and let us know if there is anything else we need to do to take advantage of the changes in https://github.com/openshift/ocp-build-data/pull/627? Thanks!

soltysh:
@damemi looks like this is breaking CI pretty badly

yselkowitz:
This cannot be merged right now at least:
1. Dynamic linking of darwin binaries is currently broken in 1.15 and currently only work with 1.14; that should be fixed in 1.15.4 but we don't have that yet.
2. Cross compilers for darwin binaries are only available on the x86_64 golang-builder, so such a build will fail on other architectures.
3. In order to use the cross-compilers, CC, CXX, and -extld need to be declared.

/hold

yselkowitz:
golang-builder 1.15 can now build Mac binaries, but the second point above still applies: only the x86_64 golang-builder has the means to dynamically build Mac binaries, so this will break the cli-artifacts build on other architectures.  Furthermore, this will break the RPM build entirely, as the RPM buildroot is not set up to build Mac binaries either for any architecture.  This needs further coordination before it can be considered.

damemi:
@yselkowitz to check up on this, are the limitations you pointed out something that the oc team can work to resolve? Or is this something that is just not currently possible to accomplish with our build and release systems?

I just ask because the linked bug has been open for quite some time, and I would like to determine whether we have the ability to move forward with this or it should be closed. Thanks!

sosiouxme:
@damemi So the two problems:

> only the x86_64 golang-builder has the means to dynamically build Mac binaries, so this will break the cli-artifacts build on other architectures.

This is something the oc team can work on; we had a similar situation getting `opm` to build in operator-registry and it required some minor kludges because Dockerfile doesn't provide logic for optional contents. Everything has to be scripted to account for toolchains and outputs that exist on one architecture and not others -- the cross-compile toolchain just literally does not exist in RH for non-x86 arches.

> Furthermore, this will break the RPM build entirely, as the RPM buildroot is not set up to build Mac binaries either for any architecture.

RPMs build in a completely different environment than an image; actually, in the case of `oc`, two different environments - it has to compile against both RHEL 7 and RHEL 8 buildroots. Both of those buildroots would need to have the necessary toolchain available (this would be something ART has to investigate; not even sure how to indicate `BuildRequires` for a subset of arches, and we have to involve EXD to configure brew), and again the RPM builds need logic to account for the part of the build that only runs on x86_64 (something for the oc team).

So I think I can see a path for this to happen, but it sure sounds like a lot of work.

damemi:
@sosiouxme thank you for the detailed response. That does seem like a lot of work, so maybe we need to do a cost-benefit analysis of this effort.

The linked BZ describes that the `oc` binary we ship for mac is completely unusable when behind certain corporate VPNs. Because this only affects mac, on some VPN configurations, this is a subset of our users. This is still a very bad look, however I do not know if the size of that subset is significant enough for us to devote the work you said to fix it.

@soltysh what should we do with this? Users that are affected do have the option of compiling `oc` themselves, but that's not really a solution. Do we have some way to measure what % of users are mac, and what % of those are using OpenShift with a VPN?

I am willing to continue driving this work if necessary, but the honest opinion I seem to get from the relevant build teams is that this is difficult, or impossible, to fix with our current build tooling. I don't want to push on something this complex if it outweighs the need to provide a fix.

yselkowitz:
still want to look into this further when I have cycles
/lifecycle frozen

yselkowitz:
so much for following directions...
/remove-lifecycle stale

PR #645, opened 2020-11-18
update discovery burst to reflect lots of CRDs on openshift clusters

deads2k:
@soltysh see if this is the source of the throttling messages.  Turns out we have lots and lots of groups and versions for CRDs, especially when people start installing operators.


soltysh:
This was moved to https://github.com/kubernetes/kubernetes/pull/96763
/close

soltysh:
And oc bit is in https://github.com/openshift/oc/pull/648

PR #663, opened 2020-12-03
oc adm release new: add check for the existence of profile annotations in release manifests

csrwng:
With the introduction of cluster profiles, only manifests with a cluster profile annotation will be recognized/applied by the cluster version operator.

It should not be possible to create a release with manifests that are not annotated for at least one profile. This PR introduces a check for this.

Related enhancement:
https://github.com/openshift/enhancements/pull/200

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -45,6 +45,8 @@ import (
 	imagemanifest "github.com/openshift/oc/pkg/cli/image/manifest"
 )
 
+const profileAnnotationPrefix = "include.release.openshift.io/"
+
 func NewNewOptions(streams genericclioptions.IOStreams) *NewOptions {
 	return &NewOptions{
 		IOStreams:       streams,
@@ -736,6 +738,24 @@ func (o *NewOptions) Run() error {
 				}
 				if s, ok := m["apiVersion"].(string); !ok || s == "" {
 					return fmt.Errorf("%s: manifests must contain Kubernetes API objects with 'kind' and 'apiVersion' set: %s", filename, s)
+				}
+				metadata, ok := m["metadata"].(map[string]interface{})
+				if !ok || metadata == nil {
+					return fmt.Errorf("%s: missing metadata; manifests must set at least one profile annotation", filename)
+				}
+				annotations, ok := metadata["annotations"].(map[string]interface{})
+				if !ok || annotations == nil {
+					return fmt.Errorf("%s: missing annotations; manifests must set at least one profile annotation", filename)
+				}
+				hasProfileAnnotation := false
+				for key := range annotations {
+					if strings.HasPrefix(key, profileAnnotationPrefix) {
+						hasProfileAnnotation = true
+						break
+					}
+				}
+				if !hasProfileAnnotation {
+					return fmt.Errorf("%s: missing profile annotation; manifests must set at least one profile annotation", filename)
 				}
 				break
 			}


pkg/cli/admin/release/new.go
```diff
@@ -737,6 +739,24 @@ func (o *NewOptions) Run() error {
...
+				}
+				hasProfileAnnotation := false
+				for key := range annotations {
+					if strings.HasPrefix(key, profileAnnotationPrefix) {
+						hasProfileAnnotation = true
+						break
+					}
+				}
+				if !hasProfileAnnotation {
+					return fmt.Errorf("%s: manifests must contain at least one profile annotation set", filename)
```
wking, line 758: nit: `... must set at least one profile annotation` or `... must have at least one profile annotation set` or some such?

wking:
[images][1]:

```
error: unable to create a release: cluster-monitoring-operator/0000_50_cluster-monitoring-operator_00_0user-priority-class.yaml: manifests must contain Kubernetes API objects with 'metadata.annotations' set
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/663/pull-ci-openshift-oc-master-e2e-aws/1334307347765399552

wking:
[That manifest][1] is indeed in need of a profile declaration.

[1]: https://github.com/openshift/cluster-monitoring-operator/blob/578598731717d3417c082e6e1f87f37aa01a684a/manifests/0000_50_cluster-monitoring-operator_00_0user-priority-class.yaml

pkg/cli/admin/release/new.go
csrwng: updated error messages

csrwng:
> [That manifest](https://github.com/openshift/cluster-monitoring-operator/blob/578598731717d3417c082e6e1f87f37aa01a684a/manifests/0000_50_cluster-monitoring-operator_00_0user-priority-class.yaml) is indeed in need of a profile declaration.

There's a pr for that :) https://github.com/openshift/cluster-monitoring-operator/pull/998

sjenning:
https://github.com/openshift/cluster-kube-scheduler-operator/pull/319 should fix the CI failure

wking:
Before actually landing this, we probably need either a flag to turn the check off, or to wait a minor or two for the existing labels to sink in.  If we land this as it stands right now, someone using a master `oc` would be unable to build 4.6 and earlier releases.  Maybe that's ok, and everyone building releases is using the `oc` that they're packaging into the release?  Probably worth checking with the test-platform, release-controller, and ART folks if we want to lean on that.

smarterclayton:
This would break backwards compatibility between new and old releases.  Why is absence of annotation not implied to mean "all profiles" in the CVO?

smarterclayton:
/hold

I have some concerns here.  CLI tooling currently works on all releases of openshift.  This changes that behavior.  I agree the use case of verifying that users specify appropriate profiles is important.  I would probably focus on adding a parameter which is an allowed set of profiles and if that is set, we reject typo'd profiles.  However, you won't be able to add that flag to all invocations of the release tool without changing release controller and ci-operator slightly.   I need to see discussion of how you'd roll that flag out there.

The remaining question is whether no profile = all or none.  I don't find the argument in the enhancement compelling (and certainly doesn't require that to prevent typos).  I would probably argue that no profile = all, and we protect against typos with the allowed profiles flag in merge PRs.

smarterclayton:
The core problem is that the enhancement change is not backwards compatible to the API of release construction, and I don't agree we should break that API.

csrwng:
@smarterclayton are you referring to https://github.com/openshift/enhancements/pull/200#discussion_r375837903 ?
/cc @deads2k

smarterclayton:
Yes.  However, what problem is THIS change solving?  Someone accidentally omitted a profile?

csrwng:
Yes that's the problem it's solving. Happy to condition it on a flag.

smarterclayton:
Without a plan for how you roll that flag out to the CI infrastructure, the flag won't do much good.  If we hadn't broken forwards behavior compatibilty on the API by requiring the profile annotation, you could have hardcoded the list of allowed profiles in oc adm release new (because old release would be the same as the new).

smarterclayton:
A more effective flag for this would be to have the CVO refuse to start if any manifest contains ANY unspecified profile.  That restricts the problem to static up front AND preserves compatibility of client tooling.  While I like detecting this in tooling, we should detect this where the object is silently ignored as well.  We can then come back and decide whether we change the default behavior in CVO and release both to preserve compatibility of the payload API.

csrwng:
I like that ... @wking ^^

wking:
openshift/cluster-kube-scheduler-operator#319 landed.

/retest

> A more effective guard (at least, more effective in being more localized and more concrete) for this would be to have the CVO refuse to start if any manifest contains ANY unspecified profile. That restricts the problem to static up front...

But when you forget to label a manifest with a profile, or you typo the profile, or whatever, the result would be that the bootstrap CVO crashloops, install hangs waiting for bootstrap complete, and folks complain to the installer?  Unless I can finally talk folks into landing automatic installer analysis of gathered log bundles (openshift/installer#2567) and find some non-brittle way to highlight the CVO issue there?  I like guarding against this at release-build time, because it's a lot easier to report quickly to the person who called `oc adm release new ...`, without having to bubble the error through multiple hosts and layers of code.

> While I like detecting this in tooling, we should detect this where the object is silently ignored as well.

Which object is silently ignored?  Because neither this `oc` approach nor a "CVO-dies-on-surprising-manifest" approach would be silently ignoring anything, would they?

> The remaining question is whether no profile = all or none.

I like the fact that the current "none" interpretation makes it easy to distinguish folks who have thought about it and decided that their manifest belongs in all profiles from folks who have not heard about profiles and may not have considered whether their manifest belongs in all of them.  Of course, even the current schema does not record whether folks have considered each manifest vs. some new profile and then rejected that profile for that manifest, so maybe "I have thought about profiles" is too coarse to be worth recording.

smarterclayton:
> But when you forget to label a manifest with a profile, or you typo the profile, or whatever, the result would be that the bootstrap CVO crashloops, install hangs waiting for bootstrap complete, and folks complain to the installer? 

You wouldn't be able to merge your code change to any of the payload repos, because you wouldn't pass e2e-aws

smarterclayton:
> I like the fact that the current "none" interpretation makes it easy to distinguish folks who have thought about it and decided that their manifest belongs in all profiles from folks who have not heard about profiles and may not have considered whether their manifest belongs in all of them. Of course, even the current schema does not record whether folks have considered each manifest vs. some new profile and then rejected that profile for that manifest, so maybe "I have thought about profiles" is too coarse to be worth recording.

The current approach buries the signal in the noise.  The signal is "this is different because it has the annotation".  Right now you have to review everything to find that.  

I think the only way we can have adm release new validate profiles is to allow "" as "any", including in the CVO, and then validate one in oc adm release new the allowed set of profiles as "no annotation is valid", or "1..N annotations with known profile strings are valid".  The release tooling then continues to work with all versions of openshift, and the noise around what is the delta is cleaned up, and we only allow a certain set of profiles at release creation time.

PR #668, opened 2020-12-11
REVERT: Bug 1812813: oc debug node: create debug namespace with empty node-selector annotation #550

sallyom:
As discussed in the bugzilla report, creating the debug namespace that can override admin configured defaultNodeSelector gives too much privilege to oc and breaks the contract of debug in that you change the context you are running in. 
We are reverting this change in favor of the original workaround, that is, for the admin to create a new debug namespace with the empty node-selector and then direct oc to use that namespace for debugging.
```
$ oc adm new-project debug --node-selector=""
$ oc debug node/<node> -n debug
```

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -26,7 +26,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
-	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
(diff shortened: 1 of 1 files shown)


sallyom:
/hold

will also add code to warn/direct users to workaround

smarterclayton:
Let's revert in a separate PR and add back in the other.  The warn should be generic anyway (pod didn't get scheduled in X time or pod was rejected can both happen for many reasons)

soltysh:
> Let's revert in a separate PR and add back in the other. The warn should be generic anyway (pod didn't get scheduled in X time or pod was rejected can both happen for many reasons)

We'll want to cherry-pick that to 4.6 so I wanted to have those together.

sallyom:
@soltysh the error returned when debug node creates a pod that doesn't satisfy the default node selector set is ambiguous, pod is created w/out error then completes, it's this: https://github.com/openshift/oc/blob/master/vendor/k8s.io/kubectl/pkg/cmd/run/run.go#L688-#L690 
To get a better error will take some refactoring, maybe it's best in a separate PR.

soltysh (approved):
/lgtm
/approve

PR #671, opened 2020-12-15
Bug 1880068: Take StatefulSets, Jobs and CronJobs into account during image prune

ricardomaraschini:
This commit makes image pruner aware of StatefulSets, Jobs and CronJobs existing in the cluster.

PR #672, opened 2020-12-16
Bug 1880068: Using image name (sha) for images indexing

ricardomaraschini:
For image indexing we don't care about the imagestream name, we should
not take it into account when deciding if a prune is adequate or not.

This commit adds a test where an image in use was being wrongly pruned,
we can achieve such scenario by using following annotation on an deploy
alpha.image.policy.openshift.io/resolve-names

### DIFF
--- a/pkg/cli/admin/prune/imageprune/prune.go
+++ b/pkg/cli/admin/prune/imageprune/prune.go
@@ -46,16 +46,6 @@ func (r imageStreamTagReference) String() string {
 	return fmt.Sprintf("%s/%s:%s", r.Namespace, r.Name, r.Tag)
 }
 
-type imageStreamImageReference struct {
-	Namespace string
-	Name      string
-	Digest    string
-}
-
-func (r imageStreamImageReference) String() string {
-	return fmt.Sprintf("%s/%s@%s", r.Namespace, r.Name, r.Digest)
-}
-
 type resourceReference struct {
 	Resource  string
 	Namespace string
@@ -276,7 +266,7 @@ type Pruner interface {
 // pruner is an object that knows how to prune a data set
 type pruner struct {
 	usedTags   map[imageStreamTagReference][]resourceReference
-	usedImages map[imageStreamImageReference][]resourceReference
+	usedImages map[string][]resourceReference
 
 	images       map[string]*imagev1.Image
 	imageStreams map[string]*imagev1.ImageStream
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
/assign @soltysh 

I have seen you looking at https://github.com/openshift/oc/pull/671, would you mind looking if this one makes sense?

pkg/cli/admin/prune/imageprune/prune.go
```diff
@@ -276,7 +266,7 @@ type Pruner interface {
 // pruner is an object that knows how to prune a data set
 type pruner struct {
 	usedTags   map[imageStreamTagReference][]resourceReference
-	usedImages map[imageStreamImageReference][]resourceReference
+	usedImages map[string][]resourceReference
```
dmage, line 269: It tracks ImageStreamImages. It should protect historical records in imagestreams.

pkg/cli/admin/prune/imageprune/prune_test.go
```diff
@@ -95,6 +95,50 @@ func TestImagePruning(t *testing.T) {
...
+				imagetest.Image(
+					"sha256:0000000000000000000000000000000000000000000000000000000000000001",
+					registryHost+"/foo/bar@sha256:0000000000000000000000000000000000000000000000000000000000000001",
+				),
+			),
+			streams: Streams(
+				imagetest.Stream(
+					registryHost,
+					"namespace",
+					"name",
```
dmage, line 115: The test is not correct, the pod uses `foo/bar@sha256:000...01`, but you want to keep `namespace/name@sha256:000...01` because of this.

dmage (changes requested):
I don't understand why you want to keep an image in `namespace/name` when you have a reference for `foo/bar`.

pkg/cli/admin/prune/imageprune/prune_test.go
ricardomaraschini: But there is no image `namespace/name@sha256:000...01`, what do you mean? Before this PR the image used by the pod was being deleted.

dmage: You have only one imagestream: `namespace/name`. And only one pod that uses `foo/bar@...`. The pod uses an imagestream that doesn't exist, so its reference is not valid and can be ignored by the pruner.

ricardomaraschini:
According to PR [feedback](https://github.com/openshift/oc/pull/672#discussion_r551462708) the old behavior is correct, these images should be deleted. Closing this one.

PR #673, opened 2020-12-16
Bug 1902299: catalog mirror improvements

ecordell:
This fixes some issues with oc adm catalog mirror, post-appregistry:

1. Images without namespaces were not generating proper mappings. Example: `registry.access.redhat.com/ubi8-minimal`
1. `--filter-by-os` was previously allowed to filter arches of mirrored content. This would break references to those images in the catalog, which point the manifestlist and not the manifest.
1. `--filter-by-os` now _only_ filters the index image that is pulled and unpacked. To clarify this, a new flag `--index-filter-by-os` was added and should be used instead.
1. Because of the way `String()` is calculated on parsed references, some valid targets were failing validation. Minimal validation for a mirroring target now only checks that a hostname has been parsed successfully.

These need to be backported to 4.6 as the address issues with mirroring index images (vs. previous workflows that assumed mirroring started with `oc adm catalog build`)


jianzhangbjz:
Depends on https://github.com/openshift/library-go/pull/973

ecordell:
/retest

Quota issues

pkg/cli/admin/catalog/mirror.go
```diff
@@ -123,9 +124,18 @@ func NewMirrorCatalog(f kcmdutil.Factory, streams genericclioptions.IOStreams) *
...
+	// image, to indicate which arch should be used to extract the catalog db (the database inside should be the same
+	// for all arches, so this flag should never need to be set explicitly for standard workflows).
+	// this flag is renamed to make it clear that the underlying images are not filtered
+	flags.StringVar(&o.FilterOptions.FilterByOS, "index-filter-by-os", o.FilterOptions.FilterByOS, "A regular expression to control which index image is picked when multiple variants are available. Images will be passed as '<platform>/<architecture>[/<variant>]'. This does not apply to images referenced by the index.")
+
+	// the old flag name is kept for backwards-compatibility.
+	// if both old and new are specified, the value of the flag coming later will be used.
+	flags.StringVar(&o.FilterOptions.FilterByOS, "filter-by-os", o.FilterOptions.FilterByOS, "Use --index-filter-by-os instead. A regular expression to control which index image is picked when multiple variants are available. Images will be passed as '<platform>/<architecture>[/<variant>]'. This does not apply to images referenced by the index.")
```
soltysh, line 137: I'd suggest deprecating this flag, not to confuse users. Can be done in a follow-up.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -144,8 +154,10 @@ func (o *MirrorCatalogOptions) Complete(cmd *cobra.Command, args []string) error
 	src := args[0]
 	dest := args[1]
 
-	if err := o.FilterOptions.Complete(cmd.Flags()); err != nil {
-		return err
+	// default to linux/arm64 for index image, which we generally expect to exist
```
soltysh, line 157: Nit: comment says default to `linux/arm64` but you default to `linux/amd64`

pkg/cli/admin/catalog/mirror.go
```diff
@@ -144,8 +154,10 @@ func (o *MirrorCatalogOptions) Complete(cmd *cobra.Command, args []string) error
...
+	pattern := o.FilterOptions.FilterByOS
+	if len(pattern) == 0 {
+		o.FilterOptions.FilterByOS = regexp.QuoteMeta(fmt.Sprintf("%s/%s", "linux", "amd64"))
```
soltysh, line 160: This is excessive, just pass `linux/amd64`.

soltysh (approved):
A few nits that can be addressed in a follow-up
/approve

PR #675, opened 2020-12-17
Bug 1812813: oc debug node: give better error for pod failure

sallyom:
This is followup of https://github.com/openshift/oc/pull/668 

/assign @soltysh 
/cc @smarterclayton 

Now, if a defaultNodeSelector is set so that debug pod cannot run on any node, user will see this:
```
$ oc debug node/ip-10-0-245-28-[node-1](https://issues.redhat.com/browse/node-1)
error: debug pod failed: NodeAffinity Pod Predicate NodeAffinity failed
```
I briefly explored matching node labels with debug-pod NodeSelector for even better error, but that might be overkill. I think the above is enough information.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -798,7 +798,9 @@ func (o *DebugOptions) createPod(pod *corev1.Pod) (*corev1.Pod, error) {
 	// create the pod
 	created, err := o.CoreClient.Pods(namespace).Create(context.TODO(), pod, metav1.CreateOptions{})
 	if err == nil || !kapierrors.IsAlreadyExists(err) {
-		return created, err
+		if err = o.checkStatusForFailure(created); err == nil {
+			return created, err
+		}
 	}
 
 	// only continue if the pod has the right annotations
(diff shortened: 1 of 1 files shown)


pkg/cli/debug/debug.go
```diff
@@ -170,13 +172,14 @@ func NewDebugOptions(streams genericclioptions.IOStreams) *DebugOptions {
...
-		IOStreams:          streams,
-		Timeout:            15 * time.Minute,
-		KeepInitContainers: true,
-		AsUser:             -1,
-		Attach:             *attachOpts,
-		LogsForObject:      polymorphichelpers.LogsForObjectFn,
+		PrintFlags:             genericclioptions.NewPrintFlags("").WithTypeSetter(scheme.Scheme),
+		IOStreams:              streams,
+		Timeout:                15 * time.Minute,
+		DebugPodRunningTimeout: 30 * time.Second,
```
sallyom, line 178: i'm not sure what time limit is appropriate to wait for a pod status of PodRunning, but in case of this bug, pod immediately fails so cmd immediately exits.

sallyom:
/hold   

still working this out

pkg/cli/debug/debug.go
```diff
@@ -798,7 +800,39 @@ func (o *DebugOptions) createPod(pod *corev1.Pod) (*corev1.Pod, error) {
...
+	}
+	return debugPod, nil
+}
+
+func (o *DebugOptions) checkStatusForFailure(pod *corev1.Pod) error {
+	var err error
+	// check for early pod failures - such as node affinity errors that will be best caught here
+	// wait 3s for better failure message, otherwise return generic failure error
+	err = wait.PollImmediate(1*time.Second, 3*time.Second, func() (bool, error) {
+		pod, err = o.CoreClient.Pods(pod.Namespace).Get(context.TODO(), pod.Name, metav1.GetOptions{})
```
sallyom, line 818: unless i'm mistaken, in testing I saw as kube-apiserver was rolling out w/ defaultNodeSelector update, the podStatus=fail precedes the message/reason so was left with 
`error: debug pod failed status: ` instead of `error: debug pod failed status: Pod Predicate NodeAffinity failed`

soltysh (changes requested):
I'm not sure that this approach will give you the desired outcome. You're checking pod status right after it is created, which in a slightly busy cluster will never fail for you. I'd probably try to create some goroutine after the pod is created and check for pod healthiness there and eventually fail the debug flow, or at least inform about issues.

PR #676, opened 2020-12-18
Bug 1909289: Should be able to debug an init container

smarterclayton:
The logic for waiting for the container to be ready before starting
the attach did not correctly take into account init containers.

The logic in PodContainerRunning is refactored to:

1. Remove the early exit if the pod is pending (pod will be pending
   for init containers)
2. Instead of exiting when a container is waiting, print a message
   to stderr that indicates what container has not been loaded. The
   exit is inappropriate because a transient pull error can happen
   at any time.
3. Only check init containers or the requested container name for
   waiting - it's irrelevant if another container is 'waiting' when
   debugging
4. When the pod is terminal, check init containers and regular
   containers and report any container with exit code != 0, not just
   regular containers
5. Stop treating unknown pod phase as an error - just use the standard checking loop (a pod could go unknown briefly because of node disruption, and then start responding later)

This ensures that debug behaves correctly for init containers and
regular containers and is more tolerant of transient errors.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -505,14 +505,30 @@ func (o *DebugOptions) RunDebug() error {
 			return false, nil
 		}
 
+		notifyFn := func(pod *corev1.Pod, container corev1.ContainerStatus) error {
+			// TODO: instead of reporting to the user a message, accumulate a certain amount of time in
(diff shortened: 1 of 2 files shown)


soltysh (approved):
/lgtm

PR #679, opened 2020-12-21
Bug 1878022: Increasing timeout for image import

ricardomaraschini:
This PR increases the timeout for an image import up to one minute,
similar to what is done by openshift-controller-manager.

### DIFF
--- a/pkg/cli/importimage/importimage.go
+++ b/pkg/cli/importimage/importimage.go
@@ -193,7 +193,7 @@ func (o *ImportImageOptions) Run() error {
 		Namespace(isi.Namespace).
 		Resource(imagev1.Resource("imagestreamimports").Resource).
 		Body(isi).
-		Timeout(time.Minute).
+		Timeout(time.Hour).
 		Do(context.TODO()).
 		Into(result)
 	if err != nil {
--- a/pkg/cli/importimage/importimage.go
+++ b/pkg/cli/importimage/importimage.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -187,7 +188,14 @@ func (o *ImportImageOptions) Run() error {
 		return err
 	}
 
-	result, err := o.imageClient.ImageStreamImports(isi.Namespace).Create(context.TODO(), isi, metav1.CreateOptions{})
+	result := &imagev1.ImageStreamImport{}
+	err = o.imageClient.RESTClient().Post().
+		Namespace(isi.Namespace).
+		Resource(imagev1.Resource("imagestreamimports").Resource).
+		Body(isi).
+		Timeout(time.Minute).
+		Do(context.TODO()).
+		Into(result)
 	if err != nil {
 		return err
 	}


pkg/cli/importimage/importimage.go
```diff
@@ -187,7 +188,14 @@ func (o *ImportImageOptions) Run() error {
 		return err
 	}
 
-	result, err := o.imageClient.ImageStreamImports(isi.Namespace).Create(context.TODO(), isi, metav1.CreateOptions{})
+	result := &imagev1.ImageStreamImport{}
+	err = o.imageClient.RESTClient().Post().
+		Namespace(isi.Namespace).
+		Resource(imagev1.Resource("imagestreamimports").Resource).
+		Body(isi).
+		Timeout(time.Minute).
```
dmage, line 196: Let's have 1 hour there, the same value that we have in OCM.

dmage:
/lgtm
I guess an approver will ask you to squash these commits.

dmage:
@soltysh can you take a look?

xiuwang:
make oc from the fix,could import image from the whole repository.

`I0115 14:41:57.651660   26786 round_trippers.go:422] POST https://api.xiuwang-custombug.qe.devcluster.openshift.com:6443/apis/image.openshift.io/v1/namespaces/build43/imagestreamimports?timeout=1h0m0s`
$oc get is  
NAME               IMAGE REPOSITORY                                                            TAGS                                                      UPDATED
myruby             image-registry.openshift-image-registry.svc:5000/xiuwang/myruby             2.5,2.5-11,2.5-13,2.5-20,2.5-20.1535038158 + 45 more...   16 minutes ago

pkg/cli/importimage/importimage.go
```diff
@@ -187,7 +188,14 @@ func (o *ImportImageOptions) Run() error {
 		return err
 	}
 
-	result, err := o.imageClient.ImageStreamImports(isi.Namespace).Create(context.TODO(), isi, metav1.CreateOptions{})
+	result := &imagev1.ImageStreamImport{}
+	err = o.imageClient.RESTClient().Post().
+		Namespace(isi.Namespace).
+		Resource(imagev1.Resource("imagestreamimports").Resource).
+		Body(isi).
+		Timeout(time.Hour).
```
soltysh, line 196: For one increasing this to 1h is not acceptable? Two, have you tried `--request-timeout`?

soltysh (changes requested):
/hold

pkg/cli/importimage/importimage.go
ricardomaraschini: I liked your idea of having this as a command line flag (I could not locate it on the current version, at least). What do you think @dmage ?

dmage: We are fixing regression, --request-timeout (if it works) may be considered as a workaround, but it used to work without additional flags. 1h is acceptable for everybody (who has supported configuration), we have the same timeout in [openshift-controller-manager](https://github.com/openshift/openshift-controller-manager/blame/e1af3a50434542be4e9fd227e5203fda41835b06/pkg/image/controller/imagestream_controller.go#L310).

ricardomaraschini: Ah, yes, I was not aware of `oc options`. Let me see if that works.

ricardomaraschini: Wow, thanks @soltysh 

```
$ /tmp/oc --request-timeout=1h import-image myruby --from=foo --all --confirm --insecure --loglevel=8
...
I0118 16:17:48.882223  101857 round_trippers.go:422] POST https://api.crc.testing:6443/apis/image.openshift.io/v1/namespaces/rmarasch/imagestreamimports?timeout=1h0m0s
...

I am inclined to just close the BZ now. What do you reckon @dmage ?
```

dmage: @soltysh is there a way to change the default value for `oc import-image`?

soltysh: It's a field on `rest.Config` so it's doable, but then again the default is no time-out, and it's the server that times out the request unless explicitly asked not to do it. I'd personally update help and docs with information about `--request-timeout` it's there, just not front of people's eyes.

dmage: The default was no time-out, but today it's 34 seconds:

https://github.com/kubernetes/kubernetes/blob/f35df2f8411ea68891c9fa829733c497002d4b97/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/rest.go#L499

https://github.com/kubernetes/kubernetes/blob/018942a92a90f91261ba63971827f2ec6b64587e/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/create.go#L61-L79

Therefore `oc import-image` stopped to work for big repositories. But given that customers didn't report about the problem and that the API server already has a bit incompatible change, maybe documenting this is sufficient.

ricardomaraschini:
Closing this one an opening a [new one](https://github.com/openshift/oc/pull/712) where help messages got improved to provide an example of `--request-timeout` flag.

PR #685, opened 2021-01-05
Bug 1904006: 'oc image extract' properly wire --dir

sallyom:
It seems the --dir flag was never working correctly, `oc image extract` always was setting the directory to the current directory. 

/assign @soltysh

### DIFF
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -188,7 +188,7 @@ type Mapping struct {
 	ConditionFn func(m *Mapping, dgst digest.Digest, imageConfig *dockerv1client.DockerImageConfig) (bool, error)
 }
 
-func parseMappings(images, paths, files []string, requireEmpty bool) ([]Mapping, error) {
+func parseMappings(images, paths, files []string, fileDir string, requireEmpty bool) ([]Mapping, error) {
 	layerFilter := regexp.MustCompile(`^(.*)\[([^\]]*)\](.*)$`)
 
 	var mappings []Mapping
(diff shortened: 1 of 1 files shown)


sallyom:
clearly i'm missing something here...extract test is failing, looking closer now

pkg/cli/image/extract/extract.go
```diff
@@ -188,7 +188,7 @@ type Mapping struct {
 	ConditionFn func(m *Mapping, dgst digest.Digest, imageConfig *dockerv1client.DockerImageConfig) (bool, error)
 }
 
-func parseMappings(images, paths, files []string, requireEmpty bool) ([]Mapping, error) {
+func parseMappings(images, paths, files []string, fileDir string, requireEmpty bool) ([]Mapping, error) {
```
smarterclayton, line 191: Hrm, dir should be applied later (IIRC, but I might be wrong), so I'm not even sure this is the correct code path.  Will have to review.

smarterclayton: FileDir has nothing to do with extract.   --dir controls "the directory on disk where image sources of type file will be located", not "the path that things are etxracted to".

/hold

smarterclayton: FileDir is already used https://github.com/openshift/oc/pull/685/files#diff-77442d06befb7d25d1cf6e73620544af7c772eac2d2dd6f457bb3062117350b9L323

If we need to update the help text to make this clear, we should so with this PR.

sallyom: So, there is no flag to control where things are extracted to? Always to the current directory?

soltysh: maybe we could similarly how `oc adm release mirror` has:
```
--to-dir='': A directory to export images to.
```
?

sallyom:
In light of the fact that 
`oc image extract image:latest --dir=some-dir` does _not_ equate to `extract the contents of image:latest into the directory some-dir`, the `--dir` flag does not at all make sense for the image extract command. I think we should remove this option from the image extract command (it doesn't work, isn't wired anyways, so no way anyone will miss it).  I believe append has the same issue, so will address both in this PR: https://github.com/openshift/oc/pull/729

pkg/cli/image/extract/extract.go
sallyom: yes, I've added to-dir to the new PR, same code as this but new flag --to-dir rather than --dir

PR #687, opened 2021-01-06
Bug 1908484: credentials-requests extract only write creds requests not all manifests

sallyom:
`oc adm release extract`: with --credentials-requests flag, only write the credentials requests manifests, don't write _all_ the manifests.


soltysh (approved):
/lgtm
/approve

PR #688, opened 2021-01-06
Bug 1914452: Fix the manifest list warning when choosing an os/arch digest from list

sallyom:
The warning issued when oc chooses an os/arch digest from a manifest list should not include the suggestion to run --keep-manifest-list=true, since that flag is only exposed during mirroring. the warning is issued with oc imageextract _and_ mirror

warning comes from `ProcessManifestList`, exercised with extract & mirror, however
`oc image extract` --keep-manifest-list is not exposed
`oc image mirror` --keep-manifest-list is exposed


soltysh (approved):
/lgtm
/approve

PR #691, opened 2021-01-08
Bug 1890995: oc new-app: provide message for unauthorized error with image lookup

sallyom:
Provide msg for UnuathorizedError with `oc new-app` 
Example: (private image quay.io/sallyom/podescape-io:test)
Before this PR:
```console
$ oc new-app --docker-image=quay.io/sallyom/podescape-io:test
error: unable to locate any local docker images with name "quay.io/sallyom/podescape-io:test"
```
With this PR:
```console
$ oc new-app --docker-image=quay.io/sallyom/podescape-io:test
W0108 17:39:59.507884 3164436 dockerimagelookup.go:240] container image remote registry lookup failed: you may not have access to the container image "quay.io/sallyom/podescape-io:test"
error: unable to locate any local docker images with name "quay.io/sallyom/podescape-io:test"
```

/assign @soltysh

### DIFF
--- a/pkg/helpers/newapp/app/dockerimagelookup.go
+++ b/pkg/helpers/newapp/app/dockerimagelookup.go
@@ -230,16 +230,16 @@ func (s ImageImportSearcher) Search(precise bool, terms ...string) (ComponentMat
 		if image.Status.Status != metav1.StatusSuccess {
 			klog.V(4).Infof("image import failed: %#v", image)
 			switch image.Status.Reason {
-			case metav1.StatusReasonInternalError:
+			case metav1.StatusReasonInternalError, metav1.StatusReasonUnauthorized:
 				// try to find the cause of the internal error
 				if image.Status.Details != nil && len(image.Status.Details.Causes) > 0 {
 					for _, c := range image.Status.Details.Causes {
-						klog.Warningf("container image registry lookup failed: %s", c.Message)
+						klog.Warningf("container image remote registry lookup failed: %s", c.Message)
 					}
 				} else {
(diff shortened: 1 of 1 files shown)


pkg/helpers/newapp/app/dockerimagelookup.go
```diff
@@ -230,16 +230,16 @@ func (s ImageImportSearcher) Search(precise bool, terms ...string) (ComponentMat
...
-						klog.Warningf("container image registry lookup failed: %s", c.Message)
+						klog.Warningf("container image remote registry lookup failed: %s", c.Message)
 					}
 				} else {
-					klog.Warningf("container image registry lookup failed: %s", image.Status.Message)
+					klog.Warningf("container image remote registry lookup failed: %s", image.Status.Message)
 				}
-			case metav1.StatusReasonInvalid, metav1.StatusReasonUnauthorized, metav1.StatusReasonNotFound:
+			case metav1.StatusReasonInvalid, metav1.StatusReasonNotFound:
 			default:
```
sallyom, line 243: These image.Status failures do not produce an error - I'm only adding the 'StatusReasonUnauthorized' to the list of reasons to print a warning, but leave the others as/is.  The unauthorized reason is pretty mild, so even if the new-app cmd succeeds to find the same image in local storage, it's not confusing to also see that the remote registry was tried first. User will see the msg: `you may not have access to the container image "registry/image"` so if the cmd ultimately fails, they'll have more idea why, but if it succeeds due to local storage, then that's ok, too, imo.

soltysh (approved):
/lgtm
/approve

Issue #647, opened 2020-11-20
oc project not works with cluster proxy.


wanghaoran1988:
We are using a proxy as the apiserver that can reverse proxy to multiple clusters, so we has config like:
```
apiVersion: v1
clusters:
- cluster:
    server: http://localhost:8001/backplane/cluster/1h2cu64jqeufe3dnk7gvpl7j3ui3msut/
  name: 1h2cu64jqeufe3dnk7gvpl7j3ui3msut
contexts:
- context:
    cluster: 1h2cu64jqeufe3dnk7gvpl7j3ui3msut
    namespace: openshift-logging
    user: haoran.openshift
  name: default/1h2cu64jqeufe3dnk7gvpl7j3ui3msut/haoran.openshift
current-context: default/1h2cu64jqeufe3dnk7gvpl7j3ui3msut/haoran.openshift
kind: Config
preferences: {}
users:
- name: haoran.openshift
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - /Users/haowang/.kube/ocm-token
      command: bash
      env: null
```
When we perform oc project <project>
It will create a new context&server&user as follows:
```
apiVersion: v1
clusters:
- cluster:
    server: http://localhost:8001/backplane/cluster/1h2cu64jqeufe3dnk7gvpl7j3ui3msut/
  name: 1h2cu64jqeufe3dnk7gvpl7j3ui3msut
- cluster:
    server: http://localhost:8001/backplane/cluster/1h2cu64jqeufe3dnk7gvpl7j3ui3msut/
  name: localhost:8001
contexts:
- context:
    cluster: 1h2cu64jqeufe3dnk7gvpl7j3ui3msut
    namespace: openshift-monitoring
    user: haoran.openshift
  name: default/1h2cu64jqeufe3dnk7gvpl7j3ui3msut/haoran.openshift
- context:
    cluster: localhost:8001
    namespace: default
    user: system:serviceaccount:openshift-backplane-srep:89bb76f8aa82917707763cfb8c4a01a5/localhost:8001
  name: default/localhost:8001/system:serviceaccount:openshift-backplane-srep:89bb76f8aa82917707763cfb8c4a01a5
current-context: default/localhost:8001/system:serviceaccount:openshift-backplane-srep:89bb76f8aa82917707763cfb8c4a01a5
kind: Config
preferences: {}
users:
- name: haoran.openshift
  user:
    exec:
[...]

PR #692, opened 2021-01-11
Bug 1944581: Preserve AuthInfo when switching projects

feichashao:
This is a possible fix for https://github.com/openshift/oc/issues/647

Before this patch, if the openshift cluster is behind a cluster proxy (eg. kubectl proxy), and the authentication method is not token, `oc project <project-name>` will create a new user in `kubeconfig`, with blank info. This will lead to further `oc` commands return authentication error.

```
- name: system:serviceaccount:openshift-xxxx-xxxx:xxxxx
  user: {}
```

With this patch, when `oc project <project-name>` creating a new `kubeconfig`, it will preserve all the `AuthInfo` from `RESTConfig`, so that it won't break the customized settings.

```
- name: system:serviceaccount:openshift-xxxxx-xxxx/xxx-xxx-xxxx-xxxxx-xxxx-xx-xxxx-com:443
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - /home/xxx/.kube/xxx-token
      command: bash
      env:
      - name: OCM_CONFIG
        value: /home/xxx/.xxxx.xxxx.stg
      provideClusterInfo: false
```

`clientcmdapi` will handle the null pointer and null strings, so it won't break the existing `oc login` or `oc project` mechanism.

### DIFF
--- a/pkg/helpers/kubeconfig/smart_merge.go
+++ b/pkg/helpers/kubeconfig/smart_merge.go
@@ -58,6 +58,9 @@ func CreateConfig(namespace, userName string, clientCfg *restclient.Config) (*cl
 
 	credentials := clientcmdapi.NewAuthInfo()
 	credentials.Token = clientCfg.BearerToken
+	credentials.TokenFile = clientCfg.BearerTokenFile
+	credentials.AuthProvider = clientCfg.AuthProvider
+	credentials.Exec = clientCfg.ExecProvider
 	credentials.ClientCertificate = clientCfg.TLSClientConfig.CertFile
 	if len(credentials.ClientCertificate) == 0 {
 		credentials.ClientCertificateData = clientCfg.TLSClientConfig.CertData


feichashao:
The `ci/prow/e2e-agnostic-cmd` failure is pretty common accross other PRs, so i think it is not related to this PR change.

The first commit preserves `Impersonate` + `Exec` + `tokenfile` to kubeconfig when swithing projects.
https://github.com/openshift/oc/pull/692/commits/f826dc2442943b8bdb94b182fd2dde7e259b19ac

The second commit remove the presevation of `Impersonate`, because users/automated-scripts might use `oc --as=<someone>` to switch project. If we put the `Impersonate` info into kubeconfig when switch projects, it is possible to break existing scripts.
https://github.com/openshift/oc/pull/692/commits/d7fd77cd0ca967e5399b00ad0b77ec79b31fd9b6


The `oc` command does not have options for `Exec`, `tokenfile` and `AuthProvider`, so it should be fine to preseve those info when swithing project, without breaking the existing behavior.

feichashao:
Testes passed. @soltysh could you take a look?

soltysh (approved):
This looks good, please squash changes and I'll tag

soltysh (approved):
/lgtm

PR #695, opened 2021-01-12
Bug 1858760: oc registry login: check that registry passed matches internal registry if using internal registry

sallyom:
If user has a valid kubeconfig and they "oc registry login --registry someother/registry:5000" the bearer token from the valid kubeconfig will be recorded as credentials for the given registry, unless the '--auth-basic' flag is also passed. This PR adds a check to prevent this. 

To reproduce this issue, while currently authenticated w/ a token to a cluster, run 'oc registry login --registry localhost:5000' (or any other registry that is not the internal registry). You'll see the config file (~/.docker/config.json) will be updated with the auth for the given registry and the bearer token from the valid but un-matching kubeconfig. However, if one instead passes the '--auth-basic user:pw' this issue will be avoided, config file will be updated with the correct user:pw auth. 

Not sure this is the only/best way to resolve this, looking for review. 

/assign @soltysh

### DIFF
--- a/pkg/cli/registry/login/login.go
+++ b/pkg/cli/registry/login/login.go
@@ -204,6 +204,30 @@ func (o *LoginOptions) Complete(f kcmdutil.Factory, args []string) error {
 	default:
 		if len(cfg.BearerToken) == 0 {
 			return fmt.Errorf("no token is currently in use for this session")
+		} else if len(o.HostPort) > 0 {
+			// check that any registry passed matches any config being used
+			// otherwise, bearertoken from valid kubeconfig can be recorded as credentials for any given registry,
+			// in absence of basic auth.
+			ns, _, err := f.ToRawKubeConfigLoader().Namespace()
+			if err != nil {
+				return err
+			}
+
+			client, err := imageclient.NewForConfig(cfg)
+			if err != nil {
+				return err
+			}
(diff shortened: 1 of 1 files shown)


smarterclayton:
This behavior was deliberate.  You want to login to the current registry, but use a different DNS name.

sallyom:
/hold

Since the behavior is intentional and issue is not a bug, I'll re-work this to add help text to clarify

pkg/cli/registry/login/login.go
```diff
@@ -204,6 +204,30 @@ func (o *LoginOptions) Complete(f kcmdutil.Factory, args []string) error {
...
+			if err != nil {
+				return err
+			}
+			registry, internal, err := findPublicHostname(client, ns, "openshift")
+			if err != nil {
+				return err
+			}
+			if len(registry) > 0 {
+				if ref, err := reference.Parse(registry); err == nil {
+					if internal && o.HostPort != ref.Registry {
```
sallyom, line 226: At this point, if using internal registry, o.HostPort should match ref.Registry. If not, there's a problem.
If a user is wanting to add a DNS name for the registry, then this will no longer be internal=true, and it would be up to the user to also pass the `--auth-basic=user:pwtoken` to `oc registery login --registry external/registry:5000 --auth-basic=user:pw`

sallyom:
opening new PR to add help text

PR #696, opened 2021-01-12
Bug 1899575: bump discovery burst to 250

soltysh:
/assign @ingvagabund


arve0:
Can the flag be set by users?

soltysh:
> Can the flag be set by users?

We don't expect that to be user-settable.

PR #697, opened 2021-01-12
Bug 1890677: oc image: Include error msg when image not found and register oci schema for mediatypes

sallyom:
/assign @soltysh 

oc was burying image not accessed error messages with "image does not exist" message
([Bug 1890677](https://bugzilla.redhat.com/show_bug.cgi?id=1890677))

This PR gives more information about why an image was not found - and clarifies error to be `image not found` rather than `image does not exist` 

Also, added the `ocischema` package to register OCI media types w/ oc  (the error from the bug report will no longer be an error)

### DIFF
--- a/pkg/cli/image/info/info.go
+++ b/pkg/cli/image/info/info.go
@@ -400,7 +400,7 @@ func (o *ImageRetriever) Run() error {
 						return callbackFn(name, nil, imagemanifest.NewImageForbidden(msg, err))
 					}
 					if imagemanifest.IsImageNotFound(err) {
-						msg := fmt.Sprintf("image %q does not exist", from)
+						msg := fmt.Sprintf("image %q not found: %s", from, err.Error())
 						return callbackFn(name, nil, imagemanifest.NewImageNotFound(msg, err))
 					}
 					return callbackFn(name, nil, fmt.Errorf("unable to read image %s: %v", from, err))
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -346,7 +346,7 @@ func (o *ExtractOptions) Run() error {
 						return imagemanifest.NewImageForbidden(msg, err)
 					}
 					if imagemanifest.IsImageNotFound(err) {
-						msg := fmt.Sprintf("image %q does not exist", from)
+						msg := fmt.Sprintf("image %q not found: %s", from, err.Error())
 						return imagemanifest.NewImageNotFound(msg, err)
 					}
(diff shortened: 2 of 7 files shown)


sallyom:
NOTE: `oc` should handle manifests of type `application/vnd.oci.image.manifest.v1+json` 
I'll open an issue/bz for that - This PR is only to stop burying that (or any other) error when an image is not found.

pkg/cli/image/info/info.go
```diff
@@ -400,7 +400,7 @@ func (o *ImageRetriever) Run() error {
 						return callbackFn(name, nil, imagemanifest.NewImageForbidden(msg, err))
 					}
 					if imagemanifest.IsImageNotFound(err) {
-						msg := fmt.Sprintf("image %q does not exist", from)
+						msg := fmt.Sprintf("image %q not found: %s", from, err.Error())
 						return callbackFn(name, nil, imagemanifest.NewImageNotFound(msg, err))
```
wking, line 404: Is it worth updating `NewImageNotFound` to suffix non-`nil` errors?  It just [serializes as the `msg` today][1], but updating it would mean we'd get the `err`-level detail in all of our call-sites, not just this one.  Although it's only one other call-site which has a chance at a non-`nil` error:

```console
$ git --no-pager grep '\.NewImageNotFound('
pkg/cli/image/extract/extract.go:                                               return imagemanifest.NewImageNotFound(msg, err)
pkg/cli/image/info/info.go:                                             return callbackFn(name, nil, imagemanifest.NewImageNotFound(msg, err))
pkg/cli/image/info/info.go:                                     return imagemanifest.NewImageNotFound(fmt.Sprintf("no manifests could be found for %q", from), nil)
```

[1]: https://github.com/openshift/oc/blob/9eb2f8f9d5b573ffeb720f17bc45cc8507f861bd/pkg/cli/image/manifest/errors.go#L17-L18

sallyom:
@sjenning @wking I have figured out how to register the oci.image.manifest  mediaType, I'll put a hold here until I push that commit, might as well fix that w/ this PR :) 

/hold

pkg/cli/image/info/info.go
sallyom: yup I did notice there are others, I'll update

sallyom:
> @sjenning @wking I have figured out how to register the oci.image.manifest mediaType, I'll put a hold here until I push that commit, might as well fix that w/ this PR :)
> 
> /hold

/hold cancel

soltysh (approved):
/lgtm
/approve
/retest

PR #707, opened 2021-01-14
Bug 1906276: oc image append|extract: clarify help for --filter-by-os

sallyom:
Clarify in help menu that `--filter-by-os=.*` is not supported with `oc image extract` and `oc image append`. 

/assign @soltysh


soltysh (approved):
/lgtm
/approve

PR #712, opened 2021-01-19
Bug 1878022: Improving import-image help messages

ricardomaraschini:
Following our discussions on https://github.com/openshift/oc/pull/679 this PR makes `oc help image-import` to contain an example of `--request-timeout` flag usage.

### DIFF
--- a/pkg/cli/importimage/importimage.go
+++ b/pkg/cli/importimage/importimage.go
@@ -39,7 +39,23 @@ var (
 		see the 'tag' command.`)
 
 	importImageExample = templates.Examples(`
+		# Import tag latest into a new image stream.
+		oc import-image mystream --from=registry.io/repo/image:latest --confirm
+
+		# Update imported data for tag latest in an already existing image stream.
 		oc import-image mystream
+
+		# Update imported data for tag stable in an already existing image stream.
+		oc import-image mystream:stable
+
+		# Update imported data for all tags in an existing image stream.
+		oc import-image mystream --all
+
+		# Import all tags into a new image stream.
+		oc import-image mystream --from=registry.io/repo/image --all --confirm
+
+		# Import all tags into a new image stream using a custom timeout.
+		oc --request-timeout=5m import-image mystream --from=registry.io/repo/image --all --confirm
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/approve

pkg/cli/importimage/importimage.go
```diff
@@ -39,7 +39,17 @@ var (
 		see the 'tag' command.`)
 
 	importImageExample = templates.Examples(`
+		# Update imported data for an already existing image stream.
 		oc import-image mystream
```
dmage, line 43: If I recall correctly, it'll update only mystream:latest.

pkg/cli/importimage/importimage.go
```diff
@@ -39,7 +39,17 @@ var (
...
+
+		# Import image tags into a new image stream.
+		oc import-image mystream --from=registry.io/repo/image --all --confirm
+
+		# Import image tags into a new image stream using a custom timeout.
+		oc --request-timeout=5m import-image mystream --from=registry.io/repo/image --all --confirm
+
+		# Import image tag into a new image stream.
+		oc import-image mystream:latest --from=registry.io/repo/image:latest --confirm
```
dmage, line 52: It'll work only if mystream:latest doesn't exist, right?

dmage (commented):
import-image have a bit funny semantics (at least I expected slightly different behavior). I'd like us to document that `mystream` means `mystream:latest` (and only `latest`). If you want to update all tags, you should use `--all`.

pkg/cli/importimage/importimage.go
ricardomaraschini: No, it works even if "mystream:latest" exists.

pkg/cli/importimage/importimage.go
ricardomaraschini: Yes, you are right. Let me add some info about it as well.

ricardomaraschini: I have adjusted, PTAL.

PR #713, opened 2021-01-22
Bug 1919032: Fix image extract from Root Directory

zerodayz:
This patch adds possibility to extract file from the root directory of an image.

Closes: BZ#1919032

### DIFF
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -387,10 +387,13 @@ func (o *ExtractOptions) Run() error {
 						alter = append(alter, newCopyFromDirectory(mapping.From))
 					default:
 						name, parent := path.Base(mapping.From), path.Dir(mapping.From)
-						if name == "." || parent == "." {
+						if name != "." && parent == "." {
+							alter = append(alter, newCopyFromPattern(parent, name, true))
+						} else if name == "." || parent == "." {
 							return fmt.Errorf("unexpected directory from mapping %s", mapping.From)
+						} else {
+							alter = append(alter, newCopyFromPattern(parent, name, false))
 						}
-						alter = append(alter, newCopyFromPattern(parent, name))
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


zerodayz:
the problem is that when only file is being a source for example /help.1, the directory "parent" is "." and it was excluded because of the name == "." || parent == ".".

Even fixing this was not enough since tar doesn't consider /. meaning the root directory is omited, for example:
usr/lib/ instead of /usr/lib.

Hence adding new functions. I am unsure about the case/switch that can be further improved. Any feedback welcome @sallyom Let me know.

pkg/cli/image/extract/extract.go
```diff
@@ -387,10 +387,13 @@ func (o *ExtractOptions) Run() error {
 						alter = append(alter, newCopyFromDirectory(mapping.From))
 					default:
 						name, parent := path.Base(mapping.From), path.Dir(mapping.From)
-						if name == "." || parent == "." {
+						if name != "." && parent == "." {
+							alter = append(alter, newCopyFromRootDir(name))
```
sallyom, line 391: you _could_ get away with adding a RootDir boolean to copyFromPattern struct, I experimented with doing that - that saves needing the separate Alter, copyFromRootDir. I pushed the branch w/ that wdyt?  https://github.com/zerodayz/oc/compare/FixImageExtract-From-RootDir...sallyom:zerodayz-FixImageExtract-From-RootDir?expand=1

sallyom:
I see this PR fixes the bug, great!!  Sharing a bit to enable easy verifying:
```console
$ touch hello
$ container=$(buildah from alpine)
$ buildah copy $container hello /hello
$ buildah commit --format docker $container localhost:5000/alpine:new
$ podman push localhost:5000/alpine:new
$ mkdir ext; cd ext
$ oc image extract localhost:5000/alpine:new --file /hello
$ ls
hello
$ rm hello && oc image extract localhost:5000/alpine:new --path /hello:/tmp/ext 
$ ls
hello
```

pkg/cli/image/extract/extract.go
zerodayz: I left a code I used for debugging there, removed and I tested your code updates, it works very well. I think bug fixed :) Thank you @sallyom here is new commit for the review: https://github.com/openshift/oc/commit/03aff152605399a5ca06ca342b2d2ead730a572c

soltysh (approved):
/approve
@sallyom owns the final tag

pkg/cli/image/extract/extract.go
```diff
@@ -593,22 +596,32 @@ func (n *copyFromDirectory) Alter(hdr *tar.Header) (bool, error) {
...
 	}
-	return &copyFromPattern{Base: dir, Name: name}
+	return &copyFromPattern{Base: dir, Name: name, RootDir: false}
 }
 
 func (n *copyFromPattern) Alter(hdr *tar.Header) (bool, error) {
-	if !changeTarEntryParent(hdr, n.Base) {
-		return false, nil
-	}
 	matchName := hdr.Name
```
sallyom, line 613: this is my bad, it was an error in the suggestion I gave you, this line `matchName := hdr.Name` is in the wrong place.
You'll have to remove it er just initialize it here like `var matchName string` then, below, add the assignment.

pkg/cli/image/extract/extract.go
```diff
@@ -593,22 +596,32 @@ func (n *copyFromDirectory) Alter(hdr *tar.Header) (bool, error) {
...
-	if i := strings.Index(matchName, "/"); i != -1 {
-		matchName = matchName[:i]
+	if n.RootDir {
+		if !changeTarEntryName(hdr, n.Name) {
+			return false, nil
+		}
```
sallyom, line 617: add line  `matchName = hdr.Name` here

pkg/cli/image/extract/extract.go
```diff
@@ -593,22 +596,32 @@ func (n *copyFromDirectory) Alter(hdr *tar.Header) (bool, error) {
...
+	} else {
+		if !changeTarEntryParent(hdr, n.Base) {
+			return false, nil
+		}
```
sallyom, line 621: and here, also, add  `matchName = hdr.Name`

sallyom: (so the fn should be this - it was my bad - otherwise extract from non-root dir is broken bc matchName set b4 hdr.Name update)
```
func (n *copyFromPattern) Alter(hdr *tar.Header) (bool, error) {
    var matchName string
    if n.RootDir {
        if !changeTarEntryName(hdr, n.Name) {
            return false, nil 
        }
        matchName = hdr.Name
    } else {
        if !changeTarEntryParent(hdr, n.Base) {
            return false, nil 
        }
        matchName = hdr.Name
        if i := strings.Index(matchName, "/"); i != -1 {
            matchName = matchName[:i]
        }
    }   
    if ok, err := path.Match(n.Name, matchName); !ok || err != nil {
        klog.V(5).Infof("Excluded %s due to filter %s", hdr.Name, n.Name)
        return false, err 
    }   
    return true, nil 
}
```

sallyom (changes requested):
@zerodayz I made a note in the file to fix the Alter fn. Also, you'll need to run
`make update-gofmt` (or `gofmt -w -s pkg/cli/image/extract/extract.go`) Then all looking good, thank you!

pkg/cli/image/extract/extract.go
zerodayz: Thought that := would initialize the variable into whatever type hdr.Name is. So I didn't do it. But I can add the Following:

```
var matchName string
matchName = hdr.Name
```

pkg/cli/image/extract/extract.go
zerodayz: Ok now I see those reviews are connected as one :) sorry. I am just updating it and will send new review request :P Thanks for your patience!!!

pkg/cli/image/extract/extract.go
zerodayz: :) Thank you

soltysh:
This test should be fixed now
/test e2e-agnostic-cmd

zerodayz:
@sallyom All tests are passing, would you be able to review again? Thanks!

smarterclayton:
Changes like these need unit tests.  I’d like to see a set added that verify the streamed function receives the correct files from a set of generated fake images before we allow more changes like this to these commands.

soltysh:
@zerodayz can you prepare tests ensuring this is working as designed in a followup?

zerodayz:
@soltysh @smarterclayton Will work on unit tests today. My bad apologies.

zerodayz:
@soltysh @smarterclayton @sallyom I started putting together Unit tests for `image extract` in https://github.com/openshift/oc/pull/736

It seems that there are no unit tests at all for `image extract`. So I am putting it together now, not just the extracting image, but parsing and other functions.

PR #719, opened 2021-01-27
Bug 1879980: Fix swallowed error message in GroupDetector

zerodayz:
ldapquery.IsQueryOutOfBoundsError(err) ||
 ldapquery.IsEntryNotFoundError(err) ||
 ldapquery.IsNoSuchObjectError(err)

Would return false but error was swallowed

### DIFF
--- a/pkg/helpers/groupsync/groupdetector/groupdetector.go
+++ b/pkg/helpers/groupsync/groupdetector/groupdetector.go
@@ -20,7 +20,7 @@ type GroupBasedDetector struct {
 func (l *GroupBasedDetector) Exists(ldapGroupUID string) (bool, error) {
 	group, err := l.groupGetter.GroupEntryFor(ldapGroupUID)
 	if ldapquery.IsQueryOutOfBoundsError(err) || ldapquery.IsEntryNotFoundError(err) || ldapquery.IsNoSuchObjectError(err) {
-		return false, nil
+		return false, err
 	}
 	if err != nil {
 		return false, err


zerodayz:
Following that, with swallowed error we set exists to false, but no err as we returned nil.

```
		exists, err := s.GroupDetector.Exists(ldapGroupUID)
		if err != nil {
			fmt.Fprintf(s.Err, "Error determining LDAP group existence for group %q: %v.\n", ldapGroupUID, err)
			errors = append(errors, err)
			continue
		}
		if exists {
			continue
		}
```

This PR fixes that and we exit with error properly.

zerodayz:
Also to mention, it has a side effect of deleting the group from OpenShift, since it doesn't find the group in AD, throws an error, that error is not passed over. no error, means skip the next block, check if exists if true, if not, then delete from OpenShift.

zerodayz:
@deads2k, @mfojtik Can you guys please review, this has been laying around since January 30th and it prevents the unnecessary delete of the AD groups from OpenShift when customer use wrong casing.

pkg/helpers/groupsync/groupdetector/groupdetector.go
```diff
@@ -20,7 +20,7 @@ type GroupBasedDetector struct {
 func (l *GroupBasedDetector) Exists(ldapGroupUID string) (bool, error) {
 	group, err := l.groupGetter.GroupEntryFor(ldapGroupUID)
 	if ldapquery.IsQueryOutOfBoundsError(err) || ldapquery.IsEntryNotFoundError(err) || ldapquery.IsNoSuchObjectError(err) {
-		return false, nil
+		return false, err
```
soltysh, line 23: Does your fix from https://github.com/go-ldap/ldap/pull/307 applies, I'm worried this is breaking expected scripts where we wanted to ignore, I'd be more inclined to log with V(4), for example, rather than hard fail.

soltysh (changes requested):

pkg/helpers/groupsync/groupdetector/groupdetector.go
zerodayz: hi @soltysh this is related to https://bugzilla.redhat.com/show_bug.cgi?id=1879980

The groupSync deleted the groups because 
OU=Groups,OU=Unix != OU=groups,OU=unix

Now this means you hit the error  ldapquery.IsQueryOutOfBoundsError(err), but that err is never returned.

Not sure which case is used for ignoring, but if you by mistake set the wrong DN, which will hit `OutOfBounds Error`. GroupSync will delete all your OpenShift groups because this error is swallowed and GroupSync doesn't know you hit the wrong DN, instead it just can't see any groups, so it will sync with nothing.

The other patch for upper/lowercase in that case you are in correct DN but just wrong case.

For example:
```
usersQuery:
        baseDN: "ou=users,dc=redhat,dc=com"

usersQuery:
        baseDN: "ou=Users,dc=redhat,dc=com"
```

With https://github.com/go-ldap/ldap/pull/307, we can then easily update oc client, especially https://github.com/openshift/library-go/blob/master/pkg/security/ldapquery/query.go#L116

```
From:
		if !baseDN.AncestorOf(dn) && !baseDN.Equal(dn) {

To:
		if !baseDN.AncestorOfFold(dn) && !baseDN.EqualFold(dn) {
```

But we would also need to bump the version of go-ldap to include that fix, I am not sure how to do that part.

soltysh: > But we would also need to bump the version of go-ldap to include that fix, I am not sure how to do that part.

Hmm... it looks like we're using quite old ldap library https://github.com/openshift/oc/blob/64b59feab183d29b263d4bea6a4a438000edf0e9/go.mod#L55 whereas the current version is 3.3 which would also require updating what we're using in oc. Lemme know if you're interested.

zerodayz: I will talk to you, I would like to take this.

PR #720, opened 2021-01-27
Bug 1900989: idle: also annotate the service

frobware:
Annotate the service with idle information as we currently do for
endpoints.

We choose to additionally annotate the service because the router in
4.6 has moved to watching endpointslices only. We don't annotate
endpointslices as they are managed by the endpointslice controller and
that doesn't propgagate annotations. The router now only checks the
service to see if it has been idled.

Previous discussion for propgagating annotations in endpointslices:

    https://github.com/kubernetes/kubernetes/pull/98379

### DIFF
--- a/pkg/cli/idle/idle.go
+++ b/pkg/cli/idle/idle.go
@@ -228,6 +228,7 @@ func scanLinesFromFile(filename string) ([]string, error) {
 // with the scalable resources that it should unidle
 type idleUpdateInfo struct {
 	obj       *corev1.Endpoints
+	service   *corev1.Service
 	scaleRefs map[unidlingapi.CrossGroupObjectReference]struct{}
 }
 
@@ -324,8 +325,14 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
 			targetScaleRefs[ref] = endpointsName
 		}
 
+		svc, err := o.ClientSet.CoreV1().Services(endpoints.Namespace).Get(context.TODO(), endpoints.Name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("unable to get service %s/%s: %v", endpoints.Namespace, endpoints.Name, err)
+		}
+
 		idleInfo := idleUpdateInfo{
 			obj:       endpoints,
+			service:   svc,
 			scaleRefs: nonNamespacedScaleRefs,
 		}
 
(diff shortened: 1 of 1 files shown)


frobware:
/hold

Looking for feedback on the approach.

pkg/cli/idle/idle.go
```diff
@@ -653,42 +705,13 @@ func (o *IdleOptions) RunIdle() error {
...
-				hadError = true
-				continue
-			}
-
-			oldData, err := json.Marshal(info.obj)
-			if err != nil {
+			if err := patchObjWithIdleAnnotations(info.service, refsWithScale, nowTime); err != nil {
 				fmt.Fprintf(o.ErrOut, "error: unable to mark service %q as idled: %v", serviceName.String(), err)
 				hadError = true
 				continue
```
aojea, line 711: do we want to continue here or do we try to patch the endpoints despite the error?

frobware: Good question; I followed the existing pattern here. Even if this fails, and endpoints succeeds, it seems we can also bail (and continue) if the scalable resource updates fail in any way. There's no atomicity across the changes we're trying to make.

Miciah:
Are we going to change oc eventually not to annotate endpoints?  I wonder if it would be worth changing the visitor function to visit services in addition to endpoints—although then we'd need to change `calculateIdlableAnnotationsByService` to allow both object kinds (and be careful not to scale the scalable resources twice, once for the endpoint and once for the service).

frobware:
> Are we going to change oc eventually not to annotate endpoints?

I was trying to avoid breaking any backwards compatibility here - this PR in some form also needs to go into 4.6 too.

frobware:
/hold

Need to decide whether we want to continue annotating endpoints at all.

frobware:
/hold cancel

We will continue to annotate endpoints.

frobware:
Are test/cmd/images and test/cmd/triggers distinct failure to test/cmd/idle?

/retest

frobware:
The origin idle test are failing because it explicitly deletes the service: https://github.com/openshift/origin/blob/master/test/extended/testdata/cmd/test/cmd/idle.sh#L31

Whereas the change here expects to find the service:

```go
		svc, err := o.ClientSet.CoreV1().Services(endpoints.Namespace).Get(context.TODO(), endpoints.Name, metav1.GetOptions{})
		if err != nil {
			return fmt.Errorf("unable to get service %s/%s: %v", endpoints.Namespace, endpoints.Name, err)
		}
```

but the tests go onto then idle it: https://github.com/openshift/origin/blob/master/test/extended/testdata/cmd/test/cmd/idle.sh#L64

Test output:

```console
executing 'oc idle idling-echo' expecting success and text 'The
service will unidle DeploymentConfig \"cmd-idle/idling-echo-pd9kj\" to
2 replicas once it receives traffic': the command returned the wrong
error code; the output content test failed\nThere was no output from
the command.\nStandard error from the command:\nWARNING: idling when
network policies are in place may cause connections to bypass network
policy entirely\nerror: no valid scalable resources found to idle:
unable to get service cmd-idle/idling-echo: services \"idling-echo\"
not found\n[ERROR] hack/lib/cmd.sh:30: `return \"${return_code}\"`
exited with status 1.\n",

```

Thanks @sgreene570!

soltysh (approved):
/lgtm
/approve

soltysh:
/hold 
for https://github.com/openshift/origin/pull/25844 to merge first

frobware:
> for [openshift/origin#25844](https://github.com/openshift/origin/pull/25844) to merge first

Merged.

/hold cancel

PR #721, opened 2021-01-28
Add help text to 'oc registry login' when using --registry flag

sallyom:
This added help text is necessary because it's easy to forget you have a connected cluster (especially when kubeconfig is unset but there's the file in the default ~/.kube/config). If you forget you are currently connected, then it's easy to forget that `oc registry login localhost:5000` will write the authentication token from the connected cluster (for current context) as the  
```
    "localhost:5000": {
          "auth": "thiswouldbewrong"
```
when what you _probably_ want is `oc registry login localhost:5000 --auth-basic=user:password`

### DIFF
--- a/pkg/cli/registry/login/login.go
+++ b/pkg/cli/registry/login/login.go
@@ -55,6 +55,11 @@ var (
 		permission to view image streams you will have to pass the --registry flag with the
 		desired hostname.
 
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/approve

PR #725, opened 2021-02-01
Add missing docs generation tools

soltysh:
Looks like we missed those when moving from origin repo.

Sample output after running `make generate-docs`: https://gist.github.com/soltysh/526b047401dad67e1558068f535ef558

/assign @sallyom


soltysh:
This is not affecting product in any way but will help our docs team so I'm manually adding valid-bug label.

PR #726, opened 2021-02-01
Remove tnozicka from OWNERS and clean those files too

soltysh:
/assign @mfojtik


soltysh:
Adding valid-label tag since this is OWNERS change and not code change.

soltysh:
This is only changing OWNERS files, so overriding failed jobs.
/override ci/prow/e2e-aws-serial 

PR #728, opened 2021-02-05
Bug 1925481: when creating AuthInfo from a restclient.Config, also configure impersonation

guillaumerose:


soltysh (changes requested):
Thanks for your contribution, but impersonation is like sudo privileged operation and as such needs to be specified explicitly. I don't expect us ever copy those values into kubeconfig. 
/hold

guillaumerose:
Thanks for your answer!

PR #729, opened 2021-02-05
Bug 1904006: update and clarify --dir --from-dir flags with 'oc image'

sallyom:
/assign @smarterclayton 

This flag `--dir` ~does not make sense~ wrt `oc image extract` and is confusing because it lends to the misconception that with it one can choose where to extract contents of an image into. ~It also is not wired properly~  This PR exposes the flag `--to-dir` to add the ability for users to designate where to extract image contents to on disk. 

`--dir` is used elsewhere and in most cases is misleading. 
This PR clarifies and updates the flags `--dir, --to-dir, --from-dir`.

For example with `oc image mirror`:
`'--dir='': The directory on disk that file:// images will be copied under'`
`oc image info`:
`--dir='': The directory on disk that file:// images will be read from.`  

And here is an example of `--dir`  with both meanings of the flag:
```console
$ oc image mirror docker.io/library/busybox:latest=file://busybox:latest --dir $(pwd)/busybox
$ oc image info file://busybox:latest --dir $(pwd)/busybox
```
Since `--dir` can mean "directory that file://image is read from" and also "directory where file://image contents will be copied under" we need to add help examples and make it more clear

This PR also resolves https://bugzilla.redhat.com/show_bug.cgi?id=1906277 by adding help examples for append

### DIFF
--- a/pkg/cli/image/append/append.go
+++ b/pkg/cli/image/append/append.go
@@ -69,6 +69,20 @@ var (
 		# Add a new layer to the image
 		oc image append --from mysql:latest --to myregistry.com/myimage:latest layer.tar.gz
 
+		# Add a new layer to the image and store the result on disk
+		# This results in $(pwd)/v2/mysql/blobs,manifests
+		oc image append --from mysql:latest --to file://mysql:local layer.tar.gz
+
+		# Add a new layer to the image and store the result on disk in a designated directory
+		# This will result in $(pwd)/mysql-local/v2/mysql/blobs,manifests
+		oc image append --from mysql:latest --to file://mysql:local --dir mysql-local layer.tar.gz
+
+		# Add a new layer to an image that is stored on disk (~/mysql-local/v2/image exists)
+		oc image append --from-dir ~/mysql-local --to myregistry.com/myimage:latest layer.tar.gz
+
+		# Add a new layer to an image that was mirrored to the current directory on disk ($(pwd)/v2/image exists)
+		oc image append --from-dir v2 --to myregistry.com/myimage:latest layer.tar.gz
+
 		# Add a new layer to a multi-architecture image for an os/arch that is different from the system's os/arch
 		# Note: Wildcard filter is not supported with append. Pass a single os/arch to append.
 		oc image append --from docker.io/library/busybox:latest --filter-by-os=linux/s390x --to myregistry.com/myimage:latest layer.tar.gz
@@ -224,6 +238,7 @@ func (o *AppendImageOptions) Run() error {
 		Insecure:        o.SecurityOptions.Insecure,
 		RegistryContext: fromContext,
 	}
+
 	if len(o.FromFileDir) > 0 {
 		fromOptions.FileDir = o.FromFileDir
 	}
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -39,7 +39,9 @@ var (
 		Download an image or parts of an image to the filesystem. Allows users to access the
 		contents of images without requiring a container runtime engine running.
 
-		Pass images to extract as arguments. The --paths flag allows you to define multiple
+		Unless the --path flag is passed, image contents will be extracted into the current directory.
+
+		Pass images to extract as arguments. The --path flag allows you to define multiple
 		source to destination directory mappings. The source section may be either a file, a
 		directory (ends with a '/'), or a file pattern within a directory. The destination
 		section	is a directory to extract to. Both source and destination must be specified.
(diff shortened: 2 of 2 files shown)


sallyom:
thinking about this more:
* `oc image extract` - remove --dir flag, might add flag to allow to extract into other directory other than current (in a followup PR)
* `oc image info` - the --dir flag should be updated to be `--from-dir` to allow to read from a directory on disk (other than `v2` in current directory) 
* `oc image append` - remove the --dir flag, keep the `--from-dir` flag to allow to read from a directory on disk (other than `v2` in current directory)

pkg/cli/image/append/append.go
```diff
@@ -220,13 +231,10 @@ func (o *AppendImageOptions) Run() error {
 		return err
 	}
 	fromOptions := &imagesource.Options{
-		FileDir:         o.FileDir,
+		FileDir:         o.FromFileDir,
```
smarterclayton, line 234: I’m confused why the PR says the flags weren’t wired when they appear here?

the convention of all image commands is —dir is the default for sources or destinations of type file://.  —from-dir overrides source only and —to-dir usually overrides target (unless the command already had a flag).

Removing consistent behavior for —dir would not be desirable.  Can you describe what you mean by “not wired”?

sallyom: with `oc image extract` the `--dir` flag does not do anything, afaict. It's an imagesource.Options.FileDir but that isn't used anywhere, since when extract, the ref.Type is Registry not File

sallyom: the PR notes --dir is not wired properly for extract

sallyom: `$ oc image extract docker.io/library/busybox:latest --dir busybox` will a) fail to run unless you are in an empty directory or pass the --confirm and b) the dir value isn't used

sallyom: and with append:
`$ oc image append --from file://busybox:latest --from-dir $(pwd)/busybox --to file://busybox-new:hello` 
this works properly, but should it be possible to allow for designating the directory for the appended image? as/is, the only choice it for $(pwd)/v2 . you're right, append isn't broken, but imo flags could be more clear

sallyom: I've removed the changes to append, since nothing is broken there, but I'm adding a few examples. I'll update the referenced bugzilla to clarify

sallyom: I've removed changes to info, also, since it's intended that --dir can either mean read from or copied to.

pkg/cli/image/extract/extract.go
```diff
@@ -166,7 +171,7 @@ func NewExtract(streams genericclioptions.IOStreams) *cobra.Command {
 	flag.BoolVarP(&o.PreservePermissions, "preserve-ownership", "p", o.PreservePermissions, "Preserve the permissions of extracted files.")
 	flag.BoolVar(&o.OnlyFiles, "only-files", o.OnlyFiles, "Only extract regular files and directories from the image.")
 	flag.BoolVar(&o.AllLayers, "all-layers", o.AllLayers, "For dry-run mode, process from lowest to highest layer and don't omit duplicate files.")
-	flag.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be copied under.")
```
soltysh, line 169: Since there might be people using it, we can't just remove it, but rather deprecate this and provide information that users should use `--to-dir` instead.

soltysh (changes requested):

pkg/cli/image/append/append.go
sallyom: OH! I finally see how the --dir flag does work with extract, it's like this (from master branch of oc):
```console
$ oc image mirror docker.io/library/busybox:latest file://busybox:local --dir busybox-local
(this results in a busybox-local/v2/busybox... contents)
$ mkdir ext; cd ext; oc image extract file://busybox:local --dir ../busybox-local
(this results in extracting image contents to current directory from the busybox-local directory where the mirrored image is stored)
```
If this is a reasonable use-case we should definitely keep this. Should the `--to-dir` option be added though, since every time I use `oc image extract` I'm annoyed that I have to make a clear directory first or run w/ `--confirm` and if I accidentally run w/ `--confirm` by arrowing up I get a slew of unwanted files extracted to my current directory that I then have to clean up.

pkg/cli/image/extract/extract.go
sallyom: and since it does currently work as intended, we should probably _not_ deprecate but add the `--to-dir` and offer examples for both.

sallyom:
@smarterclayton @soltysh 
only changes in this PR:
* added examples to `append`  - covering everything I could think of and probably more than anyone will use
* expose `--to-dir` for extract (and kept as/is `--dir`) so now users can designate where to extract into on disk
* added help examples to `extract` - covering everything I could think of and probably more than anyone will use

pkg/cli/image/extract/extract.go
```diff
@@ -63,19 +65,34 @@ var (
...
+		# this results in /tmp/yum.repos.d/*.repo on local system
+		oc image extract docker.io/library/centos:7 --path /etc/yum.repos.d/*.repo:/tmp/yum.repos.d
+
+		# Extract an image stored on disk into the current directory ($(pwd)/v2/busybox/blobs,manifests exists)
+		# --confirm is required because current directory is not empty
+		oc image extract file://busybox:local --confirm
+
+		# Extract an image stored on disk in a directory other than $(pwd)/v2 into the current directory
+		# --confirm is required because current directory is not empty ($(pwd)/busybox-mirror-dir/v2/busybox exists)
+		oc image extract file://busybox:local --dir busybox-mirror-dir
```
sallyom, line 91: oops adding confirm there

soltysh (approved):
/retest

PR #730, opened 2021-02-05
Bug 1925236: RouteDescriber: Filter endpoints on route.Spec.Port

sgreene570:
This commit resolves
https://bugzilla.redhat.com/show_bug.cgi?id=1925236.

pkg/helpers/describe/describe.go: If a route specifies
a value (integer or string type) for Spec.Port, the router only
serves endpoints that correspond to that port. In this case, `oc
describe route` should output only the endpoints corresponding to
Spec.Port, as to avoid implying that the router is serving more
backend endpoints than it should be.

This change only affects `oc describe route`.


soltysh (approved):
/lgtm
/approve

PR #738, opened 2021-02-11
Add summary to oc must-gather

deads2k:
This shows the cluster version and the state of "bad" cluster operators.  

This is NOT a place to add your pet conditions.  We have three places for components to report their errors,
  1. clusterversion - shows errors applying payloads to transition versions
  2. clusteroperators - shows whether every operand is functioning properly
  3. alerts - show whether something might be at risk

If you find yourself wanting to add an additional piece of information here, what you really want to do is add it to one of those three spots.  Doing so improves the self-diagnosis capabilities of our platform and lets *every* client benefit.

```
When opening a support case, bugzilla, or issue please include the following summary data along with any other requested information.
ClusterID: e2972a8b-7ceb-4099-98f9-a0bf082deabd
ClusterVersion: Stable at "4.8.0-0.ci.test-2021-02-12-152346-ci-op-rpqp3fqv"
ClusterOperators:
	All healthy and stable

```


sttts:
Add the output in the description so we can start the bike shedding.

deads2k:
rendered

/hold cancel

pkg/cli/admin/mustgather/summary.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// PrintBasicClusterState prints a human-readable highlight of some basic information about the openshift cluster that
+// is valuable to every caller of `oc must-gather`.  Even different products.
+// This is NOT a place to add your pet conditions.  We have three places for components to report their errors,
+//  1. clusterversion - shows errors applying payloads to transition versions
+//  2. clusteroperators - shows whether every operand is functioning properly
```
sferich888, line 20: Should the summary - show what operators are installed (IE: in addition to the core payload (from olm)? 

I ask this question (not to inflate the summary) but have we / are we considering how to display 'team specific summaries' of information that would not be surfaced by these things. 

Its often useful to know whats in/running on a cluster to look for known things that conflict/collide with one another. IE: the way we have this scope (today) may be too narrow and we may want to make room for people to provide high level summary/surfacing information to this.

pkg/cli/admin/mustgather/summary.go
```diff
@@ -0,0 +1,179 @@
...
+	fmt.Fprintf(o.RawOut, "ClusterOperators:\n")
+	fmt.Fprintf(o.RawOut, humanSummaryForInterestingClusterOperators(clusterOperators)+"\n")
+
+	// TODO gather and display firing alerts
+	fmt.Fprintf(o.RawOut, "\n\n")
+}
+
+// longExistingOperators is a list of operators that should be present on every variant of every platform and have
+// existed for all supported version.  We use this to find when things are missing (like if they fail install).
+var longExistingOperators = []string{
```
sferich888, line 48: Should we be able to pull this from the payload? I don't like having this hard coded, and would prefer this to be a fall back (should we not be able to pull a list of what should have been installed with the payload).

pkg/cli/admin/mustgather/summary.go
```diff
@@ -0,0 +1,179 @@
...
+	case available && !progressing && !degraded && upgradeable:
+		return ""
+	default:
+		return fmt.Sprintf("clusteroperator/%s is in an edge case", clusterOperator.Name)
+	}
+}
+
+func humanSummaryForClusterVersion(clusterVersion *configv1.ClusterVersion) string {
+	if clusterVersion == nil {
+		return "Cluster is in a version state we don't recognize"
```
sferich888, line 146: Can we make this message more detailed? 
If we hit this this seems 'bad' and we likely need to take some type of operation to get this cluster healthy. 

@sdodson  can the install team help in this space to outline what this error means and why one might hit it? As well as what one should look at in the must-gather to debug/diagnose this. 

IE: General catchall's like this warrant a 'this is known bad' situation to be in, and should come with some 'known next steps', like a link to a KCS?

pkg/cli/admin/mustgather/summary.go
```diff
@@ -0,0 +1,179 @@
...
+	case isInstalling:
+		return fmt.Sprintf("Installing %q for %v: %v",
+			clusterVersion.Status.Desired.Version, lastChangeHumanDuration, progressingConditionMessage)
+	case isUpdating:
+		return fmt.Sprintf("Updating to %q from %q for %v: %v",
+			clusterVersion.Status.Desired.Version, clusterVersion.Status.History[1].Version, lastChangeHumanDuration, progressingConditionMessage)
+	case isStable:
+		return fmt.Sprintf("Stable at %q", clusterVersion.Status.History[0].Version)
+	default:
+		return fmt.Sprintf("Unknown state")
```
sferich888, line 177: I make a comment above about 'default' states, and providing 'guidance on next steps', can we also address that here.

soltysh (approved):
/lgtm
/hold

soltysh:
Drop the hold at your own discretion.

sferich888 (changes requested):
In general I like where this is headed, however I would like to see (either in this PR or a future set of scoped work), where its possible to get this summary (on demand, from a collected archive). 

Relying on this to land in a 'log' makes finding this (in a mountain of data) difficult/cumbersome/awkward. I would prefer for this summary to land in its own file in the archive (so it can be extracted), or be 'collectible'/'extractable' by running an `oc` command on an existing archive. 

If this is in a separate file (as part of the archive) we can then use the information (more reliably) as part of automation (with cases, or analytic), for a variety of tools/situations.

sferich888:
This would close: https://github.com/openshift/must-gather/issues/186 if completed.

pkg/cli/admin/mustgather/summary.go
sferich888: https://github.com/openshift/must-gather/issues/186#issuecomment-705789781 support this ask.

deads2k:
> then use the information

all the information is structured in the archive.



> as part of automation

should use the json/yaml collected



> for a variety of tools

should use structured data, this is for humans to quickly assign.

pkg/cli/admin/mustgather/summary.go
deads2k: > Can we make this message more detailed?

No.  We don't know what this is.  You have the entire clusterversion object for a human/machine to look at.

pkg/cli/admin/mustgather/summary.go
deads2k: > I make a comment above about 'default' states, and providing 'guidance on next steps', can we also address that here.

I don't know why it would be in a state like this.  But someone else can work it out and potentially add something if it is common enough.

sdodson:
@sferich888 I believe all of this is extractable today, the point is that an interactive user running this command is guided to provide things that we'd want to know before even downloading the archive, this is all about triaging bugs without downloading the archive.

/approve
This looks good, thanks.

PR #741, opened 2021-02-16
Bug 1926522: oc adm catalog mirror tmp directory cleanup

zerodayz:


pkg/cli/admin/catalog/mirror.go
```diff
@@ -386,6 +389,15 @@ func (o *MirrorCatalogOptions) Run() error {
 	return WriteManifests(o.IOStreams.Out, o.SourceRef, o.DestRef, o.ManifestDir, o.IcspScope, mapping)
 }
 
+func (o *MirrorCatalogOptions) PostRun() error {
+	dir := strings.Split(o.DatabasePath, ":")
+	if err := os.RemoveAll(dir[1]); err != nil {
```
ecordell, line 394: I think we only want to do this if `--path` is not set

pkg/cli/admin/catalog/mirror.go
```diff
@@ -386,6 +389,15 @@ func (o *MirrorCatalogOptions) Run() error {
 	return WriteManifests(o.IOStreams.Out, o.SourceRef, o.DestRef, o.ManifestDir, o.IcspScope, mapping)
 }
 
+func (o *MirrorCatalogOptions) PostRun() error {
+	dir := strings.Split(o.DatabasePath, ":")
```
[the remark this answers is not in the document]
zerodayz: I added new o.TempDir, which I set to true if o.DatabasePath is "" meaning --path unset. If o.DatabasePath is set using --path, then o.TempDir remains false.
At the end based if the o.TempDir is set or not we decide to delete the directory.

pkg/cli/admin/catalog/mirror.go
zerodayz: I added new o.TempDir, which I set to true if o.DatabasePath is "" meaning --path unset. If o.DatabasePath is set using --path, then o.TempDir remains false.
At the end based if the o.TempDir is set or not we decide to delete the directory.

zerodayz:
@ecordell 

I added new `o.TempDir`, which I set to `true` if `o.DatabasePath` is `""` meaning `--path` unset. If `o.DatabasePath` is set using `--path`, then `o.TempDir` remains `false`.
At the end based if the `o.TempDir` is set or not we decide to delete the directory.

zerodayz:
We also need to update docs that when following https://docs.openshift.com/container-platform/4.6/operators/admin/olm-restricted-networks.html#olm-mirror-catalog_olm-restricted-networks

> 
> On your workstation with unrestricted network access, run the following command:
> 
> ```
> oc adm catalog mirror \
>     <index_image> \
>     <mirror_registry>:<port> \
>     [-a ${REG_CREDS}] \
>     [--insecure] \
>     --filter-by-os='.*' \
>     [--manifests-only] 
>  ```
> 
> 
> For example, to retrieve a list of images that are similar to the string jaeger:
> ```
> $ echo "select * from related_image \
>     where operatorbundle_name like '%jaeger%';" \
>     | sqlite3 -line /tmp/153048078/index.db 
> ``` 

That first command needs to have added option `--path` so we skip the deletion of the temp directory.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -386,6 +391,20 @@ func (o *MirrorCatalogOptions) Run() error {
 	return WriteManifests(o.IOStreams.Out, o.SourceRef, o.DestRef, o.ManifestDir, o.IcspScope, mapping)
 }
 
+func (o *MirrorCatalogOptions) PostRun() error {
+	// If we have NOT set --path, the TempDir is set to true and we will delete
+	// Temporary folder.
+	if o.TempDir {
+		dir := strings.Split(o.DatabasePath, ":")
+		if err := os.RemoveAll(dir[1]); err != nil {
```
timflannagan, line 399: Do we need to verify that this index exists by checking the length of the `dir` split string?

zerodayz: Right, it should but I guess nothing wrong with adding lenght check so we dont end up in  index out of bound error.

timflannagan: @zerodayz It looks like this PR needs a rebase now - I think these changes still look reasonable to me. Small nit/nothing blocking: might be worth creating a variable that's initialized from the `dir[1]` as we use that in multiple places here.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -386,6 +391,20 @@ func (o *MirrorCatalogOptions) Run() error {
 	return WriteManifests(o.IOStreams.Out, o.SourceRef, o.DestRef, o.ManifestDir, o.IcspScope, mapping)
 }
 
+func (o *MirrorCatalogOptions) PostRun() error {
+	// If we have NOT set --path, the TempDir is set to true and we will delete
+	// Temporary folder.
+	if o.TempDir {
```
timflannagan, line 397: small nit/nothing blocking: can we flip this condition to exit early:

```go
if !o.TempDir { 
    return nil
}
```

tylerslaton:
Superseded by [PR 935](https://github.com/openshift/oc/pull/935) in order to rebase.

PR #742, opened 2021-02-16
Bug 1928850: update new-app help to avoid image references without registry/repository (docker.io throttling related)

gabemontero:


gabemontero:
unrelated sig-api-machinery flake/possible permfail in e2e-metal-ipi-ovn-ipv6 ... everything else passed, including anything quasi-related to oc new-app

/skip

gabemontero:
/assign @adambkaplan 

all green tests

pkg/cli/newapp/newapp.go
```diff
@@ -134,9 +134,16 @@ To list all local templates and image streams, use:
 To search templates, image streams, and Docker images that match the arguments provided, use:
 
   oc new-app -S php
-  oc new-app -S --template=ruby
+  oc new-app -S --template=rails
   oc new-app -S --image-stream=mysql
+  # REMINDER, for docker image references, if a registry or repository are not specified, 'docker.io' and 'library' are potential defaults for the various image registry clients
   oc new-app -S --docker-image=python
+
```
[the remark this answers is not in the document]
gabemontero: > This is for oc new-app's search capability. 

Yes that is correct.

> Will this only search Docker Hub images? Are quay.io images also included in the search results?

It is a function of the imagestreamimport API actually, that is what is used from oc new-app.  It will search other registries but as it turns out with docker.io having precedence with the core docker clients used, the only image with name of "python" and nothing else is "docker.io/library/python" so that is the hit that is obtained.  There is not quay.io/library/python for example.

In addition to the warning I could have also changed the `oc new-app -S --docker-image=python` to `oc new-app -S --docker-image=<whatever python images exists on quay.io>` but that looked clunky to me.

But if you want to go that approach @adambkaplan give an explicit go ahead and we'll go from there.

pkg/cli/newapp/newapp.go
```diff
@@ -134,9 +134,16 @@ To list all local templates and image streams, use:
 To search templates, image streams, and Docker images that match the arguments provided, use:
 
   oc new-app -S php
-  oc new-app -S --template=ruby
+  oc new-app -S --template=rails
   oc new-app -S --image-stream=mysql
+  # REMINDER, for docker image references, if a registry or repository are not specified, 'docker.io' and 'library' are potential defaults for the various image registry clients
```
adambkaplan, line 139: I'd rather not have this warning - the root pain comes from the fact that we have a new thing that customers are using - the OpenShift Dev Sandbox. New users - who may be new to containers and image registries - are hitting this pain because our default example (that they are probably copy-pasting) hits docker.io. New users a) won't read this, and b) are probably not aware of the rate limiting, and c) are certainly not aware that the sandbox is on an OSD cluster, and as a result they are almost guaranteed to hit the rate limit.

So let's save them pain and have our example use a registry that doesn't impose such anonymous pull rate limits, like quay.io. @hhorak are there UBI-8 or Fedora-based images for python available on quay? Or are they available on a similar public registry (ex - registry.fedoraproject.org)

adambkaplan (changes requested):

pkg/cli/newapp/newapp.go
gabemontero: > I'd rather not have this warning - the root pain comes from the fact that we have a new thing that customers are using - the OpenShift Dev Sandbox. New users - who may be new to containers and image registries - are hitting this pain because our default example (that they are probably copy-pasting) hits docker.io. New users a) won't read this, and b) are probably not aware of the rate limiting, and c) are certainly not aware that the sandbox is on an OSD cluster, and as a result they are almost guaranteed to hit the rate limit.
> 
> So let's save them pain and have our example use a registry that doesn't impose such anonymous pull rate limits, like quay.io. @hhorak are there UBI-8 or Fedora-based images for python available on quay? Or are they available on a similar public registry (ex - registry.fedoraproject.org)

there was a discussion on this back in January @adambkaplan ... I see you were not on the email thread.  I've forwarded it to you.

I'll peruse the precise inventory wrt python.

gabemontero: I think the reader's digest from that thread @adambkaplan is:
- ubi8 related content will not sit in quay.io
- so we should go with the OKD version:  i.e. https://github.com/openshift/cluster-samples-operator/blob/master/assets/operator/okd-x86_64/python/imagestreams/python-centos.json#L51

gabemontero:
warning removed @adambkaplan 

use of precise RH/SCL python image vs. generic search string which will default to docker.io/library

pkg/cli/newapp/newapp.go
hhorak: >  @hhorak are there UBI-8 or Fedora-based images for python available on quay?

All fedora images for python are in registry.fedoraproject.org only, UBI8 images are in the unauthenticated registry.access.redhat.com. We only migrated the ubi7 images to the quay.io. HTH.

pkg/cli/newapp/newapp.go
```diff
@@ -264,7 +270,7 @@ func NewCmdNewApplication(f kcmdutil.Factory, streams genericclioptions.IOStream
 	cmd.Flags().StringSliceVar(&o.Config.SourceRepositories, "code", o.Config.SourceRepositories, "Source code to use to build this application.")
 	cmd.Flags().StringVar(&o.Config.ContextDir, "context-dir", o.Config.ContextDir, "Context directory to be used for the build.")
 	cmd.Flags().StringSliceVarP(&o.Config.ImageStreams, "image-stream", "i", o.Config.ImageStreams, "Name of an image stream to use in the app.")
-	cmd.Flags().StringSliceVar(&o.Config.DockerImages, "docker-image", o.Config.DockerImages, "Name of a Docker image to include in the app.")
+	cmd.Flags().StringSliceVar(&o.Config.DockerImages, "docker-image", o.Config.DockerImages, "Name of a Docker image to include in the app.  REMEMBER:  not specifying a registry or repository means defaults in place for client image pulls are employed.")
```
adambkaplan, line 273: nit - perhaps change to `Note:` instead of `REMEMBER`
```suggestion
	cmd.Flags().StringSliceVar(&o.Config.DockerImages, "docker-image", o.Config.DockerImages, "Name of a Docker image to include in the app. Note: not specifying a registry or repository means defaults in place for client image pulls are employed.")
```

adambkaplan (commented):
Final nit, otherwise I think we are good using the registry.access.redhat.com image.

gabemontero:
REMEMBER -> Note pushed and commits squashed @adambkaplan thanks

pkg/cli/newapp/newapp.go
```diff
@@ -134,9 +134,15 @@ To list all local templates and image streams, use:
...
+  oc new-app -S --template=rails
   oc new-app -S --image-stream=mysql
-  oc new-app -S --docker-image=python
+  oc new-app -S --docker-image=registry.access.redhat.com/ubi8/python-38
+
+For details on how to use the results from those searches, use:
+
+  oc help new-app
+
+To see how to provide images, image streams, templates, or source code locations as inputs into 'oc new-app'.
```
adambkaplan, line 145: final grammar nit  - final line here isn't a complete sentence. Perhaps we can get rid of it?

gabemontero: yeah it was suppose to read as a follow on clause to the prior sentence, but splitting up like that does seem cumbersome in hindsight.

I'm actually going to push the change where I combine a portion of this with the prior sentence.  See what you think of that.

gabemontero:
update for your latest comment pushed @adambkaplan PTAL

gabemontero:
unrelated cloud/k8s flakes

/retest

gabemontero:
the e2e-metal-ipi-ovn-ipv6 error that I've hit a few times today also showed up on sippy we a bunch of hits

as a result I opened https://bugzilla.redhat.com/show_bug.cgi?id=1931668

/retest

gabemontero:
> the e2e-metal-ipi-ovn-ipv6 error that I've hit a few times today also showed up on sippy we a bunch of hits
> 
> as a result I opened https://bugzilla.redhat.com/show_bug.cgi?id=1931668

turns out @adambkaplan opened a bug for this as well in his role as build watcher yesterday

I closed my bz as a dupe of his.

the last metal-ipi-ovn-ipv6 run hit a different failure in sig-api-machinery:

` [sig-api-machinery] CustomResourcePublishOpenAPI [Privileged:ClusterAdmin] works for CRD preserving unknown fields in an embedded object [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]`

still trying to get a clean run of that without any of these flakes

/retest
> 
> /retest

adambkaplan (approved):
/lgtm

PR #743, opened 2021-02-18
Add conntrack-tools to debug image

michaelgugino:
Very useful network debugging tool.


tssurya:
> Very useful network debugging tool.

just FYI... networking specific tools are being added into https://github.com/openshift/network-tools/blob/master/Dockerfile. We are creating a new debug network image which is based on top of tools. `conntrack-tools` is present in this image.

the integration of that image with oc client is WIP.

PR #745, opened 2021-02-19
Remove old dependencies on dockerv1client and replace with newer equivalents

smarterclayton:
* Use ImageRetriever for newapp and registry lookup
* Update to a newer fsouza/go-dockerclient, which should allow new-app flows with Docker to work on recent versions
* Turn etcd ttl command into a stub since it is deprecated and no longer in use, which removes etcd3 client
* Purge the vendor tree


go.mod
```diff
@@ -21,22 +21,24 @@ require (
 	github.com/containers/storage v0.0.0-20190726081758-912de200380a // indirect
 	github.com/davecgh/go-spew v1.1.1
 	github.com/docker/distribution v2.7.1+incompatible
-	github.com/docker/docker v17.12.0-ce-rc1.0.20200916142827-bd33bbf0497b+incompatible
+	github.com/docker/docker v20.10.3+incompatible
```
soltysh, line 24: We have a `replace` for docker/docker dependency that can be now dropped too, most likely. Wanna do it in this PR of followup?

soltysh (approved):
/lgtm
/hold
for that docker/docker question
/retest

go.mod
smarterclayton: Will do it here.

soltysh (approved):
/hold cancel
/lgtm
/retest

PR #747, opened 2021-02-23
Add json.Valid check before trying to read docker config

zerodayz:
Add json.Valid check before trying to read docker config

At the moment, if we fail during unmarshaling JSON, it doesn't
give much of hint where the failure occurred.

For example error message when we provide invalid JSON:

```
    unable to load --registry-config: error occurred while trying to unmarshal json
```

Since we have filePath and json package, before actually unmarshaling
the json, we can verify it's contents.

This patch adds a little bit more friendlier error message, before
even failing on unmarshaling

```
    unable to load --registry-config: json format is invalid: file.json
```

### DIFF
--- a/pkg/helpers/image/credentialprovider/config.go
+++ b/pkg/helpers/image/credentialprovider/config.go
@@ -147,7 +147,12 @@ func ReadSpecificDockerConfigJSONFile(filePath string) (cfg DockerConfig, err er
 	if contents, err = ioutil.ReadFile(filePath); err != nil {
 		return nil, err
 	}
-	return readDockerConfigJSONFileFromBytes(contents)
+
+	cfg, err = readDockerConfigJSONFileFromBytes(contents)
+	if err != nil {
+		return nil, fmt.Errorf("invalid config: %q: %w", filePath, err)
+	}
+	return cfg, nil
 }
 
 // ReadDockerConfigFile read a docker config file from default path


bshephar:
Nice. Looks good to me. Anything that makes it a little more intuitive as to where the error might be.

zerodayz:
Hello @soltysh and @ecordell can you please review once available?

Thank you

pkg/helpers/image/credentialprovider/config.go
```diff
@@ -147,6 +147,12 @@ func ReadSpecificDockerConfigJSONFile(filePath string) (cfg DockerConfig, err er
 	if contents, err = ioutil.ReadFile(filePath); err != nil {
 		return nil, err
 	}
+
+	ok := json.Valid(contents)
```
soltysh, line 151: Why the extra valid call, if `json.Unmarshal` invoked in `readDockerConfigJSONFileFromBytes` already does it? Why not wrapping the error returned from that method with information about file name instead?

soltysh (changes requested):

pkg/helpers/image/credentialprovider/config.go
zerodayz: That's how I did it initially, but I wanted to avoid modifying function `readDockerConfigJSONFileFromBytes` the function name says `read docker config JSON file from bytes`  so I thought that logically it's better to leave validation out of that function in `ReadSpecificDockerConfigJSONFile`.

soltysh: Right, that's why I'm suggesting to wrap the error returned from `readDockerConfigJSONFileFromBytes` with the file path, only. Iow:
```golang
cfg, err := readDockerConfigJSONFileFromBytes(contents)
if err !=  nil {
    return nil, fmt.Errorf("invalid config %q: %w", filePath, err)
}
return cfg, nil
```
wdyt?

zerodayz: Unmarshaling could fail if there is either typo in the config or the JSON is not valid. I was checking explicitly for the Validity of the JSON and not the config typo. This will cover both. I guess it's enough to say config XYZ is wrong. At least you get a path, so you can investigate more.

zerodayz: Sorry I was wrong. Unmarshal will return only if its not valid JSON.

zerodayz: Agree with your approach, at least expose the file path to the config itself, so sysadmin can investigate the problem.

soltysh (approved):
/lgtm
/approve

soltysh:
/hold
@zerodayz you need to run `make update-gofmt`, make sure to use golang 1.6

zerodayz:
Hi @soltysh will do it later today. Thanks!

zerodayz:
Done @soltysh 👍

soltysh (approved):
/hold cancel
/lgtm

soltysh:
the failure is unrelated

PR #748, opened 2021-02-23
Bug 1895053: bump(github.com/openshift/api):

adambkaplan:
Bump openshift/api to add new MountTrustedCA field to Builds and
BuildConfigs.


soltysh:
> Bump openshift/api to add new MountTrustedCA field to Builds and
> BuildConfigs.

Is there additional change to oc code missing in this PR? If not oc should work w/o this bump right away.

adambkaplan:
/close

We aren't introducing any new capability in oc, like setting this field via `new-app` or `new-build`. Perhaps in the future, but for now we can leave oc as it is.

PR #752, opened 2021-03-01
indicate how many bytes the hidden secret key was

deads2k:
still hides the secret content, but let's you know how many bytes it was.


soltysh (approved):
/lgtm
/retest

PR #755, opened 2021-03-03
Bug 1934522: Use dedicated SA when creating must-gather pod not to rely on KCM's default

soltysh:
/hold
for testing
/assign @p0lyn0mial

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -297,9 +297,16 @@ func (o *MustGatherOptions) Run() error {
 			o.PrinterDeleted.PrintObj(ns, o.LogOut)
 		}()
 	}
-
+	// ... service account ...
+	//use custom SA to not to rely on kube-controller-manager's default
+	serviceAccount, err := o.Client.CoreV1().ServiceAccounts(ns.Name).Create(context.TODO(),
+		&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{GenerateName: "openshift-must-gather-"}},
+		metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
 	// ... cluster role binding ...
-	clusterRoleBinding, err := o.Client.RbacV1().ClusterRoleBindings().Create(context.TODO(), o.newClusterRoleBinding(ns.Name), metav1.CreateOptions{})
+	clusterRoleBinding, err := o.Client.RbacV1().ClusterRoleBindings().Create(context.TODO(), o.newClusterRoleBinding(ns.Name, serviceAccount.Name), metav1.CreateOptions{})
 	if err != nil {
 		return err
 	}
@@ -323,7 +330,7 @@ func (o *MustGatherOptions) Run() error {
 			return err
 		}
 
-		pod, err := o.Client.CoreV1().Pods(ns.Name).Create(context.TODO(), o.newPod(o.NodeName, image), metav1.CreateOptions{})
+		pod, err := o.Client.CoreV1().Pods(ns.Name).Create(context.TODO(), o.newPod(o.NodeName, image, serviceAccount.Name), metav1.CreateOptions{})
 		if err != nil {
 			return err
 		}
@@ -562,7 +569,7 @@ func (o *MustGatherOptions) waitForGatherContainerRunning(pod *corev1.Pod) error
 	})
 }
 
-func (o *MustGatherOptions) newClusterRoleBinding(ns string) *rbacv1.ClusterRoleBinding {
+func (o *MustGatherOptions) newClusterRoleBinding(ns, sa string) *rbacv1.ClusterRoleBinding {
 	return &rbacv1.ClusterRoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: "must-gather-",
@@ -578,7 +585,7 @@ func (o *MustGatherOptions) newClusterRoleBinding(ns string) *rbacv1.ClusterRole
 		Subjects: []rbacv1.Subject{
 			{
 				Kind:      "ServiceAccount",
-				Name:      "default",
+				Name:      sa,
 				Namespace: ns,
 			},
 		},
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -289,9 +289,16 @@ func (o *MustGatherOptions) Run() error {
 			o.PrinterDeleted.PrintObj(ns, o.LogOut)
 		}()
 	}
-
+	// ... service account ...
+	//use custom SA to not to rely on kube-controller-manager's default
+	serviceAccount, err := o.Client.CoreV1().ServiceAccounts(ns.Name).Create(context.TODO(),
```
p0lyn0mial, line 294: an `sa` needs a token that is generated by KCM, right ?

soltysh: It's possible, I need to verify. I'll get to it :wink:

zerodayz: @soltysh @p0lyn0mial  I noticed one issue with the patch and that is the new `serviceAccountName` was not passed to a `newPod` and it failed with unable to find SA `default`.

I changed the patch a bit to pass that to `newPod` to test whether it would work or not:

```go
pod, err := o.Client.CoreV1().Pods(ns.Name).Create(context.TODO(), o.newPod(o.NodeName, image, serviceAccount.Name), metav1.CreateOptions{})
...
func (o *MustGatherOptions) newPod(node, image, sa string) *corev1.Pod {
...
Spec: corev1.PodSpec{
	NodeName:      node,
	RestartPolicy: corev1.RestartPolicyNever,
	ServiceAccountName: sa,
	Volumes: []corev1.Volume{
```

But even after that change, I can see that SA gets token generated by KCM, so no this won't work. We need to look for other way.

zerodayz: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#token-controller

TokenController runs as part of kube-controller-manager. It acts asynchronously. It:

- watches ServiceAccount creation and creates a corresponding ServiceAccount token Secret to allow API access.

zerodayz: I did similar thing as you did @soltysh + extra steps to create secret/patch the SA, at the moment it works in a way that i gets the token, but I am still trying to figure out why I am getting error:

```
[must-gather-hsss6] POD config.go:489] Expected to load root CA config from /var/run/secrets/kubernetes.io/serviceaccount/ca.crt, but got err: open /var/run/secrets/kubernetes.io/serviceaccount/ca.crt: no such file or directory
```

```go
	// ... service account ...
	//use custom SA to not to rely on kube-controller-manager's default
	serviceAccount, err := o.Client.CoreV1().ServiceAccounts(ns.Name).Create(context.TODO(),
		&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{GenerateName: "openshift-must-gather-"}},
		metav1.CreateOptions{})
	if err != nil {
		return err
	}

	// Build the secret
	saSecret := &corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			GenerateName:       serviceAccount.Name + "-token-",
			Namespace: serviceAccount.Namespace,
			Annotations: map[string]string{
				corev1.ServiceAccountNameKey: serviceAccount.Name,
				corev1.ServiceAccountUIDKey:  string(serviceAccount.UID),
			},
		},
		Type: corev1.SecretTypeServiceAccountToken,
		Data: map[string][]byte{},
	}

	tr := &authenticationv1.TokenRequest{
		Spec: authenticationv1.TokenRequestSpec{
		},
	}
	token, err := o.Client.CoreV1().ServiceAccounts(ns.Name).CreateToken(context.TODO(), serviceAccount.Name, tr, metav1.CreateOptions{})
	if err != nil {
		return err
	}
	saSecret.Data[corev1.ServiceAccountTokenKey] = []byte(token.Status.Token)
	saSecret.Data[corev1.ServiceAccountNamespaceKey] = []byte(ns.Name)

	// Save the secret
	secret, err := o.Client.CoreV1().Secrets(serviceAccount.Namespace).Create(context.TODO(), saSecret, metav1.CreateOptions{})
	if err != nil {
		return err
	}

	boolTrue := true
	serviceAccountPatchData, err := json.Marshal(corev1.ServiceAccount{
		Secrets: []corev1.ObjectReference{
			{
				Kind: "Secrets",
				Name: secret.Name,
			},
		},
		AutomountServiceAccountToken: &boolTrue,
	})

	// Patch the service account
	if _, err := o.Client.CoreV1().ServiceAccounts(ns.Name).Patch(context.TODO(), serviceAccount.Name, types.StrategicMergePatchType, serviceAccountPatchData, metav1.PatchOptions{}); err != nil {
		return err
	}

	// ... cluster role binding ...
	clusterRoleBinding, err := o.Client.RbacV1().ClusterRoleBindings().Create(context.TODO(), o.newClusterRoleBinding(ns.Name, serviceAccount.Name), metav1.CreateOptions{})
	if err != nil {
		return err
	}
```

zerodayz: Oh I see it's all of it:
```bash
sh-4.4# ls /var/run/secrets/kubernetes.io/serviceaccount/
ca.crt	namespace  service-ca.crt  token
```
So I need more than a token. Let me test it out. I guess for the other resources I could just get those and patch directly.

soltysh: You are both right, iow:
1. I missing SA wiring 
2. we need a token.

Ad. 1. it's a simple fix, I've already did that.
Ad. 2. both SA and Pod has an option not to automount token `AutomountServiceAccountToken` is the field, but for that case I still need a token. I'm considering copying one from default namespace for this case. 

I think we shouldn't do it by default, though. Only if we fail running a pod normally I'd go this route. I'll work on it more next week.

soltysh:
/reopen
I'm planning to still pick this up.

PR #746, opened 2021-02-22
Bug 1880865: Avoid TLS cert checking when login with --insecure-skip-tls-verify=true

damemi:
When using oc login with option --insecure-skip-tls-verify=true,
server certificate checkings are performed on an initial HEAD request
sent to the cluster and may lead to unexpected errors (e.g tls timeout)
preventing the user from logging eventhough he/she had intentionally
requested to skip them.

(picked from https://github.com/openshift/oc/pull/514)

PR #757, opened 2021-03-05
Bug 1935102: Revert Avoid TLS cert checking when login with --insecure-skip-tls-verify=true

damemi:
Reverts regression introduced in https://github.com/openshift/oc/pull/746 until we can determine a valid approach to solve that problem


soltysh (approved):
/lgtm
/approve

PR #759, opened 2021-03-09
Bug 1936904: Insert apiVersion and Kind into `oc adm groups sync` list output

damemi:
This attempts to get group `kind`/`apiVersion` to print in the list when using `oc adm groups sync`

Output with these changes (using sample LDAP at https://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/):
```
$ ./oc adm groups sync --sync-config=ldap-config.yaml -o yaml
apiVersion: v1
items:
- apiVersion: user.openshift.io/v1
  kind: group
  metadata:
    annotations:
      openshift.io/ldap.sync-time: 2021-03-09T11:34:59-0500
      openshift.io/ldap.uid: ou=mathematicians,dc=example,dc=com
      openshift.io/ldap.url: ldap.forumsys.com:389
    creationTimestamp: null
    labels:
      openshift.io/ldap.host: ldap.forumsys.com
    name: Mathematicians
  users: null
kind: List
metadata: {}
```

### DIFF
--- a/pkg/cli/admin/groups/sync/sync.go
+++ b/pkg/cli/admin/groups/sync/sync.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	userv1 "github.com/openshift/api/user/v1"
 	"io/ioutil"
 	"strings"
 
@@ -11,8 +12,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	kerrs "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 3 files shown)


damemi (commented):
/cc @deads2k 
I see you worked with the groups sync command before, I'm not familiar with it so just playing around here.

Is there a reason this was originally done using `Unstructured` rather than just the `GroupList` type? And, will switching to `GroupList` accurately print the group kind/version for each item in the list? (see the linked BZ for examples)

pkg/cli/admin/groups/sync/sync.go
```diff
@@ -417,6 +417,8 @@ func (o *SyncOptions) Run() error {
 			},
 		}
 		for _, item := range openshiftGroups {
+			item.APIVersion = "user.openshift.io/v1"
```
soltysh, line 420: The fix is reasonable, but how about making this where the `Group` object is actually created, iow.: https://github.com/openshift/oc/blob/f749845d71de73620f5e66553e0dd5abeb7efede/pkg/helpers/groupsync/groupsyncer.go#L151

damemi: that's actually what I originally did in this PR, but it still didn't insert these fields automatically.

I can switch back to creating the `Group` object, but we still need to manually add these fields (or figure out why they're not showing up)

damemi:
ping @soltysh any feedback on this? If it looks good I can squash, merge and move the BZ to qa

soltysh (approved):
/lgtm
/approve

PR #763, opened 2021-03-11
stop trying to gather metrics and other endpoints directly from pods

deads2k:
We now gather alerts from the monitoring stack and this takes a long
time run.  Also, when we have problems we normally also have networking
issues so these rarely work when there is something actually broken.


soltysh (approved):
/lgtm
/retest

PR #765, opened 2021-03-15
Fix unit test to use new IS

soltysh:
In https://github.com/openshift/ruby-hello-world/pull/121 IS-es were updated to `ruby-27-centos` which broke all units in oc
/cc @vrutkovs @bparees 
/assign @deads2k


vrutkovs:
looks fine to me

soltysh:
/test e2e-aws-upgrade
/override ci/prow/e2e-aws-builds 
since that's fixed in https://github.com/openshift/origin/pull/25949 and is not blocking

PR #767, opened 2021-03-17
Bug 1995291: Remove docker adjective whenever possible.

rhatdan:


rhatdan:
@adambkaplan @bparees PTAL
Is anyone working on this?

pkg/cli/newapp/newapp.go
```diff
@@ -269,7 +269,8 @@ func NewCmdNewApplication(f kcmdutil.Factory, streams genericclioptions.IOStream
 	cmd.Flags().StringSliceVar(&o.Config.SourceRepositories, "code", o.Config.SourceRepositories, "Source code to use to build this application.")
 	cmd.Flags().StringVar(&o.Config.ContextDir, "context-dir", o.Config.ContextDir, "Context directory to be used for the build.")
 	cmd.Flags().StringSliceVarP(&o.Config.ImageStreams, "image-stream", "i", o.Config.ImageStreams, "Name of an image stream to use in the app.")
-	cmd.Flags().StringSliceVar(&o.Config.DockerImages, "docker-image", o.Config.DockerImages, "Name of a Docker image to include in the app.  Note:  not specifying a registry or repository means defaults in place for client image pulls are employed.")
+	cmd.Flags().StringSliceVar(&o.Config.DockerImages, "image", o.Config.DockerImages, "Name of a container image to include in the app.  Note:  not specifying a registry or repository means defaults in place for client image pulls are employed.")
```
bparees, line 272: @rhatdan we can't drop existing command line flags.  If you want to mark the existing flag deprecated and introduce its replacement, we can do that.

bparees: ah, I see you aliased it below

bparees:
```
the build bin failed after 1m42s with reason DockerBuildFailed: Dockerfile build strategy has failed.

pkg/cli/newapp/newapp.go:273:31: undefined: newapp.AliasFlags
pkg/cli/newapp/newapp.go:1223:20: undefined: pflag
pkg/cli/newapp/newapp.go:1228:9: undefined: pflag
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk:16: build] Error 2
error: build error: error building at STEP "RUN /bin/bash ...ined-variables": error while running runtime: exit status 2
```

bparees:
seems to be missing an import

bparees:
@rhatdan fix the import and i think we can merge this

rhatdan:
/retest
@bparees Is there still something wrong with the `import`, not really sure what you mean by this?

bparees:
the pflag import was missing, you appear to have fixed it.  should be good to go now.

pkg/cli/newapp/newapp.go
soltysh: Aliasing isn't sufficient since it hides the problem. We want to force people to move off of using these flags. That's why we should deprecate the old ones and introduce new ones instead. In 3 releases or a year we'll be able to drop the old ones. Make sure to add a note in the release notes issue in https://github.com/openshift/openshift-docs/issues/33497 so that this is being tracked. 

From my conversation with @bparees since it might not be clear, marking a flag deprecated will effectively make it disappear from everywhere (docs, completions, et.) so you either need to know the name or have a pre-existing flag. The benefit of marking it deprecated is that you're getting a warning, to switch to new flag.

pkg/cli/newbuild/newbuild.go
```diff
@@ -124,13 +124,14 @@ func NewCmdNewBuild(f kcmdutil.Factory, streams genericclioptions.IOStreams) *co
 
 	cmd.Flags().StringSliceVar(&o.Config.SourceRepositories, "code", o.Config.SourceRepositories, "Source code in the build configuration.")
 	cmd.Flags().StringSliceVarP(&o.Config.ImageStreams, "image-stream", "i", o.Config.ImageStreams, "Name of an image stream to to use as a builder.")
-	cmd.Flags().StringSliceVar(&o.Config.DockerImages, "docker-image", o.Config.DockerImages, "Name of a Docker image to use as a builder.")
+	cmd.Flags().StringSliceVar(&o.Config.DockerImages, "image", o.Config.DockerImages, "Name of a container image to use as a builder.")
```
soltysh, line 127: Ditto.

soltysh (commented):
One nit to switch from aliasing to deprecating the flag and this is good to go. If you want to land it for 4.9 you'll need to open a BZ tracking this one.

rhatdan:
@bparees @soltysh Are these failures flakes or actual errors?

bparees:
the unit test job only appears to be failing on your PR:
https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-oc-master-unit

though i'm struggling to understand why just yet.

bparees:
@rhatdan ah, you changed some of the outputs, so you need to update this to match:
https://github.com/openshift/oc/blob/681248e6d3b6c3d965ea04f75b6ce03ca19671d0/pkg/cli/newbuild/newbuild_test.go#L45

bparees:
(the main tests i'd be concerned w/ are the unit test and the e2e-agnostic-cmd test)  and maybe e2e-aws-builds (i don't remember if that one has any cli tests in it, i don't think so though).

failures in the other jobs are almost certainly not related to this PR and we'll push through them as needed.

bparees:
/lgtm

may take a bit to get through the jobs but i don't think the PR itself needs anything further to pass

bparees:
@soltysh can you throw an approval on this?

soltysh (approved):
/approve

soltysh:
The main problem with test-cmd is `[sig-arch] events should not repeat pathologically` but @deejross is already working on that.

/retest

bparees:

failure there is also unrelated ([sig-network-edge] Cluster frontend ingress remain available)

PR #768, opened 2021-03-18
Bug 1925148: ImageRef Stream name priority increased

alicerum:
If `ImageRef` has `Stream` value filled, it means, that we are working with `ImageStream` that already exists.
In this case suggested name and suggested namespace for `ImageRef` should be taken from that `ImageStream`, and not from the object.

This fixes the bug in which new `ImageStream` is created when we are working with existing `ImageStream`, which has name and namespace different from our deployment name.

### DIFF
--- a/pkg/helpers/newapp/app/imageref.go
+++ b/pkg/helpers/newapp/app/imageref.go
@@ -230,11 +230,13 @@ func (r *ImageRef) SuggestName() (string, bool) {
 	if r == nil {
 		return "", false
 	}
-	if len(r.ObjectName) > 0 {
-		return r.ObjectName, true
-	}
+	// if ImageStream already exists, its name should
+	// have higher preference than object name
 	if r.Stream != nil {
 		return r.Stream.Name, true
+	}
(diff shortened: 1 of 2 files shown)


xiuwang:
Build oc client from pr code. Can't reproduce the bug issue now, when reference ImageStream already exists, no new ImageStream is created.
And no regression issue found.

$ ./oc new-app openshift/deployment-example
$ ./oc new-app --name=apptest -i deployment-example
--> Found image 1c839d8 (5 years old) in image stream "xiuwang/deployment-example" under tag "latest" for "deployment-example"


--> Creating resources ...
    deployment.apps "apptest" created
    service "apptest" created
--> Success
    Application is not exposed. You can expose services to the outside world by executing one or more of the commands below:
     'oc expose service/apptest' 
    Run 'oc status' to view your app.

alicerum:
/cc @adambkaplan 

it seems that test is not failing because of my changes, and there are many jobs failing in the similar way. @adambkaplan could you please do a review?

adambkaplan (approved):
/approve

PR #772, opened 2021-03-19
Bug 1919168: when mirroring to a file destination, mount images under the index location

ecordell:
when mirroring an index image: `quay.io/my/index:a`
that contains a reference to another image: `quay.io/other/bundle:1`
to a local file: `file:///local/index`

the referenced images are now mounted under the location for the index
itself, resulting in (for example):

`file:///local/index/my/index:a`
`file:///local/index/my/index/other/bundle:1`

this fixes a bug where the next step of an airgap mirror can't find the
images to be mirrored.

as a side effect, this keeps all catalog content stored under the same
fs location. Local file storage can be shared with other mirrored
images while keeping it simple to copy the full contents of an index
to another location.

### DIFF
--- a/pkg/cli/admin/catalog/mirrorer.go
+++ b/pkg/cli/admin/catalog/mirrorer.go
@@ -129,11 +129,21 @@ func imagesFromDb(file string) (map[string]struct{}, error) {
 }
 
 func mappingForImages(images map[string]struct{}, src, dest imagesource.TypedImageReference, maxComponents int) (mapping map[imagesource.TypedImageReference]imagesource.TypedImageReference, errs []error) {
-	// don't do any name mangling when not mirroring to a real registry
-	// this allows us to assume the names are preserved when doing multi-hop mirrors that use a file or s3 as an
-	// intermediate step
 	if dest.Type != imagesource.DestinationRegistry {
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/catalog/mirrorer.go
```diff
@@ -129,11 +129,21 @@ func imagesFromDb(file string) (map[string]struct{}, error) {
 }
 
 func mappingForImages(images map[string]struct{}, src, dest imagesource.TypedImageReference, maxComponents int) (mapping map[imagesource.TypedImageReference]imagesource.TypedImageReference, errs []error) {
-	// don't do any name mangling when not mirroring to a real registry
-	// this allows us to assume the names are preserved when doing multi-hop mirrors that use a file or s3 as an
-	// intermediate step
 	if dest.Type != imagesource.DestinationRegistry {
```
exdx, line 132: Thanks for taking this on after pairing on it!

Only thing I'm wondering is since we have `dest.Type != imagesource.DestinationRegistry` multiple times if there is a way to abstract this out by having the dest type fulfill an interface that includes a `mappingForImages` function so it's a bit more abstract. Not really worth it at this point, but maybe in the future if the mapping starts to drift more between the different destinations. 

/lgtm

PR #787, opened 2021-03-25
Remove `oc adm catalog build`

ecordell:
This command is only needed for appregistry-backed catalogs, which have not been supported for some time.

Deprecation notice published in 4.4 notes: https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-deprecation-of-operatorsources

This also means `oc adm catalog` is now the same on linux / other builds, and removes the dependency on operator-framework/operator-registry.


soltysh (approved):
/lgtm
/approve
/retest

soltysh:
Failure is real, but not related to this PR, I'm fixing it in a separate one.

PR #788, opened 2021-03-25
Fix usage error print

dastergon:
The code paths that use the `usageErrorFn` func were appending an [`%!(EXTRA ..)`](https://golang.org/pkg/fmt/) message in the output due to variadic args that were not passed properly.

Example output:
```
❯ oc process
error: Must pass a filename or name of stored template%!(EXTRA []interface {}=[])
See 'oc process -h' for help and examples
```

And output with the fix:
```
❯ oc process
error: The following parameters were provided more than once:
See 'oc process -h' for help and examples
```

PTAL


soltysh (approved):
/lgtm
/approve
/retest

PR #791, opened 2021-03-29
Fixing oc status help example message

ravitri:
Minor fix to correct the `oc status --help` message. 

Error:
```bash
$ oc --suggest
Error: unknown flag: --suggest
See 'oc --help' for usage.
$
```


soltysh (approved):
/lgtm
/approve

PR #795, opened 2021-03-30
CCCMO cluster operator addition to must-gather

Danil-Grigorev:
`cloud-controller-manager` cluster operator should report healthy conditions on cluster startup, when collected in must-gather output.

### DIFF
--- a/pkg/cli/admin/mustgather/summary.go
+++ b/pkg/cli/admin/mustgather/summary.go
@@ -68,6 +68,7 @@ var longExistingOperators = []string{
 	"network",
 	"openshift-apiserver",
 	"openshift-controller-manager",
+	"cloud-controller-manager",
 	"operator-lifecycle-manager",
 	"service-ca",
 	"storage",


soltysh (changes requested):
/hold
this is not a default component of a 4.8 that I just checked, why would it be added here?

Danil-Grigorev:
@soltysh I'm trying to figure out how we could see `cloud-controller-manager` must-gather in default CI runs. Do you have any pointers if this PR is not the right way to go? cc @elmiko

elmiko:
+1 to what @Danil-Grigorev is saying. to add a little more context, we are building an operator for the out-of-tree providers and are getting into some heavy CI work. this operator won't be in the 4.8 payload, but we would still like to see it included in the must-gather, is there a way we can do this given the state of our development?

soltysh:
You could add an optional step in CI to inspect ccm.

elmiko:
@Danil-Grigorev i think we can close this one now?

PR #797, opened 2021-03-30
Bug 1937020: `oc adm release new` should look at image stream status tags

smarterclayton:
The code that looks at spec tags when building a new release is
incorrectly assuming that a spec tag trumps the status tag, when in
fact the spec tag generation must be compared to the status tag
generation. Rearrange the code so that spec tags are skipped if the
spec tag is older than the status tag, but still errors if we are
waiting for a spec tag to be imported.

This may change behavior of OpenShift release creation in developer
environments, but the actual release tooling will now match the
behavior of the release. We were not seeing this bug impact us in
the release controller because the release controller clones the
image stream prior to creating a release, so spec/status are always
equal. This currently blocks users from creating releases off ocp/4.8
because there are spec tags that are old enough that their image has
been garbage collected in testing.

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -810,6 +810,24 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
 				continue
 			}
 
+			statusRef := findStatusTagEvents(inputIS.Status.Tags, ref.Name)
+
+			if ref.Generation != nil {
+				generation := *ref.Generation
+				if statusRef != nil && len(statusRef.Items) > 0 {
+					newest := statusRef.Items[0]
+					// spec tags that are waiting for import should prevent release building,
+					if generation > newest.Generation {
+						return fmt.Errorf("the tag %q in the source input stream has not been imported yet", statusRef.Tag)
+					}
+					// status tags that are newer should be chosen over the spec tag, otherwise we
+					// process them here depending on the type of tag
+					if generation < newest.Generation {
+						continue
+					}
+				}
+			}
+
 			if ref.From != nil && ref.From.Kind == "DockerImage" {
 				switch from, err := imagereference.Parse(ref.From.Name); {
 				case err != nil:
@@ -828,26 +846,23 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
 					covered.Insert(ref.Name)
 
 				case len(from.Tag) > 0:
-					tag := findStatusTagEvents(inputIS.Status.Tags, ref.Name)
-					if tag == nil {
+					if statusRef == nil {
+						klog.V(2).Infof("Can't use spec tag %q because the image has not been imported and all we have is a tag", ref.Name)
 						continue
 					}
-					if len(tag.Items) == 0 {
-						for _, condition := range tag.Conditions {
+					if len(statusRef.Items) == 0 {
+						for _, condition := range statusRef.Conditions {
 							if condition.Type == imageapi.ImportSuccess && condition.Status != metav1.StatusSuccess {
-								return fmt.Errorf("the tag %q in the source input stream has not been imported yet", tag.Tag)
+								return fmt.Errorf("the tag %q in the source input stream has not been imported yet", statusRef.Tag)
 							}
 						}
 						continue
 					}
-					if ref.Generation != nil && *ref.Generation != tag.Items[0].Generation {
-						return fmt.Errorf("the tag %q in the source input stream has not been imported yet", tag.Tag)
-					}
-					if len(tag.Items[0].Image) == 0 {
-						return fmt.Errorf("the tag %q in the source input stream has no image id", tag.Tag)
+					if len(statusRef.Items[0].Image) == 0 {
+						return fmt.Errorf("the tag %q in the source input stream has no image id", statusRef.Tag)
 					}
 
-					source := externalFn(tag.Items[0].DockerImageReference, tag.Items[0].Image)
+					source := externalFn(statusRef.Items[0].DockerImageReference, statusRef.Items[0].Image)
 					ref := ref.DeepCopy()
 					ref.From = &corev1.ObjectReference{Kind: "DockerImage", Name: source}
 					is.Spec.Tags = append(is.Spec.Tags, *ref)
@@ -855,7 +870,7 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
 				}
 				continue
 			}
-			// TODO: support ImageStreamTag and ImageStreamImage
+			// all other type of tags are skipped, because we will pull directly from status tags
 		}
 
 		for _, tag := range inputIS.Status.Tags {


wking:
[verify][1] seems to be mad about missing completions (`_output/verify-generated-completions/bash/.files_generated: No such file or directory`), although I'm not sure why.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/797/pull-ci-openshift-oc-master-verify/1376954473519255552#1:build-log.txt%3A304

pkg/cli/admin/release/new.go
```diff
@@ -810,6 +810,23 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
...
 			}
 
+			statusRef := findStatusTagEvents(inputIS.Status.Tags, ref.Name)
+
+			generation := int64(0)
+			if ref.Generation != nil {
+				generation = *ref.Generation
+				if statusRef != nil {
+					// spec tags that are waiting for import should prevent release building,
+					if generation > statusRef.Items[0].Generation {
```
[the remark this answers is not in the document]
wking: Hrm, [this example][1] suggests `Items[0]` might be the _oldest_ event in the history, and I'd expect us to want to use the newest.

[1]: https://docs.openshift.com/container-platform/4.7/openshift_images/image-streams-manage.html

pkg/cli/admin/release/new.go
```diff
@@ -828,34 +845,30 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
 					covered.Insert(ref.Name)
 
 				case len(from.Tag) > 0:
-					tag := findStatusTagEvents(inputIS.Status.Tags, ref.Name)
-					if tag == nil {
+					if statusRef == nil {
 						continue
```
wking, line 849: Can we add:

```go
klog.V(2).Infof("Can't use spec tag %q because it had no acceptable status tag events", ref.Name)
```

or some such?

pkg/cli/admin/release/new.go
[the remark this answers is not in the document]
smarterclayton: Yeah, validation allows zero items, i'll guard that

pkg/cli/admin/release/new.go
smarterclayton: added a message

soltysh:
> [verify](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/797/pull-ci-openshift-oc-master-verify/1376954473519255552#1:build-log.txt%3A304) seems to be mad about missing completions (`_output/verify-generated-completions/bash/.files_generated: No such file or directory`), although I'm not sure why.

There was a different problem, which should be already fixed by now related with switch to golang 1.16

pkg/cli/admin/release/new.go
```diff
@@ -810,6 +810,24 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
...
+				generation = *ref.Generation
+				if statusRef != nil && len(statusRef.Items) > 0 {
+					newest := statusRef.Items[0]
+					// spec tags that are waiting for import should prevent release building,
+					if generation > newest.Generation {
+						return fmt.Errorf("the tag %q in the source input stream has not been imported yet", statusRef.Tag)
+					}
+					// status tags that are newer should be chosen over the spec tag
+					if generation < newest.Generation {
+						continue
```
wking, line 826: When the status ref's generation is larger than the generation we see in the spec tag, isn't that fine?  I don't understand what this block is protecting us from.  I guess I would expect something like "if `Items[0]` is too new (because it has too large a generation), and for some reason we are worried about that large generation (not clear to me why that would be), then walk `Items[1...]` to see if we can find an exact match, and use that".  Not the current "`continue`, I guess we want to silently ignore this `ref`".

wking: [This][1] is me stumbling around in the ImageStream data which lead to the bug.  I still don't understand this spec generation < newest status generation case, but it looks like there are an awful lot of those tags, including things like the CVO, that we need to get a functional release image:

```console
$ yaml2json <ocp.4.8.is.yaml | jq '.[].spec.tags[] | select(.name == "cluster-version-operator")'
{
  "annotations": null,
  "from": {
    "kind": "DockerImage",
    "name": "registry.ci.openshift.org/ocp/4.8@sha256:f1582d5e229f03ea2ee995e6813392695b85a7b7e6c539cacb49261aa9d9879c"
  },
  "generation": 2716,
  "importPolicy": {},
  "name": "cluster-version-operator",
  "referencePolicy": {
    "type": "Source"
  }
}
$ yaml2json <ocp.4.8.is.yaml | jq '.[].status.tags[] | select(.tag == "cluster-version-operator")'
{
  "items": [
    {
      "created": "2021-02-18T23:29:19Z",
      "dockerImageReference": "image-registry.openshift-image-registry.svc:5000/ocp/4.8@sha256:96124bee9c94ffcc36bd60348b2b6cadd235c501b91e972fbbda6e15160e916a",
      "generation": 3057,
      "image": "sha256:96124bee9c94ffcc36bd60348b2b6cadd235c501b91e972fbbda6e15160e916a"
    },
    {
      "created": "2021-02-16T23:23:12Z",
      "dockerImageReference": "image-registry.openshift-image-registry.svc:5000/ocp/4.8@sha256:ef8c687b2b64cf9f1c849ade80cddf014dd44be3111330ac23bfa9638860c5b3",
      "generation": 3057,
      "image": "sha256:ef8c687b2b64cf9f1c849ade80cddf014dd44be3111330ac23bfa9638860c5b3"
    },
    {
      "created": "2021-02-16T02:06:20Z",
      "dockerImageReference": "image-registry.openshift-image-registry.svc:5000/ocp/4.8@sha256:a7cac3991b20066042e8df874c0d6673d5ed82d8cf3d1a2d62f3f28de44c3aa3",
      "generation": 3057,
      "image": "sha256:a7cac3991b20066042e8df874c0d6673d5ed82d8cf3d1a2d62f3f28de44c3aa3"
    }
  ],
  "tag": "cluster-version-operator"
}
```

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1937020#c3

smarterclayton: Status tags are created by a) tagging in spec, b) pushes (which don't change spec).  At the time a status tag is created the current metadata.generation is taken.  Any time a spec tag is changed the tag gets the generation of metadata.generation.  If status tag generation > spec tag generation, someone pushed.  If status tag generation < spec tag generation, we're waiting to import that tag.  If spec tag generation == status tag generation, we've imported the spec tag.

smarterclayton: Everything besides item[0] is history.  So generation comparison is about waiting for import, picking the pushed tag, or taking the spec tag.

wking: So:

* `generation > newest.Generation`, we're waiting for an import, error out (as you are currently doing).
* `generation == newest.Generation`, we've imported a spec tag.  But we may also have had subsequent pushes, as long as nobody else has touched something to get `metadata.generation` bumped.  For example, see the `machine-os-content` entries in [my bug comment][1].  I expect we want to use the status tag, because it's at least as new as the spec tag.
 * `generation < newest.Generation`, someone has definitely pushed since the last time the spec tag was bumped.  Use the status tag.

Doesn't that mean we want to drop this `generation < newest.Generation` `continue` block this thread is attached to?

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1937020#c3

wking: Ah, I see the intention is to fall through to [the `status.tags` iterator][1] in the following block.  `continue` makes sense for that, but then I think it should be `<=` here to fall back to status for the `machine-os-content` pushed-since-the-last-spec-bump case.

[1]: https://github.com/openshift/oc/blob/be64dc73d95b407c2b0b36026fd0c38198b5be3e/pkg/cli/admin/release/new.go#L861-L866

pkg/cli/admin/release/new.go
```diff
@@ -810,6 +810,24 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
 				continue
 			}
 
+			statusRef := findStatusTagEvents(inputIS.Status.Tags, ref.Name)
+
+			generation := int64(0)
```
wking, line 815: nit: I think you can drop this and use `generation := *ref.Generation` inside the subsequent `if`.

pkg/cli/admin/release/new.go
smarterclayton: this can only be a digest, and so by design they would be equal.  if they’re equal, would prefer to emphasize we’re choosing the spec tag (by handling it here)

smarterclayton: also the handling of “not yet imported” is inside this loop, so <= would break that handling

smarterclayton: added a bit more comment

smarterclayton:
If there are no other comments this is getting tagged

/retest

pkg/cli/admin/release/new.go
```diff
@@ -810,6 +810,24 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
 				continue
 			}
 
+			statusRef := findStatusTagEvents(inputIS.Status.Tags, ref.Name)
```
soltysh, line 813: Nit: `imageutil.LatestTaggedImage` does just that.

soltysh: I'll open followup to change these, I see similarly we have methods searching through spec.

soltysh (approved):
/lgtm
/retest

pkg/cli/admin/release/new.go
smarterclayton: It's not quite the same though iirc.

smarterclayton: IT's not the same, we need generation here.

soltysh: Both are identical, both iterate over `.status.tags` and you're getting a `TagEvent` which contains the generation. We don't do more comparison than just simple iteration and name matching. After which we in both cases pick the 0th element from tag events.

PR #807, opened 2021-04-12
[WIP] Updating doc generation to separate dev and admin commands

bergerhoffer:
@soltysh This is a WIP PR for seeing if we can separate the dev commands from the oc adm commands and produce two separate files. In the current version of the docs, we put them onto separate pages (dev [here](https://docs.openshift.com/container-platform/4.7/cli_reference/openshift_cli/developer-cli-commands.html) and admin [here](https://docs.openshift.com/container-platform/4.7/cli_reference/openshift_cli/administrator-cli-commands.html)). We'd like to do the same with this if possible.

This is mostly a copy/paste with adjustments. I'm assuming that it might be possible to do this a cleaner way without duplicating, but this is what I came up with to see if it could work. Let me know what you think.

I also know that I'll need a BZ for this if we want to get it in now - will open one if we want to move forward with this.

For comparison, this is a preview that has all commands in a single file, and is really long: https://deploy-preview-31403--osdocs.netlify.app/openshift-enterprise/latest/cli_reference/openshift_cli/cli-command-reference.html

And here are the commands broken into two separate files (preview is internal):
* Dev: http://file.rdu.redhat.com/~ahoffer/2021/OSDOCS-1790/OSDOCS-1790/cli_reference/openshift_cli/developer-cli-commands.html
* Admin: http://file.rdu.redhat.com/~ahoffer/2021/OSDOCS-1790/OSDOCS-1790/cli_reference/openshift_cli/administrator-cli-commands.html

/assign @soltysh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /oc
 /clicheck
 /gendocs
+/gendocs-admin
 /genman
 /_output
 .idea/
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,10 @@ generate-docs:
 	go run ./tools/gendocs
 .PHONY: generate-docs
 
+generate-docs-admin:
+	go run ./tools/gendocs-admin
+.PHONY: generate-docs-admin
+
 generate-versioninfo:
 	SOURCE_GIT_TAG=$(SOURCE_GIT_TAG) hack/generate-versioninfo.sh
 .PHONY: generate-versioninfo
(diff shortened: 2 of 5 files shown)


bergerhoffer:
@soltysh Can you take a look at this and see if you have a better way that we can accomplish this? Would like to get an update in to separate the commands for 4.8.

soltysh:
Replaced with #827 
/close

PR #811, opened 2021-04-21
oc: add debug collection back to inspect

mfojtik:


mfojtik:
/hold

because David does not want this :-)

PR #815, opened 2021-04-29
Bug 1937020: Releases from image streams must prefer status tag

smarterclayton:
In the previous change a key scenario was omitted. An image stream
tag might have the form:

```
spec:
  tags:
  - name: foo
    generation: 5
    from: {name: "image1"}
status:
  tags:
  - tag: foo
    items:
    - generation: 5
      image: image2
    - generation: 5
      image: image1
```

because a push happened immediately after a spec tagging (generation
only changes on import). Switch oc adm release new to always prefer
status tag, and to let the image stream decide what pull spec should
be used.


wking:
Following up on #797.

wking (approved):
/bugzilla refresh
/lgtm

soltysh (approved):
/lgtm

PR #816, opened 2021-04-30
release: extract Linux binaries for multiple architectures

yselkowitz:
Previously, one could extract x86_64 Windows and Mac binaries (the only 
architectures for those OSes until recently, and still the only ones 
fully supported by Go at the moment) and the Linux binary for the 
payload architecture, but there was no way to e.g. extract an x86_64 
Linux binary to use as a client with a server of a different another 
architecture.  This fixes that, and also extracts binaries for the 
(client) architecture on which oc is run even on Linux.

This also adds the ability to extract an x86_64 Linux installer from a 
payload of any architecture, on the basis that a remote install should 
only require what the user has on their desk, and they may not have (and 
should not need) a matching system for that purpose.  This uses a new 
addition to installer-artifacts.

This is blocked by https://github.com/openshift/installer/pull/4891 and 
whatever degree of backwards compatibility is required.

/cc @soltysh

### DIFF
--- a/pkg/cli/admin/release/extract_tools.go
+++ b/pkg/cli/admin/release/extract_tools.go
@@ -38,8 +38,10 @@ import (
 // extractTarget describes how a file in the release image can be extracted to disk.
 type extractTarget struct {
 	OS       string
+	Arch     string
 	Command  string
 	Optional bool
+	NewArch  bool
 
 	InjectReleaseImage   bool
 	InjectReleaseVersion bool
@@ -136,6 +138,9 @@ var (
 	OpenShift is licensed under the Apache Public License 2.0. The source code for this
 	program is [located on github](https://github.com/openshift/origin).
 	`)
+
+	// indicates that the architecture of the binary matches the release payload
+	targetReleaseArch = "release-arch"
 )
 
 // extractTools extracts specific commands out of images referenced by the release image.
@@ -147,6 +152,7 @@ func (o *ExtractOptions) extractCommand(command string) error {
 	availableTargets := []extractTarget{
 		{
 			OS:      "darwin",
+			Arch:    "amd64",
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli-artifacts", From: "usr/share/openshift/mac/oc"},
 
@@ -157,6 +163,7 @@ func (o *ExtractOptions) extractCommand(command string) error {
 		},
 		{
 			OS:      "linux",
+			Arch:    targetReleaseArch,
 			Command: "oc",
 			Mapping: extract.Mapping{Image: "cli", From: "usr/bin/oc"},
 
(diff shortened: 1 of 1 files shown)


yselkowitz:
This is the commit we discussed last month.  The installer part is in now, so we're ready to proceed with this.
/assign @soltysh

pkg/cli/admin/release/extract_tools.go
```diff
@@ -157,6 +160,7 @@ func (o *ExtractOptions) extractCommand(command string) error {
 		},
 		{
 			OS:      "linux",
+			Arch:    "release",
```
soltysh, line 163: `amd64` I expect these to be copies of the new locations, which we'll keep for 1-2 releases for backwards compatibility only.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -187,6 +204,7 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
```
soltysh, line 207: Ditto.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -195,8 +213,21 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
+			Mapping: extract.Mapping{Image: "installer-artifacts", From: "usr/share/openshift/linux_amd64/openshift-install"},
+
+			Readme:               readmeInstallUnix,
+			InjectReleaseImage:   true,
+			InjectReleaseVersion: true,
+			ArchiveFormat:        "openshift-install-linux-amd64-%s.tar.gz",
+		},
 		{
 			OS:       "linux",
+			Arch:     "release",
```
soltysh, line 230: Ditto.

soltysh (changes requested):
The other thing missing there is exposing `--command-arch` to be able to overwrite the architecture.

yselkowitz:
The architecture of `/usr/bin/*` depends on the payload; in an X86 payload they are amd64, but in an ARM payload they are arm64, etc.  Since, even in the long term (iow after 1.17 when the arm64 linker should be fixed in golang), the baremetal-installer is not cross-compilable (as it depends on libvirt), there still needs to be a concept of a payload-dependent binary.

Ultimately, though, the reason I changed approaches in this PR is that the names "openshift-client-linux" and "openshift-install-linux" are relied upon by SDA as they are used in the download pages (e.g [[1](https://cloud.redhat.com/openshift/install/aws/installer-provisioned)]).  Once I realized that making these names unambiguous -- by using only cli-artifacts instead of cli -- would require cross-team coordination, I decided to focus only on the immediate requirements in this PR.

Similarly, I didn't want to add --command-arch, as this wasn't covering all the different permutations, I wasn't convinced that there was a use case (after all, is anyone really going to e.g. want an s390x client for their x86 payload?), and of course it's a lot harder to remove the flag once it's added.

Hence the change to a much more minimalistic approach in this PR, to only add linux/amd64 binaries to non-x86 payloads, for which we have immediate need.

We should indeed have further conversations about what this should look like long-term in a multi-arch world (for not only Linux, but Mac, and maybe one day Windows) in conjunction with SDA, and probably ART, as there are also other binaries which are or perhaps should be extracted from the payload nowadays too.

/cc @LorbusChris

soltysh (approved):
/lgtm
/approve

yselkowitz:
e2e-aws-upgrade failure is https://bugzilla.redhat.com/show_bug.cgi?id=1970315
e2e-metal-ipi-ovn-ipv6 failure is https://bugzilla.redhat.com/show_bug.cgi?id=1975865

pkg/cli/admin/release/extract_tools.go
```diff
@@ -333,7 +384,7 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			target.Mapping.To = filepath.Join(dir, target.Mapping.Name)
 		} else {
 			target.Mapping.To = filepath.Join(dir, target.Command)
-			target.Mapping.Name = fmt.Sprintf("%s-%s", target.OS, target.Command)
+			target.Mapping.Name = fmt.Sprintf("%s-%s-%s", target.OS, target.Arch, target.Command)
```
smarterclayton, line 387: Did this change the output of extract?  That's a breaking change (we can't change filenames on disk that might be used as part of automation).

If so, this needs to be reverted and replaced with a change that only changes the names on the net new components.

yselkowitz: No, that bit is used only for internal tracking of what has been extracted.

PR #818, opened 2021-05-05
Bug 1951203: Allow users to set a limit on ICSP file size

awgreene:


pkg/cli/admin/catalog/mirror.go
```diff
@@ -71,7 +71,10 @@ var (
 	`)
 )
 
-const IndexLocationLabelKey = "operators.operatorframework.io.index.database.v1"
+const (
+	IndexLocationLabelKey = "operators.operatorframework.io.index.database.v1"
+	mbSize                = 1000000
```
kevinrizza, line 76: Nit: Should this just be `maxBites`? mb is confusing because my first thought is megabytes

pkg/cli/admin/catalog/mirror.go
```diff
@@ -145,6 +149,7 @@ func NewMirrorCatalog(f kcmdutil.Factory, streams genericclioptions.IOStreams) *
 	flags.StringVar(&o.FromFileDir, "from-dir", o.FromFileDir, "The directory on disk that file:// images will be read from. Overrides --dir")
 	flags.IntVar(&o.MaxPathComponents, "max-components", 2, "The maximum number of path components allowed in a destination mapping. Example: `quay.io/org/repo` has two path components.")
 	flags.StringVar(&o.IcspScope, "icsp-scope", o.IcspScope, "Scope of registry mirrors in imagecontentsourcepolicy file. Allowed values: repository, registry. Defaults to: repository")
+	flags.IntVar(&o.MaxICSPSize, "max-icsp-size", mbSize, "The maximum size of an ICSP yaml")
```
kevinrizza, line 152: Maybe we should define the unit (bytes) here? And also mention and set the default in the flag rather than as a global variable?

pkg/cli/admin/catalog/mirror.go
```diff
@@ -485,35 +522,26 @@ func generateICSP(out io.Writer, name string, icspScope string, mapping map[imag
...
 
-	// Create an unstructured object for removing creationTimestamp
-	unstructuredObj := unstructured.Unstructured{}
-	var err error
-	unstructuredObj.Object, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&icsp)
-	if err != nil {
-		return nil, fmt.Errorf("error converting to unstructured: %v", err)
+		if len(y) > byteLimit {
+			if len(icsp.Spec.RepositoryDigestMirrors) > 0 {
+				icsp.Spec.RepositoryDigestMirrors = icsp.Spec.RepositoryDigestMirrors[:len(icsp.Spec.RepositoryDigestMirrors)-1]
```
kevinrizza, line 537: Won't this just limit the number of mappings and then return? Don't we need to generate multiple files for the overflow?

awgreene: > Won't this just limit the number of mappings and then return?
Yes, but this function is called until all entries in the map are removed.

> Don't we need to generate multiple files for the overflow?
The output of this change is a single file with multiple yamls separated by `---`.

awgreene: Example Yaml:
```apiVersion: operator.openshift.io/v1alpha1
metadata:
  name: redhat-operator-index-0
spec:
  repositoryDigestMirrors:
  - mirrors:
    - quay.io/agreene/new-index-openshift4-ose-sriov-network-operator
    source: registry.redhat.io/openshift4/ose-sriov-network-operator
  - mirrors:
    - quay.io/agreene/new-index-integration-service-registry-infinispan-rhel8
    source: registry.redhat.io/integration/service-registry-infinispan-rhel8
  - mirrors:
    - quay.io/agreene/new-index-openshift-serverless-1-tech-preview-serving-webhook-rhel8
    source: registry.redhat.io/openshift-serverless-1-tech-preview/serving-webhook-rhel8
---
metadata:
  name: redhat-operator-index-1
spec:
  repositoryDigestMirrors:
  - mirrors:
    - quay.io/agreene/new-index-distributed-tracing-jaeger-es-rollover-rhel8
    source: registry.redhat.io/distributed-tracing/jaeger-es-rollover-rhel8
  - mirrors:
    - quay.io/agreene/new-index-openshift-pipelines-tech-preview-pipelines-triggers-webhook-rhel8
    source: registry.redhat.io/openshift-pipelines-tech-preview/pipelines-triggers-webhook-rhel8
  - mirrors:
    - quay.io/agreene/new-index-openshift-service-mesh-grafana-rhel8
    source: registry.redhat.io/openshift-service-mesh/grafana-rhel8
---
metadata:
  name: redhat-operator-index-2
spec:
  repositoryDigestMirrors:
  - mirrors:
    - quay.io/agreene/new-index-openshift-serverless-1-tech-preview-serving-autoscaler-rhel8
    source: registry.redhat.io/openshift-serverless-1-tech-preview/serving-autoscaler-rhel8
  - mirrors:
    - quay.io/agreene/new-index-openshift-pipelines-tech-preview-pipelines-entrypoint-rhel8
    source: registry.redhat.io/openshift-pipelines-tech-preview/pipelines-entrypoint-rhel8
  - mirrors:
    - quay.io/agreene/new-index-codeready-workspaces-jwtproxy-rhel8
    source: registry.redhat.io/codeready-workspaces/jwtproxy-rhel8```

File: pkg/cli/admin/catalog/mirror.go (hunk 1 above)
Comment (awgreene) [reply pkg/cli/admin/catalog/mirror.go:77]: I'll remove it as requested in https://github.com/openshift/oc/pull/818/files#r627375702

File: pkg/cli/admin/catalog/mirror.go (hunk 2 above)
Comment (awgreene) [reply pkg/cli/admin/catalog/mirror.go:153]: Can do.

**Comment by kevinrizza** [comment]
Aside from test failures, this lgtm

**Comment by awgreene** [comment]
Unit tests will keep failing until #821 is merged.

**Comment by kevinrizza** [comment]
/hold cancel

@mfojtik @ecordell @bparees Please review :)

File: pkg/cli/admin/catalog/mirror.go [hunk 4]
```diff
@@ -68,10 +69,15 @@ var (
 		# Edit the mirroring mappings and mirror with "oc image mirror" manually
 		oc adm catalog mirror --manifests-only quay.io/my/image:latest myregistry.com
 		oc image mirror -f manifests/mapping.txt
+
+		# Delete all ImageContentSourcePolicies generated with this command
```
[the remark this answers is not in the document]
awgreene: Accepted!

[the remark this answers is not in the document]
bparees: hm, i don't see the update?

bparees:
/lgtm cancel

doesn't look like my text suggestion made it in

mfojtik:
LGTM

awgreene:
>/lgtm cancel
doesn't look like my text suggestion made it in

Sorry @bparees, the change is now present.

bparees:
@mfojtik @soltysh need an approval from one of you

soltysh (changes requested):
/hold
Why do we need to artificially limit the size of ICSP, since I've been working with those over the past week I'm pretty familiar and I know that they rarely are bigger than 1kB, why would a single mirror command against one repository generate ICSP which on 2nd apply exceeds etcd limit? 

My guess for 2nd apply failure is that `oc apply` applies `last-applied-configuration` annotation which with bigger resources tends to exceed allowed annotation limit. 

The current approach of setting limit seems rather artificial and unnecessary.

bparees:
it's limiting it to 1meg, not 1k.

the linked bug summarizes why this is needed:
https://bugzilla.redhat.com/show_bug.cgi?id=1951203

```
Any ICSP that is generated by oc adm catalog mirror should be less than 50% of the default size limit of etcd (which is 2mb) or just under 1mb.
```

soltysh:
> it's limiting it to 1meg, not 1k.

Right, I'm asking why this limit is needed in the first place, since most of the ICSP files I've worked with were less than 1k in size. What is causing the generated files to be that big that we need to expose a flag for users to control. Not to mention that users rarely should know or understand this kind of functionality, they trust us we do the right thing for them.

bparees:
> Right, I'm asking why this limit is needed in the first place, since most of the ICSP files I've worked with were less than 1k in size. What is causing the generated files to be that big that we need to expose a flag for users to control. Not to mention that users rarely should know or understand this kind of functionality, they trust us we do the right thing for them.

the command defaults to do the right thing, so they can trust us to do the right thing.

Is your concern that its user settable? 

As for how it can happen, it can happen because the mirror process can generate repository-scoped mirror rules (which customers need/want), and that can generate a massive ICSP if you have hundreds or thousands of images in a catalog that you are mirroring.

soltysh:
> Is your concern that its user settable?

Yeah, mostly that. 

> As for how it can happen, it can happen because the mirror process can generate repository-scoped mirror rules (which customers need/want), and that can generate a massive ICSP if you have hundreds or thousands of images in a catalog that you are mirroring.

If a single entry is ~100 bytes, example:
```
  - mirrors:
    - quay.io/soltysh/release
    source: docker.io/soltysh/release
```
to get to 1M, I'd need more than 10k entries like the one above, is our mirror actually creating more than that?

bparees:
>>    Is your concern that its user settable?

> Yeah, mostly that.

the setting probably isn't necessary, but it's nice to have the escape hatch in case things go wrong here.  I don't really see that an extra flag w/ a sane default creates a big UX issue.  People who don't have a reason to care can/will ignore that it exists.

soltysh:
> the ability for users to set it probably isn't necessary, but it's nice to have the escape hatch in case things go wrong here. I don't really see that an extra flag w/ a sane default creates a big UX issue. People who don't have a reason to care can/will ignore that it exists.

Why then expose unless we know it's needed.

bparees:
> Why then expose unless we know it's needed.

so that if someone needs it, we have an answer for them instead of "sorry, we didn't consider that and you'll have to wait for a fix".  it's an escape hatch.

soltysh:
> so that if someone needs it, we have an answer for them instead of "sorry, we didn't consider that and you'll have to wait for a fix". it's an escape hatch.

That's why I'm pushing towards fixing the root cause of generating this many ICSP entries over exposing a flag which is confusing and might introduce a set of other problems.

bparees:
> That's why I'm pushing towards fixing the root cause of generating this many ICSP entries over exposing a flag which is confusing and might introduce a set of other problems.

it doesn't matter why it happened.  Whether or not there is another bug, it is perfectly valid to fix this command so that it cannot/does not generate ICSPs which are too big to be applied to the cluster.

So we can get back to you on "why did it happen/is there a bug in the generation"(I would assume/hope the OLM team did that investigation before producing this fix, but sure, we can sanity check that) but that is not sufficient reason to block a change that makes the generation logic safer: "we know generated content larger than 1meg is invalid, so break that content up into pieces".  We know the limit exists, we should update the command to respect the limit.

And if we agree that it it's reasonable to protect ourselves from etcd limits, then the next question is, "should we also make that protection configurable" for which i again say "i can think of no good reason not to give ourselves that escape hatch. The addition of the flag(which has a sane default value) does not meaningfully impact users (oc adm catalog mirror has 14 flags already, many of which are not "normally" needed)"

pkg/cli/admin/catalog/mirror.go
```diff
@@ -472,36 +519,39 @@ func generateCatalogSource(source imagesource.TypedImageReference, mapping map[i
...
-		} else {
-			registryMapping[k.Ref.AsRepository().String()] = v.Ref.AsRepository().String()
-		}
-	}
 	for key := range registryMapping {
 		icsp.Spec.RepositoryDigestMirrors = append(icsp.Spec.RepositoryDigestMirrors, operatorv1alpha1.RepositoryDigestMirrors{
 			Source:  key,
 			Mirrors: []string{registryMapping[key]},
 		})
+		y, err := yaml.Marshal(icsp)
```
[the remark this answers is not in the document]
awgreene: I can run a performance check but I did not see a significant change when running the tool locally with ~600 entries.

awgreene:
Thanks for all the conversation @bparees and @soltysh. The crux of the conversation seems to focus primarily around this point:

> That's why I'm pushing towards fixing the root cause of generating this many ICSP entries over exposing a flag which is confusing and might introduce a set of other problems.

The oc adm mirror catalog is generating the correct number of ICSPs. When generating an ICSP for a catalog with the `--icsp-scope` flag set to `repository` a single operator may account for multiple 
entries because of the "Related Images" they define which typically exist in their own repository. For context, I've seen a single operator define over 50 Related Images.

I tested the existing oc command generated from the master branch against the 4.8 RH Catalog to better understand the current size of the generated ICSP using the following command:
```
$ oc adm catalog mirror --icsp-scope=repository registry.redhat.io/redhat/redhat-operator-index:v4.8 quay.io/agreene/new-index --manifests-only`
```

The ICSP is 93857 bytes large and contains 579 entries, this file will continue to grow as more operators are added to the catalog. 

To @soltysh's point, users will probably hit the 262144 byte annotation limit, which might make 250000 bytes a more sane default. However, if any other annotations exist on the ICSP (possibly by means
of a mutating webhook or some other controller), I believe users would still desire to set the ICSP limit using the flag introduced in this PR.

bparees:
> To @soltysh's point, users will probably hit the 262144 byte annotation limit, which might make 250000 bytes a more sane default. However, if any other annotations exist on the ICSP (possibly by means
of a mutating webhook or some other controller), I believe users would still desire to set the ICSP limit using the flag introduced in this PR.

currently this PR defaults it to 1,000,000 bytes, right?  If we know that a 300,000 byte ICSP will break (because the entire ICSP shows up as an annotation, aiui?), then shouldn't we default it to a 250,000 limit?

In my mind the default should be the lowest value we know will prevent breakage in a standard customer environment.  The point of making it user configurable is for the case where a customer environment has higher or lower limits (e.g. they tuned their etcd object size limit).

awgreene:
>currently this PR defaults it to 1,000,000 bytes, right? If we know that a 300,000 byte ICSP will break (because the entire ICSP shows up as an annotation, aiui?), then shouldn't we default it to a 250,000 limit?

Correct, I did not know that this annotation size limit existed until I saw @soltysh's review.

> In my mind the default should be the lowest value we know will prevent breakage in a standard customer environment. The point of making it user configurable is for the case where a customer environment has higher or lower limits (e.g. they tuned their etcd object size limit).

Agreed.

awgreene:
@bparees I set the default to 250000 in the latest version of the PR.

bparees:
/lgtm

@soltysh if you still have concerns we can discuss on slack tomorrow, i'd like to get this fix merged before 4.8 hits code freeze EOD tomorrow.

soltysh:
> And if we agree that it it's reasonable to protect ourselves from etcd limits, then the next question is, "should we also make that protection configurable" for which i again say "i can think of no good reason not to give ourselves that escape hatch. The addition of the flag(which has a sane default value) does not meaningfully impact users (oc adm catalog mirror has 14 flags already, many of which are not "normally" needed)"

My default thinking is we know better than user how to properly limit these, rarely do users need to change that limit. From my experience with oc and kubectl I know that the more flags the more confusion we introduce to the users. I'm very reluctant adding more flags just in case, which is how this sounds. 

Having said that, I'll give you a free pass on this one :smile:

soltysh (approved):
/approve
/retest

bparees:
/hold cancel
(i'm taking @soltysh's approval as implicit hold cancelation intent)

Thanks @soltysh, i think it's always a worthwhile discussion when adding more flags/configurability as to whether the added complexity is worth it/necessary.

soltysh:
> (i'm taking @soltysh's approval as implicit hold cancelation intent)

sorry I forgot :sweat_smile: thx 

> Thanks @soltysh, i think it's always a worthwhile discussion when adding more flags/configurability as to whether the added complexity is worth it/necessary.

:100:

pkg/cli/admin/catalog/mirror.go
```diff
@@ -382,10 +390,49 @@ func (o *MirrorCatalogOptions) Run() error {
...
+	}
+	return registryMapping
+}
+
+func generateICSPs(out io.Writer, source string, icspScope string, maxICSPSize int, mapping map[imagesource.TypedImageReference]imagesource.TypedImageReference) ([][]byte, error) {
+	registryMapping := getRegistryMapping(out, icspScope, mapping)
+	icsps := [][]byte{}
+
+	for i := 0; len(registryMapping) != 0; i++ {
+		icsp, err := generateICSP(out, source+"-"+strconv.Itoa(i), maxICSPSize, registryMapping)
```
orenc1, line 417: @awgreene , is there a reason to add a suffix for the name ICSP object we are creating?
That way, the generated object name is different than the name the user requested.

/cc @tiraboschi

InbarRose: this may have another unintended consequence. 
If we create an ICSP with name "myname" it will be created with "myname-0"
if we check for an ICSP with name "myname" it will not be there. So if we then create another ICSP with the name "myname" it will clash on "myname-0" already existing.

awgreene: >@awgreene , is there a reason to add a suffix for the name of the ICSP object we are creating?
That way, the generated object name is different than the name the user requested.

Good morning @orenc1, thanks for taking a look at this PR. This is an intentional change. The ICSPs generated when mirroring each repository can easily exceed the:
- Maximum annotation key size when using the `oc apply` command
- Maximum ETCD object store size

Using the user provided name would fail to handle either of the situations described above as the generated ICSP would be rejected when applied to the cluster. 

>this may have another unintended consequence.
If we create an ICSP with name "myname" it will be created with "myname-0"
if we check for an ICSP with name "myname" it will not be there. So if we then create another ICSP with the name "myname" it will clash on "myname-0" already existing.

This is true but I fail to understand how this is an issue. The second attempt to create the ICSP should notify the user that the ICSP already exists and that the ICSP has already been applied to the cluster. Furthermore, if you are concerned about removing the ICSPs, the `oc adm catalog mirror --help` command provides an example command for cleaning up the ICSPs it has previously generated..

awgreene: cc @InbarRose ^^

PR #819, opened 2021-05-06
[WIP] Use gokrb5 for GSSAPI auth

LorbusChris:
https://issues.redhat.com/browse/ARM-9


pkg/helpers/tokencmd/negotiator_gssapi.go
```diff
@@ -1,197 +1,77 @@
...
-			return nil, convertGSSAPIError(err)
+			return nil, err
 		}
-		defer nameBuf.Release()
 
-		name, err := nameBuf.Name(lib.GSS_C_NT_HOSTBASED_SERVICE)
+		// TODO check whether REALM can be safely omitted (2nd param)
+		g.client = krb5client.NewWithKeytab(g.principalName, "", keytab, config)
+
+		err = g.client.Login()
```
LorbusChris, line 48: TODO: Move this out of the condition

pkg/helpers/tokencmd/negotiator_gssapi.go
```diff
@@ -1,197 +1,77 @@
-// +build gssapi
```
stlaz, line 1: Does this work on windows, then? You couldn't use `/etc/krb5.conf` ever as the default location, then. What happens when both this handler and `SSPI` are active?

pkg/helpers/tokencmd/negotiator_gssapi.go
```diff
@@ -1,197 +1,77 @@
...
 		if err != nil {
-			return nil, convertGSSAPIError(err)
+			return nil, err
 		}
-		g.name = name
-		g.ctx = lib.GSS_C_NO_CONTEXT
 	}
 
-	incomingTokenBuffer, err := lib.MakeBufferBytes(challengeToken)
+	servicePrincipleName, err := getServiceName('/', requestURL)
```
stlaz, line 54: `servicePrincipalName`?

stlaz: Why did you change the principal separator?

stlaz (changes requested):
/hold
you'll have to prove this works against a cluster that is set up with IdP responding with Negotiate challenges

pkg/helpers/tokencmd/negotiator_gssapi.go
LorbusChris: It's the format the lib expects, see https://github.com/jcmturner/gokrb5/blob/master/v8/USAGE.md#generic-kerberos-client

LorbusChris: > `servicePrincipalName`?

Yes, I'm reusing the vocabulary that SPNEGO uses here.

stlaz: I meant `s/Principle/Principal` to match the Kerberos terms

stlaz: As for the separator: https://web.mit.edu/kerberos/krb5-1.5/krb5-1.5.4/doc/krb5-user/What-is-a-Kerberos-Principal_003f.html

pkg/helpers/tokencmd/negotiator_gssapi.go
LorbusChris: I'd expect it to, yes. With this it won't be required to run `kinit` beforehand to get the ticket.
I think it might be easiest if we require the user to set the path to the conf file as an env var and read it from there, which should work on all OSes.

PR #820, opened 2021-05-07
Bug 1957295: Must gather pod should have priority class

ravisantoshgudimetla:
Must-gather pod should not preemptable but it
is ok to be OOMKilled if there is resource
crunch on the node. So, it should have a
priority class of `system-cluster-critical`

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -606,8 +606,12 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 			},
 		},
 		Spec: corev1.PodSpec{
-			NodeName:      node,
-			RestartPolicy: corev1.RestartPolicyNever,
+			NodeName: node,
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -607,6 +607,7 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 		},
 		Spec: corev1.PodSpec{
 			NodeName:      node,
+			PriorityClassName: "system-cluster-critical",
```
deads2k, line 610: can you provide the comment you gave me in slack about the three options, what they mean and why I chose this one?

As I recall, I wanted this pod to be oomkillable, but not pre-emptable.

deads2k:
needs a comment

/approve

soltysh (approved):
/lgtm

soltysh:
/override ci/prow/unit
this PR should help solve issues in origin tests, and unit is not related, @ravisantoshgudimetla will look into unit failure separately

PR #827, opened 2021-05-18
Generate docs separately for admin and non-admin users

soltysh:
@bergerhoffer this will replace #807, it builds on your PR. In 2nd commit I tweaked the generation so it leaves in a single file and the next few is just some minor cleanups that I stumbled upon. 

Now you should still have `make generate-docs` which will produce file in `docs/generated/oc-by-example-content.adoc` and `make generate-docs-admin` which will produce `docs/generated/oc-adm-by-example-content.adoc`. 

/assign @bergerhoffer


soltysh:
This is only touching docs, so I'm adding valid-bug label to it.

pkg/cli/admin/policy/modify_roles.go
```diff
@@ -29,10 +29,10 @@ import (
 var (
 	addRoleToUserExample = templates.Examples(`
 		# Add the 'view' role to user1 for the current project
-		oc adm policy add-role-to-user view user1
+		oc policy add-role-to-user view user1
```
bergerhoffer, line 32: @soltysh Unrelated to doc generation, but I'm just curious why `adm` is being removed from these examples. Aren't these examples for the `oc adm policy *` commands?

bergerhoffer:
Thanks @soltysh this works great! Once this is merged, that should be all we need from you all to be able to get this in for 4.8.

Do you have thoughts on when we should run this, and against which branch? I'm assuming that we'd want to wait until further in the release so that we make sure we're capturing all changes. Would waiting until after feature freeze be a good enough target, or do you think we should wait until code freeze?

And does it make sense for us run it against the appropriate version branch (like release-4.8) as opposed to master?

/lgtm

pkg/cli/admin/policy/modify_roles.go
soltysh: @bergerhoffer they are used for both `oc policy` and `oc adm policy`, there are 2 possibilities how to fix this, either have a flag which presents the appropriate prefix, or just use `oc policy` which will be available for every user (admin and non-admin). I went with the latter, for simplicity :sweat_smile:

bergerhoffer: Thanks @soltysh. Do the commands do the exact same things whether an admin uses "oc adm policy" or "oc policy"?

PR #829, opened 2021-05-20
Bug 1823143: wire ICSP lookups to oc image info

soltysh:
This shows https://github.com/openshift/library-go/pull/1084 applied to `oc image info`

/assign @smarterclayton 

/hold
for real bump


pkg/cli/image/strategy/simplelookup.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+	}
+	uniqueMirrors := make([]reference.DockerImageReference, 0, len(imageSources))
+	uniqueMap := make(map[reference.DockerImageReference]bool)
+	for _, imageSourceMirror := range imageSources {
+		if _, ok := uniqueMap[imageSourceMirror]; !ok {
+			uniqueMap[imageSourceMirror] = true
+			uniqueMirrors = append(uniqueMirrors, imageSourceMirror)
+		}
+	}
```
mtrmac, line 125: This does not follow https://github.com/openshift/api/blob/a99ffa1cac6709edf8f502b16890b16f9a557e00/operator/v1alpha1/types_image_content_source_policy.go#L34-L39 .

If at all possible, this should share the implementation in https://github.com/openshift/runtime-utils/blob/master/pkg/registries/registries.go (used e.g. in machine-config-* to generate CRI-O configuration from ICSP), and perhaps even the consumer from https://github.com/containers/image/blob/97b3ffa7bb92a7778238b58010454031b0c2cbee/pkg/sysregistriesv2/system_registries_v2.go#L815 + https://github.com/containers/image/blob/97b3ffa7bb92a7778238b58010454031b0c2cbee/pkg/sysregistriesv2/system_registries_v2.go#L130 , OTOH that would require API additions to make it usable with in-memory configs) instead of maintaining an independent implementation. There are RFEs to add hostname wildcards and the like to the configuration; of course it’s possible to maintain independent implementations but there are enough corner cases that sharing an implementation is the easiest way to remain consistent.

---

(I haven’t read the code in detail, but the behavior of ICSP is not “fallback alternatives”; the mirrors, if any are tried _before_ the primary location, e.g. so that in a disconnected environment there isn’t a firewall/DNS noise, and delays, for attempts to reach the global internet for images that are available on mirrors.)

pkg/cli/image/strategy/simplelookup.go
```diff
@@ -0,0 +1,166 @@
...
+	klog.V(5).Infof("%v ImageReference added to potential ImageSourcePrefixes from ImageContentSourcePolicy", imageRef.AsRepository())
+	for _, icsp := range icspList {
+		repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
+		for _, rdm := range repoDigestMirrors {
+			var err error
+			rdmSourceRef, err := reference.Parse(rdm.Source)
+			if err != nil {
+				return nil, err
+			}
+			if imageRef.AsRepository() != rdmSourceRef.AsRepository() {
```
mtrmac, line 104: The way ICSP actually works when pulling, configuring {Source: `quay.io/foo`, Mirrors: `example.local/mirror-of-quay-foo`} will also match `quay.io/foo/bar` and rewrite to `example.local/mirror-of-quay-foo/bar`.

djzager:
Just providing the link. This PR is blocked by https://github.com/openshift/library-go/pull/1084

pkg/cli/image/strategy/simplelookup.go
smarterclayton: this is intentional, oc is not a cluster.  For cli commands not tied to the behavior of a cluster the appropriate behavior is fallback.  Not the least of which is that ICSP defines alternatives,  and the tradeoffs different clients make drives different benefits.  A cli is low latency on the happy path.

mtrmac: (To be explicit, my _primary_ concern is about sharing the implementation: the primary vs. mirrors ordering is far less important to me.)

If the user wants to hit the primary location directly, it’s trivial not to submit an ICSP to `oc`; all of this is only invoked on explicit user action to involve ICSP.

The linked bug also motivates this with a situation where the primary location is not available, and even suggests that `oc extract` is used during cluster installation. Both of these would benefit from using the mirrors first.

The `--cluster-icsp` option documentation also says:
> honor the ordering of those sources

pkg/cli/image/info/info.go
```diff
@@ -78,6 +80,9 @@ func NewInfo(streams genericclioptions.IOStreams) *cobra.Command {
 	o.SecurityOptions.Bind(flags)
 	flags.StringVarP(&o.Output, "output", "o", o.Output, "Print the image in an alternative format: json")
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be read from.")
+	flags.BoolVar(&o.ClusterICSP, "cluster-icsp", o.ClusterICSP, "When set to true, look for alternative image sources from ImageContentSourcePolicy objects in cluster, honor the ordering of those sources, and fail if an ImageContentSourcePolicy is not found in cluster.")
```
smarterclayton, line 83: I don't like that this fails if there is no ICSP.  As a user, if I want ICSP to be used, I'll specify the flag in scripts and tolerate when a cluster doesn't have an ICSP.  The flag should be `use-cluster-icsp`.  There's no real disadvantage for a user using an ICSP, or there not being one on some clusters.   If we had an explicit ICSP locator flag, then I would agree we would fail.

pkg/cli/image/info/info.go
```diff
@@ -78,6 +80,9 @@ func NewInfo(streams genericclioptions.IOStreams) *cobra.Command {
...
+	flags.StringVar(&o.ICSPFile, "icsp-file", o.ICSPFile, "Path to an ImageContentSourcePolicy file.  If set, data from this file will be used to set alternative image sources.")
```
smarterclayton, line 84: > data from this file will be used to find alternative locations for images.

pkg/cli/image/info/info.go
```diff
@@ -116,6 +136,10 @@ func (o *InfoOptions) Run() error {
 	if err != nil {
 		return err
 	}
+	if len(o.ICSPFile) > 0 || o.ClusterICSP {
+		registryContext = registryContext.WithAlternateBlobSourceStrategy(
+			strategy.NewSimpleLookupICSPStrategy(o.ICSPFile, o.operatorClient.ImageContentSourcePolicies()))
```
smarterclayton, line 141: I'd prefer to have two strategies - one for file (`NewExplicitICSPStrategy`) and one for cluster lookup (`NewICSPFromClusterOnError`).

pkg/cli/image/strategy/simplelookup.go
smarterclayton: I expect a user to provide `--use-cluster-icsp` in scripting, and for it not to fail if one exists, and for us to "do the right thing for the user".  Since ICSP lookup on some commands (anything not doing bulk retrieval) is better suited for latency, I think we're reserving the right to "do the right thing" by underspecifying.

The clause `honor the ordering` should be removed from the flag.

smarterclayton:
Some things we missed (and need to fix before this is correct):

1. If PING fails against the registry

```
I0602 11:37:32.225230   24632 round_trippers.go:432] GET https://registry-1.docker.io/v2/
I0602 11:37:32.225235   24632 round_trippers.go:438] Request Headers:
I0602 11:37:47.226765   24632 round_trippers.go:457] Response Status:  in 15001 milliseconds
I0602 11:37:47.226796   24632 round_trippers.go:460] Response Headers:
I0602 11:37:47.226862   24632 workqueue.go:143] about to send work queue error: unable to read image docker.io/openshift/origin-cli@sha256:d0795436a770d7ad2b7b01c92f551f01b378d472f6034d2f3a97ed957125ef3d: Get "https://registry-1.docker.io/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

we don't fail over correctly (we just exit).  This is a bug in the Repository ping behavior and is going to require changes to library-go

2. If Auth fails against the repository (getting the token) because the current user doesn't have access

```
I0602 11:31:41.622119   24437 round_trippers.go:432] GET https://registry-1.docker.io/v2/foo/bar/manifests/sha256:d0795436a770d7ad2b7b01c92f551f01b378d472f6034d2f3a97ed957125ef3d
I0602 11:31:41.622151   24437 round_trippers.go:438] Request Headers:
I0602 11:31:41.622164   24437 round_trippers.go:442]     Accept: application/vnd.docker.distribution.manifest.list.v2+json
I0602 11:31:41.622175   24437 round_trippers.go:442]     Accept: application/vnd.docker.distribution.manifest.v2+json
I0602 11:31:41.622186   24437 round_trippers.go:442]     Accept: application/vnd.oci.image.manifest.v1+json
I0602 11:31:41.622198   24437 round_trippers.go:442]     Authorization: Bearer <masked>
I0602 11:31:41.972523   24437 round_trippers.go:457] Response Status: 401 Unauthorized in 350 milliseconds
```

we don't fail over.

For both of these we need to move "ping" into the actual repository loader lazily, which I believe was an early comment on the PR and I forgot to bring it back up.  I think it's potentially ok if our test impl doesn't have it, but it will absolutely block 'oc' being used with ICSP to solve the original bugs here (where you are running oc in a disabled context).  I'll suggest a fix later.

smarterclayton:
Tested this locally, didn't seem to work:

```
$ oc image info docker.io/openshift/origin-cli@sha256:d0795436a770d7ad2b7b01c92f551f01b378d472f6034d2f3a97ed957125ef3d -v=8 --icsp-file=contrib/icsp.yaml

I0602 11:42:11.471694   24967 client_mirrored.go:372] get manifest for sha256:d0795436a770d7ad2b7b01c92f551f01b378d472f6034d2f3a97ed957125ef3d served from registryclient.retryManifest{ManifestService:registryclient.manifestServiceVerifier{ManifestService:(*client.manifests)(0xc000e04c30)}, repo:(*registryclient.retryRepository)(0xc000e6bf00)}: manifest unknown: manifest unknown
```

Did not try the alternate at all:

```
spec:
  repositoryDigestMirrors:
  - source: docker.io/openshift/origin-cli
    mirrors:
    - quay.io/openshift/origin-cli
```

Also, did this PR add extra logging to the client stack?  Seeing this:

```
○ oc explain imagecontentsourcepolicy.spec.repositoryDigestMirrors
E0602 11:44:33.296923   25032 request.go:1027] Unexpected error when reading response body: context deadline exceeded (Client.Timeout or context cancellation while reading body)
error: unexpected error when reading response body. Please retry. Original error: context deadline exceeded (Client.Timeout or context cancellation while reading body)
```

We should not be printing that...

smarterclayton:
Not sure why docker.io wasn't selected as an alternate, but it could be because of the legacy lookup behavior (in which case that's a workaround that should be solved when the ICSP is loaded

smarterclayton:
Failover did work, but didn't prefer the calculated ordering on subsequent calls.  

```
I0602 11:48:42.299108   25197 client_mirrored.go:372] get manifest for sha256:d0795436a770d7ad2b7b01c92f551f01b378d472f6034d2f3a97ed957125ef3d served from registryclient.retryManifest{ManifestService:registryclient.manifestServiceVerifier{ManifestService:(*client.manifests)(0xc000ae7350)}, repo:(*registryclient.retryRepository)(0xc000e62600)}: name unknown: repository name not known to registry
I0602 11:48:42.299135   25197 client_mirrored.go:166] Attempting to connect to quay.io/openshift/origin-cli
...
I0602 11:48:43.470154   25197 client_mirrored.go:372] get manifest for sha256:d0795436a770d7ad2b7b01c92f551f01b378d472f6034d2f3a97ed957125ef3d served from registryclient.retryManifest{ManifestService:registryclient.manifestServiceVerifier{ManifestService:(*client.manifests)(0xc0002450e0)}, repo:(*registryclient.retryRepository)(0xc000e62c00)}: <nil>
...
now tries to load the blob:
...
I0602 11:48:43.470236   25197 client_mirrored.go:166] Attempting to connect to quay.io/openshift/origin-test
^ the blob lookup should have been shared with the manifest lookup, didn't expect to go back to the source
```

It's possible the error lookup strategy should be returning an order that prefers the alternate first (if you get an error, on subsequent calls change the order).  But need to think about it a bit.

pkg/cli/image/strategy/simplelookup.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// OnFailure returns a list of possible image references for the locator image reference, gathered from image content source policies, or an error
+func (s *simpleLookupICSP) OnFailure(ctx context.Context, locator reference.DockerImageReference) (alternateRepositories []reference.DockerImageReference, err error) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	if err := s.resolve(ctx, locator); err != nil {
+		return nil, err
+	}
+	if len(s.alternates) == 0 {
```
smarterclayton, line 49: Having the resolve method set a global is weird (just return the list).  Also, why is this a list?  If two calls are made to the same strategy with different locators, then the strategy has to get a new alternates list.  So if you want to cache, you have to cache by locator, and you probably need to maintain the size of the cache (at least bound it to something high like 1024 entries).

In general, for this strategy you want to add the locator as the last option, not the first (also if it's not duplicated) when you're handling the "on error" strategy.

pkg/cli/image/strategy/simplelookup.go
soltysh: > The clause `honor the ordering` should be removed from the flag.

Removed

pkg/cli/image/info/info.go
soltysh: Roger that.

pkg/cli/image/info/info.go
soltysh: Updated.

pkg/cli/image/info/info.go
soltysh: Although we do fail when you specify a file with `--icsp-file` and we have issues reading the file.

pkg/cli/image/info/info.go
soltysh: Should both of them have similar behavior or they will behave differently, as in:
- `NewExplicitICSPStrategy` reads always first from alternative (just like Miroslav is saying)
- `NewICSPFromClusterOnError` is similar to current `SimpleLookup` in that it fails first and tries alternates only then?

soltysh:
> Also, did this PR add extra logging to the client stack? Seeing this:

This might have been introduced in https://github.com/kubernetes/kubernetes/pull/102217

>     If PING fails against the registry

Yeah, I've noticed that when testing against missing registry. 

>     If Auth fails against the repository (getting the token) because the current user doesn't have access

I guess we want a similar behavior as above. 


> Tested this locally, didn't seem to work:

It will test alternates only if it fails, not by default, although it should read the ICSP in.

pkg/cli/image/strategy/simplelookup.go
soltysh: Fixed. 

> In general, for this strategy you want to add the locator as the last option, not the first (also if it's not duplicated) when you're handling the "on error" strategy.

I'm not quite sure I understand what you mean by that?

soltysh:
> Not sure why docker.io wasn't selected as an alternate, but it could be because of the legacy lookup behavior (in which case that's a workaround that should be solved when the ICSP is loaded

Yeah, `docker.io` needs to be translated to `registry-1.docker.io` we have `AsV2` in the reference.

soltysh:
We need https://github.com/openshift/library-go/pull/1096 too to land this

pkg/cli/image/strategy/simplelookup.go
soltysh: I've added a TODO there, we'll tackle this in the next revisions.

soltysh:
@smarterclayton split into 2 separate strategies:
- `NewICSPOnErrorStrategy` which falls back to ICSP only after getting an error, that's what's being used in `oc image info` 
- `NewICSPExplicitStrategy` which will be for `oc adm release` commands
I've added TODO from @mtrmac about supporting partial matches.

This should be ready for final review.

jparrill:
Hey folks, this give me 2 issues:

- First, the GLIBC version on the Downstream bits is set for `glibc-2.28-151.el8.x86_64` and this oc client needs at least `glibc-2.33-8.fc34.x86_64` so if you execute this on an image based on RHEL8 or UBI will fail:

```
sh-4.4$ /tmp/oc version                                                                                                                                                                                                                                                         
/tmp/oc: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/oc)
/tmp/oc: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/oc)
```

**NOTE:** This works fine in Fedora

- Second, I've tried this on a disconnected environment and it still tries to reach the external registry:

```
bash-5.1# /tmp/oc adm release extract -v=4 --registry-config=./lol --command=openshift-baremetal-install --to=. xenomorph.localdomain:5000/ocp4/release:4.8.0-fc.8-x86_64 

I0610 08:57:05.575739     261 config.go:128] looking for config.json at /root/.docker/config.json
I0610 08:57:05.575840     261 config.go:94] looking for .dockercfg at /root/.dockercfg
I0610 08:57:05.641289     261 manifest.go:310] Raw image config json:
...
...
...
I0610 08:57:05.683333     261 info.go:802] Found manifest 0000_50_service-ca-operator_04_sa.yaml
I0610 08:57:05.683342     261 info.go:802] Found manifest 0000_50_service-ca-operator_05_deploy-ibm-cloud-managed.yaml
I0610 08:57:05.683350     261 info.go:802] Found manifest 0000_50_service-ca-operator_05_deploy.yaml
I0610 08:57:05.683361     261 info.go:802] Found manifest 0000_50_service-ca-operator_07_clusteroperator.yaml
I0610 08:57:05.683390     261 workqueue.go:143] about to send work queue error: <nil>
I0610 08:57:05.683431     261 workqueue.go:54] worker 0 stopping
I0610 08:57:05.683488     261 workqueue.go:60] work queue exiting
I0610 08:57:05.683500     261 extract_tools.go:323] Will extract usr/bin/openshift-install from quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5b1c59f1139438f2b899569850d7664d59f7199d57c486e8a9264ec7cc52d4b1
I0610 08:57:05.683559     261 config.go:128] looking for config.json at /root/.docker/config.json
I0610 08:57:05.683580     261 config.go:94] looking for .dockercfg at /root/.dockercfg
I0610 08:57:06.342546     261 workqueue.go:143] about to send work queue error: unable to read image quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5b1c59f1139438f2b899569850d7664d59f7199d57c486e8a9264ec7cc52d4b1: unauthorized: access to the requested resource is not authorized
I0610 08:57:06.342580     261 workqueue.go:54] worker 0 stopping
error: unable to read image quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5b1c59f1139438f2b899569850d7664d59f7199d57c486e8a9264ec7cc52d4b1: unauthorized: access to the requested resource is not authorized
```

And the flag you mention of `--icsp-file` is not there. The implementation should be done into `oc adm release extract` at least.

regards

soltysh:
>     * First, the GLIBC version on the Downstream bits is set for `glibc-2.28-151.el8.x86_64` and this oc client needs at least `glibc-2.33-8.fc34.x86_64` so if you execute this on an image based on RHEL8 or UBI will fail:

The problem is where you compile this, there is a problem with rhel7 vs rhel8 due to glibc versions mismatch, that's why we build it on rhel7 for now. 

>     * Second, I've tried this on a disconnected environment (using fedora container image to workaround the first issue) and it still tries to reach the external registry:

Correct, `oc image info` by default will try original location, and fallback to mirrors after getting an error. Although, there are some problems when ping from original registry doesn't work, but that's being worked on.

pkg/cli/image/info/info.go
```diff
@@ -87,35 +91,67 @@ type InfoOptions struct {
...
-	return nil
-}
 
-func (o *InfoOptions) Validate() error {
-	return o.FilterOptions.Validate()
+	clientConfig, err := f.ToRESTConfig()
+	if err != nil {
+		return err
+	}
+	o.operatorClient, err = operatorv1alpha1client.NewForConfig(clientConfig)
```
smarterclayton, line 112: This has to be lazy - we don't want to fail an `oc image` command because we couldn't create a client.  So you have to pass a `ToRESTConfig` method into your strategy.

smarterclayton: Actually, if you're not defaulting to the cluster strategy, you don't need this anymore, so you can remove.

pkg/cli/image/info/info.go
```diff
@@ -87,35 +91,67 @@ type InfoOptions struct {
...
 	if len(o.Images) == 0 {
 		return fmt.Errorf("must specify one or more images as arguments")
 	}
+	if len(o.ICSPFile) > 0 {
+		warned := false
+		for _, location := range o.Images {
+			if warned {
+				break
+			}
+			sources, err := imagesource.ParseSourceReference(location, nil)
```
smarterclayton, line 130: You have to handle wildcard expansion here.  Do this inline in Run() (otherwise you have to lift the entire loop out of Run which is a major refactoring (and likely to break other stuff).

pkg/cli/image/strategy/explicit.go
```diff
@@ -0,0 +1,81 @@
...
+	icspList, err := s.readICSPsFromFileFunc(s.icspFile)
+	if err != nil {
+		return nil, err
+	}
+	imageRefList, err := alternativeImageSources(imageRef, icspList)
+	if err != nil {
+		return nil, err
+	}
+	// always add the original as the last reference
+	imageRefList = append(imageRefList, imageRef.AsRepository())
```
smarterclayton, line 79: The strategy doesn't do duplicate protection, you might have to verify that you're not double adding the original.

pkg/cli/image/strategy/onerror.go
```diff
@@ -0,0 +1,143 @@
...
+		repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
+		for _, rdm := range repoDigestMirrors {
+			var err error
+			rdmSourceRef, err := reference.Parse(rdm.Source)
+			if err != nil {
+				return nil, err
+			}
+			// AsV2 in the right call is required to ensure we transform docker registry
+			// from docker.io to registry-1.docker.io
+			// TODO: support partial match, iow. quay.io/foo should match quay.io/foo/bar
```
smarterclayton, line 94: Hrm, that wouldn't be correct.  quay.io/foo is simply incorrect as quay.io/foo is not a repository.  Remove this TODO.

pkg/cli/image/strategy/onerror.go
```diff
@@ -0,0 +1,143 @@
...
+			if imageRef.AsRepository() != rdmSourceRef.AsRepository().AsV2() {
```
smarterclayton, line 95: This seems incorrect, I expected all refs used in the strategy to be `.AsV2()` which then normalizes the alternative list.

pkg/cli/image/strategy/explicit.go
smarterclayton: Actually, `alternativeImageSources` does duplicate protection, so you need to add a boolean to it `addSourceAsLastAlternate bool` and pass true above and false in onerror.  You'd also need to normalize imageref to v2.

pkg/cli/image/strategy/explicit_test.go
```diff
@@ -0,0 +1,221 @@
...
+			imageSourcesExpected: []string{"someregistry/mirrors/match", "quay.io/another/release", "quay.io/andanother/release", "quay.io/ocp-test/release"},
+		},
+		{
+			name: "docker.io vs registry-1.docker.io",
+			icspList: []operatorv1alpha1.ImageContentSourcePolicy{
+				{
+					Spec: operatorv1alpha1.ImageContentSourcePolicySpec{
+						RepositoryDigestMirrors: []operatorv1alpha1.RepositoryDigestMirrors{
+							{
+								Source: "docker.io/ocp-test/release",
```
smarterclayton, line 154: Add a test case where source mirrors include both `index.docker.io` and `registry-1.docker.io`.

pkg/cli/image/strategy/onerror.go
soltysh: Changed all to `.AsV2()`

eranco74:
Any ETA for this PR?
I'm not sure what's holding it back.

smarterclayton:
lgtm

mtrmac:
(meanwhile, https://github.com/openshift/api/pull/874 has added two new CRDs representing mirrors — see https://github.com/openshift/runtime-utils/pull/15 and https://github.com/openshift/machine-config-operator/pull/3037 )

soltysh:
> (meanwhile, [openshift/api#874](https://github.com/openshift/api/pull/874) has added two new CRDs representing mirrors — see [openshift/runtime-utils#15](https://github.com/openshift/runtime-utils/pull/15) and [openshift/machine-config-operator#3037](https://github.com/openshift/machine-config-operator/pull/3037) )

I'll update that in a followup, since we will want to support both. 
/hold cancel

pkg/cli/image/info/info.go
```diff
@@ -132,6 +138,10 @@ func (o *InfoOptions) Run() error {
 			if len(src.Ref.Tag) == 0 && len(src.Ref.ID) == 0 {
 				return fmt.Errorf("--from must point to an image ID or image tag")
 			}
+			if !icspWarned && len(o.ICSPFile) > 0 && len(src.Ref.Tag) > 0 {
+				fmt.Fprintf(o.ErrOut, "warning: --icsp-file only applies to images referenced by digest and will be ignored for tags\n")
```
eranco74, line 142: Why it will be ignored for tags?
Our (assisted-service) users are using tags https://github.com/openshift/hive/blob/master/docs/using-hive.md#openshift-version

PR #831, opened 2021-05-21
Bug 1949050: Bump k8s to 1.21

soltysh:


soltysh:
is not related with this change, and it's already being worked on elsewhere

PR #833, opened 2021-05-24
Bug 1960758: use recent pull spec for must-gather and debug

soltysh:
/assign @smarterclayton 

This builds on top of methods introduced in https://github.com/openshift/library-go/pull/1078/

/cc @dmage @ricardomaraschini 
since you two will be also interested in usage

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -210,9 +210,8 @@ func (o *MustGatherOptions) resolveImageStreamTag(namespace, name, tag string) (
 		return "", err
 	}
 	var image string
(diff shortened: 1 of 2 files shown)


smarterclayton:
/lgtm

The failures in metal-ipi indicate there's another bug, but the core bug is fixed (we pull the image from the payload)

/retest

smarterclayton:

We need to fix https://github.com/openshift/origin/blob/b8eab051c232f3a2909aefa1d0311ffe3d2d232a/test/extended/cli/mustgather.go#L356 to use the same method, @soltysh will do that

wking:
openshift/origin#26185 has the test-side fix.

PR #835, opened 2021-05-26
Bug 1965330: exclude security during exctraction

soltysh:


pkg/cli/image/archive/archive.go
```diff
@@ -391,6 +391,11 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 
 	var errors []string
 	for key, value := range hdr.Xattrs {
+		// exclude these capabilities, since they are also omitted during
```
soltysh, line 394: Note to self: we should probably try setting those and if we fail just skip it and warn users about it, but don't break.

PR #837, opened 2021-05-27
Re-use DialContext and Proxy configuration from original http.Transport

guillaumerose:
crc uses tokencmd package to generate the kubeconfig directly after the VM is started.
To avoid DNS issues on the host, we force the http.Transport to use the IP of the VM by passing a function to DialContext (see below).

It appears that, for root CAs verification, a http.Client is built from scratch and doesn't use the given http.Transport. 
Since it doesn't use the custom DialContext function, a DNS resolution is performed in my case.

```go
	tokencmd.RequestToken(&restclient.Config{
		Transport: &http.Transport{
			DialContext: func(ctx context.Context, network, address string) (net.Conn, error) {
				port := strings.SplitN(address, ":", 2)[1]
				dialer := net.Dialer{
					Timeout:   30 * time.Second,
					KeepAlive: 30 * time.Second,
				}
				return dialer.Dial(network, fmt.Sprintf("127.0.0.1:%s", port))
			},
		},
	}, nil, username, password)
```

### DIFF
--- a/pkg/helpers/tokencmd/request_token.go
+++ b/pkg/helpers/tokencmd/request_token.go
@@ -427,7 +427,7 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 	// perform the retrieval with insecure transport, otherwise oauth-server
 	// logs remote tls error which is confusing during troubleshooting
 	client := http.Client{
-		Transport: insecureTransport(issuerURL),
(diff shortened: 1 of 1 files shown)


cfergeau:
This is causing some issues in `crc` for customers who are behind a proxy, could this PR be looked at?

PR #821, opened 2021-05-07
Fix unit test failure

ravisantoshgudimetla:
Seems like repo update error

PR #838, opened 2021-06-02
[release-4.7] Fix unit test failure

soltysh:
I couldn't trigger automatic cherry-pick, so this is a manual one from https://github.com/openshift/oc/pull/821 which should fix units in 4.7 branch.


soltysh:
Adding valid-bug label, since this is unblocking the queue.

derekwaynecarr:
(Patch-Manager) approved to give test signal.

Issue #641, opened 2020-11-13
[RFE] oc debug node: Use `unconfined_t` as SELinux context for debug container


travier:
The SELinux policyhas been changed to allow container runtimes to create unconfined container for host debugging ([BZ#1839065](https://bugzilla.redhat.com/show_bug.cgi?id=1839065)) instead of using the `spc_t` semi-restricted context that leads to weird error cases ([BZ#1896369](https://bugzilla.redhat.com/show_bug.cgi?id=1896369)).

This is released with [RHSA-2020:3053](https://access.redhat.com/errata/RHSA-2020:3053).

For OCP, this can be enabled starting with 4.6 as it includes the fixed SELinux policy in the release image.
For OKD, this can be enabled starting with 4.6 as Fedora CoreOS has included the fix for a while.

travier:
We encountered this issue again in https://bugzilla.redhat.com/show_bug.cgi?id=1924926. Can I get an initial review? Should I start working on it?

travier:
/remove-lifecycle stale
/lifecycle frozen
We really should be doing that at some point or figuring out a way out (via an SSH proxy for example) so marking as frozen.

codespearhead:
According to [Daniel Walsh 2020-05-29 17:41:48 UTC](https://bugzilla.redhat.com/show_bug.cgi?id=1839065#c7), the source problem was fixed in [container-selinux v2.135.0](https://github.com/containers/container-selinux/releases/tag/v2.135.0).

Should this PR be closed then?

travier:
The last time I check, it wasn't fixed. I had made: https://github.com/openshift/oc/pull/842

PR #842, opened 2021-06-08
OCPBUGS-10615: debug node: Run under an unconfined SELinux context

travier:
Run the node debug pod under an unconfined SELinux context
(unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023) to avoid issues
with the `spc_t` context.

Fixes: https://github.com/openshift/oc/issues/641
See also:
 - https://bugzilla.redhat.com/show_bug.cgi?id=1839065
 - https://bugzilla.redhat.com/show_bug.cgi?id=1896369

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -1045,6 +1045,12 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 						SecurityContext: &corev1.SecurityContext{
 							Privileged: &isTrue,
 							RunAsUser:  &zero,
+							SELinuxOptions: &corev1.SELinuxOptions{
+								User:  "unconfined_u",
+								Role:  "unconfined_r",
+								Type:  "unconfined_t",
+								Level: "s0-s0:c0.c1023",
+							},
 						},
 						VolumeMounts: []corev1.VolumeMount{
 							{


travier:
Looks like this is effectively requesting an unconfined pod but does not result in one:

```
$ ./oc debug node/ip-10-0-140-177.us-east-2.compute.internal -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    debug.openshift.io/source-container: container-00
    debug.openshift.io/source-resource: /v1, Resource=nodes/ip-10-0-140-177.us-east-2.compute.internal
  creationTimestamp: null
  name: ip-10-0-140-177us-east-2computeinternal-debug
  namespace: default
spec:
  containers:
  - command:
    - /bin/sh
    image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ac752e15c3b5a557a8c4e12d02279ab97f531c56381b92f75cbc0c68205f561e
    name: container-00
    resources: {}
    securityContext:
      privileged: true
      runAsUser: 0
      seLinuxOptions:
        role: unconfined_r
        type: unconfined_t
        user: unconfined_u
    stdin: true
    stdinOnce: true
    tty: true
    volumeMounts:
    - mountPath: /host
      name: host
  hostNetwork: true
  hostPID: true
  nodeName: ip-10-0-140-177.us-east-2.compute.internal
  restartPolicy: Never
  volumes:
  - hostPath:
      path: /
      type: Directory
    name: host
status: {}
```

Still running as `spec_t` and nothing stands out in the logs so far. Investigating.

cgwalters:
Thanks for working on this!  One tricky aspect is we likely want to handle older cluster versions that have an older selinux-policy where this will fail.

cgwalters:
One alternative flow I'd like to enable is something like typing `hostdo` in the debug shell switches you into something that should feel absolutely indistinguishable from `ssh`.  (For example, we do run through PAM, you become `unconfined_t`, you're actually in the host mount namespace, etc.)  We could actually implement this just in the support-tools container without changing `oc`.

travier:
Another potential idea here is to make an option for `oc debug node` to only be a proxy from the local ssh client to the remote node ssh server and then let SSH, PAM, SSSD, etc. handle everything else, just like on a classic system.

travier:
/remove-lifecycle stale
/lifecycle frozen

I would still be interesting to do that one. I need to figure out what's blocking it.

travier:
See https://bugzilla.redhat.com/show_bug.cgi?id=2180937.

sdodson:
/retitle OCPBUGS-10615: debug node: Run under an unconfined SELinux context

travier:
Note that I'm not sure about what the tests here exercise in functionality so I'm not sure this change is tested.

travier:
This does not work for me but I'm not sure how to debug this:

```
$ ./oc debug node/ip-10-0-139-210.ec2.internal
Temporary namespace openshift-debug-hmvzf is created for debugging node...
Starting pod/ip-10-0-139-210ec2internal-debug ...
Pod IP: 10.0.139.210
sh-4.4# chroot /host
sh-5.1# id -Z
system_u:system_r:spc_t:s0
sh-5.1#
```

saschagrunert (commented):
I think setting `unconfined_t` is not the right choice here, the container already runs privileged so it should run as `spc_t`.

atiratree:
It seems that the `SELinuxOptions` are effectively skipped by cri-o when we use HostPID, HostIPC, HostNetwork in our debug pod (https://github.com/openshift/origin/issues/25488#issuecomment-864410067)

https://github.com/cri-o/cri-o/issues/5501 and https://github.com/openshift/origin/issues/25488 summarize the issue nicely and the cri-o PR has a discussion about possible future solution.

This PR does not solve the issue as it is still reproducible with these changes. It seems this should be addressed on the cri-o side first.
/hold

saschagrunert:
@atiratree I opened a PR for discussion a possible solution in CRI-O: https://github.com/cri-o/cri-o/pull/6780

rhatdan:
Setting the labels should be allowed always. The namespaces should have no effect. I don't know why looking at a share network namespace is ever a problem, but the IPC and pid would potentially cause issues with a confined container process not being able to see or work with a spc_t or unconfined_t process.

PR #851, opened 2021-06-15
Bug 1972962: Add validation for max-icsp-size flag

awgreene:
Problem: It is possible to set a --max-icsp-size for the oc adm
catalog mirror command that is too small to create a valid icsp.

Solution: If a user sets the max icsp size to less than 1, return
an error. If a single mirror entry exceeds the length of the
provided icsp size limit, return an error.


dinhxuanvu (commented):
/lgtm

dinhxuanvu (commented):
/lgtm

PR #852, opened 2021-06-15
Support ibmcloud provider in release extract

BobbyRadford:
This PR enables "ibmcloud" to be specified as part of the cloud argument to `oc adm release extract". An example would look like this:
```sh
oc adm release extract --credentials-requests --cloud=ibmcloud --to=./credrequests <RELEASE_IMAGE>
```

This is related to this CCO PR: https://github.com/openshift/cloud-credential-operator/pull/356 
...and this enhancement: https://github.com/openshift/enhancements/pull/773


soltysh (approved):
/lgtm
/approve

PR #869, opened 2021-06-25
[release-4.7] Bug 1976284: skip-multiple-scopes with adm catalog mirror

soltysh:
Manually picking part of https://github.com/openshift/oc/pull/780

'oc adm catalog mirror' often fails with server error
'414 Request-URI Too Large' This is because mirroring the
catalog pushes a lot of content in parallel. To avoid this,
set the mirroring flag --skip-multiple-scopes to true.

/assign @damemi


damemi (commented):
/lgtm

PR #878, opened 2021-07-07
contrib/completions/OWNERS: Delegate to all approver aliases

wking:
So folks can touch commands/options within their sections and not need to reach out to root apporovers.  To make it convenient to continue to include the root approvers here, I've shifted them into a new alias.


soltysh (approved):
/lgtm
/approve

Issue #879, opened 2021-07-12
`oc debug pod --as-root` fails with `Error: container's runAsUser breaks non-root policy` when `.spec.securityContext.runAsNonRoot` is true


orimanabu:
I ran `oc debug pod --as-root` with cluster-network-operator, then got `CreateContainerConfigError`.
CNO Pod has `.spec.securityContext.runAsNonRoot: true`, then should we set it nil?

- How to reproduce

Run `oc debug --as-root` with cluster-network-operator.

```
$ oc -n openshift-network-operator debug pod/network-operator-7855cfd575-wmrlz --as-root --dry-run=client -o yaml | oc apply -f -
$ oc -n openshift-network-operator get pod
NAME                                      READY   STATUS                       RESTARTS   AGE
network-operator-7855cfd575-wmrlz         1/1     Running                      0          29d
network-operator-7855cfd575-wmrlz-debug   0/1     CreateContainerConfigError   0          5s
$ oc -n openshift-network-operator describe pod/network-operator-7855cfd575-wmrlz-debug | tail -n 5
Events:
  Type     Reason  Age                    From     Message
  ----     ------  ----                   ----     -------
  Warning  Failed  2m55s (x12 over 5m8s)  kubelet  Error: container's runAsUser breaks non-root policy
  Normal   Pulled  6s (x25 over 5m8s)     kubelet  Container image "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:79e60721e7b5362d4de26275281ca8b7208fef8a0a6830bab477b798c84faa3f" already present on machine
```

PR #880, opened 2021-07-12
runAsNonRoot in Pod SCC should be nil when oc debug run with `--as-root` 

orimanabu:
When running oc debug with `--as-root`, Pod with `.spec.securityContext.runAsNonRoot: true` fails with `Error: container's runAsUser breaks non-root policy`.
See also: #879

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -799,6 +799,13 @@ func (o *DebugOptions) transformPodForDebug(annotations map[string]string) (*cor
 		container.SecurityContext.RunAsNonRoot = nil
 	}
 
+	// if DebugOptions set container.SecurityContext.RunAsNonRoot to nil,
+	// pod.Spec.SecurityContext.runAsNonRoot should be nil also.
+	if container.SecurityContext.RunAsNonRoot == nil &&
+		pod.Spec.SecurityContext != nil {
+		pod.Spec.SecurityContext.RunAsNonRoot = nil
+	}
+
 	switch {
 	case o.OneContainer:
 		pod.Spec.InitContainers = nil


pkg/cli/debug/debug.go
```diff
@@ -785,6 +785,7 @@ func (o *DebugOptions) transformPodForDebug(annotations map[string]string) (*cor
 		zero := int64(0)
 		container.SecurityContext.RunAsUser = &zero
 		container.SecurityContext.RunAsNonRoot = nil
+		pod.Spec.SecurityContext.RunAsNonRoot = nil
```
soltysh, line 788: According to docs either this or the above should be set as such. Can you verify if both are present which takes precedence, it's possible that we'll need to apply this rule not just here but in the accompanying places accordingly.

soltysh: When updating please add a comment explaining why we're doing so.

soltysh (changes requested):

pkg/cli/debug/debug.go
orimanabu: According to doc, if set in both Container SecurityContext and Pod SecurityContext, the value specified in Container SecurityContext takes precedence.
So we need to take care of cases where Pod SecurityContext has some value and `oc debug` set Container SecurityContext to nil.

soltysh (approved):
/lgtm
/approve
/ok-to-test

soltysh:
/hold
looks like the failure from https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/880/pull-ci-openshift-oc-master-e2e-aws/1445324381986230272 is legit @orimanabu can you address that

orimanabu:
Sure, looking...

PR #881, opened 2021-07-12
Make sure to set the namespace for routes resources

tareqmamari:
When doing dry run, the namespace will not be part of the output
manifests ( i.e. yaml or json, etc. ). This commit adds the namespace
to the generated route resource if it is not the default.

How to reproduce it?
```
$ oc create route passthrough namex --service servicex -n namespacex --dry-run=client -o yaml
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/component: identity-provider
    app.kubernetes.io/instance: e2e-test
    app.kubernetes.io/name: namex
    app.kubernetes.io/version: v2.24.0
  name: namex
spec:
  port:
    targetPort: http
  tls:
    termination: passthrough
  to:
    kind: ""
    name: servicex
    weight: null
status: {}
```

### DIFF
--- a/pkg/cli/create/routeedge.go
+++ b/pkg/cli/create/routeedge.go
@@ -99,15 +99,9 @@ func (o *CreateEdgeRouteOptions) Run() error {
 	if err != nil {
 		return err
 	}
-	route, err := route.UnsecuredRoute(o.CreateRouteSubcommandOptions.CoreClient, o.CreateRouteSubcommandOptions.Namespace, o.CreateRouteSubcommandOptions.Name, serviceName, o.Port, false)
+	route, err := route.UnsecuredRoute(o.CreateRouteSubcommandOptions.CoreClient, o.CreateRouteSubcommandOptions.Namespace, o.CreateRouteSubcommandOptions.Name, serviceName, o.Port, false, o.CreateRouteSubcommandOptions.EnforceNamespace)
 	if err != nil {
 		return err
-	}
-
-	// If the namespace is not the default, add it. This makes sure it will eventually
-	// appear in the manifest if the dry-run is enabled.
-	if o.CreateRouteSubcommandOptions.EnforceNamespace {
-		route.Namespace = o.CreateRouteSubcommandOptions.Namespace
 	}
 
 	if len(o.WildcardPolicy) > 0 {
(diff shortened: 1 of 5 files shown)


tareqmamari:
@mfojtik @soltysh Hi folks, any thing is needed from our sides to get this reviewed and merged please? Thank you!

pkg/cli/create/routeedge.go
```diff
@@ -104,6 +104,12 @@ func (o *CreateEdgeRouteOptions) Run() error {
 		return err
 	}
 
+	// If the namespace is not the default, add it. This makes sure it will eventually
+	// appear in the manifest if the dry-run is enabled.
+	if o.CreateRouteSubcommandOptions.Namespace != corev1.NamespaceDefault {
```
soltysh, line 109: This is not correct, you should be setting the namespace only when explicitly requested, iow. you should check `o.CreateRouteSubcommandOptions.EnforceNamespace` and set it only then. I'd expect this to happen inside `UnsecuredRoute` method, though. See https://github.com/kubernetes/kubernetes/blob/80b602299ff0b8ae13596ad6d4634aa89cd99697/staging/src/k8s.io/kubectl/pkg/cmd/create/create_job.go#L248-L250 as a good example.

soltysh (changes requested):
/ok-to-test

pkg/cli/create/routeedge.go
tareqmamari: @soltysh I just updated it, it should look better now. Thanks for the comment.

pkg/cli/create/routeedge.go
```diff
@@ -104,6 +104,12 @@ func (o *CreateEdgeRouteOptions) Run() error {
 		return err
 	}
 
+	// If the namespace is not the default, add it. This makes sure it will eventually
+	// appear in the manifest if the dry-run is enabled.
+	if o.CreateRouteSubcommandOptions.EnforceNamespace {
```
soltysh, line 109: Better, but still I mentioned to move this if-clause to `UnsecuredRoute` method invoked above. This way it'll apply to all invocations.

PR #883, opened 2021-07-13
Bug 1957668: show console URL when asking for Authentication

atiratree:
~There is no easy access for not logged in users to the console.~

~We have to guess the console URL, since we are not authenticated yet and we cannot be sure of the real URL (configurability docs: https://docs.openshift.com/container-platform/4.7/web_console/configuring-web-console.html)~

show console URL when asking for Authentication
- the URL will redirect to the real URL
- added check for a server version since this feature is not generally available before OpenShift 4.11 (Kubernetes 1.24)

The output of the command:

```bash
> oc login -u test
Authentication required for https://api.my-cluster0.com:6443 (openshift)
Console URL: https://api.my-cluster0.com:6443/console
Username: test
Password:

```

### DIFF
--- a/pkg/helpers/tokencmd/basicauth.go
+++ b/pkg/helpers/tokencmd/basicauth.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/oc/pkg/helpers/term"
+	"github.com/openshift/oc/pkg/version"
 )
 
 // BasicAuthNoUsernameError when basic authentication challenge handling was attempted
@@ -31,6 +32,9 @@ func NewBasicAuthNoUsernameError() error {
 type BasicChallengeHandler struct {
 	// Host is the server being authenticated to. Used only for displaying messages when prompting for username/password
 	Host string
+
+	// serverVersionRetriever is used for fetching server version
+	serverVersionRetriever version.ServerVersionRetriever
 
 	// Reader is used to prompt for username/password. If nil, no prompting is done
 	Reader io.Reader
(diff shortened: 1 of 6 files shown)


atiratree:
@soltysh 
I am not really sure if this is that good of a solution for such common use case, since we are guessing the URL.

pkg/helpers/tokencmd/basicauth.go
```diff
@@ -87,6 +89,10 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 		} else {
 			fmt.Fprintf(w, "Authentication required for %s\n", c.Host)
 		}
+		consoleURL, err := guessConsoleURL(c.Host)
+		if err == nil {
+			fmt.Fprintf(w, "Default console URL: %s\n", consoleURL)
```
soltysh, line 94: Maybe say it's possibly at that URL, because we allow configuring that through https://github.com/openshift/api/blob/0a6cab75f657a12860a97faaf37ec71cd92530fa/kubecontrolplane/v1/types.go#L36

soltysh (changes requested):

pkg/helpers/tokencmd/basicauth.go
atiratree: I am not sure if that is a good choice, ie `$API_URL/console` . Even though it can be set I have not found this documented anywhere. I am worried that for most of the users this will not work.

My solution is not great either and would prefer not to merge it. Because we are guessing the apps subdomain and the domain  of the cluster is configurable through ingress. So this will not work in many cases as well.

I think the best consistent solution would be for apiserver to detect a console URL by default if not specified , so the user would always get redirected to the right URL.

Thoughts?

atiratree: ping @soltysh

soltysh: @atiratree maybe we should have a hardcoded `$API_URL/console` which will redirect to appropriate console URL as configured. This would allow us to use that, but still would leave users with a functioning mechanism for exposing the URL where they wish. wdyt?

atiratree:
should be replaced by a new approach of redirecting to the console url by API server in https://github.com/openshift/kubernetes/pull/1110

/hold

soltysh (approved):
/lgtm
/approve
/bugzilla refresh
feel free to drop the hold once https://github.com/openshift/kubernetes/pull/1110 merges

atiratree:
this PR still has the older approach and needs to be rewritten to use https://github.com/openshift/kubernetes/pull/1110 so /hold till then

atiratree:
I have updated the implementation to point to the newly implemented api server console URL (`${API_SERVER}/console`).

I have also added version detection since this feature was not generally available before and could cause issues to the users who are connecting to older clusters.  Unfortunately  this can cause issues when connecting to vanilla 1.24+ Kubernetes.    

@soltysh please rereview

soltysh (approved):
/lgtm

PR #884, opened 2021-07-18
Add release version annotation to extracted credentials requests

akhil-rane:
The ccoctl tools needs to know the release version from the credentials
request manifests extracted using `oc adm release extract` in order to
create required infrastructure in the cloud provider.

x-ref: https://issues.redhat.com/browse/CCO-84


akhil-rane:
/hold

CCO team is yet to finalize on this approach

PR #886, opened 2021-07-19
WIP: Example collection of SOSReport using OC tool

MichaelWasher:
NOTE: This is PoC and does not have enough code re-use / testing

### DIFF
--- a/pkg/cli/admin/sosreport/sosreport.go
+++ b/pkg/cli/admin/sosreport/sosreport.go
@@ -19,7 +19,6 @@ import (
 
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	imagev1client "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
-	"github.com/openshift/library-go/pkg/image/imageutil"
 	imagereference "github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/library-go/pkg/operator/resource/retry"
 	"github.com/openshift/oc/pkg/cli/admin/inspect"
@@ -72,13 +71,7 @@ func NewSOSReportCommand(f kcmdutil.Factory, streams genericclioptions.IOStreams
 	}
 
 	cmd.Flags().StringVar(&o.NodeName, "node-name", o.NodeName, "Set a specific node to use - by default a random master will be used")
-	// cmd.Flags().StringSliceVar(&o.Images, "image", o.Images, "Specify a must-gather plugin image to run. If not specified, OpenShift's default must-gather image will be used.")
-	// cmd.Flags().StringSliceVar(&o.ImageStreams, "image-stream", o.ImageStreams, "Specify an image stream (namespace/name:tag) containing a must-gather plugin image to run.")
-	// cmd.Flags().StringVar(&o.DestDir, "dest-dir", o.DestDir, "Set a specific directory on the local machine to write gathered data to.")
-	// cmd.Flags().StringVar(&o.SourceDir, "source-dir", o.SourceDir, "Set the specific directory on the pod copy the gathered data from.")
-	// cmd.Flags().StringVar(&o.timeoutStr, "timeout", "10m", "The length of time to gather data, like 5s, 2m, or 3h, higher than zero. Defaults to 10 minutes.")
-	// cmd.Flags().BoolVar(&o.Keep, "keep", o.Keep, "Do not delete temporary resources when command completes.")
-	cmd.Flags().MarkHidden("keep")
+	cmd.Flags().StringSliceVar(&o.Images, "image", o.Images, "Specify a must-gather plugin image to run. If not specified, OpenShift's default must-gather image will be used.")
 
 	return cmd
 }
@@ -90,6 +83,7 @@ func NewSOSReportOptions(streams genericclioptions.IOStreams) *SOSReportOptions 
 		LogOut:    newPrefixWriter(streams.Out, "[sosreport      ] OUT"),
 		RawOut:    streams.Out,
 		Timeout:   10 * time.Minute,
+		Images:    []string{"registry.redhat.io/rhel8/support-tools:8.4-10"},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/sosreport/sosreport.go
```diff
@@ -0,0 +1,687 @@
...
+					// TODO: Add optional SOSReport flags here. (such as -k crio.all=on -k crio.logs=on)
+					// always force disk flush to ensure that all data gathered is accessible in the copy container
+					Command: []string{"/bin/bash", "-c", fmt.Sprintf("/usr/sbin/sos report --batch --tmp-dir=%s %s; sync", o.SourceDir, sosreportAdditionalArgs)},
+
+					// SOSReport requires additional securityContexts and mounts
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: &privileged,
+					},
+					// Env: []corev1.EnvVar{
+					// 	{
```
MichaelWasher, line 598: Currently the SOSReport is collected of the container itself. This is for quicker testing. To collect the SOSReport of the Node, uncomment these environment variables.

pkg/cli/admin/sosreport/sosreport.go
```diff
@@ -0,0 +1,687 @@
...
+		Experimental: This command is under active development and may change without notice.
+	`)
+
+	sosReportExample = templates.Examples(`
+		# Collect a SOSReport using the default image configuration, writing the SOSReport into ./<host-name>.sosreport
+		  oc adm sos
+	`)
+)
+
+func NewSOSReportCommand(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
```
sferich888, line 59: @TurboTurtle please review! 
I don't think we want to be calling sosreport directly on N nodes; but instead call `sos collect` in these situations; to pull data from N nodes. Is that correct?

pkg/cli/admin/sosreport/sosreport.go
```diff
@@ -0,0 +1,687 @@
...
+			fmt.Fprint(o.ErrOut, "Flag --timeout's value in seconds has been deprecated, use duration like 5s, 2m, or 3h, instead\n")
+			i, err := strconv.ParseInt(o.timeoutStr, 0, 64)
+			if err != nil {
+				return fmt.Errorf(`invalid argument %q for "--timeout" flag: %v`, o.timeoutStr, err)
+			}
+			o.Timeout = time.Duration(i) * time.Second
+		}
+	}
+	if len(o.DestDir) == 0 {
+		// TODO: Add node-name to the directory-name
```
sferich888, line 126: You may not want to do that. In some situations 'host names' or 'node names' are 'sensitive' data; and 'obfuscation' becomes an issue if your naming things this way.

pkg/cli/admin/sosreport/sosreport.go
```diff
@@ -0,0 +1,687 @@
...
+		Example: sosReportExample,
+		Run: func(cmd *cobra.Command, args []string) {
+			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			kcmdutil.CheckErr(o.Validate())
+			kcmdutil.CheckErr(o.Run())
+		},
+	}
+
+	cmd.Flags().StringVar(&o.NodeName, "node-name", o.NodeName, "Set a specific node to use - by default a random master will be used")
+	cmd.Flags().StringSliceVar(&o.Images, "image", o.Images, "Specify a must-gather plugin image to run. If not specified, OpenShift's default must-gather image will be used.")
```
sferich888, line 74: Why would we use a must-gather image? Why not just use the RHEL tools image? 

cc @control-d @TurboTurtle

sferich888 (changes requested):
I think we want to see if we should use `soscollecor` vs `sos` to collect the reports; but ultimately this isn't vastly different from must-gather's approach to getting data from 'pods' (in this case pods bound to nodes).

pkg/cli/admin/sosreport/sosreport.go
TurboTurtle: Correct - the preferred way (from an `sos` perspective anyways) would be to run `sos collect` on a single node.

There are 2 approaches here:

1. Run `sos collect` bare on a node that can return a list of nodes via `oc get nodes`. This can either be done by configuring the node such that `root` can run `oc get nodes` successfully, or by passing an access token to `sos collect` that allows `oc get nodes` to be successful. This token can either be passed via the `-c ocp.token=<token>` option, or could be set in the `SOSOCPTOKEN` env var, the latter of which prevents the token from appearing in process listings.

2. Run `sos collect --nodes=node1,node2,etc...` where the calling image directly provides the nodes to collect from.

pkg/cli/admin/sosreport/sosreport.go
TurboTurtle: It'd be `support-tools`, but yes I'd agree here to use an existing image. Additionally, the `support-tools` image contains the `RUN` label which defines how to launch the container so that `sos` collects from the node and not the container it is running in.

soltysh (changes requested):
I wonder if re-using must-gather wouldn't be simpler for this:
`oc adm must-gather -- /usr/bin/sosreport`
the less commands users need to remember the better for the users. And having a consistent way of gathering data in the cluster is definitely a better UX than several almost identical commands.

sferich888:
@soltysh to be more 'explicit' something like: `oc adm must-gather --image registry.redhat.io/rhel8/support-tools --node-name worker-1 --source-dir=/var/tmp/ -- /usr/bin/sosreport` is what you are trying to suggest.

deads2k:
A one-off sosreport command built into `oc` looks like a mistake to me.  Instead of trying to include such information in client tool, it makes more sense to produce an image to use and either use `oc debug node` or `oc adm must-gather` to use the image which collects what you want and can be coupled to the correct ocp levels instead binding an additional one-off in `oc`.

/hold

MichaelWasher:
@deads2k @sferich888 @soltysh 
Thanks for all the feedback.

Is something like this more desirable?
https://github.com/openshift/oc/compare/master...MichaelWasher:priv_mustgather

Then an image could be pushed into a registry (similar to what is here)
https://github.com/MichaelWasher/SOSReportImage

And put it all together like this:
`./oc adm must-gather --max-privilege --node-name worker-1.sharedocp4upi48.lab.upshift.rdu2.redhat.com --image=quay.io/mwasher/sos-gather`

soltysh:
> @deads2k @sferich888 @soltysh
> Thanks for all the feedback.
> 
> Is something like this more desirable?
> [master...MichaelWasher:priv_mustgather](https://github.com/openshift/oc/compare/master...MichaelWasher:priv_mustgather)
> 
> Then an image could be pushed into a registry (similar to what is here)
> https://github.com/MichaelWasher/SOSReportImage
> 
> And put it all together like this:
> `./oc adm must-gather --max-privilege --node-name worker-1.sharedocp4upi48.lab.upshift.rdu2.redhat.com --image=quay.io/mwasher/sos-gather`

That's definitely much better than what is proposed in this PR :+1:

PR #887, opened 2021-07-20
Bug 1980118: Keep workload annotations during the `oc debug` call.

cynepco3hahue:
During the call to `oc debug` for the pod with management resources
it possible that the debug pod will lose workload annotations and
the admission will reject the debug pod because of it.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>


soltysh (approved):
/lgtm
/approve

PR #737, opened 2021-02-11
Bug 1949050: Update containers/image

mtrmac:
… in two steps:
- Revert the only user of openshift/containers-image to the original repo (still using a very old v3.0.2)
- Update to v5.10.2 (which is relevant to `oc adm verify-image-signature` by adding a new identity mapping feature); sadly, this drags in various compression implementations that are not actually used.

The first part might affect development on Mac, because `go test` and the like (not mediated through `Makefile`) would require manually specifying the `containers_image_openpgp` tag, or installing `gpgme` (`brew install gpgme`).

Implemented as two separate commits, so you can choose to include only the first one. See individual commit messages for details.

PR #888, opened 2021-07-27
Update containers/image

mtrmac:
... to remove the dependency on compression implementations.

This removes most of the cruft added in #737 .

<s>This is vendoring an unreleased commit from a main branch; a release should happen fairly soon, I'm filing this now so
that I don't forget.</s>


mtrmac:
Updated to use released containers/image v5.15.0 now.

soltysh:
Picked in https://github.com/openshift/oc/pull/890.

PR #894, opened 2021-08-09
Bug 1786835: Check for out of range condition

deejross:
When attempting to mirror a registry using `oc adm catalog mirror`, an out of range panic occurs.

This fixes the panic by ensuring we stay in the correct range of the slice that generates image name.


soltysh (approved):
/lgtm
/approve
/retest

PR #895, opened 2021-08-09
Bug 1989391: Revert to UnstructuredList to fix yaml output

deejross:
When writing out a list of Group objects, the resulting YAML was unusable and causing issues. Reverting back to an UnstructuredList should resolve issues surrounding missing resource fields.

### DIFF
--- a/pkg/cli/admin/groups/sync/sync.go
+++ b/pkg/cli/admin/groups/sync/sync.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	userv1 "github.com/openshift/api/user/v1"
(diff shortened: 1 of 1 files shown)


soltysh (commented):
Can you make sure that this change doesn't bring back the problem fixed in https://github.com/openshift/oc/pull/759?

deejross:
Yes, the lines that were added in #759 that fixed that problem are still intact: https://github.com/openshift/oc/pull/759/files#diff-0e9da589427fb76bf92f5a899aa2833a9c16c24ba9183197b9245498a3f807cbR182-R183

soltysh:
@deejross will look into test-cmd failures

PR #899, opened 2021-08-16
Bug 1992680: pkg/cli/admin/upgrade/upgrade: Copy edits, including "assists with cluster upgrades"

wking:
The previous:

> This command will request that the cluster begin an upgrade...

and similar [confused some readers][1], because the no-argument invocation is just a read query that does not request an update.  This commit makes the wording more generic, to match the subcommand's actual generic functionality (reading from and optionally writing to ClusterVersion).

I've also tweaked the "No updates available" message to avoid folks misreading as:

> ... doing so may not... result in downtime or data loss.

when we are trying to convey that updating despite a lack of recommendations might be dangerous.

In both cases, the wording I'm altering dates back to the creation of this subcommand in cd30f2f864 (openshift/origin#21605).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1992680#c2

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -38,9 +38,9 @@ func New(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command
 		Use:   "upgrade --to=VERSION",
 		Short: "Upgrade a cluster",
 		Long: templates.LongDesc(`
-			Upgrade the cluster to a newer version
+			Check on upgrade status or upgrade the cluster to a newer version
 
-			This command will request that the cluster begin an upgrade. If no arguments are passed
+			This command assists with cluster upgrades. If no arguments are passed
 			the command will retrieve the current version info and display whether an upgrade is
 			in progress or whether any errors might prevent an upgrade, as well as show the suggested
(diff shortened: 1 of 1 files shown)


wking:
CC @bparees, do you find the wording I'm proposing here more clear than what `oc adm upgrade --help` is using today?  Or do you have further tweaks or alternatives you'd like to float?

I'm leaving the main "`Upgradeable=False` confused me" bit alone for now, while we sort out if/how we can make that more clear.  I don't think we need to block the easy, orthogonal improvements on that particular aspect.

bparees:
it looks good to me, thank you @wking 

/lgtm

wking:
I'm pretty sure [`events should not repeat pathologically` failing on some `DeploymentAwaitingCancellation` thing][1] is unrelated, if a root approver wants to:

But I also expect we'll eventually sneak in on retests if they're too busy.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/899/pull-ci-openshift-oc-master-e2e-agnostic-cmd/1427333317564305408

bparees:
> I'm pretty sure events should not repeat pathologically failing on some DeploymentAwaitingCancellation thing is unrelated

yeah that is on my list to chase elsewhere.

PR #900, opened 2021-08-17
OCPBUGS-3714: pkg/cli/admin/upgrade: Report on Failing!=False conditions

wking:
Spun out of [this comment][0].

cae0b5e096 (openshift/origin#22644) moved this code from Failing to Degraded, likely inspired by openshift/api#287.  But Degraded is only used in ClusterOperator. ClusterVersion kept using Failing, as seen in openshift/cluster-version-operator#191.  This commit returns us to watching for Failing (the condition the CVO has been setting the whole time), and informing the caller for any non-happy statuses (or the lack of a Failing condition at all).

Even though the issue causing `Failing=True` may block the current update from progressing, it should not block admins from requesting a new update target.  For some bugs, retargeting is the recommended way to resolve the issue that is currently sticking the update ([one example][1]).

[0]: https://bugzilla.redhat.com/show_bug.cgi?id=1992680#c9
[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1988576#c30

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -28,6 +28,13 @@ import (
 	"github.com/openshift/oc/pkg/cli/admin/upgrade/channel"
 )
 
+const (
+	// clusterStatusFailing is set on the ClusterVersion status when a cluster
+	// cannot reach the desired state. It is considered more serious than Degraded
+	// and indicates the cluster is not healthy.
+	clusterStatusFailing = configv1.ClusterStatusConditionType("Failing")
+)
+
 var upgradeExample = templates.Examples(`
 	# Review the available cluster updates
 	oc adm upgrade
@@ -340,15 +347,12 @@ func (o *Options) Run() error {
 		return nil
 
 	default:
-		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorDegraded); c != nil && c.Status == configv1.ConditionTrue {
-			prefix := "No upgrade is possible due to an error"
-			if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil && c.Status == configv1.ConditionTrue && len(c.Message) > 0 {
-				prefix = c.Message
+		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil {
+			if c.Status != configv1.ConditionFalse {
+				fmt.Fprintf(o.Out, "%s=%s:\n\n  Reason: %s\n  Message: %s\n\n", c.Type, c.Status, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 			}
-			if len(c.Message) > 0 {
-				return fmt.Errorf("%s:\n\n  Reason: %s\n  Message: %s\n\n", prefix, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
-			}
-			return fmt.Errorf("The cluster can't be upgraded, see `oc describe clusterversion`")
+		} else {
+			fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details.\n", clusterStatusFailing)
 		}
 
 		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil && len(c.Message) > 0 {
@@ -358,12 +362,12 @@ func (o *Options) Run() error {
 				fmt.Fprintln(o.Out, c.Message)
 			}
 		} else {
-			fmt.Fprintln(o.ErrOut, "warning: No current status info, see `oc describe clusterversion` for more details")
+			fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details.\n", configv1.OperatorProgressing)
 		}
 		fmt.Fprintln(o.Out)
 
 		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorUpgradeable); c != nil && c.Status == configv1.ConditionFalse {
-			fmt.Fprintf(o.Out, "Upgradeable=False\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
+			fmt.Fprintf(o.Out, "%s=%s\n\n  Reason: %s\n  Message: %s\n\n", c.Type, c.Status, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 		}
 
 		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, "ReleaseAccepted"); c != nil && c.Status != configv1.ConditionTrue {
@@ -560,20 +564,27 @@ func findClusterOperatorStatusCondition(conditions []configv1.ClusterOperatorSta
 func checkForUpgrade(cv *configv1.ClusterVersion) error {
 	results := []string{}
 	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, "Invalid"); c != nil && c.Status == configv1.ConditionTrue {
-		results = append(results, fmt.Sprintf("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
+		results = append(results, fmt.Sprintf("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: %s\n  Message: %s", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
 	}
-	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorDegraded); c != nil && c.Status == configv1.ConditionTrue {
-		results = append(results, fmt.Sprintf("the cluster is experiencing an upgrade-blocking error:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
+	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil && c.Status == configv1.ConditionTrue {
+		target := cv.Status.Desired.Version
+		if target == "" {
+			target = cv.Status.Desired.Image
+			if target == "" {
+				target = "ClusterVersion status.desired has neither a version nor image; please open a bug"
+			}
+		}
+		results = append(results, fmt.Sprintf("the cluster is experiencing an error reconciling %q:\n\n  Reason: %s\n  Message: %s", target, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
 	}
 	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil && c.Status == configv1.ConditionTrue {
-		results = append(results, fmt.Sprintf("the cluster is already upgrading:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
+		results = append(results, fmt.Sprintf("the cluster is already upgrading:\n\n  Reason: %s\n  Message: %s", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
 	}
 
 	if len(results) == 0 {
 		return nil
 	}
 
-	return errors.New(strings.Join(results, ""))
+	return errors.New(strings.Join(results, "\n\n"))
 }
 
 // targetMatch returns true if the target release matches the target
--- a/pkg/cli/admin/upgrade/upgrade_test.go
+++ b/pkg/cli/admin/upgrade/upgrade_test.go
@@ -48,26 +48,33 @@ func TestCheckForUpgrade(t *testing.T) {
 				Reason:  "SomeReason",
 				Message: "Some message.",
 			}},
-			expected: errors.New("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: SomeReason\n  Message: Some message.\n\n"),
+			expected: errors.New("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: SomeReason\n  Message: Some message."),
 		},
 		{
-			name: "degraded and progressing",
+			name: "failing and progressing",
 			conditions: []configv1.ClusterOperatorStatusCondition{{
 				Type:    configv1.OperatorProgressing,
 				Status:  configv1.ConditionTrue,
 				Reason:  "RollingOut",
 				Message: "Updating to v2.",
 			}, {
-				Type:    configv1.OperatorDegraded,
+				Type:    clusterStatusFailing,
 				Status:  configv1.ConditionTrue,
 				Reason:  "BadStuff",
 				Message: "The widgets are slow.",
 			}},
-			expected: errors.New("the cluster is experiencing an upgrade-blocking error:\n\n  Reason: BadStuff\n  Message: The widgets are slow.\n\nthe cluster is already upgrading:\n\n  Reason: RollingOut\n  Message: Updating to v2.\n\n"),
+			expected: errors.New("the cluster is experiencing an error reconciling \"4.1.0\":\n\n  Reason: BadStuff\n  Message: The widgets are slow.\n\nthe cluster is already upgrading:\n\n  Reason: RollingOut\n  Message: Updating to v2."),
 		},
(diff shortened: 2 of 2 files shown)


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -307,15 +314,12 @@ func (o *Options) Run() error {
...
+		if c := findCondition(cv.Status.Conditions, ClusterStatusFailing); c != nil {
+			if c.Status != configv1.ConditionFalse {
+				fmt.Fprintf(o.Out, "%s=%s:\n\n  Reason: %s\n  Message: %s\n\n", c.Type, c.Status, c.Reason, c.Message)
 			}
-			if len(c.Message) > 0 {
-				return fmt.Errorf("%s:\n\n  Reason: %s\n  Message: %s\n\n", prefix, c.Reason, c.Message)
-			}
-			return fmt.Errorf("The cluster can't be upgraded, see `oc describe clusterversion`")
+		} else {
+			fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details", ClusterStatusFailing)
```
LalatenduMohanty, line 322: Why is this a warning? Also it should be s/`No current %s info`/ `No information on current %s`/

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/upgrade.go
wking: It's a warning, because we should always be setting `Failing`.  Ideally to `Failing=False`, but sometimes to `Failing=True`.  A ClusterVersion with no `Failing` at all is a sign that something is probably wrong with the CVO.

`...No current Failing info, see...` sounds pretty clear to me.  I'd be fine rephrasing to something like `...No current conditions with type Failing, see...` here and below for `Progressing` if you prefer.

LalatenduMohanty:
@wking Do you have a sample output of old code vs this PR?

wking:
Before this PR, `oc` ignored `Failing=True`.  So to use the unit-test dummy data, `Progressing=True` `Failing=True` would have looked like:

```
the cluster is already upgrading:

  Reason: RollingOut
  Message: Updating to v2.
```

With this PR, it will look like:

```
the cluster is experiencing a possibly-upgrade-blocking error:

  Reason: BadStuff
  Message: The widgets are slow.

the cluster is already upgrading:

  Reason: RollingOut
  Message: Updating to v2.
```

wking:
[`TestNewAppBuildOutputCycleDetection`][1] seems unrelated:

/test unit

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/900/pull-ci-openshift-oc-master-unit/1491503230864068608

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -569,8 +573,8 @@ func checkForUpgrade(cv *configv1.ClusterVersion) error {
 	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, "Invalid"); c != nil && c.Status == configv1.ConditionTrue {
 		results = append(results, fmt.Sprintf("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
 	}
-	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorDegraded); c != nil && c.Status == configv1.ConditionTrue {
-		results = append(results, fmt.Sprintf("the cluster is experiencing an upgrade-blocking error:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
+	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil && c.Status == configv1.ConditionTrue {
+		results = append(results, fmt.Sprintf("the cluster is experiencing a possibly-upgrade-blocking error:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
```
LalatenduMohanty, line 577: I am not comfortable with the string `experiencing a possibly-upgrade-blocking error` . Either it is a upgrade blocker or it is not . By suggesting that it is possibly an upgrade blocker we are creating confusion for the user.

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/upgrade.go
wking: it might be an update blocker (e.g. a ClusterOperator is [`Degraded=True` on a broken webhook, and as you update it stays mad, and the update wedges][1]).  Or it might not be an update blocker (e.g. a ClusterOperator is `Degraded=True` on a webhook that it thinks is more broken than it is, and as you update the ClusterOperator [realizes it was a false-positive, becomes happy][2], and the update completes).  It's hard for the CVO to distinguish between those two cases, but what we can do is inform the user that the cluster is `Failing=True` now (those `Degraded=True` are two of many possible reasons we could be `Failing=True`), and leave it up to them to distinguish "yes, I think this update might fix that" from "yikes, I'd better fix that before updating".

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2052513#c1
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=2052744

wking:
I'm still pretty sure we want this.

/reopen
/remove-lifecycle rotten
/lifecycle frozen

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -351,15 +358,12 @@ func (o *Options) Run() error {
 		return nil
 
 	default:
-		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorDegraded); c != nil && c.Status == configv1.ConditionTrue {
-			prefix := "No upgrade is possible due to an error"
-			if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil && c.Status == configv1.ConditionTrue && len(c.Message) > 0 {
-				prefix = c.Message
+		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil && c.Status == configv1.ConditionTrue {
+			if c.Status != configv1.ConditionFalse {
```
shellyyang1989, line 362: The condition check seems not to be duplicate with the condition in line#361 `c != nil && c.Status == configv1.ConditionTrue`.

wking: Thanks :).  Rerolled in f95fea08d -> 33c93e15c to drop the `c.Status == configv1.ConditionTrue` guard.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -562,8 +566,8 @@ func checkForUpgrade(cv *configv1.ClusterVersion) error {
 	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, "Invalid"); c != nil && c.Status == configv1.ConditionTrue {
 		results = append(results, fmt.Sprintf("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
 	}
-	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorDegraded); c != nil && c.Status == configv1.ConditionTrue {
-		results = append(results, fmt.Sprintf("the cluster is experiencing an upgrade-blocking error:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
+	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil && c.Status == configv1.ConditionTrue {
+		results = append(results, fmt.Sprintf("the cluster is experiencing a possibly-upgrade-blocking error:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
```
LalatenduMohanty, line 570: instead of `The cluster is experiencing a possibly-upgrade-blocking error` what about `Here are the errors on the cluster which might impact upgrades`

pkg/cli/admin/upgrade/upgrade.go
wking: Reworded to `the cluster is experiencing an error reconciling the current desired release` with 33c93e15c -> 54eca896f.

pkg/cli/admin/upgrade/upgrade.go
wking: I've pushed d0030fcfd where I use `the cluster is experiencing an error reconciling ${CURRENT_TARGET_VERSION}`.  And it falls back to the target pullspec in the unlikely event of not knowing the version.

evakhoni:
pre-merge verified in https://issues.redhat.com/browse/OCPBUGS-3714#comment-21273892  
/label qe-approved

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -340,15 +347,12 @@ func (o *Options) Run() error {
...
-			}
-			if len(c.Message) > 0 {
-				return fmt.Errorf("%s:\n\n  Reason: %s\n  Message: %s\n\n", prefix, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
+		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil {
+			if c.Status != configv1.ConditionFalse {
+				fmt.Fprintf(o.Out, "%s=%s:\n\n  Reason: %s\n  Message: %s\n\n", c.Type, c.Status, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 			}
-			return fmt.Errorf("The cluster can't be upgraded, see `oc describe clusterversion`")
+		} else {
+			fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details", clusterStatusFailing)
```
LalatenduMohanty, line 355: Looks like this is a new addition from last time I had reviewed the PR or I had missed it. I am not sure what would be the output for this.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -560,20 +564,24 @@ func findClusterOperatorStatusCondition(conditions []configv1.ClusterOperatorSta
...
 	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, "Invalid"); c != nil && c.Status == configv1.ConditionTrue {
-		results = append(results, fmt.Sprintf("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
+		results = append(results, fmt.Sprintf("the cluster version object is invalid, you must correct the invalid state first:\n\n  Reason: %s\n  Message: %s", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
 	}
-	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorDegraded); c != nil && c.Status == configv1.ConditionTrue {
-		results = append(results, fmt.Sprintf("the cluster is experiencing an upgrade-blocking error:\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  ")))
+	if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil && c.Status == configv1.ConditionTrue {
+		target := cv.Status.Desired.Version
+		if target == "" {
+			target = cv.Status.Desired.Image
```
LalatenduMohanty, line 572: The value of cv.Status.Desired.Image might be empty too, right? But we are not checking for it.

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/upgrade.go
wking: if `version` and `image` are both empty, we have no idea where we're going.  Seems unlikely, and also like there's not much we could do to mitigate.  What would you fall back to?

pkg/cli/admin/upgrade/upgrade.go
wking: not new, this is from f42948c1d0, my original attempt in this PR:

```console
$ git --no-pager show --pretty=fuller f42948c
commit f42948c1d04d2edef347f28c22d0b95611be360c
Author:     W. Trevor King <wking@tremily.us>
AuthorDate: Mon Aug 16 17:29:58 2021 -0700
Commit:     W. Trevor King <wking@tremily.us>
CommitDate: Mon Aug 16 17:29:58 2021 -0700

    pkg/cli/admin/upgrade: Report on Failing!=False conditions
...
-                       return fmt.Errorf("The cluster can't be upgraded, see `oc describe clusterversion`")
+               } else {
+                       fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details", ClusterStatusFailing)
                }
 
                if c := findCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil && len(c.Message) > 0 {
@@ -325,12 +329,12 @@ func (o *Options) Run() error {
                                fmt.Fprintln(o.Out, c.Message)
                        }
                } else {
-                       fmt.Fprintln(o.ErrOut, "warning: No current status info, see `oc describe clusterversion` for more details")
+                       fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details.\n", configv1.OperatorProgressing)
                }
...
```

The output would be:

    warning: No current Failing info, see `oc describe clusterversion` for more details`

when there was no `Failing` condition.  But I see I'm missing a trailing newline.  Fixed with ea3dd7068 -> 1900577fa.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -358,12 +362,12 @@ func (o *Options) Run() error {
 				fmt.Fprintln(o.Out, c.Message)
 			}
 		} else {
-			fmt.Fprintln(o.ErrOut, "warning: No current status info, see `oc describe clusterversion` for more details")
+			fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details.\n", configv1.OperatorProgressing)
 		}
 		fmt.Fprintln(o.Out)
```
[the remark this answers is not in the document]
wking: this is just asking for a blank line after the `Progressing` output.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -340,15 +347,12 @@ func (o *Options) Run() error {
...
-			}
-			if len(c.Message) > 0 {
-				return fmt.Errorf("%s:\n\n  Reason: %s\n  Message: %s\n\n", prefix, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
+		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, clusterStatusFailing); c != nil {
+			if c.Status != configv1.ConditionFalse {
+				fmt.Fprintf(o.Out, "%s=%s:\n\n  Reason: %s\n  Message: %s\n\n", c.Type, c.Status, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 			}
-			return fmt.Errorf("The cluster can't be upgraded, see `oc describe clusterversion`")
+		} else {
+			fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details.\n", clusterStatusFailing)
```
LalatenduMohanty, line 355: Because we have removed the return statement , the upgrade might be able to start with the warning?

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: We should check and put an appropriate error message. I do not think we can do more than that.

pkg/cli/admin/upgrade/upgrade.go
wking: those `return` were just for the dead-code `Degraded` processing, that never did anything, because we never have `Degraded` in ClusterVersion.  Previously we ignored `Failing` completely.  Now we'll tell you about it, and if you're asking for an update, we'll ask you to set `--allow-upgrade-with-warnings`.

pkg/cli/admin/upgrade/upgrade.go
wking: I've added a fallback to `<ClusterVersion status.desired has neither a version nor image; please open a bug>` with 1900577fa -> 438e88491.

pkg/cli/admin/upgrade/upgrade.go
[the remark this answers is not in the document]
LalatenduMohanty: We can mark it resolved

LalatenduMohanty (approved):
/lgtm

PR #917, opened 2021-09-01
build: Replace github.com/apcera/gssapi with openshift/gssapi

cfergeau:
At this point, github.com/apcera/gssapi is unlikely to come back online.


cfergeau:
Rebased, and still relevant.

cfergeau:
@deads2k @mfojtik could this get a review?

cfergeau:
The reason why I want to remove it is because we have to carry this replace in `crc` as well, and I'd like to get rid of it:
https://github.com/code-ready/crc/blob/a5684193cd084d93024ea9246d8752bd95d3cd89/go.mod#L190-L195
```
replace (
	github.com/apcera/gssapi => github.com/openshift/gssapi v0.0.0-20161010215902-5fb4217df13b

	k8s.io/kubectl => github.com/openshift/kubernetes-kubectl v0.0.0-20220124124213-df787b5b471c
)
```

cfergeau:
After further discussion, I'm going to close this.
Keeping the `replace` hints that `github.com/openshift/gssapi` is not something developed/maintained by the openshift org, but a workaround for a repository which is gone.
The right course of action would be to switch to a different gssapi implementation (maybe https://github.com/golang-auth/go-gssapi ?), this is not a lot of code to replace, but this would need very careful testing, which is going to be the hard part of the work.

PR #918, opened 2021-09-01
Bug 1999246: adm catalog mirror should ignore .indexignore files

joelanford:
When loading file-based catalogs, oc adm catalog mirror should
automatically ignore .indexignore files it finds rather than requiring
an explicit pattern in the .indexignore file to ignore itself. There is
no circumstance where a .indexignore file would be expected to be
understood as an catalog file.

Signed-off-by: Joe Lanford <joe.lanford@gmail.com>


soltysh (approved):
/approve
/retest

PR #919, opened 2021-09-06
Bug 1977414: check BuildConfig Secrets and ConfigMaps before start-build creates a build

alicerum:
In case of a binary build, we want to create for resources `BuildConfig` is expecting to exist, before creating the `Build` object.

This way, we'll save cluster resources by not creating objects that are destined to fail.

### DIFF
--- a/pkg/cli/startbuild/startbuild.go
+++ b/pkg/cli/startbuild/startbuild.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/third_party/forked/golang/netutil"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
+	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/scheme"
@@ -91,6 +92,8 @@ var (
 )
 
 type StartBuildOptions struct {
+	ctx context.Context
+
 	PrintFlags *genericclioptions.PrintFlags
 
 	Printer printers.ResourcePrinter
@@ -124,6 +127,7 @@ type StartBuildOptions struct {
 
 	Mapper         meta.RESTMapper
 	BuildClient    buildv1client.BuildV1Interface
+	KubeClient     kubernetes.Interface
 	BuildLogClient buildclientmanual.BuildLogInterface
 	ClientConfig   *restclient.Config
 
@@ -155,9 +159,9 @@ func NewCmdStartBuild(f kcmdutil.Factory, streams genericclioptions.IOStreams) *
 		Example:    startBuildExample,
 		SuggestFor: []string{"build", "builds"},
 		Run: func(cmd *cobra.Command, args []string) {
-			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			kcmdutil.CheckErr(o.Complete(cmd.Context(), f, cmd, args))
 			kcmdutil.CheckErr(o.Validate())
-			kcmdutil.CheckErr(o.Run())
+			kcmdutil.CheckErr(o.Run(cmd.Context()))
 		},
 	}
 	cmd.Flags().StringVar(&o.LogLevel, "build-loglevel", o.LogLevel, "Specify the log level for the build log output")
@@ -187,7 +191,7 @@ func NewCmdStartBuild(f kcmdutil.Factory, streams genericclioptions.IOStreams) *
 	return cmd
 }
 
-func (o *StartBuildOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
+func (o *StartBuildOptions) Complete(ctx context.Context, f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
 	var err error
 	o.Git = git.NewRepository()
 	o.ClientConfig, err = f.ToRESTConfig()
@@ -271,6 +275,11 @@ func (o *StartBuildOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 		return err
 	}
 
+	o.KubeClient, err = kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
 	var (
 		name     = buildName
 		resource = build.Resource("builds")
@@ -299,7 +308,7 @@ func (o *StartBuildOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 
 	// when listing webhooks, allow --from-build to lookup a build config
 	if build.Resource("builds") == resource && len(o.ListWebhooks) > 0 {
-		build, err := o.BuildClient.Builds(o.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
+		build, err := o.BuildClient.Builds(o.Namespace).Get(ctx, name, metav1.GetOptions{})
 		if err != nil {
 			return err
 		}
@@ -378,12 +387,12 @@ func (o *StartBuildOptions) Validate() error {
 }
 
 // Run contains all the necessary functionality for the OpenShift cli start-build command
-func (o *StartBuildOptions) Run() error {
+func (o *StartBuildOptions) Run(ctx context.Context) error {
 	if len(o.FromWebhook) > 0 {
 		return o.RunStartBuildWebHook()
 	}
 	if len(o.ListWebhooks) > 0 {
-		return o.RunListBuildWebHooks()
+		return o.RunListBuildWebHooks(ctx)
 	}
 
 	buildRequestCauses := []buildv1.BuildTriggerCause{}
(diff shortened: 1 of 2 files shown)


gabemontero (changes requested):
I think there is an opportunity for unit tests in https://github.com/openshift/oc/blob/master/pkg/cli/startbuild/startbuild_test.go @alicerum using the k8s fake  client

basic pass/fail based on the presence or lack there of of secrets/configmaps

Otherwise LGTM

gabemontero (commented):
I might have had the comment explaining that you've created a partial list of the specified configmaps and secrets up in `TestBuildConfigWithMissingResources` itself, but I won't block/force and update on that detail.

Also, I *think* it is OK to assume that the "all secrets / configmaps are present" would pass, but am less sure there.  Ideally having a permutation for that scenario would be best.

I'll add 

/lgtm

for now, but can reapply if you agree with me about explicitly validating the "all configmaps/secrets" scenarios is warranted.

And of course, if @adambkaplan is less lenient with me wrt this as part of deciding whether to do the approve, you'll have that to contend with.

pkg/cli/startbuild/startbuild.go
```diff
@@ -513,6 +526,112 @@ func (o *StartBuildOptions) streamBuildLogs(build *buildv1.Build) error {
...
+	}
+
+	var volumes []buildv1.BuildVolume
+	switch buildConfig.Spec.Strategy.Type {
+	case buildv1.DockerBuildStrategyType:
+		volumes = buildConfig.Spec.Strategy.DockerStrategy.Volumes
+	case buildv1.SourceBuildStrategyType:
+		volumes = buildConfig.Spec.Strategy.SourceStrategy.Volumes
+	}
+
```
coreydaley, line 555:
```suggestion
if len(volumes) == 0 {
    return nil
}

You could return quickly here if there are no volumes that need to be processed.
```

pkg/cli/startbuild/startbuild.go
```diff
@@ -513,6 +526,112 @@ func (o *StartBuildOptions) streamBuildLogs(build *buildv1.Build) error {
...
+	return nil
+}
+
+// checkNonExistantConfigMaps takes a list of ConfigMap names and returns
+// list of all of those that do not exist in the namespace
+// or return error in case of unexpected behavior
+func (o *StartBuildOptions) checkNonExistantConfigMaps(cms []string) ([]string, error) {
+	result := make([]string, 0, 3)
```
coreydaley, line 597:
```suggestion
        if len(cms) == 0 {
            return []string{}, nil
        }
	result := make([]string, 0, 3)
```

Return quickly if there are no configMaps

pkg/cli/startbuild/startbuild.go
```diff
@@ -513,6 +526,112 @@ func (o *StartBuildOptions) streamBuildLogs(build *buildv1.Build) error {
...
+	}
+
+	return result, nil
+}
+
+// checkNonExistantSecrets takes a list of Secret names and returns
+// list of all of those that do not exist in the namespace
+// or return error in case of unexpected behavior
+func (o *StartBuildOptions) checkNonExistantSecrets(secrets []string) ([]string, error) {
+	result := make([]string, 0, 3)
```
coreydaley, line 618: Same thing here, return quickly if there are no secrets passed in

gabemontero:
@coreydaley has been driving the review in a better fashion than me on this one

/assign @coreydaley 
/unassign @gabemontero 

of course still leaving @adambkaplan on for the approve label

pkg/cli/startbuild/startbuild.go
```diff
@@ -513,6 +526,120 @@ func (o *StartBuildOptions) streamBuildLogs(build *buildv1.Build) error {
 	return err
 }
 
+// checkNonExistantResources checks for Resources (config maps, secrets, etc) listed in the BuildConfig
+// and returns error in case any of them are missing, or in case of unexpected behavior
+func (o *StartBuildOptions) checkNonExistantResources() error {
+	buildConfig, err := o.BuildClient.BuildConfigs(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
```
adambkaplan, line 532: This isn't a blocking issue, more an observation that we have a lot of `context.TODO()` calls. Ideally we should have a parent context for the start-build command. We can then create sub-contexts which then give us flexibility to add things like cancellation and timeouts.

pkg/cli/startbuild/startbuild.go
```diff
@@ -513,6 +526,120 @@ func (o *StartBuildOptions) streamBuildLogs(build *buildv1.Build) error {
...
+	if err != nil {
+		return err
+	}
+
+	cms := make([]string, 0, 3)
+	secrets := make([]string, 0, 3)
```
adambkaplan, line 538: Is there a compelling reason to set an initial capacity here? I don't think we lose much with an initial capacity of 0:
```suggestion
	cms := []string{}
	secrets := []string{}
```

pkg/cli/startbuild/startbuild.go
```diff
@@ -513,6 +526,120 @@ func (o *StartBuildOptions) streamBuildLogs(build *buildv1.Build) error {
...
+	if err != nil {
+		return err
+	}
+	missingSecrets, err := o.checkNonExistantSecrets(secrets)
+	if err != nil {
+		return err
+	}
+
+	// join all the missing resources names together to form an error message
+	resourceMessages := make([]string, 0, 2)
```
adambkaplan, line 578: Ditto on the initial capacity:
```suggestion
	resourceMessages := []string{}
```

pkg/cli/startbuild/startbuild.go
adambkaplan: I'm not in favor of setting the capacity:
```suggestion
	result := []string{}
```

pkg/cli/startbuild/startbuild.go
```diff
@@ -513,6 +526,120 @@ func (o *StartBuildOptions) streamBuildLogs(build *buildv1.Build) error {
...
+func (o *StartBuildOptions) checkNonExistantConfigMaps(cms []string) ([]string, error) {
+	if len(cms) == 0 {
+		return []string{}, nil
+	}
+
+	result := make([]string, 0, 3)
+
+	for _, name := range cms {
+		_, err := o.KubeClient.CoreV1().ConfigMaps(o.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
+		if err != nil && kerrors.IsNotFound(err) {
```
adambkaplan, line 605: Not a bad thing here, but I believe `kerrors.IsNotFound` returns false if `err` is nil. Feel free to discard this suggestion:
```suggestion
		if kerrors.IsNotFound(err) {
```

pkg/cli/startbuild/startbuild.go
adambkaplan:
```suggestion
	result := []string{}
```

pkg/cli/startbuild/startbuild_test.go
```diff
@@ -390,3 +395,156 @@ func TestStreamBuildLogs(t *testing.T) {
 		})
 	}
 }
+
+func TestBuildConfigWithMissingResources(t *testing.T) {
```
adambkaplan, line 399: I would like to see more test cases here - there are a lot of permutations we should check:

1. BuildConfig with no Secrets or ConfigMaps
2. BuildConfig with build input Secrets only (available and missing)
3. BuildConfig with build input ConfigMaps only (available and missing)
4. BuildConfig with build input Secrets and ConfigMaps (available and missing)
5. BuildConfig with build volume Secrets in source strategy (available and missing)
6. BuildConfig with build volume ConfigMaps in source strategy (available and missing)
7. BuildConfig with build volume secrets in Docker strategy (available and missing)
8. BuildConfig with build volume ConfigMaps in Docker strategy (available and missing)
9. BuildConfig with build volume Secrets and ConfigMaps in Docker strategy (available and missing)
10. BuildConfig with build input and volumes - Secrets and ConfigMaps (basically everything)

I would be OK with combining some of these, as you would hit 19 test cases here doing each of these individually.

adambkaplan (changes requested):
Ironing out the details here - core logic looks good.

adambkaplan:
/retitle Bug 1977414: check BuildConfig Secrets and ConfigMaps before start-build creates a build

adambkaplan:
/approve

Signaling that this change should go in once code issues are addressed.

pkg/cli/startbuild/startbuild.go
coreydaley: We have a lot of code that does a nil check before using `kerrors.IsNotFound()` (https://github.com/openshift/oc/search?q=kerrors.IsNotFound), I don't think it is a strictly bad idea since it future proofs the code. But, we also have some code that does not do that first. Is there any doc that points to whether or not that function handles nil ok?

pkg/cli/startbuild/startbuild.go
alicerum: I'll do it for the `startbuild.go` then. Ideally we should create an issue and work on that through as much as possible, at least whole `build` part of `oc`. I don't want to do more here in this PR at least, in order to not make it bigger that it has to be.

adambkaplan (approved):
/lgtm

PR #922, opened 2021-09-08
Bug 1970805: Replace slashes in suggested ImageStream name

alicerum:
When figuring out an image stream name for created build objects, `SuggestName` function can return name with slashes in some situations.

It is important to hold correct data (that including slashes, too) in the `ImageRef` object, because it is used in other places and can lead to errors.

`SuggestName` method of the `ImageRef` however, should return name without slashes when suggesting name for an `ImageStream`.


gabemontero (approved):
/lgtm

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,16 +230,27 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
-	if len(r.Reference.Name) > 0 {
-		return r.Reference.Name, true
+	if len(name) > 0 {
+		if r.AsImageStream {
+			// in some certain situations, suggested ImageStream name might contain slashes
+			// in this case we want to replace it with dashes, referenced object might
+			// contain said dashes, but imagestream must cannot
+			// example: https://bugzilla.redhat.com/show_bug.cgi?id=1970805
+			name = strings.ReplaceAll(name, "/", "-")
+		}
```
coreydaley, line 251: I think we could future-proof this some by eliminating the `r.AsImageStream` check and just always replacing slashes with dashes.
```suggestion

			// in some certain situations, suggested ImageStream name might contain slashes
			// in this case we want to replace it with dashes, referenced object might
			// contain said dashes, but imagestream must cannot
			// example: https://bugzilla.redhat.com/show_bug.cgi?id=1970805
			name = strings.ReplaceAll(name, "/", "-")
```

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,16 +230,27 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
+
+		return name, true
 	}
 	return "", false
```
coreydaley, line 255: I think that we could eliminate the `if len(name)) > 0` above and just return `name, len(name) > 0` here?
```suggestion
	return name, len(name) > 0
```

pkg/helpers/newapp/app/imageref_test.go
```diff
@@ -89,6 +89,83 @@ func TestBuildConfigOutput(t *testing.T) {
...
+			},
+			expectedSuggestedName: expectedObjectNameNoSlashes,
+		},
+	}
+
+	for _, test := range tests {
+		suggestedName, _ := test.imageRef.SuggestName()
+
+		if suggestedName != test.expectedSuggestedName {
+			t.Errorf("Wrong suggested name, expected '%s', actual '%s'", test.expectedSuggestedName, suggestedName)
```
coreydaley, line 164:
```suggestion
			t.Errorf("Wrong suggested name, expected %q, actual %q", test.expectedSuggestedName, suggestedName)
```

pkg/helpers/newapp/app/imageref_test.go
```diff
@@ -89,6 +89,83 @@ func TestBuildConfigOutput(t *testing.T) {
...
+	objectName := "some/test/name-1"
+	host := "test-host.org"
+	tag := "latest"
+	namespace := "namespace-name"
+
+	expectedObjectNameNoSlashes := "some-test-name-1"
+
+	tests := []struct {
+		imageRef              *ImageRef
+		expectedSuggestedName string
```
coreydaley, line 102: nit: can just use `expected` since only one this is being checked.

pkg/helpers/newapp/app/imageref_test.go
```diff
@@ -89,6 +89,83 @@ func TestBuildConfigOutput(t *testing.T) {
...
+					},
+				},
+				AsImageStream: true,
+			},
+			expectedSuggestedName: expectedObjectNameNoSlashes,
+		},
+	}
+
+	for _, test := range tests {
+		suggestedName, _ := test.imageRef.SuggestName()
```
coreydaley, line 161: i think we should also be checking the returned `bool` like the test below does

pkg/helpers/newapp/app/imageref_test.go
coreydaley, line 164: Since all of your tests are using the same `expectedSuggestedName` it is going to be very difficult for someone to determine which test case is failing. I would rather see the test format used where each case has a `name` which is output when the test runs if it fails: Here is an example: https://gobyexample.com/testing

coreydaley (changes requested):

pkg/helpers/newapp/app/imageref.go
alicerum: At first I was going to argue that this is only the `ImageStream` issue, and other types of image references must be left intact, but after giving it a thought, I realized that you are right.
`SuggestedName` is used to generate names or part of names for internal `k8s` objects, and those names must follow the DNS standard, hence, no slashes. I think it is safe to assume we will never need slashes in any names here.

Do you think we might go even further and replace all the non[alphanumeric, dash or dot] symbols here?

pkg/helpers/newapp/app/imageref_test.go
alicerum: It takes some branching to figure out type of `ImageRef` test is running against, so I just added test names.

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,18 +230,25 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
-	}
-	if len(r.Reference.Name) > 0 {
-		return r.Reference.Name, true
+	switch {
+	case r.Stream != nil:
+		name = r.Stream.Name
+	case len(r.ObjectName) > 0:
+		name = r.ObjectName
+	case len(r.Reference.Name) > 0:
+		name = r.Reference.Name
```
coreydaley, line 242: +1, love seeing switch statements instead of multiple `if` or `if/else`

pkg/helpers/newapp/app/imageref.go
coreydaley: So, basically replace this with a regexp that replaces all non-valid characters with dashes? I think that sounds like a great idea.

alicerum: @coreydaley yeah, `[^0-9a-zA-Z-.]` should do the trick. From what I read, `k8s` names should go along with rfc1123, which references rfc952, which says, only alphanumeric, dots and dashes are allowed in a hostname. So, I thought it might be better to go with this more uniform approach rather than just slashes.

pkg/helpers/newapp/app/imageref.go
```diff
@@ -44,6 +45,14 @@ type SecretAccessor interface {
 
 type imageRefGenerator struct{}
 
+var (
+	dnsNamePattern *regexp.Regexp
+)
+
+func init() {
+	dnsNamePattern = regexp.MustCompile("[^a-zA-Z0-9-.]")
```
coreydaley, line 53: https://github.com/kubernetes/apimachinery/blob/v0.22.1/pkg/util/validation/validation.go#L177-L196

alicerum: ah, so no dots either. thanks.

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,18 +239,25 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
+	case len(r.Reference.Name) > 0:
+		name = r.Reference.Name
 	}
-	return "", false
+
+	// in some certain situations, ref name might contain slashes or other symbols
+	// k8s naming goes along rfc952, which states that only alphanumeric characters,
+	// dots and dashes are allowed in a name
+	// bug example: https://bugzilla.redhat.com/show_bug.cgi?id=1970805
+	name = dnsNamePattern.ReplaceAllString(name, "-")
```
coreydaley, line 258: I would swear that we have existing code somewhere that does this (maybe not in this repository). Maybe in library-go?

pkg/helpers/newapp/app/imageref.go
```diff
@@ -44,6 +45,14 @@ type SecretAccessor interface {
...
 type imageRefGenerator struct{}
 
+var (
+	dnsNamePattern *regexp.Regexp
+)
+
+func init() {
+	dnsNamePattern = regexp.MustCompile("[^a-zA-Z0-9-]")
+}
+
```
coreydaley, line 55:
```suggestion
var (
	dnsNamePattern *regexp.Regexp =  regexp.MustCompile("[^a-zA-Z0-9-]")
)

```

I don't think you need this `init` here, the regexp can just be a var, unless there is a specific reason to use `init` that I am missing?

pkg/helpers/newapp/app/imageref_test.go
```diff
@@ -89,6 +89,67 @@ func TestBuildConfigOutput(t *testing.T) {
 	}
 }
 
+func TestImageRefAsImageStreamSuggestedNameContainsSlashes(t *testing.T) {
```
coreydaley, line 92:
```suggestion
func TestSuggestName(t *testing.T) {
```
This test and the below `TestImageStreamRefNaming` test seem like they should be combined since they are both testing the same function.

pkg/helpers/newapp/app/imageref.go
```diff
@@ -44,6 +45,10 @@ type SecretAccessor interface {
 
 type imageRefGenerator struct{}
 
+var (
+	dnsNamePattern = regexp.MustCompile("[^a-zA-Z0-9-]")
```
adambkaplan, line 49: Upstream the DNS pattern for object names (DNS_SUBDOMAIN) is more nuanced:

https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L178
https://github.com/kubernetes/apimachinery/blob/71bf7ba06cd490392e05cd6828ae24a69cb0ccaa/pkg/util/validation/validation.go#L199

I would perhaps rename this to "dnsSubdomainAllowedCharacters"

```suggestion
	// Imagestream names must comply with k8s DNS1123 Subdomains
	dnsSubdomainAllowedCharacters = regexp.MustCompile("[^a-zA-Z0-9-]")
```

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,18 +235,25 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
+	case len(r.Reference.Name) > 0:
+		name = r.Reference.Name
 	}
-	return "", false
+
+	// in some certain situations, ref name might contain slashes or other symbols
+	// k8s naming goes along rfc952, which states that only alphanumeric characters,
+	// dots and dashes are allowed in a name
+	// bug example: https://bugzilla.redhat.com/show_bug.cgi?id=1970805
+	name = dnsNamePattern.ReplaceAllString(name, "-")
```
adambkaplan, line 254: Do we need to also check that we comply with the DNS subdomain length requirement?

If yes, add `validation.IsDNS1123Subdomain`

coreydaley: Here is an example of how we do the names for build volumes: https://github.com/openshift/builder/blob/master/pkg/build/builder/util.go#L158-L162

alicerum: @coreydaley this function does not replace nonalphanumerics, and adds hash

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,18 +236,45 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
-		return r.ObjectName, true
+
+	name = checkNameIsValidSubdomain(name)
+
+	return name, len(name) > 0
+}
+
+// checkNameIsValidSubdomain checks that ImageRef suggested name is a valid
+// subdomain name according to DNS1123
+func checkNameIsValidSubdomain(name string) string {
```
adambkaplan, line 258:
```suggestion
// toValidImageRefName checks that ImageRef suggested name is a valid
// subdomain name according to DNS1123, and transforms the name if necessary.
func toValidImageRefName(name string) string {
```

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,18 +236,45 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
+		name = r.Stream.Name
+	case len(r.ObjectName) > 0:
+		name = r.ObjectName
+	case len(r.Reference.Name) > 0:
+		name = r.Reference.Name
 	}
-	if len(r.ObjectName) > 0 {
-		return r.ObjectName, true
+
+	name = checkNameIsValidSubdomain(name)
```
adambkaplan, line 251:
```suggestion
	name = toValidImageRefName(name)
```

adambkaplan (commented):
/approve

Suggesting that the name of the function be clarified, otherwise code and tests look good.
Leaving approve so others can verify the function name suggestion.

pkg/helpers/newapp/app/imageref.go
```diff
@@ -230,18 +236,45 @@ func (r *ImageRef) SuggestName() (string, bool) {
...
 	}
-	if len(r.ObjectName) > 0 {
-		return r.ObjectName, true
+
+	name = toValidImageRefName(name)
+
+	return name, len(name) > 0
+}
+
+// checkNameIsValidSubdomain checks that ImageRef suggested name is a valid
```
adambkaplan, line 256: @alicerum missed the update to the godoc here.

alicerum: oops

alicerum: thanks for noticing!

adambkaplan (approved):
/lgtm

gabemontero:
@alicerum I would suggest researching on sippy (like I described in office hours a couple of weeks ago) the 

: [sig-cli][Feature:LegacyCommandTests][Disruptive][Serial] test-cmd: test/cmd/basicresources.sh 

and some of the other recent ones you've gotten with e2e-agnostic-cmd

either open new defects or post your failures in the existing ones

adambkaplan:
/hold

Something appears to be broken;

```
FAILURE after 1.000s: test/cmd/printer.sh:8: executing 'oc new-app openshift/nodejs' expecting success: the command returned the wrong error code
Standard output from the command:
warning: Cannot find git. Ensure that it is installed and in your path. Git is required to work with git repositories.

Standard error from the command:
W0924 14:21:58.794600     126 dockerimagelookup.go:300] container image remote registry lookup failed: you may not have access to the container image "docker.io/openshift/nodejs:latest"
error: multiple images or templates matched "openshift/nodejs"

Argument 'openshift/nodejs' was classified as an image, image~source, or loaded template reference.

The argument "openshift/nodejs" could apply to the following container images, OpenShift image streams, or templates:

* Template "nodejs-postgresql-example" in project "openshift"
  Use --template="openshift/nodejs-postgresql-example" to specify this image or template

* Template "nodejs-postgresql-persistent" in project "openshift"
  Use --template="openshift/nodejs-postgresql-persistent" to specify this image or template

* Image stream "jenkins-agent-nodejs" (tag "v4.0") in project "openshift"
  Use --image-stream="openshift/jenkins-agent-nodejs:v4.0" to specify this image or template

* Image stream "nodejs" in project "openshift"
  Use --image-stream="openshift/nodejs" --allow-missing-imagestream-tags to specify this image or template
```

soltysh (approved):
/hold cancel

PR #930, opened 2021-09-15
OWNERS: Prune crawford

wking:
Alex has left Red Hat :(.


sdodson:
Need to chase all who have left but since these PRs are already open I'll merge.

Issue #941, opened 2021-09-26
Must-Gather pod scheduled on NotReady nodes.


apoorvajagtap:
We have observed some intermittent behavior while the collection of must-gather, where the must-gather pod gets scheduled to one of the master nodes which is actually in `NotReady` state with `SchedulingDisabled`. (when --node-name is not specified).

This seems to be due to the [`tolerations`](https://github.com/openshift/oc/blob/master/pkg/cli/admin/mustgather/mustgather.go#L653-L657) set in the MG pod's template.
As it just specifies `Operator: Exists`, this check seems to overlook the situation when the node might have additional taints like `node.kubernetes.io/unreachable` or `node.kubernetes.io/not-ready` existing.

PR #942, opened 2021-09-26
Modified MG pod tolerations

apoorvajagtap:
Fixes: https://github.com/openshift/oc/issues/941
As per the current tolerations, `Operator: Exists`, the must-gather pod is spawned with following `spec.tolerations`:

> $ oc get pods must-gather-2cnh8 -ojson -n openshift-must-gather-strbd | jq .spec.tolerations
> [
>   {
>     "operator": "Exists"
>   }
> ]

After specifying the `Key: "node-role.kubernetes.io/master"`, the pod comes up with these additional tolerations (which helps to take the `not-ready` and `unreachable` conditions into consideration):

> $ oc get pods must-gather-lcq28 -ojson -n openshift-must-gather-j245c | jq .spec.tolerations
> [
>   {
>     "key": "node-role.kubernetes.io/master",
>     "operator": "Exists"
>   },
>   {
>     "effect": "NoExecute",
>     "key": "node.kubernetes.io/not-ready",
>     "operator": "Exists",
>     "tolerationSeconds": 300
>   },
>   {
>     "effect": "NoExecute",
>     "key": "node.kubernetes.io/unreachable",
>     "operator": "Exists",
>     "tolerationSeconds": 300
>   }
> ]

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -594,8 +594,12 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 	nodeSelector := map[string]string{
 		corev1.LabelOSStable: "linux",
 	}
+
+	tolerationKey := ""
+
 	if node == "" {
 		nodeSelector["node-role.kubernetes.io/master"] = ""
+		tolerationKey = "node-role.kubernetes.io/master"
 	}
 
 	ret := &corev1.Pod{
@@ -649,7 +653,7 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 			TerminationGracePeriodSeconds: &zero,
 			Tolerations: []corev1.Toleration{
 				{
-					Key:      "node-role.kubernetes.io/master",
+					Key:      tolerationKey,
 					Operator: "Exists",
 				},
 			},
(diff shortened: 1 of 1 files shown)


apoorvajagtap:
Hi @deads2k @soltysh, I would really like to know your inputs on this one, whenever possible.
In case I have missed any specific part of the workflow, please let me know and I'll abide by that. Thank you!

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -649,6 +649,7 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 			TerminationGracePeriodSeconds: &zero,
 			Tolerations: []corev1.Toleration{
 				{
+					Key:      "node-role.kubernetes.io/master",
 					Operator: "Exists",
```
soltysh, line 653: This particular use-case is to ensure that we match all possible taints, see Note section in https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ I'd suggest adding another toleration like the one you're proposing, but only when node wasn't specified. Your current code will break that last assumption.

soltysh (changes requested):

pkg/cli/admin/mustgather/mustgather.go
apoorvajagtap: Thank you @soltysh. I have modified the toleration considering the conditions. Let me know if this fits better or not.

soltysh (approved):
/lgtm
/approve

soltysh:
This PR is being reverted in #1003 since it's blocking k8s bump, also upon further investigation I see the current approach with tolerations as intended and we won't change it.

PR #924, opened 2021-09-10
Bug 1858760: allow Authorization Service URL to differ from Registry URL when retrieving tokens

atiratree:
Since we are trying to switch to podman backend (https://github.com/openshift/oc/pull/893) it makes sense to start using similar logic here (please see https://github.com/containers/image/blob/main/pkg/docker/config/config.go for reference). This fixes most of our auth issues.

We cannot fully switch to a podman backend yet, since this comes up with some incomptability issues: eg wrong auth config path order.

The implementation is also leveraging new CredentialStoreFactory interface, but for that we need the following PRs.

TODO:
- [x] https://github.com/openshift/library-go/pull/1224
- [x] https://github.com/openshift/oc/pull/949

PR #947, opened 2021-09-30
use CredentialStoreFactory instead of static CredentialStore

atiratree:
to use a newer approach as mentioned: https://github.com/openshift/library-go/pull/1205#issuecomment-925840746

TODO:

- [ ] https://github.com/openshift/oc/pull/949


pkg/cli/image/manifest/dockercredentials/credentials.go
```diff
@@ -15,29 +15,22 @@ import (
...
-		DockerKeyring:     keyring,
-		RefreshTokenStore: registryclient.NewRefreshTokenStore(),
+// NewCredentialStore creates a new credential store that uses the authFilePath file
+// (defaults to a docker locations) to find a valid
+// authentication for registry targets.
+func NewCredentialStore(authFilePath string) (auth.CredentialStore, error) {
+	var cfg credentialprovider.DockerConfig
+	var err error
+
+	if authFilePath != "" {
```
atiratree, line 25: I can still move this logic to the Basic method but that would mean we would have to load the config twice because we need to notify the user when invalid config is specified (on CLI).

@dmage do you still think this is a problem? From what I see `CredentialStoreFor` is called just before making the connection

atiratree:
unit tests failing because of openshift/api update. https://github.com/openshift/oc/pull/949 needs to merge first

atiratree:
closing in favor of https://github.com/openshift/oc/pull/924

PR #952, opened 2021-10-07
pkg/cli/admin/upgrade: Indent multi-line messages

wking:
Replace output like:

```console
$ oc adm upgrade
Cluster version is 4.9...

Upgradeable=False

  Reason: MultipleReasons
  Message: Cluster should not be upgraded between minor versions for multiple reasons: ClusterVersionOverridesSet,Testing
* Disabling ownership via cluster version overrides prevents upgrades. Please remove overrides before continuing.
* Cluster operator testing should not be upgraded between minor versions: Testing upgradeable https://example.com/a.

warning: Cannot display available updates:
  Reason: NoChannel
  Message: The update channel has not been configured.
```

With:

```
...
  Message: Cluster should not be upgraded between minor versions for multiple reasons: ClusterVersionOverridesSet,Testing
  * Disabling ownership via cluster version overrides prevents upgrades. Please remove overrides before continuing.
  * Cluster operator testing should not be upgraded between minor versions: Testing upgradeable https://example.com/a.
...
```

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -314,7 +314,7 @@ func (o *Options) Run() error {
 				prefix = c.Message
 			}
 			if len(c.Message) > 0 {
-				return fmt.Errorf("%s:\n\n  Reason: %s\n  Message: %s\n\n", prefix, c.Reason, c.Message)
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

PR #957, opened 2021-10-20
Bug 2010258: add printable-group-name validation

ibihim:
# What

It prevents users from creating groups with unprintable names.

# Why

- It is impossible to delete such groups after the fact.
- Some users seem to have done so, so it would be useful to prevent them from doing so ever again: https://bugzilla.redhat.com/show_bug.cgi?id=2010258

# Notes

There will be also an equivalent for the server side.

# Questions

I didn't see any tests in the group pkg, is it optional?

### DIFF
--- a/pkg/cli/admin/groups/new/new.go
+++ b/pkg/cli/admin/groups/new/new.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"unicode"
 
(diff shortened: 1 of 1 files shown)


ibihim:
@soltysh, I will close this PR as I am currently working with the hypothesis from @sttts, that the bug is rather on the "can't delete" side of things. Thank you a lot for your support.

PR #958, opened 2021-10-27
Introduce an option to retrieve rotated log files for a pod

kyrtapz:
Currently the pod logs API does not expose [rotated logs](https://github.com/kubernetes/kubernetes/blob/b0bc8adbc2178e15872f9ef040355c51c45d04bb/pkg/kubelet/kuberuntime/logs/logs.go#L42). This means that we are not getting all the pod logs that might be available on the node.

The introduced hidden flag enables getting the rotated log files available on the node for a specific container.
It uses the same approach as `node-logs`, it reads the HTML and extracts file names that represent specific log files.

Example:
```
./oc adm inspect ns/default --rotated-pod-logs

tree namespaces/default/pods/bbx/bbx/bbx/
namespaces/default/pods/bbx/bbx/bbx/
└── logs
    ├── current.log
    ├── previous.insecure.log
    ├── previous.log
    └── rotated
        ├── 0.log.20211110-131044.gz
        └── 0.log.20211110-131209
```

This flag can be removed once kubernetes supports getting the rotated logs through the API:
https://github.com/kubernetes/kubernetes/issues/59902

Signed-off-by: Patryk Diak <pdiak@redhat.com>

### DIFF
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -75,6 +75,7 @@ type InspectOptions struct {
 	namespace      string
 	sinceTime      string
 	allNamespaces  bool
+	rotatedPodLogs bool
 	sinceInt       int64
 	sinceTimestamp metav1.Time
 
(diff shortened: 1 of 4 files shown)


pkg/cli/admin/inspect/inspect.go
```diff
@@ -113,6 +114,7 @@ func NewCmdInspect(streams genericclioptions.IOStreams) *cobra.Command {
 	cmd.Flags().StringVar(&o.DestDir, "dest-dir", o.DestDir, "Root directory used for storing all gathered cluster operator data. Defaults to $(PWD)/inspect.local.<rand>")
 	cmd.Flags().StringVar(&o.eventFile, "events-file", o.eventFile, "A path to an events.json file to create a HTML page from")
 	cmd.Flags().BoolVarP(&o.allNamespaces, "all-namespaces", "A", o.allNamespaces, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
+	cmd.Flags().BoolVar(&o.podLogFiles, "pod-log-files", o.podLogFiles, "If present, retrieve all log files(including rotated logs) that are available for selected pods.")
```
soltysh, line 117: How this flag plays with the next two? I don't see any interaction, but that will be confusing to a lot of users. You should properly wire both `--since` and `--since-time` in this logic, especially that the pod logs can be big at times.

soltysh: Make it explicit in the flag description that this is best-effort and doesn't guarantee any value.

contrib/completions/bash/oc
```diff
@@ -2545,6 +2545,8 @@ _oc_adm_inspect()
     two_word_flags+=("--events-file")
     local_nonpersistent_flags+=("--events-file")
     local_nonpersistent_flags+=("--events-file=")
+    flags+=("--pod-log-files")
```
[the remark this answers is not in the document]
Kyrtapz: I changed the approach to download the additional rotated logs instead of all log files, renamed the flag to `--rotated-pod-logs`

pkg/cli/admin/inspect/pod.go
```diff
@@ -78,6 +87,58 @@ func filterContainerLogsErrors(err error) error {
 	return err
 }
 
+func (o *InspectOptions) gatherContainerLogFiles(destDir string, pod *corev1.Pod, container *corev1.Container) error {
```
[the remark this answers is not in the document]
Kyrtapz: I changed the approach and used `x/net` to make it a little more readable.

pkg/cli/admin/inspect/inspect.go
Kyrtapz: Supporting `--since/--since-time` would require implementing what is there in kubelet and possibly extracting multiple archives.
I've hidden the option as discussed offline and provided a better help message.

soltysh (approved):
/lgtm
/approve

PR #961, opened 2021-11-03
Display conditional and not-recommended versions

Davoska:
Implementing the `--include-not-recommended` option for the command `oc adm upgrade` proposed in the enhancement for [Targeted Update Edge Blocking](https://github.com/openshift/enhancements/blob/master/enhancements/update/targeted-update-edge-blocking.md).

Displaying a prompt if there are available not-recommended versions for the cluster but the user uses the `oc adm upgrade` command.
Modifying the printed-out versions of the command `oc adm upgrade` to be indented, adding a few new lines for an aesthetic purpose, and changing "Available updates" to "Recommended updates".
Changing the order of the printed-out version using the `oc adm upgrade` command to a decreasing SemVer order.

Example of having no not-recommended updates and using the `oc adm upgrade` command.
```bash
$ ./oc adm upgrade
Cluster version is 4.7.30

Upstream is unset, so the cluster will use an appropriate default.
Channel: stable-4.7 (available channels: candidate-4.7, candidate-4.8, fast-4.7, fast-4.8, stable-4.7, stable-4.8)

Recommended updates:

  VERSION     IMAGE
  4.7.34      quay.io/openshift-release-dev/ocp-release@sha256:e121118406de30f9a92168808979d9363c1a576401c399bf6f528fb47c48b16c
  4.7.33      quay.io/openshift-release-dev/ocp-release@sha256:d3c0d73bd5c519f8969b7d598d287a85c69eb1979a7f7fb78508ca98929fd548
  4.7.32      quay.io/openshift-release-dev/ocp-release@sha256:96f00e5b0cde92488c42cd9395ff5a755c77f4273eb10aa83bc2707a8badf93f
  4.7.31      quay.io/openshift-release-dev/ocp-release@sha256:4046047beed84bbba2c1762f130c402f9d05d584cb1dc3e8440f5809b2bb587e
```

Example of having no not-recommended updates and using the `oc adm upgrade --include-not-recommended` command.
```
$ ./oc adm upgrade --include-not-recommended
Cluster version is 4.7.30

Upstream is unset, so the cluster will use an appropriate default.
Channel: stable-4.7 (available channels: candidate-4.7, candidate-4.8, fast-4.7, fast-4.8, stable-4.7, stable-4.8)

Recommended updates:

  VERSION     IMAGE
  4.7.34      quay.io/openshift-release-dev/ocp-release@sha256:e121118406de30f9a92168808979d9363c1a576401c399bf6f528fb47c48b16c
  4.7.33      quay.io/openshift-release-dev/ocp-release@sha256:d3c0d73bd5c519f8969b7d598d287a85c69eb1979a7f7fb78508ca98929fd548
  4.7.32      quay.io/openshift-release-dev/ocp-release@sha256:96f00e5b0cde92488c42cd9395ff5a755c77f4273eb10aa83bc2707a8badf93f
  4.7.31      quay.io/openshift-release-dev/ocp-release@sha256:4046047beed84bbba2c1762f130c402f9d05d584cb1dc3e8440f5809b2bb587e

No updates which are not recommended based on your cluster configuration are available.
```

Example of having not-recommended updates available and using the `oc adm upgrade` command.
```
$ ./oc adm upgrade 
Cluster version is 4.10.0-0.ci-2021-11-03-034804

Upstream is unset, so the cluster will use an appropriate default.
Channel: stable-4.7

Recommended updates:

  VERSION     IMAGE
  4.7.34      quay.io/openshift-release-dev/ocp-release@sha256:e121118406de30f9a92168808979d9363c1a576401c399bf6f528fb47c48b16c
  4.7.33      quay.io/openshift-release-dev/ocp-release@sha256:d3c0d73bd5c519f8969b7d598d287a85c69eb1979a7f7fb78508ca98929fd548
  4.7.32      quay.io/openshift-release-dev/ocp-release@sha256:96f00e5b0cde92488c42cd9395ff5a755c77f4273eb10aa83bc2707a8badf93f
  4.7.31      quay.io/openshift-release-dev/ocp-release@sha256:4046047beed84bbba2c1762f130c402f9d05d584cb1dc3e8440f5809b2bb587e

Additional updates which are not recommended based on your cluster configuration are available, to view those re-run the command with --include-not-recommended.
```

Example of having not-recommended updates available and using the `oc adm upgrade --include-not-recommended` command.
```
$ ./oc adm upgrade --include-not-recommended
Cluster version is 4.10.0-0.ci-2021-11-03-034804

Upstream is unset, so the cluster will use an appropriate default.
Channel: stable-4.7

Recommended updates:

  VERSION     IMAGE
  4.7.34      quay.io/openshift-release-dev/ocp-release@sha256:e121118406de30f9a92168808979d9363c1a576401c399bf6f528fb47c48b16c
  4.7.33      quay.io/openshift-release-dev/ocp-release@sha256:d3c0d73bd5c519f8969b7d598d287a85c69eb1979a7f7fb78508ca98929fd548
  4.7.32      quay.io/openshift-release-dev/ocp-release@sha256:96f00e5b0cde92488c42cd9395ff5a755c77f4273eb10aa83bc2707a8badf93f
  4.7.31      quay.io/openshift-release-dev/ocp-release@sha256:4046047beed84bbba2c1762f130c402f9d05d584cb1dc3e8440f5809b2bb587e

Supported but not recommended updates:

  Version: 4.7.4
  Image: example image
  Recommended: Unknown
  Reason: MultipleReasons
  Message:
    Clusters on vSphere Virtual Hardware Version 14 and later may experience cross-node networking issues. https://access.redhat.com/solutions/5896081
    
    vSphere clusters leveraging the vSphere cloud provider may lose node names which can have serious impacts on the stability of the control plane and workloads. https://bugzilla.redhat.com/show_bug.cgi?id=1942207#c3
    
    On clusters with a Proxy configured, the authentication operator may keep many oauth-server connections open... operator and router pods. https://bugzilla.redhat.com/show_bug.cgi?id=1941840#c33
    
  Version: 4.6.99-example
  Image: example image
  Recommended: False
  Reason: ExampleReason
  Message: Example message

  Version: 4.6.30
  Image: example image
  Recommended: False
  Reason: ExampleReason
  Message: Example message
```

Pull request related to https://issues.redhat.com/browse/OTA-482


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -431,21 +431,22 @@ func updateIsEquivalent(a configv1.Update, b configv1.Release) bool {
 	}
 }
 
-// sortSemanticVersions sorts the input slice in increasing order.
-func sortSemanticVersions(versions []configv1.Release) {
+// sortReleasesBySemanticVersions sorts the input slice in decreasing order.
```
vrutkovs, line 475: Not sure why would we want to change the sort order - just for the sake of using `cv.Status.AvailableUpdates[0]` instead of `len(cv.Status.AvailableUpdates)-1`?

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -340,20 +340,20 @@ func (o *Options) Run() error {
...
+				fmt.Fprintf(o.Out, "Channel: %s\n\n", cv.Spec.Channel)
 			}
 		}
 
 		if len(cv.Status.AvailableUpdates) > 0 {
-			fmt.Fprintf(o.Out, "Available Updates:\n\n")
-			w := tabwriter.NewWriter(o.Out, 0, 2, 1, ' ', 0)
-			fmt.Fprintf(w, "VERSION\tIMAGE\n")
+			fmt.Fprintf(o.Out, "Recommended updates:\n\n")
+			w := tabwriter.NewWriter(o.Out, 14, 2, 1, ' ', 0)
```
vrutkovs, line 353: Not sure what `14` means here - is this padding?

vrutkovs (changes requested):
Looks fine overall, see comments inline.

Some note on commits - they should describe why the change is made, not what change is done (its visible in diff)

Davoska:
/hold
Making a few modifications and improving the commits.

Davoska:
I have tried to improve the commit messages.

Added new changes in `/contrib/completions/bash/oc` generated by the `/hack/update-generated-completions.sh` script in the commit https://github.com/openshift/oc/pull/961/commits/95d742fe4bae0d97a400053937dc134dcbd586d7.

Dropped the initial changes for modifying the sorting functions to sort in a decreasing order.
Added functions to reverse the sorted slices.

pkg/cli/admin/upgrade/upgrade.go
Davoska: I have dropped the initial changes for modifying the sorting functions.
(The initial thinking was to have a decreasing order of the semantic versions. And then have the following invalid semantic versions in ascending lexicographic order. Which would not be possible with reversing the current sorting function returned slice.)
And I have added functions for reversing the relevant slices to be in decreasing order.

pkg/cli/admin/upgrade/upgrade.go
Davoska: I have initially forgotten to add this change to the commit message.

The argument of `14` represents the `minwidth` ([minimal cell width including any padding](https://pkg.go.dev/text/tabwriter#NewWriter)). Which results in having the column have bigger minimal width than before. And since the ordinal versions have an approximate length of 6. The columns seem further apart.

But for longer version lengths, there won't be an unnecessary big empty space.
But a similar behavior could be achieved with fixed bigger `padding`.

vrutkovs: Ooh, I see, that makes sense

Lets add a code comment above this (`git blame` would show this, but the fewer WTF per code line the better)

Davoska:
Improved the commit message for the commit https://github.com/openshift/oc/pull/961/commits/30f708aadf0ead208b5d3888d80056a92a378036.

Added a comment in the source code as discussed in https://github.com/openshift/oc/pull/961/files#r742706668.

pkg/cli/admin/upgrade/upgrade.go
Davoska: I have added the comment:
`// set the minimal cell width to 14 to have a larger space between the columns for shorter versions`

In the new commit https://github.com/openshift/oc/commit/30f708aadf0ead208b5d3888d80056a92a378036.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -341,39 +343,81 @@ func (o *Options) Run() error {
...
-			w := tabwriter.NewWriter(o.Out, 0, 2, 1, ' ', 0)
-			fmt.Fprintf(w, "VERSION\tIMAGE\n")
+			fmt.Fprintf(o.Out, "Recommended updates:\n\n")
+			// set the minimal cell width to 14 to have a larger space between the columns for shorter versions
+			w := tabwriter.NewWriter(o.Out, 14, 2, 1, ' ', 0)
+			fmt.Fprintf(w, "  VERSION\tIMAGE\n")
 			// TODO: add metadata about version
-			sortSemanticVersions(cv.Status.AvailableUpdates)
+			sortReleasesBySemanticVersions(cv.Status.AvailableUpdates)
+			reverseReleases(cv.Status.AvailableUpdates)
```
wking, line 359: Instead of sorting and then reversing, can we just flip the sense of the sort?  `sortReleasesBySemanticVersions` is an internal helper with this one use-case, and we can just adjust the godocs to say it's becoming `decreasing`.  That would allow this call site to say "sort these however we like them sorted" without having to worry about whether it was increasing or decreasing.

vrutkovs (approved):
/approve
/lgtm

Looks good to me, lets sort out latest comments from Trevor

Davoska:
In the new commit https://github.com/openshift/oc/pull/961/commits/ae4725ef19378254d2f90f2a47c8762a9d126799.

- I have modified the function `sortReleasesBySemanticVersions` to sort in decreasing order as discussed in https://github.com/openshift/oc/pull/961#discussion_r747010161 as opposed to reversing the slice to a decreasing order.
- Modified the use-case of the function accordingly.
- Modified the `pkg/cli/admin/upgrade/upgrade_test.go` accordingly.

I have also dropped the reversing of the slice of `conditionalUpdates`. And I have modified the `sortConditionalUpdatesBySemanticVersions` function to sort in decreasing order in the new commit https://github.com/openshift/oc/pull/961/commits/d747e5b6fac643db738e665da857d2d891e9bda6.

pkg/cli/admin/upgrade/upgrade.go
Davoska: Yes, that seems very reasonable. 
Fixed in the new commits ae4725e and d747e5b (see: https://github.com/openshift/oc/pull/961#issuecomment-968759024).

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -341,39 +343,79 @@ func (o *Options) Run() error {
...
+		return
+	}
+
+	fmt.Fprintf(o.Out, "  Version: %s\n  Image: %s\n", update.Release.Version, update.Release.Image)
+	fmt.Fprintf(o.Out, "  Recommended: %s\n  Reason: %s\n", c.Status, c.Reason)
+	if c.Reason == "MultipleReasons" {
+		indented_message := "    " + strings.Replace(c.Message, "\n", "\n    ", -1)
+		fmt.Fprintf(o.Out, "  Message:\n%s\n", indented_message)
+	} else {
+		fmt.Fprintf(o.Out, "  Message: %s\n", c.Message)
```
wking, line 415: `MultipleReasons` is more likely to be multi-line, but even single-reason messages could be multi-line.  Can we use `strings.ReplaceAll(c.Message, "\n", "\n    ")`, to follow the pattern from #952, regardless of `Reason`?

Davoska:
I have changed the formatting of the printed conditional update's `Message` to be the same regardless of its `Reason` in the new commit https://github.com/openshift/oc/pull/961/commits/73ac3b3ac24ae78c65b1a1a9b44bcfe0cae46ced (as discussed in https://github.com/openshift/oc/pull/961#discussion_r749889906).

I have also moved the blank line that was separating `Recommended updates` from its previous block.  
The blank line was previously associated with the outgoing block, and it would be better for it to be associated with the incoming block. Fixed in the new commit https://github.com/openshift/oc/pull/961/commits/21f3636bd8ead6e9f574b0f2d464887f67f00309 (as discussed in https://github.com/openshift/oc/pull/961#discussion_r747015207)

pkg/cli/admin/upgrade/upgrade.go
Davoska: Yes, it's definitely better to follow the same style.

Since the previous function `printNotRecommendedUpdate` would now have no logic, I have moved its modified body to the place of its original calling.  Every other printing is in the function `Run` also. Let me know if I should change that.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -341,39 +343,81 @@ func (o *Options) Run() error {
 				fmt.Fprintf(o.Out, "Upstream: %s\n", cv.Spec.Upstream)
 			}
 			if len(cv.Status.Desired.Channels) > 0 {
-				fmt.Fprintf(o.Out, "Channel: %s (available channels: %s)\n", cv.Spec.Channel, strings.Join(cv.Status.Desired.Channels, ", "))
+				fmt.Fprintf(o.Out, "Channel: %s (available channels: %s)\n\n", cv.Spec.Channel, strings.Join(cv.Status.Desired.Channels, ", "))
 			} else {
-				fmt.Fprintf(o.Out, "Channel: %s\n", cv.Spec.Channel)
+				fmt.Fprintf(o.Out, "Channel: %s\n\n", cv.Spec.Channel)
```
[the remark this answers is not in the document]
Davoska: Ah, yes. I have no problem changing that. And...

>  it would be good to have blank lines associated with the incoming block instead of the outgoing block

Is definitely a valid point, and I would also agree it is a better way.

I have removed the previous blank lines from the `Channels` block, and added a blank line in: `fmt.Fprintf(o.Out, "\nRecommended updates:\n\n")`.
Fixed in the new commit 21f3636.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -348,32 +350,59 @@ func (o *Options) Run() error {
...
 				fmt.Fprintf(o.ErrOut, "warning: Cannot display available updates:\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 			} else {
 				fmt.Fprintf(o.Out, "No updates available. You may force an upgrade to a specific release image, but doing so may not be supported and may result in downtime or data loss.\n")
 			}
 		}
 
+		if o.IncludeNotRecommended {
+			if containsNotRecommendedUpdate(cv.Status.ConditionalUpdates) {
+				sortConditionalUpdatesBySemanticVersions(cv.Status.ConditionalUpdates)
+				fmt.Fprintf(o.Out, "\nSupported but not recommended updates:\n\n")
```
wking, line 377: I think you want one less trailing `\n` here and one more below with `\n  Version...`, to give a blank line between each target-version block.

Davoska: Yes. I have forgotten I was testing with messages that were all ending with the character `\n`.

Is there any guarantee that the messages end or they do not end with the character `\n`?

Davoska: If we can't be sure,  I can add the [func TrimSuffix](https://pkg.go.dev/strings#TrimSuffix) for the `Message` and add the newlines as you have proposed.

Davoska: I have modified the new-lines as you have proposed and added `strings.ReplaceAll(strings.TrimSpace(c.Message), "\n", "\n  ")` when printing update's `Message`.

Fixed in https://github.com/openshift/oc/pull/961/commits/8b225ea2e172938fb0bf6e250d40185c920a85a2.

wking (approved):
/lgtm

PR #850, opened 2021-06-14
[must-gather] Add options for multi-node / host-network capture

alexanderConstantinescu:
This is must-gather feature enhancement coming from https://issues.redhat.com/browse/SDN-1760. 

In networking (possibly other components as well) we would need an ad-hoc ability to run must-gather commands against all/set of cluster nodes, as to facilitate customer/support in tasks that sometimes require several steps / manual interactions to perform. The most prevalent example of this is packet captures using `tcpdump`. Currently to run a packet capture on multiple nodes a customer/user needs to:

- Open X shell sessions
- In each: exec / run a pod with packet capture capabilities, specifying a file name and location for each
- Open a X+1 shell session which will be used to copy all `.pcap` files from the pods used in step 2, to the user's local location, by doing `oc cp` for each pod in step 2.  

Given that the [networking-dedicate debug image](https://github.com/openshift/network-tools/) already contains `tcpdump`: this process could be reduced to doing:

`oc adm must-gather --dest-dir ./tmp --source-dir '/tmp/tcpdump/' --image quay.io/openshift/origin-network-tools:latest --node-selector 'kubernetes.io/os=linux,node-role.kubernetes.io/master' --host -- timeout 30 tcpdump -i any -w /tmp/tcpdump/\$POD_NAME-%Y-%m-%dT%H:%M:%S.pcap -W 1 -G 300`

That single command would allow a user to run a packet capture on selected nodes (linux + master nodes) for 30 seconds and download all of the generated files. That is a significant improvement in usability from what currently exists.

PR #962, opened 2021-11-05
Bug 2020377: add linux capability for host network packet captures

alexanderConstantinescu:
This PR addresses two issues: 

- the bug mentioned in the title, i.:e currently we can't run packet captures using must-gather because we are missing the proper security capability to do so. I am limiting the capability to `CAP_NET_RAW` which will allow a user to perform packet captures, while also not granting too broad permissions.  
- nits from my previous PR: https://github.com/openshift/oc/pull/850 

/assign @soltysh


soltysh (approved):
/lgtm
/approve

PR #965, opened 2021-11-08
mustgather: add NodeName to destDir folder name

npinaeva:
Since we've added option to run must-gather container on multiple nodes
(https://github.com/openshift/oc/pull/850, --node-selector), folder name
should be different for pods running on different nodes with the same
image name.

Signed-off-by: Nadia Pinaeva <npinaeva@redhat.com>

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -452,7 +452,13 @@ func (o *MustGatherOptions) logTimestamp() error {
 func (o *MustGatherOptions) copyFilesFromPod(pod *corev1.Pod) error {
 	streams := o.IOStreams
 	streams.Out = newPrefixWriter(streams.Out, fmt.Sprintf("[%s] OUT", pod.Name))
-	destDir := path.Join(o.DestDir, regexp.MustCompile("[^A-Za-z0-9]+").ReplaceAllString(pod.Status.ContainerStatuses[0].ImageID, "-"))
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -452,7 +452,7 @@ func (o *MustGatherOptions) logTimestamp() error {
 func (o *MustGatherOptions) copyFilesFromPod(pod *corev1.Pod) error {
 	streams := o.IOStreams
 	streams.Out = newPrefixWriter(streams.Out, fmt.Sprintf("[%s] OUT", pod.Name))
-	destDir := path.Join(o.DestDir, regexp.MustCompile("[^A-Za-z0-9]+").ReplaceAllString(pod.Status.ContainerStatuses[0].ImageID, "-"))
+	destDir := path.Join(o.DestDir, regexp.MustCompile("[^A-Za-z0-9]+").ReplaceAllString(pod.Status.ContainerStatuses[0].ImageID+pod.Spec.NodeName, "-"))
```
soltysh, line 455: Can we make this optional when you actually specify node selector such that the structure will look like:
- destDir/
    - node1/image-name
    - node2/image-name
instead of appending those. It's already pretty lengthy name and adding node name at the end won't be helpful :/

soltysh (changes requested):

pkg/cli/admin/mustgather/mustgather.go
npinaeva: Done, ptal

soltysh (approved):
/lgtm
/approve
/retest-required

PR #973, opened 2021-11-18
Fix docs for oc adm policy add-role-to-user vs oc policy add-role-to-user

soltysh:
/assign @bergerhoffer

### DIFF
--- a/pkg/cli/policy/policy.go
+++ b/pkg/cli/policy/policy.go
@@ -18,10 +18,10 @@ func NewCmdPolicy(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobr
 	}
 
 	cmds.AddCommand(adminpolicy.NewCmdWhoCan(f, streams))
-	cmds.AddCommand(adminpolicy.NewCmdSccSubjectReview(f, streams))
-	cmds.AddCommand(adminpolicy.NewCmdSccReview(f, streams))
+	cmds.AddCommand(adminpolicy.NewCmdSccSubjectReview(f, streams, false))
+	cmds.AddCommand(adminpolicy.NewCmdSccReview(f, streams, false))
 
-	cmds.AddCommand(adminpolicy.NewCmdAddRoleToUser(f, streams))
+	cmds.AddCommand(adminpolicy.NewCmdAddRoleToUser(f, streams, false))
 	cmds.AddCommand(adminpolicy.NewCmdRemoveRoleFromUser(f, streams))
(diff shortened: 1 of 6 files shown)


pkg/cli/policy/policy.go
```diff
@@ -21,7 +21,7 @@ func NewCmdPolicy(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobr
 	cmds.AddCommand(adminpolicy.NewCmdSccSubjectReview(f, streams))
 	cmds.AddCommand(adminpolicy.NewCmdSccReview(f, streams))
```
bergerhoffer, line 22: It looks like we'll need to do the same thing for `oc adm policy scc-review` and `oc adm policy scc-subject-review`. They are still showing `oc policy` in their examples:

```
[ahoffer@ahoffer oc] (pr-973) $ ./oc adm policy scc-review --help
Check which service account can create a pod. The pod is inferred from the pod template spec in the provided resource.
If no service account is provided the one specified in podTemplateSpec.spec.serviceAccountName is used, unless it is
empty, in which case "default" is used. If service accounts are provided, the podTemplateSpec.spec.serviceAccountName is
ignored.

Usage:
  oc adm policy scc-review [flags]

Examples:
  # Check whether service accounts sa1 and sa2 can admit a pod with a template pod spec specified in my_resource.yaml
  # Service Account specified in myresource.yaml file is ignored
  oc policy scc-review -z sa1,sa2 -f my_resource.yaml
...
```
and 

```
[ahoffer@ahoffer oc] (pr-973) $ ./oc adm policy scc-subject-review --help
Check whether a user, service account or group can create a pod. It returns a list of security context constraints that
will admit the resource. If user is specified but not groups, it is interpreted as "what if the user is not a member of
any groups". If user and groups are empty, then the check is performed using the current user.

Usage:
  oc adm policy scc-subject-review [flags]

Examples:
  # Check whether user bob can create a pod specified in myresource.yaml
  oc policy scc-subject-review -u bob -f myresource.yaml
...
```

bergerhoffer (commented):
Thanks a lot for this! One comment about needing to do this for `scc-review` and `scc-subject-review` too.

None of the other shared subcommands have examples, so they don't have this problem yet. But if any examples are added in the future for them, then we'd need to do this there as well.

pkg/cli/policy/policy.go
soltysh: Good catch - updated.

bergerhoffer (commented):
Thanks a lot, this looks great!

/lgtm

PR #976, opened 2021-11-19
Drop patches for apimachinery and client-go forks

soltysh:
I've manged to get rid of the patches from:
1. https://github.com/openshift/kubernetes-apimachinery/commits/oc-4.9-kubernetes-1.22.0-rc.0
    SCC are currently a CRD so that's not needed anymore. 
2. https://github.com/openshift/kubernetes-client-go/commits/oc-4.9-kubernetes-1.22.0-rc.0
    I've move the carry patch over to the shim, and newer client-go has the fix from you Filip. 

/assign @atiratree


atiratree:
@soltysh unit tests failing

soltysh:
> @soltysh unit tests failing

Should be fixed now, it was caused by restmapper used in tests in `ObjectTracker`, which doesn't handle external types quite fully. It can be easily fixed when instead of `Add` being used when creating fake client you explicitly call `Create` on the resources, see last commit.

atiratree:
I see
/lgtm

PR #977, opened 2021-11-22
Adding an example for oc adm upgrade

bergerhoffer:
Commands need an example in order to show up in the autogenerated CLI docs. Adding an example for `oc adm upgrade` so that it'll show up.


soltysh (approved):
/lgtm
/approve

PR #893, opened 2021-08-09
Bug 1982498: default to podman credential configs

atiratree:
we should first try podman location for detecting credentials and then check the docker ones 

relevant docs:
https://man.archlinux.org/man/community/containers-common/containers-auth.json.5.en
http://docs.podman.io/en/latest/markdown/podman-login.1.html

PR #983, opened 2021-12-01
fix registry-config messages and imports

atiratree:
followup to https://github.com/openshift/oc/pull/893


soltysh (approved):
/lgtm
/approve

PR #986, opened 2021-12-03
Adding the flag --allow-not-recommended to oc adm upgrade

LalatenduMohanty:
So that users can upgrade to a version listed in Supported but not
recommended updates. The list of versions can be seen with oc adm upgrade --include-not-recommended
For details refer to https://github.com/openshift/enhancements/blob/2cc2d9b331532c852878a7c793f3a754914c824e/enhancements/update/targeted-update-edge-blocking.md

https://issues.redhat.com/browse/OTA-528

Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -96,6 +96,7 @@ func New(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command
 	flags.BoolVar(&o.AllowExplicitUpgrade, "allow-explicit-upgrade", o.AllowExplicitUpgrade, "Upgrade even if the upgrade target is not listed in the available versions list.")
 	flags.BoolVar(&o.AllowUpgradeWithWarnings, "allow-upgrade-with-warnings", o.AllowUpgradeWithWarnings, "Upgrade even if an upgrade is in process or a cluster error is blocking the update.")
 	flags.BoolVar(&o.IncludeNotRecommended, "include-not-recommended", o.IncludeNotRecommended, "Display additional updates which are not recommended based on your cluster configuration.")
+	flags.BoolVar(&o.AllowNotRecommended, "allow-not-recommended", o.AllowNotRecommended, "Allows upgrade to a version when it is supported but not recommended for updates")
 
 	cmd.AddCommand(channel.New(f, streams))
 
@@ -114,6 +115,7 @@ type Options struct {
 	Force                    bool
 	Clear                    bool
 	IncludeNotRecommended    bool
+	AllowNotRecommended      bool
 
 	Client configv1client.Interface
 }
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -247,13 +249,29 @@ func (o *Options) Run() error {
...
 				}
 			}
+			if o.AllowNotRecommended {
+				for _, upgrade := range cv.Status.ConditionalUpdates {
+					if c := findCondition(upgrade.Conditions, "Recommended"); c != nil && c.Status != metav1.ConditionTrue {
+						if upgrade.Release.Version == o.To {
+							update = &configv1.Update{
+								Version: upgrade.Release.Version,
+								Image:   upgrade.Release.Image,
+							}
```
wking, line 259: We should add additional complaining here like:

```console
fmt.Fprintf(o.ErrOut, "warning: --allow-not-recommended is bypassing %s=%s %s: %s\n", c.Type, c.Status, c.Reason, c.Message)

File: pkg/cli/admin/upgrade/upgrade.go [hunk 2]
```diff
@@ -247,13 +249,29 @@ func (o *Options) Run() error {
...
+			}
 			if update == nil {
 				if len(cv.Status.AvailableUpdates) == 0 {
 					if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.RetrievedUpdates); c != nil && c.Status == configv1.ConditionFalse {
 						return fmt.Errorf("Can't look up image for version %s. %v", o.To, c.Message)
 					}
 					return fmt.Errorf("No available updates, specify --to-image or wait for new updates to be available")
 				}
+				if o.AllowNotRecommended {
+					return fmt.Errorf("The update %s is not availale as conditional  updates or supported but not recommended updates", o.To)
```
wking, line 273: spelling/rewording: `The update %s is not one of the conditional or recommended updates`.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -247,13 +249,29 @@ func (o *Options) Run() error {
...
+				}
 				return fmt.Errorf("The update %s is not one of the available updates: %s", o.To, strings.Join(versionStrings(cv.Status.AvailableUpdates), ", "))
```
wking, line 275: "available" here should change to "recommended"

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -247,13 +249,29 @@ func (o *Options) Run() error {
...
+						}
+					}
+				}
+			}
 			if update == nil {
 				if len(cv.Status.AvailableUpdates) == 0 {
 					if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.RetrievedUpdates); c != nil && c.Status == configv1.ConditionFalse {
 						return fmt.Errorf("Can't look up image for version %s. %v", o.To, c.Message)
 					}
 					return fmt.Errorf("No available updates, specify --to-image or wait for new updates to be available")
```
wking, line 270: We probably have to shard this out into a switch or something, because I see more cases now than we're currently handling:

* If `RetrievedUpdates!=True`, then we should complain about that.
* If there are no recommended or conditional updates, then `--to-image` or wait.
* If there are no recommended updates, but there are conditional updates, and...
  * `--allow-not-recommended`, so the user already knows they're living dangerously, and we just tell them that their requested `To` wasn't found, and maybe list the available conditional choices so they can easily fix typos.
  * no `--allow-not-recommended`, so `--to-image` or wait, but also possibly mention viewing with `--include-not-recommended` to see the conditional updates.
* If there are recommended updates but no conditional updates, then list the recommended choices so they can easily fix typos.
* If there are both recommended and conditional updates, and...
  * `--allow-not-recommended`, so the user already knows they're living dangerously, and we just tell them that their requested `To` wasn't found, and list the available choices (recommended and conditional) so they can easily fix typos.
  * no `--allow-not-recommended`, list the recommended choices so the can easily fix typos.

If that sounds sane, one way to simplify in this block would be to collect a `[]string` slice of choices as we walk above, so by the time we get down here we can use a `Join` to easily format the choices regardless of `--allow-not-recommended`-ness (or see that that choice set is empty).

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: Unless we are  waiting for user confirmation to wait before moving to the version the additional complaining does not give any significantly value.

LalatenduMohanty: I could come up with `fmt.Fprintf(o.ErrOut, "warning: with --allow-not-recommended you have accepted the risk with %s and bypassing %s=%s %s: %s\n", o.To, c.Type, c.Status, c.Reason, c.Message)` which I felt would be more appropriate .

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: For `--to-image` there is a different code block below the current one. i.e. `if len(o.ToImage) > 0 {` which looks fine to me.

LalatenduMohanty: So I trying to simplify 
```
	if update == nil {
				if len(cv.Status.AvailableUpdates) == 0 {
					if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.RetrievedUpdates); c != nil && c.Status == configv1.ConditionFalse {
						return fmt.Errorf("Can't look up image for version %s. %v", o.To, c.Message)
					}
					return fmt.Errorf("No available updates, specify --to-image or wait for new updates to be available")
				}
				if o.AllowNotRecommended {
					return fmt.Errorf("The update %s is not one of the conditional or recommended updates", o.To)
				}
				return fmt.Errorf("The update %s is not one of the available updates: %s", o.To, strings.Join(versionStrings(cv.Status.AvailableUpdates), ", "))
			}
```

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: Fixed https://github.com/openshift/oc/commit/7316d191bb03657921a4680fdb4a12106be33c81

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: fixed https://github.com/openshift/oc/commit/7316d191bb03657921a4680fdb4a12106be33c81

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: PTAL https://github.com/openshift/oc/commit/7316d191bb03657921a4680fdb4a12106be33c81 and see if looks better.

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: fixed in https://github.com/openshift/oc/commit/7316d191bb03657921a4680fdb4a12106be33c81

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -247,16 +250,48 @@ func (o *Options) Run() error {
...
+			// Check if the --to version present in condiional update
+			for _, upgrade := range cv.Status.ConditionalUpdates {
+				if c := findCondition(upgrade.Conditions, "Recommended"); c != nil && c.Status != metav1.ConditionTrue {
+					if upgrade.Release.Version == o.To {
+						conditionalUpdate = &configv1.Update{
+							Version: upgrade.Release.Version,
+							Image:   upgrade.Release.Image,
+						}
+						if o.AllowNotRecommended {
+							fmt.Fprintf(o.ErrOut, "warning: with --allow-not-recommended you have accepted the risks with %s and bypassing %s=%s %s: %s\n", o.To, c.Type, c.Status, c.Reason, c.Message)
```
wking, line 263: nit: "you have accepted ... and bypassing ..." should be "... bypassed..." to match "accepted".

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -247,16 +250,48 @@ func (o *Options) Run() error {
...
+				}
+			case update != nil && conditionalUpdate == nil:
+				if o.AllowNotRecommended {
+					return fmt.Errorf("The update %s is not available as a conditional update but available as a recommended update. To proceed with the update do not use --allow-not-recommended flag", o.To)
 				}
-				return fmt.Errorf("The update %s is not one of the available updates: %s", o.To, strings.Join(versionStrings(cv.Status.AvailableUpdates), ", "))
+			}
+
+			//if user has used --allow-not-recommended and the version is present in conditional edges then the update should proceed
+			if o.AllowNotRecommended {
```
wking, line 290: if we get down here and `o.AllowNotRecommended` is not set, we have a bug in the preceeding `switch`.  Can we drop this guard, and maybe add a `default` that fails on an unrecognized situation to the `switch`?

LalatenduMohanty: This can be moved to

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: fixed with https://github.com/openshift/oc/pull/986/commits/55b2321b9028f17aa2475cf75c650727b6539ab7

pkg/cli/admin/upgrade/upgrade.go
LalatenduMohanty: fixed with https://github.com/openshift/oc/pull/986/commits/55b2321b9028f17aa2475cf75c650727b6539ab7

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -260,7 +260,7 @@ func (o *Options) Run() error {
 							Image:   upgrade.Release.Image,
 						}
 						if o.AllowNotRecommended {
-							fmt.Fprintf(o.ErrOut, "warning: with --allow-not-recommended you have accepted the risks with %s and bypassing %s=%s %s: %s\n", o.To, c.Type, c.Status, c.Reason, c.Message)
+							fmt.Fprintf(o.ErrOut, "warning: with --allow-not-recommended you have bypassed %s=%s %s: %s and accepted the risks with %s\n", c.Type, c.Status, c.Reason, c.Message, o.To)
```
wking, line 263: I think you want to keep `Message` at the tail of this string, because it might run to multiple paragraphs, and by the time folks get around to `and accepted the risks with ${VERSION}`, they may have forgotten the initial line of the `warning`.

wking (approved):
/lgtm

PR #987, opened 2021-12-06
oc adm prune deployments for ReplicaSets

deejross:
All unit tests are passing, so that's a good start.


pkg/cli/admin/prune/deployments/data.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"time"
 
+	k8sappsv1 "k8s.io/api/apps/v1"
```
soltysh, line 7: Nit: `kappsv1` is the preferred alias for k8s-based aliases that will collide with ours.

pkg/cli/admin/prune/deployments/prune.go
```diff
@@ -46,6 +47,10 @@ type PrunerOptions struct {
 	DeploymentConfigs []*appsv1.DeploymentConfig
 	// Deployments is the entire list of deployments across all namespaces in the cluster.
 	Deployments []*corev1.ReplicationController
+	// V1Deployments is the entire list of v1 Deployments across all namespaces in the cluster.
+	V1Deployments []*k8sappsv1.Deployment
```
soltysh, line 51: Deployments (+RS) and DeploymentConfigs (+RC) are two distinct sets, so I'd suggest creating entirely separate Pruner just for the former.

soltysh: Going through pruner and especially resolver I see they are very tightly coupled with openshift types, you could try to generlize the code to work with both but it'll be similar effort to writing entirely separate just for k8s types. The latter will be easier to understand, maintain and test. In the `PruneDploymentOptions`'s `Run` method you'd just call one and then the other, let's start with this simpler approach and we can iterate on it as we go.

pkg/cli/admin/prune/deployments/deployments.go
```diff
@@ -80,6 +81,7 @@ func NewCmdPruneDeployments(f kcmdutil.Factory, streams genericclioptions.IOStre
 
 	cmd.Flags().BoolVar(&o.Confirm, "confirm", o.Confirm, "If true, specify that deployment pruning should proceed. Defaults to false, displaying what would be deleted but not actually deleting anything.")
 	cmd.Flags().BoolVar(&o.Orphans, "orphans", o.Orphans, "If true, prune all deployments where the associated DeploymentConfig no longer exists, the status is complete or failed, and the replica size is 0.")
+	cmd.Flags().BoolVar(&o.ReplicaSets, "replica-sets", o.ReplicaSets, "If true, prune ReplicaSets associated with Deployments.")
```
soltysh, line 84: Let's mark this as `experimental` to send appropriate signal that it's not fully tested.

pkg/cli/admin/prune/deployments/deployments.go
```diff
@@ -80,6 +81,7 @@ func NewCmdPruneDeployments(f kcmdutil.Factory, streams genericclioptions.IOStre
 
 	cmd.Flags().BoolVar(&o.Confirm, "confirm", o.Confirm, "If true, specify that deployment pruning should proceed. Defaults to false, displaying what would be deleted but not actually deleting anything.")
 	cmd.Flags().BoolVar(&o.Orphans, "orphans", o.Orphans, "If true, prune all deployments where the associated DeploymentConfig no longer exists, the status is complete or failed, and the replica size is 0.")
+	cmd.Flags().BoolVar(&o.ReplicaSets, "replica-sets", o.ReplicaSets, "EXPIRIMENTAL: If true, ReplicaSets will be included in the pruning process.")
```
soltysh, line 84: s/EXPIRIMENTAL/EXPERIMENTAL/g

soltysh (approved):
Small nits
/approve

soltysh (approved):
/lgtm

Issue #968, opened 2021-11-11
`oc` completion not working despite running the required command `oc completion zsh` or saving output to zsh `fpath`


robbyki:
Not understanding why my `oc` completion isn't working despite all of my other completions working perfectly in zsh environment. Downloaded the latest binary for `oc` and have it installed in my `/usr/local/bin/` space and then trying the following scenarios with zero luck:
 `source [<oc completion zsh)`  and saving that in `.zshrc` with no success. 
or simply saving the output to my `fpath` custom location along with all of my other completions such as `kubectl` which work fine. 
`oc completion zsh > ~/.oh-my-zsh/completions/_oc`
After running any of the above I still get nothing for tab completion against `oc`.

eifrach:
try to edit `~/.oh-my-zsh/completions/_oc`
and add this change on the top of the file

```console
# compdef _kubectl kubectl
compdef _oc oc
```

i see that the completion doesn't work on mac without this

robbyki:
thank you. This worked when I tried a version different from the binary I originally tried from redhat portal. When I tried `Client Version: 4.7.36` the completion from `fpath` worked finally. Didn't have a chance to look into why the latest version from redhat didn't work.

eifrach:
yes so i see this change was added on the `4.9` release 
i was using the `4.8` without an issue.

https://github.com/openshift/oc/blob/master/contrib/completions/zsh/oc#L2-L4

paulbarfuss:
There are a couple of other workarounds for this, such as downgrading your client for the `oc completion zsh` command, or replacing doing a `s/kubectl/oc/` as a one-liner.

```bash
# Load the oc completion code for zsh[1] into the current shell
source <(oc completion zsh | sed -e 's/kubectl/oc/g')
# Set the oc completion code for zsh[1] to autoload on startup
oc completion zsh | sed -e 's/kubectl/oc/g' > "${fpath[1]}/_oc"
```

paulbarfuss:
@eifrach  It appears https://github.com/openshift/oc/blob/master/contrib/completions/zsh/oc is just a generated file for testing purposes and not a template. The `oc completion` command is just a wrapper for `kubectl completion` so the issue is going to be a change upstream to how this command produces the zsh completions and would not be as simple as fixing the generated file.

It looks like the `compdef kubectl _kubectl` line is now being hardcoded in the upstream `kubectl completion zsh` function:

[...]

PR #989, opened 2021-12-07
Unwrap completion command to fix oc completion zsh

paulbarfuss:
CC: eifrach@redhat.com

Related Issue/Bugzilla:

https://github.com/openshift/oc/issues/968
https://bugzilla.redhat.com/show_bug.cgi?id=2024427

`oc completion zsh` is currently hardcoded to add a `kubectl` compdef header.

This PR is one possible solution where we could unwrap the `oc completion` command to properly generate completions for `oc`

Ideally, the upstream could also use the `*cobra.Command.Name` value instead of hard-coding the command.

Current upstream function:

```go
func runCompletionZsh(out io.Writer, boilerPlate string, kubectl *cobra.Command) error {
	zshHead := "#compdef kubectl\ncompdef _kubectl kubectl\n"
	out.Write([]byte(zshHead))
        // Removed some boilerplate
        return kubectl.GenZshCompletion(out)
}
```

Proposed:

```go
func runCompletionZsh(streams genericclioptions.IOStreams, oc *cobra.Command) error {
        // EDIT: Removed zshHead
	return oc.GenZshCompletion(streams.Out)
}
```
This should generate a working `zsh` completion regardless of what is wrapping the original command.

Note that the `contrib` was generated by the scripts in `hack`

Thoughts?

### DIFF
--- a/contrib/completions/bash/oc
+++ b/contrib/completions/bash/oc
@@ -1,4 +1,3 @@
-
 # bash completion for oc                                   -*- shell-script -*-
 
 __oc_debug()
--- a/contrib/completions/zsh/oc
+++ b/contrib/completions/zsh/oc
@@ -1,5 +1,3 @@
-#compdef oc
-compdef _oc oc
 #compdef _oc oc
 
 # zsh completion for oc                                   -*- shell-script -*-
--- a/contrib/completions/zsh/oc
+++ b/contrib/completions/zsh/oc
@@ -1,3 +1,5 @@
+#compdef oc
+compdef _oc oc
 #compdef _oc oc
 
 # zsh completion for oc                                   -*- shell-script -*-
--- a/contrib/completions/zsh/oc
+++ b/contrib/completions/zsh/oc
@@ -1,6 +1,5 @@
-#compdef kubectl
-compdef _kubectl kubectl
-
+#compdef oc
+compdef _oc oc
 #compdef _oc oc
 
 # zsh completion for oc                                   -*- shell-script -*-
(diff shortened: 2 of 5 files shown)


paulbarfuss:
Looking closer at this, it is possible that the compdef string added by zshHead is not even necessary and can be removed completely. Possibly a leftover from the old way upstream `kubectl` did zsh completion.

eifrach:
@paulbarfuss no it not necessary for `zsh`, you can comment it out and it will still work

`source <(oc completion zsh | sed -e 's/compdef kubectl _kubectl/#compdef kubectl _kubectl/' )`

paulbarfuss:
Updated to remove the extra `#compdef` bits (zshHead in the upstream code) and removed WIP label.

marckhouzam:
> Updated to remove the extra `#compdef` bits (zshHead in the upstream code) and removed WIP label.

Be aware that without the second part of the `zshHead` upstream code (`compdef _oc oc`), you will need to put the generated script in `fpath`, but that simply sourcing the script won't work.  This behaviour is the recommended approach from the `zsh` documentation, which is why `Cobra` generates the script in such a fashion.  However, for backwards-compatibility `kubectl` adds the command `compdef _kubectl kubectl` so that sourcing the script still works.

If you wish to support sourcing the script, you can add that extra line at the end of the script instead of the beginning, so as not to mess up the first line generated by Cobra (`#compdef oc`)

paulbarfuss:
Hey @marckhouzam thank you for the review. Your comment matches the behavior I found when testing for this upstream pull request: https://github.com/kubernetes/kubernetes/pull/106969

This just remove the hardcoded `kubectl` and uses `*cobra.Command.Name()` instead.

The good news is that upstream PR had some good movement over the last couple of weeks and works with both sourcing and fpath.

I have updated this PR to match the upstream fix. I am happy to close this if it is best to wait for the upstream fix, or could always wrap it back up again later.

paulbarfuss:
Closing in favor up using upstream fix: https://github.com/kubernetes/kubernetes/pull/106969

PR #990, opened 2021-12-07
Add darwin/arm64 oc

yselkowitz:
This will allow users of the newest M1 MacOS hardware to use a native oc binary, rather than relying on Rosetta2 translation of the darwin/amd64 oc.

Extraction of a darwin/arm64 openshift-install will follow separately.
﻿


soltysh (approved):
/lgtm
/approve

PR #995, opened 2021-12-14
Bug 2032111: requestproject: use agnhost serve-hostname

yselkowitz:
k8s.gcr.io/serve_hostname is deprecated and single-arch.  agnhost is the literal replacement in the E2E test suite.


soltysh (approved):
/lgtm
/approve

PR #1003, opened 2021-12-23
Bug 2033751: Revert "Modified MG pod tolerations"

soltysh:
This reverts #942 

cc @apoorvajagtap

This reverts commit b4ac3d30b5524f0e84774dbfc2663102b0fba8ec.

This reverts commit 5ad360cf7cc3321afdd35544af2aea1db48c2172.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -622,12 +622,8 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 	nodeSelector := map[string]string{
 		corev1.LabelOSStable: "linux",
 	}
-
-	tolerationKey := ""
-
 	if node == "" {
 		nodeSelector["node-role.kubernetes.io/master"] = ""
-		tolerationKey = "node-role.kubernetes.io/master"
 	}
 
 	ret := &corev1.Pod{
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -704,7 +700,6 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 			TerminationGracePeriodSeconds: &zero,
 			Tolerations: []corev1.Toleration{
 				{
-					Key:      tolerationKey,
```
ravisantoshgudimetla, line 707: This is supposed to be blanket toleration, now we changed it to just master :(. 

@@apoorvajagtap
If you want something like you should list all the taints that you would like to see tolerations against

ravisantoshgudimetla (commented):
/approve

/lgtm

pkg/cli/admin/mustgather/mustgather.go
apoorvajagtap: Hi @ravisantoshgudimetla my perspective was more inclined to the fact that by default we do not have any custom taints added to master nodes, and considering that when `--node-name` is not specified the pod would schedule on one of the master nodes, I opted for this single toleration.
But, now when I think of it, I agree if there exists a cluster that's not in an ideal situation, this might be a problem. Let me know in case I am missing anything else too.

soltysh:
e2e-aws is an unrelated failure to this change
/override ci/prow/e2e-aws

PR #1004, opened 2021-12-23
Comment tolerations used in must-gather

soltysh:
Followup to #1003 explains why we need empty tolerations. 

/assign @ravisantoshgudimetla

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -703,6 +703,9 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 			TerminationGracePeriodSeconds: &zero,
 			Tolerations: []corev1.Toleration{
 				{
+					// An empty key with operator Exists matches all keys,
+					// values and effects which means this will tolerate everything.
+					// As noted in https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
 					Operator: "Exists",
 				},
 			},


soltysh:
/hold
we're discussing options

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -704,7 +700,9 @@ func (o *MustGatherOptions) newPod(node, image string) *corev1.Pod {
 			TerminationGracePeriodSeconds: &zero,
 			Tolerations: []corev1.Toleration{
 				{
-					Key:      tolerationKey,
+					// An empty key with operator Exists matches all keys,
+					// values and effects which means this will tolerate everything.
+					// As noted in https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
```
wking, line 705: This comment clarifies what `operator: Exists` means, but doesn't currently explain why we want it.  Perhaps add something like:

> Even if all the control-plane nodes are unschedulable (for example, because an admin has cordoned them), we still want to be able to run the must-gather, to help debug issues.  With this aggressive toleration, users don't have to think about setting `--node-name` to bypass the scheduler.  When some subset of the target nodes are schedulable, the scheduler will prefer them, and it will fall back to unschedulable nodes when no schedulable nodes are available.

Where I'm optimistically guessing about how the scheduler handles that situation, it would be good to have someone with more scheduler knowledge confirm or refute.  And if that's too chatty for an inline comment, we could tuck it or something similar into the commit message where folks could find it via `git blame ...`.

PR #1005, opened 2021-12-23
Bug 2034889: Fix prune deployment panic

deejross:
Specifically, this fixes a panic when using `--replica-sets`.


soltysh (approved):
/lgtm
/approve

PR #1008, opened 2021-12-31
Bug 2034889: Fix another prune panic

deejross:


pkg/cli/admin/prune/deployments/data.go
```diff
@@ -165,8 +165,12 @@ func (d *dataSet) GetDeployment(replica metav1.Object) (metav1.Object, bool, err
 func (d *dataSet) ListDeployments() ([]metav1.Object, error) {
 	results := []metav1.Object{}
 	for _, item := range d.deploymentStore.List() {
-
-		results = append(results, item.(*appsv1.DeploymentConfig))
+		switch v := item.(type) {
+		case *appsv1.DeploymentConfig:
+			results = append(results, v)
+		case *kappsv1.Deployment:
+			results = append(results, v)
```
soltysh, line 172: `results` is `metav1.Object` list so you don't need to cast separately to deployment or deployment config, but just `metav1.Object` should be sufficient.

soltysh (changes requested):

pkg/cli/admin/prune/deployments/data.go
soltysh: Can you also double check if there are similar places further down the code, given this is 2nd similar fix in a row?

deejross: I did another look through, and this was the only one that looked problematic. Also, the casting is to filter out any undesired objects that might have been added. I can take that out if we don't feel like that's a condition we need to worry about.

soltysh: Fair.

soltysh (approved):
/lgtm
/approve

Issue #985, opened 2021-12-01
Login fails if target file is empty


zaneb:
Since #893, the command `oc registry login --kubeconfig=aaa --to=bbb` fails if the file `bbb` is empty:

```
+(utils.sh:530): write_pull_secret(): oc registry login --kubeconfig=/tmp/kubeconfig--0dZLmzoYmT --to=/tmp/pullsecret--Sm82grIvD0
info: Using registry public hostname registry.ci.openshift.org
error: 1 error occurred:
	* reading JSON file "/tmp/pullsecret--Sm82grIvD0": unmarshaling JSON at "/tmp/pullsecret--Sm82grIvD0": unexpected end of JSON input
```

atiratree:
being solved in #1011

atiratree:
@zaneb the regression should be fixed but nevertheless this behaviour is getting deprecated so It is advisable to keep your fix openshift-metal3/dev-scripts#1318 in

PR #1011, opened 2022-01-04
Bug 2036940: allow empty files and stdout in oc registry login

atiratree:
fixes https://github.com/openshift/oc/issues/985

- allow empty files in oc registry login
- return support for stdout output in oc registry login 
- add deprecation notices

@soltysh although, this is a regression I would prefer having less validations and passing this change in behaviour / regression to the user. Reasons:
- less complexity
- same behaviour as podman

### DIFF
--- a/pkg/cli/registry/login/login.go
+++ b/pkg/cli/registry/login/login.go
@@ -3,6 +3,7 @@ package login
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


atiratree:
also fixed regression for stdout `-` files

soltysh (approved):
/lgtm
/approve

PR #1012, opened 2022-01-04
Bug 2035717: Enhancing the output provided when backup collections are attempted.

sferich888:
@soltysh @deads2k this is to address the complaints in https://bugzilla.redhat.com/show_bug.cgi?id=2035717 where the 'flow' of the output doesn't really inform the user that a 'backup' collection is happening, and that the archive you 'end up with' isn't a true 'must-gather' (if you will). 

Some of the re-ordering, of functions also helps with ensuring the 'state' information is presented at the proper times.


soltysh (approved):
/lgtm
/approve

wking:
nit: probably squash acfa5f4c6fb into e77eba50f2; it's a fixup, not a separate idea.

soltysh:
/close
Replaced with #1013.

PR #1014, opened 2022-01-04
Bug 1992596: Remove admin bashcomp func

deejross:
This cleans up some bash completion code for commands that no longer exist.


soltysh (approved):
/lgtm
/aprove

PR #1015, opened 2022-01-07
Bug 2036826: Fix prune deploy resolver

deejross:


pkg/cli/admin/prune/deployments/resolvers.go
```diff
@@ -129,7 +129,9 @@ func (o *perDeploymentResolver) Resolve() ([]metav1.Object, error) {
 			case *corev1.ReplicationController:
 				status = appsutil.DeploymentStatusFor(v)
 			case *kappsv1.ReplicaSet:
-				status = appsutil.DeploymentStatusFor(v)
+				if v.Status.Replicas == 0 {
+					status = "Complete"
```
atiratree, line 133: what if this is the latest revision rs with just 0 replicas? We probably don't want to prune this one.

pkg/cli/admin/prune/deployments/data.go
```diff
@@ -76,9 +76,9 @@ func FilterDeploymentsPredicate(item metav1.Object) bool {
 	case *corev1.ReplicationController:
 		return len(appsutil.DeploymentConfigNameFor(v)) > 0
 	case *kappsv1.ReplicaSet:
-		return len(appsutil.DeploymentNameFor(v)) > 0
+		return len(v.OwnerReferences) > 0 && v.OwnerReferences[0].Kind == "Deployment"
 	default:
-		return false
+		return true
```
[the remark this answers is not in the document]
deejross: Both Deployments and ReplicaSets are now going through the same filter, so it made sense to just pass through an unexpected type. But I suppose checking for those types in both filters and allowing pass through would be safer than ignoring the types altogether.

pkg/cli/admin/prune/deployments/data.go
```diff
@@ -23,11 +23,11 @@ func ReplicaByDeploymentIndexFunc(obj interface{}) ([]string, error) {
 		}
 		return []string{v.Namespace + "/" + name}, nil
 	case *kappsv1.ReplicaSet:
-		name := appsutil.DeploymentNameFor(v)
-		if len(name) == 0 {
+		if len(v.OwnerReferences) == 0 || v.OwnerReferences[0].Kind != "Deployment" {
```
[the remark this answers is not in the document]
deejross: I couldn't find a case where there was ever more than one Owner, but the fact that it is a list means I can't rule that out. I'll see if I can adjust this to include a loop looking for the Deployment.

pkg/cli/admin/prune/deployments/resolvers.go
deejross: There only seemed to be two status fields on ReplicaSets, so determining complete vs failed, like for ReplicationControllers, seems much more difficult. I'm open to suggestions on this one.

deejross:
Made some changes based on comments from @atiratree

pkg/cli/admin/prune/deployments/data.go
```diff
@@ -76,7 +77,12 @@ func FilterDeploymentsPredicate(item metav1.Object) bool {
 	case *corev1.ReplicationController:
 		return len(appsutil.DeploymentConfigNameFor(v)) > 0
```
soltysh, line 78: I'm inclined to also apply here similar rule as below wrt using `OwnerRferences` since that's what garbage collector uses and it seems reasonable for the prune command to follow as well. This can be made as a followup.

pkg/cli/admin/prune/deployments/data.go
```diff
@@ -90,7 +96,7 @@ func FilterZeroReplicaSize(item metav1.Object) bool {
 	case *kappsv1.ReplicaSet:
 		return *v.Spec.Replicas == 0 && v.Status.Replicas == 0
 	default:
-		return true
+		return false
```
soltysh, line 99: Honestly both `true` and `false` will be fine here, since a) theoretically we shouldn't get a different resource than RS or RC here and b) even if we did it will be filtered out elsewhere. Still it would be nice to see a comment here why one was picked over the other.

pkg/cli/admin/prune/deployments/data.go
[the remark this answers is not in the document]
atiratree: I think it is better to filter out objects we are not expecting - we could also log to be sure, but we control our use case so probably no need to do that. The current code looks ok.

pkg/cli/admin/prune/deployments/resolvers.go
```diff
@@ -121,15 +121,19 @@ func (o *perDeploymentResolver) Resolve() ([]metav1.Object, error) {
...
-		for _, replica := range replicas {
+		for i, replica := range replicas {
 			var status appsv1.DeploymentStatus
 
 			switch v := replica.(type) {
 			case *corev1.ReplicationController:
 				status = appsutil.DeploymentStatusFor(v)
 			case *kappsv1.ReplicaSet:
-				status = appsutil.DeploymentStatusFor(v)
+				if v.Status.Replicas == 0 && i < len(replicas)-1 {
```
atiratree, line 134: @deejross latest revision RS, does not have to be the newest one. You can check the annotations of the RS, also you have the deployment which you can use to filter out the current revision.

pkg/cli/admin/prune/deployments/resolvers.go
```diff
@@ -121,15 +121,19 @@ func (o *perDeploymentResolver) Resolve() ([]metav1.Object, error) {
 			return nil, err
 		}
 
+		sort.Sort(ByMostRecent(replicas))
```
soltysh, line 124: Why we need this sorted?

soltysh (approved):
Left some comments, but overall 
/lgtm
/approve

pkg/cli/admin/prune/deployments/data.go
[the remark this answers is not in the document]
soltysh: That'd be my personal preference as well, tbh :wink:

PR #1016, opened 2022-01-10
Bug 2038930: properly handle --dry-run=server 

soltysh:
All of the oc set subcommands did not have a proper wiring for
--dry-run=server flag. They entirely ignored the value being provided
by user. Additionally, I switched all the commands to use
ResourceBuilder rather than directly using dynamic client.

Manual cherry-pick of #1010

/assign @deejross


soltysh:
/label backport-risk-assessed
it's being worked on elsewhere

PR #1018, opened 2022-01-10
Ignore VSCode files

deejross:
This will ignore any VSCode-specific configuration files from accidentally being added to source control.


soltysh (approved):
/lgtm
/approve
Manually adding valid-bug, since that's a minor fix that doesn't affect product.

PR #1019, opened 2022-01-10
Bug 2036826: Improved prune deployments

deejross:
This is a followup to https://github.com/openshift/oc/pull/1015 that improves a few aspects of pruning ReplicaSets and switches to matching replicas and deployments using OwnerReferences rather than Annotations.


soltysh (approved):
/lgtm
/approve

PR #1026, opened 2022-01-17
Bug 1976894: Idling a StatefulSet seems to work however accessing the Services Route does not wake up the application ("Application is not available" error page is returned).

miheer:
Idling a StatefulSet seems to work however accessing the Services Route does not wake up the application ("Application is not available" error page is returned).
So, with this fix we will return an error saying that we don't support idling of stateful sets as the accessing the routes for the service which is idled does not work. For that a feature work will be required. But, for now we need to add a error whenever a customer tries to idle a stateful service. This commit fixes Bug 1976894.

### DIFF
--- a/pkg/cli/idle/idle.go
+++ b/pkg/cli/idle/idle.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	kappsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -268,6 +269,9 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
 		}
 		// just get the unversioned version of this
 		gk := schema.GroupKind{Group: gv.Group, Kind: ref.Kind}
+		if kappsv1.SchemeGroupVersion.WithKind("StatefulSet").GroupKind() == gk {
+			return nil, fmt.Errorf("idling StatefulSet is not supported yet.")
+		}
 		helper, ok := helpers[gk]
 		if !ok {
 			var mapping *meta.RESTMapping
@@ -427,7 +431,7 @@ func findScalableResourcesForEndpoints(endpoints *corev1.Endpoints, getPod func(
 	controllerRefs := make(map[namespacedCrossGroupObjectReference]struct{})
 	for controllerRef := range immediateControllerRefs {
 		controller, err := getController(controllerRef)
-		if err != nil && errors.IsNotFound(err) {
+		if err != nil && !errors.IsNotFound(err) {
 			return nil, fmt.Errorf("unable to load %s %q: %v", controllerRef.Kind, controllerRef.Name, err)
 		}
 


pkg/cli/idle/idle.go
```diff
@@ -268,6 +268,10 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
 		}
 		// just get the unversioned version of this
 		gk := schema.GroupKind{Group: gv.Group, Kind: ref.Kind}
+		if gk.Kind == "StatefulSet" {
+			fmt.Println("Idling StatefulSet is not supported yet.")
+			return nil, nil
```
Miciah, line 273: Is there a reason you need to print here instead of returning an error?
```suggestion
			return nil, fmt.Errorf("idling StatefulSets is not supported")
```

Miciah: Also, don't you need three return values here?  
```suggestion
			return nil, nil, fmt.Errorf("idling StatefulSets is not supported")
```

miheer: Thanks @Miciah ! return  is inside 	getController := func(ref namespacedOwnerReference) (metav1.Object, error) {

so I will add return nil, fmt.Errorf("idling StatefulSets is not supported") as can't return 3 values.

miheer: > Is there a reason you need to print here instead of returning an error?

It works fine however the right approach will be to add an error in the return as you mentioned as the return type of the function is error.

Miciah: > return is inside getController := func(ref namespacedOwnerReference) (metav1.Object, error) {

Ah, I see.  GitHub doesn't show the indentation correctly.

miheer: > Thanks @Miciah ! return is inside getController := func(ref namespacedOwnerReference) (metav1.Object, error) {
> 
> so I will add return nil, fmt.Errorf("idling StatefulSets is not supported") as can't return 3 values.



@Miciah after adding this I don't see the error message like I do see with this commit as follows -

[miheer@localhost oc]$ ./oc idle nginx
WARNING: idling when network policies are in place may cause connections to bypass network policy entirely
Idling StatefulSet is not supported yet.
error: unable to mark the service "nginx/nginx" as idled.
Make sure that the service is not already marked as idled and that it is associated with resources that can be scaled.
See 'oc idle -h' for help and examples.
[miheer@localhost oc]$ 

After adding the return  nil, fmt.Errorf("idling StatefulSets is not supported") it does not show the error

[miheer@localhost oc]$ ./oc idle nginx
WARNING: idling when network policies are in place may cause connections to bypass network policy entirely
error: unable to mark the service "nginx/nginx" as idled.
Make sure that the service is not already marked as idled and that it is associated with resources that can be scaled.
See 'oc idle -h' for help and examples.

miheer: BTW I had to make the code changes where get controller is called and where it's error is captured if you want to return error.

miheer: [miheer@localhost oc]$ ./oc idle nginx
WARNING: idling when network policies are in place may cause connections to bypass network policy entirely
error: no valid scalable resources found to idle: unable to calculate scalable resources for service nginx/nginx: Idling StatefulSet is not supported yet.
[miheer@localhost oc]$

pkg/cli/idle/idle.go
```diff
@@ -427,7 +430,9 @@ func findScalableResourcesForEndpoints(endpoints *corev1.Endpoints, getPod func(
 	controllerRefs := make(map[namespacedCrossGroupObjectReference]struct{})
 	for controllerRef := range immediateControllerRefs {
 		controller, err := getController(controllerRef)
-		if err != nil && errors.IsNotFound(err) {
+		if err != nil {
+			return nil, err
+		} else if errors.IsNotFound(err) {
 			return nil, fmt.Errorf("unable to load %s %q: %v", controllerRef.Kind, controllerRef.Name, err)
 		}
```
Miciah, line 437: @miheer, after your change, the `else` branch is unreachable, isn't it?  

Anyway, it seems the logic was broken by an earlier change: <https://github.com/openshift/origin/pull/22919/files#diff-037051bf889204d52030b00c7a2754f939bb8653b22842a3dc53db1259f433d6L420-R420>.  The original intention was any errors _except_ not-found errors should be reported.
```suggestion
		if err != nil && !errors.IsNotFound(err) {
			 if  {
				return nil, fmt.Errorf("unable to load %s %q: %v", controllerRef.Kind, controllerRef.Name, err)
			}
			return nil, err
		}
```
@mfojtik, can you double-check that my analysis of your change and suggestion here is correct?

Miciah:
@soltysh, can you approve this PR?

pkg/cli/idle/idle.go
```diff
@@ -268,6 +268,9 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
...
 		// just get the unversioned version of this
 		gk := schema.GroupKind{Group: gv.Group, Kind: ref.Kind}
+		if gk.Kind == "StatefulSet" {
+			return nil, fmt.Errorf("Idling StatefulSet is not supported yet.")
```
soltysh, line 272: All errors should be lowercased, ie. `idling StatefulSet...`

pkg/cli/idle/idle.go
```diff
@@ -268,6 +268,9 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
 		}
 		// just get the unversioned version of this
 		gk := schema.GroupKind{Group: gv.Group, Kind: ref.Kind}
+		if gk.Kind == "StatefulSet" {
```
soltysh, line 271: I'd suggest being more accurate ie.:
```golang
if kappsv1.SchemeGroupVersion.WithKind("StatefulSet").GroupKind() == gk {
...
```
this will ensure that only k8s apps StatefulSet will be ignored.

soltysh (changes requested):

soltysh (approved):
/lgtm
/approve

PR #907, opened 2021-08-23
Bug 1990014: Use cmd for Windows pods

deejross:

PR #1028, opened 2022-01-18
Bug 1990014: Throw error when debugging Windows nodes

deejross:


soltysh (approved):
/lgtm
/approve

pkg/cli/debug/debug.go
```diff
@@ -968,6 +968,10 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 			// TODO: allow --as-root=false to skip all the namespaces except network
 			return nil, fmt.Errorf("can't debug nodes without running as the root user")
 		}
+		if t.Labels["kubernetes.io/os"] == "windows" {
+			// Windows nodes don't yet support privileged containers, which is required for debug
```
aravindhp, line 972: Why do you need priviledged containers for debugging a pod? Can't you run `cmd` like you were doing in #907?

aravindhp:
/hold

Adding a hold until we figure out why this is being disabled.

pkg/cli/debug/debug.go
deejross: It doesn't work, and the docs explained that debug needs privileged containers, and that Windows nodes don't currently support that: https://docs.openshift.com/container-platform/4.9/support/troubleshooting/troubleshooting-windows-container-workload-issues.html#accessing-windows-node

This PR is to address a bug where running `oc debug node` currently hangs attempting to debug a Windows node. If we can support debugging Windows nodes, that would be great though.

aravindhp: That is correct. `oc debug node` does not work for Windows nodes. The main confusion here is the association with [Bug 1990014  - oc debug <pod-name> does not work for Windows pods](https://bugzilla.redhat.com/show_bug.cgi?id=1990014) which has to be debugging Windows Pods. I want to ensure that we are not disabling that.

deejross: Correct, in that BZ, someone tried both nodes and pods. The intent is to provide users with better feedback when `oc debug node` is attempted on Windows nodes. Once this is merged, I plan to set the bug back to NEW to address the issue with pods, as that is a valid bug.

selansen: @deejross, Thank you for the detailed explanation.

aravindhp:
/hold cancel

Confirmed with @deejross on Slack that this will not affect pod debugging

deejross:
As a side note, I confirmed that you can debug Windows pods in oc 4.9.15. I was using 4.8.2 and realized #907 hadn't been included in that version.

aravindhp:
> As a side note, I confirmed that you can debug Windows pods in oc 4.9.15. I was using 4.8.2 and realized #907 hadn't been included in that version.

You can `/cherry-pick release-4.8` and get #907 backported given it is a bug fix. It will get picked up in the next 4.8.z.

PR #1032, opened 2022-01-24
adm inspect: delete unused pod URL getting code

markmc:
#763 deleted all use of the PortForwardURLGetter type, but left its implementation in place.


soltysh (approved):
/lgtm
/approve

PR #1041, opened 2022-01-30
Bug 2044140: Updated the error message to include the suggestion

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

PR #1050, opened 2022-02-03
Bug 2044140: pkg/cli/admin/upgrade: Fix nextStep option sense

wking:
As [Evgeni points out][1], there is a bug in a0f1832b91 (#1041).  When `--to-image` is not set, we want to mention that the user may want to set it.  When `--to-image` is set, we want to mention `--allow-explicit-update`.  The previous check had the reversed sense, and mentioned `--allow-explicit-upgrade` when `--to-image` was unset.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2044140#c8


LalatenduMohanty (approved):
/lgtm

PR #1051, opened 2022-02-04
Show managedFields in inspect

soltysh:
/assign @deads2k

### DIFF
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/inspect/inspect.go
```diff
@@ -176,6 +177,14 @@ func (o *InspectOptions) Complete(args []string) error {
 	if err != nil {
 		return err
 	}
+	// TODO (soltysh) change this to set WithManagedFields in PrintFlags creation
+	// above when https://github.com/kubernetes/kubernetes/pull/107947 merges
+	if tsp, ok := printer.(*printers.TypeSetterPrinter); ok {
+		printer = tsp.Delegate
```
deads2k, line 183: Seems like we only want to overwrite `printer` on line 185 after we've removed the `OmitManagedFieldsPrinter`.  This line as-is could result in us losing the typesetting capability which is needed for proper serialization.

soltysh: I'm setting back the typer below, just like what we get from invoking `WithTypeSetter` in https://github.com/openshift/oc/blob/3aedcfc3ba5afaa0b7993c3e1824b179a9c241a3/pkg/cli/admin/inspect/inspect.go#L93

deads2k: could we do

```go
if tsp, ok := printer.(*printers.TypeSetterPrinter); ok {
		if omp, ok := tsp.Delegate.(*printers.OmitManagedFieldsPrinter); ok {
			printer = &printers.TypeSetterPrinter{Typer: scheme.Scheme, Delegate: omp.Delegate}
		}
	}
```

to avoid writing `printer` early?

deads2k:
/lgtm
/hold

holding so you can confirm that it works

/cherrypick release-4.10
/cherrypick release-4.9

please pick this back to the point where we lost that data.

soltysh:
confirmed
/hold cancel
/retest-required

PR #1058, opened 2022-02-09
Bug 2049133: Fix catalog mirror from files

vitus133:
This commit fixes `oc adm catalog mirror` from
file to a disconnected registry.
For operators with manifest media type
`application/vnd.oci.image.config.v1+json`
mediaType was incorrectly unmarshalled.


soltysh (approved):
/approve

PR #1068, opened 2022-02-11
Bug 2053222: Fix mirroring images that have dots in their namespace

dmage:
Backport of #1067.

This PR bumps library-go to get a new registryclient that can correctly
handle image references like

    registry.example.com/namespace.with.dot/foo

PR #1069, opened 2022-02-11
Bug 2053223: Fix mirroring images that have dots in their namespace

dmage:
Backport of #1068.

This PR bumps library-go to get a new registryclient that can correctly
handle image references like

    registry.example.com/namespace.with.dot/foo


dmage:
Note, client_mirrored.go doesn't exist in 4.8 yet (because library-go hasn't been bumped), so we don't really need to bump library-go in oc 4.8.

xiuwang:
/label qe-approved
https://bugzilla.redhat.com/show_bug.cgi?id=2053223#c1

PR #1078, opened 2022-02-21
Bug 2056893: pkg/cli/admin/upgrade: Drop --to-image help warning

wking:
We've had the warning at least since this repo was created in 3abf60a31e.  But since we grew `--allow-explicit-upgrade` in 0501d04ff1 (#109), `--to-image` has been just as restricted to recommended updates as `--to`.  Removing the warning here avoids unnecessary alarm.  We can save the warning for `--allow-explicit-upgrade`, whose help text I'm consolidating in this
commit.  The new structure is:

* `--to` and `--to-image` together, since in the absence of `--allow-explicit-upgrade` those are equally safe.

* The paragraph about unsupported updates or broken CVO <-> upstream update service communication.  Before this paragraph had focused on `--to-image`, but I've adjusted it to focus on `--allow-explicit-upgrade`.

* The `--allow-upgrade-with-warnings` paragraph.  I've moved the sentence about retargeting updates over to this paragraph, since `--allow-upgrade-with-warnings` is the relevant flag for that issue.

* The `--force` paragraph, because that's about a cluster-side check that occurs after the oc-side `--allow-*` checks, so having it last in the help text gives us the same order in help as we get for check application.

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -57,30 +57,30 @@ func New(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command
 			retrieved from the update server and cached on the cluster - these are updates that are
 			known to be supported as upgrades from the current version.
 
-			Passing --to=VERSION will upgrade the cluster to one of the available updates or report
-			an error if no such version exists. The cluster will then upgrade itself and report
-			status that is available via "oc get clusterversion" and "oc describe clusterversion".
+			Passing --to=VERSION or --to-image=IMAGE will upgrade the cluster to one of the available
+			updates or report an error if no such version exists. The cluster will then upgrade
+			itself and report status that is available via "oc get clusterversion" and "oc describe
+			clusterversion".
 
-			If the desired upgrade from --to-image is not in the list of available versions, you must
-			pass --allow-explicit-upgrade to allow upgrade to proceed. If the cluster is
-			already being upgraded, or if the cluster is reporting a failure or other error, you
-			must pass --allow-upgrade-with-warnings to proceed (see note below on the implications).
+			If there are no versions available, or a bug in the cluster version operator prevents
+			updates from being retrieved, --to-image may be combined with the more powerful and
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -57,30 +57,30 @@ func New(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command
...
+			of the specified release image, regardless of whether that upgrade is known to be
+			supported for the current version. While rolling back to a previous micro version
+			(4.1.2 -> 4.1.1) may be safe, upgrading more than one minor version ahead (4.1 -> 4.3) or
+			downgrading one minor version (4.2 -> 4.1) is likely to cause data corruption or to
+			completely break a cluster.
+
+			If the cluster is already being upgraded, or if the cluster is reporting a failure or
+			other error, you must pass --allow-upgrade-with-warnings to proceed (see note below on
+			the implications). Avoid upgrading when the cluster is reporting errors or when another
+			upgrade is in progress unless the upgrade cannot make progress.
```
LalatenduMohanty, line 77: `Avoid upgrading when the cluster is reporting errors or when another upgrade is in progress unless the upgrade cannot make progress.`  is confusing and do not communicate the context properly.  Here is my suggestion `You should understand risks from the error/warnings before considering to use --allow-upgrade-with-warnings` oe something like that. I am looking for a smaller sentence here.

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/upgrade.go
wking: Addressed with b6bb202a8 -> 4e3339fe2:

> If the cluster is already being upgraded, or if the cluster is reporting a failure or other error, the update will not be triggered.  It is usually best to give these conditions time to resolve, or to actively work to resolve them.  But if you decide to trigger the update regardless of these concerns, use --allow-upgrade-with-warnings.

LalatenduMohanty (approved):
/lgtm

PR #1085, opened 2022-03-04
Bug 2056122: expose --keep-startup flag for oc debug

soltysh:
/assign @deejross


jeffnowicki:
@soltysh do you intend to `cherry-pick` back to `4.101`?

jmcmeek:
@soltysh From an IBM Cloud perspective, 4.11 is two releases in the future. 4.9 is our latest release. Given the client / server version skew requirements, that would basically mean we have to tell customers they cannot properly debug pods with startupProbes until they have moved their clusters to 4.10 or later.

PR #1087, opened 2022-03-07
Bug 2057633: add validations for a pod & container to rsync

atiratree:
The oc rsync fails, complaining that rsync and tar are missing in the
container:

```
$ oc rsync --container foo -n openshift-storage pod/rook-ceph-tools-56f88ff6cb-tvpg5:/etc/redhat-release /tmp
WARNING: cannot use rsync: rsync not available in container
WARNING: cannot use tar: tar not available in container
error: No available strategies to copy.
```

### DIFF
--- a/pkg/cli/rsync/util.go
+++ b/pkg/cli/rsync/util.go
@@ -4,13 +4,16 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os/exec"
 	"runtime"
 
 	"k8s.io/klog/v2"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubectl/pkg/cmd/util/podcmd"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


atiratree:
@soltysh @deejross  @ardaguclu please review

pkg/cli/rsync/util.go
```diff
@@ -127,13 +130,36 @@ func rsyncSpecificFlags(o *RsyncOptions) []string {
...
+	if pod.DeletionTimestamp != nil {
+		return fmt.Errorf("pod %v is getting deleted", p.podName)
+	}
+
+	container, err := podcmd.FindOrDefaultContainerByName(pod, p.containerName, p.quiet, p.stdErr)
+	if err != nil {
+		return err
+	}
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.Name == container.Name && containerStatus.State.Running == nil {
```
ardaguclu, line 160: If we already find our container in here, maybe we should break loop in any case?, because it seems that there is no need to look other container statuses.

atiratree: makes sense, thanks

atiratree:
/assign @soltysh 
Can you please take a look?

soltysh (approved):
/approve

PR #1091, opened 2022-03-17
Obtain OpenShift version from ClusterVersion resource

csrwng:
In clusters with hosted control planes, neither the OpenShift
apiserver nor its operator are present inside the cluster. Switching the
`oc version` command to work with the ClusterVersion resource instead of
the openshift-apiserver ClusterOperator resource makes it possible for
`oc version` to work with this topology as well.


soltysh (approved):
/lgtm
/approve

PR #1093, opened 2022-03-18
Bug 1999891: bubble errors which happen prior to collection to BackupGathering

soltysh:
/assign @deejross @atiratree 
/cc @kasturinarra

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -266,6 +266,10 @@ func (o *MustGatherOptions) Validate() error {
 	if o.NodeName != "" && o.NodeSelector != "" {
 		return fmt.Errorf("--node-name and --node-selector are mutually exclusive: please specify one or the other")
 	}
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -329,26 +337,39 @@ func (o *MustGatherOptions) Run() error {
 		_, err := imagereference.Parse(image)
 		if err != nil {
 			o.log("unable to parse image reference %s: %v", image, err)
+			// ensure the errors bubble up to BackupGathering method for display
+			errs = []error{err}
```
atiratree, line 341: might be better to have `"unable to parse image reference %s: %v"` extra info in the error as well instead of just `error: invalid reference format`

atiratree (commented):
@soltysh seems we are getting some errors in CI after this change

```
Error running must-gather collection:
    resource name may not be empty
```

pkg/cli/admin/mustgather/mustgather.go
soltysh: Done

PR #1100, opened 2022-03-30
Introduce Jira defects into release info

ardaguclu:
Currently, `oc adm release info` retrieves and describes bugs only from Bugzilla.
But in the near future, Jira will be the primary source for managing bugs.

This PR adds handler for Jira bugs whose start with `OCPBUGS-` prefix and combining
Jira defects as well as Bugzilla bugs whose start with `BUG `.

### DIFF
--- a/pkg/cli/admin/release/bug.go
+++ b/pkg/cli/admin/release/bug.go
@@ -1,0 +1,309 @@
+package release
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+
+	"k8s.io/klog/v2"
+)
+
+type BugType int
+
+const (
+	Bugzilla BugType = iota
+	Jira
+)
+
+const (
+	maxRetries = 2
+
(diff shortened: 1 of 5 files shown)


pkg/cli/admin/release/bug.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// BugRemoteInfo stores the detail information of bug retrieved
+// from remote url like bugzilla or jira.
+type BugRemoteInfo struct {
+	ID       int    `json:"id"`
+	Status   string `json:"status"`
+	Priority string `json:"priority"`
+	Summary  string `json:"summary"`
+	Source   BugType
```
ardaguclu, line 63: From the reviewer's point of view, to make review easy, this PR might be split into different commits. However, this `Source` field, which is used to discriminate bugs according to it's source, demands one commit by changing everywhere at one time.

ardaguclu:
Currently, there is no bug in Jira containing prefix `OCPBUGS-`(and of course, not in release payloads), thus it is not possible to test on live. 
I have tested for existing Bugzilla bugs and this PR does not break any existing features. What I'm thinking is maybe we can move forward for review or even can merge.

What do you think @soltysh @deejross ?

Job failures look not related to this PR;

/retest

PR #1103, opened 2022-03-31
Fix `oc set env` example

ammachado:
Example was showing a command line where it executes `oc set env` without parameters, while the correct is `env` only.


pkg/cli/set/env.go
```diff
@@ -78,7 +78,7 @@ var (
 		oc set env -f dc.json ENV-
 
 		# Set some of the local shell environment into a deployment config on the server
-		oc set env | grep RAILS_ | oc env -e - dc/myapp
+		env | grep RAILS_ | oc env -e - dc/myapp
```
soltysh, line 81: Can we instead just drop this example, it doesn't gives us too much value but only clutters the overall help page for this command?

soltysh (changes requested):
/ok-to-test

pkg/cli/set/env.go
soltysh: @ammachado I'm open to get that also removed from `kubectl` as well.

ammachado: Example removed.

PR #1106, opened 2022-04-08
Bug 2073113: do not report docker conf deprecation warning when the docker is not available

atiratree:
Also specify a subject of the docker conf deprecation message as noted in the BZ.


soltysh (approved):
/lgtm
/approve

PR #1108, opened 2022-04-11
Bug 2041454: Validate reference-policy for import-image command

ardaguclu:
This PR adds validation for reference-policy flag for import-image
command. It will return error when reference-policy is other than
source or local. In addition to that upper or lower cases will be treated
as valid(ie. Local, loCal, Source, etc.)


soltysh (approved):
/lgtm
/approve

ardaguclu:
/hold

Tests might be needed to change according to this PR.

soltysh (approved):
/lgtm

PR #1109, opened 2022-04-11
Add --keep-manifest-list support to `oc adm release *` commands

soltysh:
/assign @ricardomaraschini @deejross


pkg/cli/image/append/append.go
```diff
@@ -259,26 +263,87 @@ func (o *AppendImageOptions) Run() error {
...
+		fromRepo          distribution.Repository
+	)
+	if repo != nil || srcManifest != nil {
+		fromRepo = repo
+
 		base, layers, err = imagemanifest.ManifestToImageConfig(ctx, srcManifest, repo.Blobs(ctx), manifestLocation)
 		if err != nil {
-			return fmt.Errorf("unable to parse image %s: %v", from, err)
+			// return fmt.Errorf("unable to parse image %s: %v", from, err)
+			return err
```
deejross, line 346: Did you forget to remove or un-comment this line?

pkg/cli/image/append/append.go
```diff
@@ -260,26 +263,87 @@ func (o *AppendImageOptions) Run() error {
...
+func (o *AppendImageOptions) appendManifestList(ctx context.Context, createdAt *time.Time,
+	from *imagesource.TypedImageReference, to imagesource.TypedImageReference,
+	repo distribution.Repository, srcManifest distribution.Manifest, manifestLocation imagemanifest.ManifestLocation,
+	toRepo distribution.Repository, toManifests distribution.ManifestService) error {
+	// process manifestlist
+	// oldDigest:newDigest mappging so that we can create a new manifestlist
+	newDigests := make(map[digest.Digest]digest.Digest)
+	manifestMap, oldList, _, err := imagemanifest.AllManifests(ctx, from.Ref, repo)
+	for digest, srcManifest := range manifestMap {
+		err = o.append(ctx, createdAt, from, to, repo, srcManifest, manifestLocation, toRepo, toManifests)
```
ricardomaraschini, line 298: Regarding the `to` argument here: IIUC it points to an image by `tag` and, as a consequence, each time we call `append()` the tag `latest` will be updated remotely to point to a single manifest (i.e. it will overwrite the previous manifest list reference with a reference to a single manifest).

Shouldn't `to` be updated to point to an image by `sha` before calling `append()` ? If we could do that then we would assure we would not stomp with the remote tag until the very last moment (when we upload the new manifest list).

soltysh: Good point, lemme try to figure something out.

soltysh: The caller needs to define if we should push a tag or not, which can be skipped in `PutManifestInCompatibleSchema` so I went with a boolean passed to `append` which tells if tag should or not be skipped, which I set in `appendManifestlList`.

ricardomaraschini (commented):
/lgtm

PR #1110, opened 2022-04-14
Bug 2071614: Remove network CRDs scheme registration

ardaguclu:
Api server does not permit strategic merge patch type for CRDs during
patch operations. This is ensured by not registering CRDs schemes.

Therefore, `oc` must only register non-crds into scheme. This PR removes
`network.openshift.io` registration because it does not contain any non-crd.
ref: https://github.com/openshift/api/blob/2f6f4d85d80b3f39d66e7e1674aeff71f8a68b3d/network/v1/register.go#L33-L40


soltysh (approved):
/approve

PR #1111, opened 2022-04-14
Bug 2075647: pkg/cli/admin/upgrade: Use PATCH instead of POST for spec updates

wking:
ClusterVersion's spec can evolve over time, growing additional properties like [capabilities][1].  With the old `POST`/`Update(...)` approach, that could lead to older `oc` (who didn't know about the new properties) naively clearing them when they washed the in-cluster resource through their local ClusterVersion Go type.  With this commit, we transition to `PATCH`, so we can touch just `spec.desiredUpdate`, regardless of what else is going on in spec.

Also, while I'm touching this code:

* I set things up so `--allow-not-recommended` works with `--to-latest` as well.  This also consolidated so there was only a single `Update` to convert to a `Patch`.
* I adjusted `--allow-explicit-upgrade --to-image ...` logging to only complain about `--allow-explicit-upgrade` when we needed to use it, and to not log complaints when the requested image was known to ClusterVersion `status`.

Details about those two in their commit messages.

[1]: https://github.com/openshift/api/pull/1106/commits/5b82635ef101e7c10b5fcbbcfb81315bb7a0da20

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -3,6 +3,7 @@ package upgrade
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


wking:
[`The openshift-etcd pods should be scheduled on different nodes`][1] is unrelated, but I won't bother kicking off retests until this PR has been reviewed.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1111/pull-ci-openshift-oc-master-e2e-agnostic-cmd/1514682136743907328

wking:
[`etcd-guard` healthz connections][1] are unrelated.  @soltysh, can we get an `/override ci/prow/e2e-agnostic-cmd`?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1111/pull-ci-openshift-oc-master-e2e-agnostic-cmd/1516066991490732032

PR #1112, opened 2022-04-19
pkg/cli/debug: set pod security level for debug pod

s-urbaniak:
This is necessary as otherwise the (privileged) debug pod would violate restricted pod security admission:

Example:
```
oc debug node/rgangwar-18de2-x6zpk-control-plane-0
Warning: would violate PodSecurity "restricted:latest": host namespaces (hostNetwork=true, hostPID=true), hostPath volumes (volume "host"), privileged (container "container-00" must not set securityContext.privileged=true), allowPrivilegeEscalation != false (container "container-00" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "container-00" must set securityContext.capabilities.drop=["ALL"]), restricted volume types (volume "host" uses restricted volume type "hostPath"), runAsNonRoot != true (pod or container "container-00" must set securityContext.runAsNonRoot=true), runAsUser=0 (container "container-00" must not set runAsUser=0), seccompProfile (pod or container "container-00" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
Starting pod/rgangwar-18de2-x6zpk-control-plane-0-debug ...
Pod IP: 172.31.248.77
```

/cc @soltysh @stlaz

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -32,6 +32,7 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
 	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/kubectl/pkg/cmd/attach"
 	"k8s.io/kubectl/pkg/cmd/logs"
 	krun "k8s.io/kubectl/pkg/cmd/run"
@@ -41,6 +42,7 @@ import (
 	"k8s.io/kubectl/pkg/util/interrupt"
 	"k8s.io/kubectl/pkg/util/templates"
 	"k8s.io/kubectl/pkg/util/term"
+	admissionapi "k8s.io/pod-security-admission/api"
 
 	appsv1 "github.com/openshift/api/apps/v1"
 	dockerv10 "github.com/openshift/api/image/docker10"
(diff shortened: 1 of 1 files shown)


pkg/cli/debug/debug.go
```diff
@@ -446,6 +448,22 @@ func (o *DebugOptions) RunDebug() error {
 	pod.Name, pod.Namespace = fmt.Sprintf("%s-debug", generateapp.MakeSimpleName(infos[0].Name)), ns
 	o.Attach.Pod = pod
 
+	// set pod security label on the namespace as the debug pod is privileged.
```
soltysh, line 451: That's not good approach, you're updating namespace for all `oc debug` operations and that's not always recommended. `oc debug` can and should be able to be invoked for regular users, since they can invoked `oc debug mypod`, not every `oc debug` invocation is `oc debug node` like in your example. Most importantly, the `oc debug` should not modify the namespace w/o prior ack from the owner of the namespace. It should most likely fail and inform user what the right action should be, here adding correct labels.

soltysh (changes requested):
/hold
this isn't good approach for `oc debug`

pkg/cli/debug/debug.go
s-urbaniak: Thank you @soltysh! I will try to find if there is a dedicated error I can parse, or, i will simply always print out an instruction message in case of error saying something like "starting debug pod failed, error=..., please make sure you set the restricted pod security on the namespace xyz using oc ..."

deads2k: > Most importantly, the oc debug should not modify the namespace w/o prior ack from the owner of the namespace. It should most likely fail and inform user what the right action should be, here adding correct labels.

Agree.  I would go so far as to suggest that instead of suggesting a label change, you suggest the user create a dedicated namespace and give that command (`oc create namespace --dry-run | oc label -f - --dry-run name=value| oc create -f -`

PR #1113, opened 2022-04-20
Bug 2065507: Add the ReleaseAccepted condition to the oc adm upgrade command

Davoska:
To accurately display the upgrade's status add the `ReleaseAccepted` condition to the `oc adm upgrade` command.

With the new logic introduced by the bug fix for the CVO (https://bugzilla.redhat.com/show_bug.cgi?id=1822752), it is necessary to accurately display the current status of the upgrade in a case where a release payload wasn't accepted.

This change will display (using the `oc adm upgrade` command) the reason and the message containing information on why the given release payload wasn't accepted and thus why the update hasn't started.

Pull request related to https://issues.redhat.com/browse/OTA-589

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -377,6 +377,10 @@ func (o *Options) Run() error {
 			fmt.Fprintf(o.Out, "Upgradeable=False\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 		}
 
+		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, "ReleaseAccepted"); c != nil && c.Status != configv1.ConditionTrue {
+			fmt.Fprintf(o.Out, "ReleaseAccepted=%s\n\n  Reason: %s\n  Message: %s\n\n", c.Status, c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
+		}
+
 		if cv.Spec.Channel != "" {
 			if cv.Spec.Upstream == "" {
 				fmt.Fprint(o.Out, "Upstream is unset, so the cluster will use an appropriate default.\n")


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -377,6 +377,10 @@ func (o *Options) Run() error {
 			fmt.Fprintf(o.Out, "Upgradeable=False\n\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 		}
 
+		if c := findClusterOperatorStatusCondition(cv.Status.Conditions, "ReleaseAccepted"); c != nil && c.Status == configv1.ConditionFalse {
```
jottofar, line 380: I'm thinking we want to simply display `ReleaseAccepted` status, not just when it's failed. It may not be set though so something like CVO does https://github.com/openshift/cluster-version-operator/blob/484ebf2c299bdf03764ce18974525c8276507740/pkg/cvo/status.go#L395

/cc @LalatenduMohanty 
/cc @wking

wking: Or trigger on !=True, so we display on Unknown?  I dunno if we should distract users in the happy =True case

wking:
/retitle Bug 2065507: Add the ReleaseAccepted condition to the oc adm upgrade command

pkg/cli/admin/upgrade/upgrade.go
Davoska: Ah, I forgot about the `configv1.ConditionUnknown` status, thanks :+1:.

I am leaning more towards informing the users only when things don't go as planned in this case: triggering on `ReleaseAccepted != True`. I would guess that the average user doesn't know about the update being a two-phase process (loading, applying). In the case of `ReleaseAccepted == True`, things go well as always (as before introducing the `ReleaseAccepted` to the CVO). So the printing could be a little bit  distracting maybe?

But I have no problem printing the `ReleaseAccepted` condition whenever running `oc adm upgrade`, not just when it's failed. It is an interesting piece of information.

jottofar: Just go with showing the user when "things don't go as planned"

Davoska: I have changed the logic to trigger the print on `ReleaseAccepted != True` in the new commit https://github.com/openshift/oc/commit/bf761c3443dcd88934d08ef87cf9a8f270d74b27.

Davoska:
Thanks, Trevor! :+1:

PR #1115, opened 2022-04-21
Pass non-zero exit code to debug command

ardaguclu:
`oc debug node/nodename -- chroot /host invalidcommand` returns failure
as expected but does not pass exit code to caller.

This PR handles non-zero code cases and if debug pod is terminated,
it passes to it.


soltysh (approved):
/approve

PR #1116, opened 2022-04-26
Bug 2007647: Add squash-merge support into oc adm release info

ardaguclu:
Repositories, like insights-operator, use squash-merge and
`oc adm release info --changes-from` does not show changes for these repos.

This PR adds handling also squash merging and shows changes.

PR #1118, opened 2022-04-28
Fix squash merge regex

ardaguclu:
As stated in here https://github.com/openshift/oc/pull/1116#discussion_r859881400,
current squash-merge regex does not cover the cases when PR subject also
contains some other PR reference(#123).

This PR fixes squash merge regex to also cover the cases where
PR subject also contains some other PR number.


soltysh (approved):
/approve

PR #1124, opened 2022-05-05
Remove long-deprecated commands oc adm migrate etcd-ttl|image-references|legacy-hpa|storage

soltysh:
These commands where deprecated since 4.2 (see https://github.com/openshift/oc/pull/97), since they don't make any sense in 4.x at all. It's about time to remove them entirely. 

/assign @ardaguclu


ardaguclu:
This looks good to me but according to the `verify-deps` job, this PR needs `go mod tidy`?

PR #1125, opened 2022-05-05
Bug 2080416: Fix project command auto completion

ardaguclu:
Currently if user enters <tab> after `oc project`, it auto completes
as `oc project project`. This PR fixes that and after entering <tab>,
it lists all projects.


soltysh (approved):
/lgtm
/approve

PR #1128, opened 2022-05-09
Bug 2023295: [inspect] Add namespace-scoped networking resources to inspect

npinaeva:
add `egressfirewalls` and `networkpolicies` to the namespace resources.
Required for networking debugging.
`egressfirewalls` are only registered for ovn-kubernetes network plugin, but inspect will just skip them for other clusters and log
```
error: errors occurred while gathering data:
    the server doesn't have a resource type "egressfirewalls"
```

Signed-off-by: Nadia Pinaeva <npinaeva@redhat.com>

### DIFF
--- a/pkg/cli/admin/inspect/namespace.go
+++ b/pkg/cli/admin/inspect/namespace.go
@@ -21,9 +21,11 @@ func namespaceResourcesToCollect() []schema.GroupResource {
 		// this is actually a group which collects most useful things
 		{Resource: "all"},
 		{Resource: "configmaps"},
+		{Resource: "egressfirewalls"},
 		{Resource: "events"},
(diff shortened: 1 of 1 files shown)


soltysh (changes requested):
/hold
I don't think adding these resources will achieve what you want, looking at https://github.com/openshift/cluster-network-operator/pull/1432 and its associated BZ https://bugzilla.redhat.com/show_bug.cgi?id=2023295 you should be more changing what relatedObject your operator reports, including a namespace. This place serves as a generic bag of all resources that should be gathered for every namespace, which won't solve your BZ.

soltysh (approved):
Spoke with @npinaeva on slack
/lgtm
/approve
/hold cancel

PR #1134, opened 2022-05-12
[WIP] Bug 2075638: Specify an architecture when retrieving release payloads 

jeffdyoung:
https://bugzilla.redhat.com/show_bug.cgi?id=2075638

Currently ,if a user is not connected to an arm64/s390x/ppc64le cluster they can only search x86 release payloads when using `oc adm release info|extract|mirror`

We would like to be able to pass `--arch` to specify which cincinnati graph to query from `https://api.openshift.com/api/upgrades_info/v1/graph` 

cc: @Prashanth684

### DIFF
--- a/pkg/cli/admin/release/extract.go
+++ b/pkg/cli/admin/release/extract.go
@@ -106,6 +106,10 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 			# Use git to check out the source code for the current cluster release to DIR from linux/s390x image
 			# Note: Wildcard filter is not supported. Pass a single os/arch to extract
 			oc adm release extract --git=DIR quay.io/openshift-release-dev/ocp-release:4.2.2 --filter-by-os=linux/s390x
+
+			# Query the cincinnati graph for a specific version and cpu architecture and extract to another directory (default is amd64)
+			oc adm release extract 4.10.8 --to=/tmp/example --arch=arm64
+
 		`),
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Complete(f, cmd, args))
@@ -117,6 +121,7 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 	o.SecurityOptions.Bind(flags)
 	o.FilterOptions.Bind(flags)
 	o.ParallelOptions.Bind(flags)
+	o.ArchOptions.Bind(flags)
 
 	flags.StringVar(&o.ICSPFile, "icsp-file", o.ICSPFile, "Path to an ImageContentSourcePolicy file. If set, data from this file will be used to find alternative locations for images.")
 
@@ -145,7 +150,7 @@ type ExtractOptions struct {
 	SecurityOptions imagemanifest.SecurityOptions
 	FilterOptions   imagemanifest.FilterOptions
 	ParallelOptions imagemanifest.ParallelOptions
-	FilterOptions   imagemanifest.FilterOptions
+	ArchOptions     ArchOptions
 
 	ICSPFile string
 
@@ -185,7 +190,10 @@ func (o *ExtractOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args [
 	if len(o.From) > 0 {
 		args = []string{o.From}
 	}
-	args, err := findArgumentsFromCluster(f, args, o.FilterOptions)
+	if err := o.ArchOptions.Validate(); err != nil {
+		return err
+	}
+	args, err := findArgumentsFromCluster(f, args, o.ArchOptions.Arch)
 	if err != nil {
 		return err
 	}
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -61,7 +61,7 @@ func NewInfoOptions(streams genericclioptions.IOStreams) *InfoOptions {
 func NewInfo(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
 	o := NewInfoOptions(streams)
 	cmd := &cobra.Command{
-		Use:   "info IMAGE [--changes-from=IMAGE] [--verify|--commits|--pullspecs]",
+		Use:   "info IMAGE [--changes-from=IMAGE] [--verify|--commits|--pullspecs|--arch=ARCH]",
 		Short: "Display information about a release",
 		Long: templates.LongDesc(`
 			Show information about an OpenShift release.
@@ -103,8 +103,8 @@ func NewInfo(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Com
 			# Show information about the cluster's current release
 			oc adm release info
 
-			# Search information about a specific architecture's release (default is linux/amd64)
-			oc adm release info 4.10.0 --filter-by-os=linux/arm64
+			# Search the cincinnati graph for a specific version and cpu architecture for a release and architecture (default is amd64)
+			oc adm release info 4.10.0 --arch=arm64
 
 			# Show the source code that comprises a release
 			oc adm release info 4.2.2 --commit-urls
@@ -131,6 +131,7 @@ func NewInfo(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Com
 	o.FilterOptions.Bind(flags)
 	o.ParallelOptions.Bind(flags)
 	o.KubeTemplatePrintFlags.AddFlags(cmd)
+	o.ArchOptions.Bind(flags)
 
 	flags.StringVar(&o.From, "changes-from", o.From, "Show changes from this image to the requested image.")
 
@@ -179,6 +180,7 @@ type InfoOptions struct {
 	ParallelOptions imagemanifest.ParallelOptions
 	SecurityOptions imagemanifest.SecurityOptions
 	FilterOptions   imagemanifest.FilterOptions
+	ArchOptions     ArchOptions
 }
 
 func findSemanticVersionArgs(args []string) map[string]semver.Version {
@@ -224,7 +226,7 @@ const defaultGraphURL = "https://api.openshift.com/api/upgrades_info/v1/graph"
 // replaceStableSemanticArgs attempts to look up known major versions in existing public stable
 // channels.
 // TODO: perfom graph lookups from the cluster's graph endpoint and channel in preference
-func replaceStableSemanticArgs(args []string, semanticArgs map[string]semver.Version, graphURL string, filterByOs imagemanifest.FilterOptions) error {
+func replaceStableSemanticArgs(args []string, semanticArgs map[string]semver.Version, graphURL string, arch string) error {
 	if len(graphURL) == 0 {
 		graphURL = defaultGraphURL
 	}
@@ -256,10 +258,7 @@ func replaceStableSemanticArgs(args []string, semanticArgs map[string]semver.Ver
 		var found bool
 		for _, stream := range []string{"fast", "stable", "candidate"} {
 			params := url.Values{"channel": []string{fmt.Sprintf("%s-%d.%d", stream, v.Major, v.Minor)}}
-			//the api only accpets the arch portion of os/arch privided by filter-by-os parameters, so we filter out linux/ prefix
-			if filterByOs.FilterByOS != "" {
-				arch := filterByOs.FilterByOS
-				arch = strings.ReplaceAll(arch, "linux/", "")
+			if arch != "" {
 				params.Add("arch", arch)
 			}
 			u.RawQuery = params.Encode()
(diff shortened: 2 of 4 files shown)


pkg/cli/admin/release/info.go
```diff
@@ -328,24 +337,35 @@ func replaceClusterSemanticArgs(f kcmdutil.Factory, args []string, semanticArgs
...
+	//if graph-arch isn't defined, check the cluster then the graph
+	if arch == "" {
+		// attempt to find semantic args from the cluster
+		args, clusterErr := replaceClusterSemanticArgs(f, args, semanticArgs)
+
+		if len(semanticArgs) == 0 {
+			return args, clusterErr
+		}
+		// if any semantic args remain, try to fetch them from the api endpoint out of a stable channel
+		err := replaceStableSemanticArgs(args, semanticArgs, defaultGraphURL, arch)
```
Prashanth684, line 356: can get rid of this and remove the else bloock below and just have the `replaceStableSemanticArgs` outside?

pkg/cli/admin/release/extract.go
```diff
@@ -98,7 +98,14 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 
 			# Extract cloud credential requests for AWS
 			oc adm release extract --credentials-requests --cloud=aws
+
+			# Extract a specifc version and cpu architecture for a release and architecture
```
yselkowitz, line 102:
```suggestion
			# Extract a specific version and cpu architecture for a release and architecture
```

pkg/cli/admin/release/extract.go
```diff
@@ -98,7 +98,14 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
...
+			# Default architecture is the amd64
```
yselkowitz, line 103:
```suggestion
			# Default architecture is amd64
```

pkg/cli/admin/release/info.go
```diff
@@ -60,7 +60,7 @@ func NewInfoOptions(streams genericclioptions.IOStreams) *InfoOptions {
 func NewInfo(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
 	o := NewInfoOptions(streams)
 	cmd := &cobra.Command{
-		Use:   "info IMAGE [--changes-from=IMAGE] [--verify|--commits|--pullspecs]",
+		Use:   "info IMAGE [--changes-from=IMAGE] [--verify|--commits|--pullspecs|--graph-arch]",
```
yselkowitz, line 63:
```suggestion
		Use:   "info IMAGE [--changes-from=IMAGE] [--verify|--commits|--pullspecs|--graph-arch=ARCH]",
```

yselkowitz (changes requested):

jeffdyoung:
@Prashanth684 @yselkowitz I think we are good here? Who else should review?

pkg/cli/admin/release/info.go
```diff
@@ -328,23 +337,29 @@ func replaceClusterSemanticArgs(f kcmdutil.Factory, args []string, semanticArgs
...
-	args, clusterErr := replaceClusterSemanticArgs(f, args, semanticArgs)
-	if len(semanticArgs) == 0 {
-		return args, clusterErr
-	}
-	// if any semantic args remain, try to fetch them from the api endpoint out of a stable channel
-	err := replaceStableSemanticArgs(args, semanticArgs, defaultGraphURL)
-	if len(semanticArgs) == 0 || err != nil {
+
+	//if graph-arch isn't defined, attempt to find semantic args from the cluster
+	if grapharch == "" {
```
atiratree, line 348: shouldn't we also try the cluster and check if the arch is equal to grapharch?

pkg/cli/admin/release/release.go
```diff
@@ -20,9 +24,25 @@ func NewCmd(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Comm
 			registries.
 			`),
 	}
+	cmd.PersistentFlags().StringVar(&GraphArch, "graph-arch", "", "Specify the architecture to find in the release graph (amd64|arm64|ppc64le|s390x) when not connected to a cluster")
```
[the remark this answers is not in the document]
atiratree: and why not just `--arch` ?

pkg/cli/admin/release/release.go
```diff
@@ -20,9 +24,25 @@ func NewCmd(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Comm
...
 	cmd.AddCommand(NewExtract(f, streams))
 	cmd.AddCommand(NewMirror(f, streams))
 	return cmd
 }
+
+//Check to ensure that --graph-arch is a valid input
+func checkReleaseArch() error {
+	switch GraphArch {
+	case "x86_64":
+		GraphArch = "amd64"
```
atiratree, line 39: we could do the mapping in place when building the request and just check for invalid formats here

pkg/cli/admin/release/info.go
Prashanth684: the reasoning was that if a user specifies the arch, it makes sense to directly retrieve from the graph?

pkg/cli/admin/release/release.go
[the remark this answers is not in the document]
atiratree: From user perspective that is just an implementation detail. IMO `arch` would be more straightforward and remember-able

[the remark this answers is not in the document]
atiratree: or is there an advantage for the user to know this?

pkg/cli/admin/release/info.go
Prashanth684: that is true, but in a sense, --arch doesn't make sense when querying a cluster because it will be only a single arch. so either it matches or it doesn't. moreover, there is no easy way to detect the architecture of a cluster...the only way is to look at node labels to figure out their architecture.

atiratree: ok, I suppose if we describe it clearly in help

pkg/cli/admin/release/release.go
jeffdyoung: Hey @atiratree What do you mean by "mapping in place" I didn't see a way to rename/remap values passed in flags? I did pull remapping out of the case logic, but I don't think that is what you meant?

pkg/cli/admin/release/info.go
jeffdyoung: How does this look?: 

"Specify a release's architecture (amd64|arm64|ppc64le|s390x). When used, it will bypass searching the current connected cluster for the release payload."

atiratree: +1 sounds good

pkg/cli/admin/release/release.go
atiratree: I meant resolving the special cases (x86_64, aarch64) in the replaceStableSemanticArgs function, but It is matter of preference - we can leave it here as well

pkg/cli/admin/release/info.go
```diff
@@ -366,7 +380,13 @@ func findArgumentsFromCluster(f kcmdutil.Factory, args []string) ([]string, erro
 }
 
 func (o *InfoOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
-	args, err := findArgumentsFromCluster(f, args)
+
+	if err := o.ArchOptions.Validate(); err != nil {
```
atiratree, line 384: superfluous space

pkg/cli/admin/release/release.go
```diff
@@ -26,3 +30,29 @@ func NewCmd(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Comm
...
+}
+
+type ArchOptions struct {
+	Arch string
+}
+
+// Validating Arch, if --arch is passed
+func (o *ArchOptions) Validate() error {
+
+	if len(o.Arch) > 0 {
```
atiratree, line 46: too many spaces

atiratree (commented):
looks good

pkg/cli/admin/release/info.go
```diff
@@ -103,14 +103,17 @@ func NewInfo(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Com
...
-			oc adm release info 4.2.2 --commit-urls
+			oc adm release info 4.10.0 --commit-urls
 
 			# Show the source code difference between two releases
-			oc adm release info 4.2.0 4.2.2 --commits
+			oc adm release info 4.10.0 4.10.6 --commits
 
 			# Show where the images referenced by the release are located
-			oc adm release info quay.io/openshift-release-dev/ocp-release:4.2.2 --pullspecs
+			oc adm release info quay.io/openshift-release-dev/ocp-release:4.10.0 --pullspecs
```
wking, line 116: There's no `4.10.0` tag; you need an arch suffix:

```console
$ oc image info quay.io/openshift-release-dev/ocp-release:4.10.0
error: image "quay.io/openshift-release-dev/ocp-release:4.10.0" not found: manifest unknown: manifest unknown
$ oc image info quay.io/openshift-release-dev/ocp-release:4.10.0-x86_64
Name:        quay.io/openshift-release-dev/ocp-release:4.10.0-x86_64
Digest:      sha256:a63c470411c087c0568729fa5faf32cd8d5fa1db3c73e28f3989f17f5c458351
...
```

pkg/cli/admin/release/info.go
```diff
@@ -61,7 +61,7 @@ func NewInfoOptions(streams genericclioptions.IOStreams) *InfoOptions {
 func NewInfo(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
 	o := NewInfoOptions(streams)
 	cmd := &cobra.Command{
-		Use:   "info IMAGE [--changes-from=IMAGE] [--verify|--commits|--pullspecs]",
+		Use:   "info IMAGE [--changes-from=IMAGE] [--verify|--commits|--pullspecs|--arch=ARCH]",
```
wking, line 64: This seems to overlap with `--filter-by-os`?  Do we need a new option?

pkg/cli/admin/release/release.go
```diff
@@ -26,3 +30,29 @@ func NewCmd(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Comm
...
+
+		o.Arch = strings.ReplaceAll(o.Arch, "x86_64", "amd64")
+		o.Arch = strings.ReplaceAll(o.Arch, "aarch64", "arm64")
+
+		switch o.Arch {
+		case "", "amd64", "arm64", "ppc64le", "s390x":
+			return nil
+		}
+		return fmt.Errorf("%s is not a valid arch", o.Arch)
```
wking, line 55: I don't think we need to vet these before making the update service request.  We can just report "release `${VERSION}` not found in `${URI}`" or "no releases found in `${URI}`" as hints that maybe the update service on the other end doesn't like the query parameters.  Failing fast doesn't win much, and blocking access to new architectures that the update service might know about seems risky.

pkg/cli/admin/release/info.go
jeffdyoung: @wking agree, removed --arch logice, using --filter-by-os instead.

pkg/cli/admin/release/info.go
```diff
@@ -249,7 +251,14 @@ func replaceStableSemanticArgs(args []string, semanticArgs map[string]semver.Ver
...
 		for _, stream := range []string{"fast", "stable", "candidate"} {
-			u.RawQuery = url.Values{"channel": []string{fmt.Sprintf("%s-%d.%d", stream, v.Major, v.Minor)}}.Encode()
+			params := url.Values{"channel": []string{fmt.Sprintf("%s-%d.%d", stream, v.Major, v.Minor)}}
+			//the api only accpets the arch portion of os/arch privided by filter-by-os parameters, so we filter out linux/ prefix
+			if filterByOs.FilterByOS != "" {
+				arch := filterByOs.FilterByOS
+				arch = strings.ReplaceAll(arch, "linux/", "")
+				params.Add("arch", arch)
+			}
+			u.RawQuery = params.Encode()
```
jeffdyoung, line 261: @Prashanth684 @atiratree 
@wking 's comment about the overlap between `filter-by-os` and `arch` made me rethink this. After release payloads are manifest listed by default, we'll still want to get old releases by arch from the api. We probably don't want users to have to know which option to use.

For this to work, I think we'll need a plan for what will be returned from these endpoints:

`https://api.openshift.com/api/upgrades_info/v1/graph`
and 
`https://api.openshift.com/api/upgrades_info/v1/graph?arch=<arch> `

I assume that older releases (like 4.10) that aren't manifest listed will still be available? In other words, we aren't going to retroactively MF older releases?

Also, does it make sense to add a `filterbyos` param on the api? Feels wrong to filter `linux` here.

Prashanth684: i'm wondering how `--filter-by-os` will work for multi arch payloads. if the graph name is "multi", it doesn't make sense to pass in filter-by-os i guess. the arch option would probably be better in that case?

jeffdyoung: Prashanth and I talked offline, reverting back to using `--arch` to avoid confusion between single arch payloads and `multi` release payloads.

PR #1137, opened 2022-05-13
Bug 2086459: Continue inspection when some resources are not found

ardaguclu:
Purpose of inspection is to get whatever data it can in best effort.
In case of some resources are not found, inspect still can continue
for other resources.

This PR handles isnotfound error for resources and continues
to inspect. After inspection is completed, it prints error message
to warn user about non-existed resources.

### DIFF
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/inspect/inspect.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
soltysh, line 13: Nit: I like those imports nicely grouped into the following groups:
1. golang imports
2. 3-rd party
3. k8s imports
4. openshift imports

soltysh (approved):
/approve

soltysh (approved):
/lgtm

ardaguclu:
/hold

Will add bugzilla id.

PR #1143, opened 2022-05-16
Bug 2083999: Delete recently created images when --prune-over-size-limit is used

dmage:
When `--prune-over-size-limit` is used, all oversized images should be deleted, even if they are freshly created.


ricardomaraschini (approved):
/lgtm

PR #1148, opened 2022-05-18
Bug 2087679: [inspect] Add EgressQoS to inspect

oribon:
Another namespace-scoped networking resource we need to fetch
https://github.com/openshift/oc/pull/1128

Signed-off-by: Ori Braunshtein <obraunsh@redhat.com>


soltysh (approved):
/approve

PR #1150, opened 2022-05-23
Bug 2040654: Pass pod exit error codes to user

ardaguclu:
When must-gather pod exits with the value other than 0, it should
pass to terminal and `echo $?` should show that.

This PR handles exit errors and returns `CodeExitError` to achieve that.


soltysh (approved):
/lgtm
/approve

PR #1152, opened 2022-05-27
Bug 2088483: update 'oc adm catalog mirror' command to accept --continue-on-error flag

tylerslaton:


pkg/cli/admin/catalog/mirror.go
```diff
@@ -173,6 +174,7 @@ func NewMirrorCatalog(f kcmdutil.Factory, streams genericclioptions.IOStreams) *
 	flags.IntVar(&o.MaxPathComponents, "max-components", 2, "The maximum number of path components allowed in a destination mapping. Example: `quay.io/org/repo` has two path components.")
 	flags.StringVar(&o.IcspScope, "icsp-scope", o.IcspScope, "Scope of registry mirrors in imagecontentsourcepolicy file. Allowed values: repository, registry. Defaults to: repository")
 	flags.IntVar(&o.MaxICSPSize, "max-icsp-size", maxICSPSize, "The maximum number of bytes for the generated ICSP yaml(s). Defaults to 250000")
+	flags.BoolVar(&o.ContinueOnError, "continue-on-error", o.ContinueOnError, "If an error occurs while mirroring, keep going and attempt to mirror as much as possible.")
```
kevinrizza, line 177: should we explicitly call the default out as true here? otherwise we're losing that context when someone else comes along later

ardaguclu: agree, this changes the default behavior from true to false

pkg/cli/admin/catalog/mirror.go
```diff
@@ -322,10 +324,10 @@ func (o *MirrorCatalogOptions) Complete(cmd *cobra.Command, args []string) error
 		a.Mappings = mappings
 		a.SkipMultipleScopes = true
 		if err := a.Validate(); err != nil {
-			fmt.Fprintf(o.IOStreams.ErrOut, "error configuring image mirroring: %v\n", err)
+			return fmt.Errorf("error configuring image mirroring: %w", err)
```
grokspawn, line 327:
```suggestion
			return fmt.Errorf("error configuring image mirroring: %v", err)
```
Unless you're expecting that a recipient of this error will want to un-wrap it and use either the `.As()` or `e.Err == constantType` constructions to get more info or to have different handling, then it's better to continue to use `%v`.

pkg/cli/admin/catalog/mirror.go
```diff
@@ -322,10 +324,10 @@ func (o *MirrorCatalogOptions) Complete(cmd *cobra.Command, args []string) error
...
 		}
 		if err := a.Run(); err != nil {
-			fmt.Fprintf(o.IOStreams.ErrOut, "error mirroring image: %v\n", err)
+			return fmt.Errorf("error mirroring image: %w", err)
```
grokspawn, line 330:
```suggestion
			return fmt.Errorf("error mirroring image: %v", err)
```
and here

pkg/cli/admin/catalog/mirror.go
```diff
@@ -559,7 +561,12 @@ func (o *MirrorCatalogOptions) Run() error {
 	}
 	mapping, err := indexMirrorer.Mirror()
 	if err != nil {
-		fmt.Fprintf(o.IOStreams.ErrOut, "errors during mirroring. the full contents of the catalog may not have been mirrored: %s\n", err.Error())
+		err = fmt.Errorf("errors during mirroring. the full contents of the catalog may not have been mirrored: %w", err)
```
grokspawn, line 564:
```suggestion
		err = fmt.Errorf("errors during mirroring. the full contents of the catalog may not have been mirrored: %v", err)
```
and here

grokspawn (changes requested):

pkg/cli/admin/catalog/mirror.go
```diff
@@ -559,7 +561,12 @@ func (o *MirrorCatalogOptions) Run() error {
...
+		if !o.ContinueOnError {
+			return err
+		} else {
+			fmt.Fprintln(o.IOStreams.ErrOut, err.Error())
+		}
```
timflannagan, line 569: nit: this can be consolidated into a single condition:

```go
if !o.ContinueOnError {
    return err
}
fmt.Fprintln(o.IOStreams.ErrOut, err.Error())
```

grokspawn (approved):

PR #1161, opened 2022-06-03
Bug 1880865: Avoid TLS cert checking when login with --insecure-skip-tls-verify=true

ardaguclu:
In order to fix that bug https://bugzilla.redhat.com/show_bug.cgi?id=1880865, https://github.com/openshift/oc/pull/746 patch was proposed. However, it was reverted in https://github.com/openshift/oc/pull/757 because of failing in here https://github.com/openshift/oc/blob/f7835631278b03e4388f65145b56e1fd93bf4e37/vendor/k8s.io/client-go/transport/transport.go#L64-L66

The reason of failure is `CAFile` and `CAData` needs to be empty, if `Insecure` flag is set which has already been tackled by here https://github.com/openshift/oc/blob/f712166787e5cce66492d9c6434dfc7d14808cdb/pkg/cli/login/loginoptions.go#L152-L154

To fix the main bug, this PR(continuation of https://github.com/openshift/oc/pull/746) sets `CAFile` and `CAData` empty, if `--insecure-skip-tls-verify` flag is set to true.

### DIFF
--- a/pkg/cli/login/loginoptions.go
+++ b/pkg/cli/login/loginoptions.go
@@ -123,14 +123,17 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 	}
 	o.Server = serverNormalized
 	clientConfig.Host = o.Server
+	clientConfig.Insecure = o.InsecureTLS
 
-	// use specified CA or find existing CA
-	if len(o.CAFile) > 0 {
-		clientConfig.CAFile = o.CAFile
-		clientConfig.CAData = nil
-	} else if caFile, caData, ok := findExistingClientCA(clientConfig.Host, *o.StartingKubeConfig); ok {
-		clientConfig.CAFile = caFile
-		clientConfig.CAData = caData
+	if !o.InsecureTLS {
+		// use specified CA or find existing CA
+		if len(o.CAFile) > 0 {
+			clientConfig.CAFile = o.CAFile
+			clientConfig.CAData = nil
+		} else if caFile, caData, ok := findExistingClientCA(clientConfig.Host, *o.StartingKubeConfig); ok {
+			clientConfig.CAFile = caFile
+			clientConfig.CAData = caData
+		}
 	}
 
 	// try to TCP connect to the server to make sure it's reachable, and discover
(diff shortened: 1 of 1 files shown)


ardaguclu:
@stlaz @soltysh this patch fixes bug. But if you have any concerns in terms of security, I'll close that PR.

IINW, it is safe to do that because we have already done this in here https://github.com/openshift/oc/blob/f712166787e5cce66492d9c6434dfc7d14808cdb/pkg/cli/login/loginoptions.go#L152-L154

soltysh (approved):
/approve
@stlaz has final tag

pkg/cli/login/loginoptions.go
```diff
@@ -124,13 +124,18 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
...
+		if len(o.CAFile) > 0 {
+			clientConfig.CAFile = o.CAFile
+			clientConfig.CAData = nil
+		} else if caFile, caData, ok := findExistingClientCA(clientConfig.Host, *o.StartingKubeConfig); ok {
+			clientConfig.CAFile = caFile
+			clientConfig.CAData = caData
+		}
+	} else {
+		// if user has selected option --insecure-skip-tls-verify=true, TLS server certificate verification should not be attempted
+		clientConfig.Insecure = true
```
stlaz, line 138: this alone should work, no? If not, explain in another comment why you specifically leave the above CA data unpopulated

ardaguclu: Previous [attempt ](https://github.com/openshift/oc/pull/746) was exactly doing this but it failed in here https://github.com/openshift/oc/blob/f7835631278b03e4388f65145b56e1fd93bf4e37/vendor/k8s.io/client-go/transport/transport.go#L64-L66.

I think client-go transport wants CA fields are empty if `insecure` is set.

ardaguclu:
@stlaz would you mind looking at this patch one more time?, thanks.

pkg/cli/login/loginoptions.go
```diff
@@ -182,6 +187,10 @@ func (o *LoginOptions) gatherAuthInfo() error {
 		return err
 	}
 
+	if directClientConfig.Insecure && o.InsecureTLS {
```
stlaz, line 190: should this be `||`?

PR #1167, opened 2022-06-10
Bug 2090266: add --filter-by-os support in oc adm release extract and linux/amd64 for getting IS during mirror

soltysh:
/assign @ardaguclu

### DIFF
--- a/pkg/cli/admin/release/extract.go
+++ b/pkg/cli/admin/release/extract.go
@@ -91,6 +91,10 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 			if the component is not associated with source code. The command will not perform
 			any destructive actions on your behalf except for executing a 'git checkout' which
 			may change the current branch. Requires 'git' to be on your path.
+
+			If the specified image supports multiple operating systems, the image that matches the
+			current operating system will be chosen. Otherwise you must pass --filter-by-os to
+			select the desired image.
 		`),
 		Example: templates.Examples(`
 			# Use git to check out the source code for the current cluster release to DIR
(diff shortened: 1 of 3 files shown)


Prashanth684:
the extract works as expected the release image pullspec that is embedded in the install binary is incorrect. for a multi release image:

```
[psundara@fedora oc]$ ./oc image info quay.io/openshift-release-dev/ocp-release@sha256:2eb25ebb203fd613d831499ae95235224533b019e3b620f27af31cde4e159ba8
  OS            DIGEST
  linux/amd64   sha256:3b22e750d071da20ce179359f1e9b979e62e72ef5db0dd9415a8cab8418c94c0
  linux/ppc64le sha256:2184fb3393cc238234b733113d5d9f61b63cd9e26e654a3500763337b4579bb8
  linux/s390x   sha256:5e76edc70d5d026c2dc7f845e5614817b62f58e2e6840cc20224543e57a96fdc
  linux/arm64   sha256:222949cfa1426f69c91ba9c5f07ed73dc6d3917d87fd1b8d487de74f43a37504
```

the pull spec embedded in the install binary is the x86 image pullspec:

```
[psundara@fedora oc]$ ./openshift-install version
./openshift-install 4.11.0-0.nightly-multi-2022-06-07-203352
built from commit 2ca5014d0840189256709c396bf9bf28474cca97
release image quay.io/openshift-release-dev/ocp-release@sha256:3b22e750d071da20ce179359f1e9b979e62e72ef5db0dd9415a8cab8418c94c0
release architecture amd64
```

This would be a problem when installing a cluster with a control plane that is of a differing architecture than the install binary arch. the check here: https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template#L41 would fail. instead the release pull spec embedded would have to be the manifestlisted release image spec.

the release architecture also needs to be set to "multi" rather than amd64 to indicate that it is a multi-arch image

soltysh:
@Prashanth684 we've already embed all the supported platforms' binaries inside the x86 image, see [here](https://github.com/openshift/oc/blob/35c7eeb5274e1dfd8aa860107607d01d557459be/pkg/cli/admin/release/extract_tools.go#L152). By default when you invoke `oc adm release extract` we pick one that matches your current OS, but you can force a different one with `--command-os`.

PR #1169, opened 2022-06-14
Bug 1823143: add --icsp-file option to adm release subcommands

soltysh:
Replaces #1165
/assign @ardaguclu 
@zaneb fyi

### DIFF
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -29,6 +29,7 @@ import (
 	"github.com/openshift/oc/pkg/cli/image/archive"
 	"github.com/openshift/oc/pkg/cli/image/imagesource"
 	imagemanifest "github.com/openshift/oc/pkg/cli/image/manifest"
+	"github.com/openshift/oc/pkg/cli/image/strategy"
 	"github.com/openshift/oc/pkg/cli/image/workqueue"
 )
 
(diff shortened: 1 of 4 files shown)


zaneb:
Thanks! Looks good to me.

pkg/cli/image/extract/extract.go
```diff
@@ -351,9 +358,14 @@ func (o *ExtractOptions) Run() error {
 	defer close(stopCh)
 	q := workqueue.New(o.ParallelOptions.MaxPerRegistry, stopCh)
 	return q.Try(func(q workqueue.Try) {
+		icspWarned := false
 		for i := range o.Mappings {
 			mapping := o.Mappings[i]
 			from := mapping.ImageRef
+			if !icspWarned && len(o.ICSPFile) > 0 && len(from.Ref.Tag) > 0 {
+				fmt.Fprintf(o.ErrOut, "warning: --icsp-file only applies to images referenced by digest and will be ignored for tags\n")
```
eranco74, line 366: @soltysh why ICSP is ignored for images referenced by tags?

ardaguclu: If I understand it correctly, because of this https://github.com/openshift/library-go/blob/9b950a41acdc4f4ad3a6026bbb55c0569fa6b020/pkg/image/registryclient/client_mirrored.go#L77-L81

soltysh: It's not supported with tags, only the new version of ICSP will have that functionality added.

PR #1173, opened 2022-06-20
Bug 2097334: ensure kubectl name is replaced in plugin cmd

soltysh:
/assign @ardaguclu @bergerhoffer


soltysh:
> @soltysh is there any need to do the same thing for these

@ardaguclu if you find not replaced kubectl occurences in `--help` go for it.

PR #1176, opened 2022-06-22
Bug 2096855: Use filter options only when keep-manifest-list is true

ardaguclu:
Filter options are used only when `keep-manifest-list` flag is set to true for manifestlist images. 
This PR adds this conditional to raise error when `keep-manifest-list` is set to false for 
manifest list images.


soltysh (approved):
/lgtm
/approve

PR #1178, opened 2022-06-23
Bug 2015321: Add destdir special character validation in must-gather

ardaguclu:
Colon character is used by rsync command to extract pod name and
determine the type of path is local or remote.

This PR adds validation to ensure that dest-dir option does not
contain any special character possibly affecting rsync.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -270,6 +270,9 @@ func (o *MustGatherOptions) Validate() error {
 	if len(o.NodeName) > 0 && strings.ContainsAny(o.NodeName, "/%") {
 		return fmt.Errorf("--node-name may not contain '/' or '%%'")
 	}
+	if strings.Contains(o.DestDir, ":") {
+		return fmt.Errorf("--dest-dir may not contain special characters such as colon(:)")
+	}
 	return nil
 }
 


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -270,6 +270,10 @@ func (o *MustGatherOptions) Validate() error {
 	if len(o.NodeName) > 0 && strings.ContainsAny(o.NodeName, "/%") {
 		return fmt.Errorf("--node-name may not contain '/' or '%%'")
 	}
+	destDirRe := regexp.MustCompile(":")
+	if destDirRe.MatchString(o.DestDir) {
+		return fmt.Errorf("--dest-dir may contain special characters such as colon(:)")
+	}
```
deejross, line 276: I think we should use `strings.Contains` here instead of `regexp` as it's far lighter/cheaper. Also, the error message appears to be missing the word `not`. Should be `--dest-dir may NOT contain special characters...".

ardaguclu: @deejross thanks, done

omertuc:
This is a breaking change that broke our automation, `:` is a perfectly valid character in Linux

ardaguclu:
> This is a breaking change that broke our automation, `:` is a valid path character in Linux

@omertuc it is valid path for Linux but it is special for `rsync` (see https://bugzilla.redhat.com/show_bug.cgi?id=2015321#c10). This PR just adds a validation for that.

omertuc:
I'm not sure how rsync is involved in this context but shouldn't we escape the colon for rsync rather than blocking users from outputting to a target dir with `:`?

ardaguclu:
must-gather is tightly coupled with `rsync` command to copy pod logs from source to `destDir`. Do you use must-gather by without copying pod logs? because otherwise that's means, automation system does not gather pod logs?

PR #1182, opened 2022-06-26
TRT-179 Extend event-filter.html with shareable url & filters

pmtk:
Changes:
- formats html
- bumps `bootstrap` to `5.2.0-beta1`
- bumps `bootstrap-table` to `1.21.1`
- adds a text based time filter (a guide is added to the file)
- adds a regex based filter for other columns
- disables advanced search as it does not support regex filtering (it runs before column-filtering clearing out data causing empty results)
- adds an URL parameters containing filters, paging, and sorting (based on existing `bootstrap-table`'s `addrbar` extension)

### DIFF
--- a/pkg/cli/admin/inspect/event_filter_html.go
+++ b/pkg/cli/admin/inspect/event_filter_html.go
@@ -139,11 +139,11 @@ const eventHTMLPage = `
     integrity="sha384-pprn3073KE6tl6bjs2QrFaJGz5/SUsLqktiwsUTF55Jfv3qYSDhgCecCxMW52nD2"
     crossorigin="anonymous"></script>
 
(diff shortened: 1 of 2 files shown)


dgoodwin:
This looks terrific @pmtk thanks once again.

It seems to work fine to me.

I am not able to review the code, do we have someone familiar?

dgoodwin:
/assign @soltysh 

Maciej what do you think of this change? This would substantially improve debugging with this tool for TRT.

pmtk:
I had another look at my creation, I think that I need to increase debounce so the url isn't refresh so often that it conflict with text editing. I'll take a look at this during ShiftWeek

PR #1183, opened 2022-06-27
Bug 2100702: Use from-release as based image when base digest is invalid

ardaguclu:
In order to default base image, base digest set in `io.openshift.release.base-image-digest` annotation is used. However, in some cases, this digest can be invalid and can cause failing of `release new` command. Because base digest does not refer to any valid image can be used as base image.

For these cases, we can safely use `from-release` image as base image. This PR adds this.


soltysh (approved):
/lgtm
/approve

mfojtik:
overriding invalid-bug as we don't have 4.12 target release yet.

PR #1184, opened 2022-06-28
Bug 2100708: Lower chosen platform architecture/OS log level

ardaguclu:
"Chose %s/%s manifest from the manifest list." log creates noise
especially for the commands calling image extraction in a multi-threaded way.

This PR lowers of this log's level. Users will still be able to see
these logs by passing `-v=2` but log will be disabled by default.


ardaguclu:
@soltysh could you please take a look at this patch?. Thanks.

PR #1185, opened 2022-06-29
Bug 2100166: Add new IsManifestList flag into ReleaseInfo struct

ardaguclu:
`release info` and `release extract` commands need manifestlist image's
sha value instead filtered by os image sha.

Thanks to that, extracted tools will be used for all archs. This PR
adds IsManifestList field into ReleaseInfo and uses it when digest needs
to be set for these commands.


soltysh (approved):
/lgtm
/approve

Prashanth684:
can we get this fix backported for 4.11 please? it is needed for the heterogeneous 4.11 tech preview

PR #1186, opened 2022-06-30
Bug 2101885: Set completion function for get command

ardaguclu:
Completion function in get command was cut out from it and instead it is being set in [here](https://github.com/kubernetes/kubernetes/blob/e89a87f708a03c5a8331639739982d426077ad3a/staging/src/k8s.io/kubectl/pkg/cmd/cmd.go#L325). This change was done in [this PR](https://github.com/kubernetes/kubernetes/pull/108493).

In order to support completion for get command, we need to set explicitly this in oc. This PR does this.

/cc @soltysh


soltysh (approved):
/lgtm
/approve

PR #1192, opened 2022-07-05
Bug 2100702: Set hasMedataOverrides to true when ToImageBase is set

ardaguclu:
If baseDigest can not be found in current images, image passed by
from-release flag is used for --to-image-base. But in this case,
we are overriding metadata and in order to skip verifications,
this PR sets hasMetadataOverrides to true.


soltysh (approved):
/lgtm
/approve

PR #1197, opened 2022-07-06
Add unit tests for release new package

ardaguclu:
Currently, there is no unit tests running for the `release new` command. As a starter, this PR adds a couple of simple tests. In the future, there will be more complex unit tests are being added to increase coverage.

### DIFF
--- a/pkg/cli/admin/release/new_test.go
+++ b/pkg/cli/admin/release/new_test.go
@@ -1,0 +1,97 @@
+package release
+
+import (
+	"strings"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/release/new_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
+	imageapi "github.com/openshift/api/image/v1"
```
soltysh, line 11: Nit: `imagev1` is the correct alias for these APIs

soltysh: And also you're mixing openshift and k8s ones, I'd keep each in a separate group :wink:

soltysh (approved):
/approve

PR #1203, opened 2022-07-08
Bug 2105325: Add ManifestListDigest field to release info struct

Prashanth684:
When performing oc adm release info or extract commands on heterogeneous images with tags, the release image reference is incorrect. infer the manifestlist digest and add that to the release image when referencing by tag

### DIFF
--- a/pkg/cli/admin/release/extract_tools.go
+++ b/pkg/cli/admin/release/extract_tools.go
@@ -373,8 +373,11 @@ func (o *ExtractOptions) extractCommand(command string) error {
 	refExact.Ref.Tag = ""
 	// if the release image is manifestlist image, we'll not change digest with
 	// arch based sha. Because we want that the extracted tool can be used for all archs.
-	if !release.IsManifestList {
+	if len(release.ManifestListDigest) == 0 {
 		refExact.Ref.ID = release.Digest.String()
+	} else {
+		// if the image is manifestlisted, use the manifestlist digest.
+		refExact.Ref.ID = release.ManifestListDigest.String()
 	}
 	exactReleaseImage := refExact.String()
 
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -692,16 +692,14 @@ type ReleaseManifestDiff struct {
 }
 
 type ReleaseInfo struct {
-	Image          string                          `json:"image"`
-	ImageRef       imagesource.TypedImageReference `json:"-"`
-	Digest         digest.Digest                   `json:"digest"`
-	ContentDigest  digest.Digest                   `json:"contentDigest"`
-	IsManifestList bool
-	// TODO: return the list digest in the future
-	// ListDigest    digest.Digest                       `json:"listDigest"`
-	Config     *dockerv1client.DockerImageConfig `json:"config"`
-	Metadata   *CincinnatiMetadata               `json:"metadata"`
-	References *imageapi.ImageStream             `json:"references"`
+	Image              string                            `json:"image"`
+	ImageRef           imagesource.TypedImageReference   `json:"-"`
+	Digest             digest.Digest                     `json:"digest"`
+	ContentDigest      digest.Digest                     `json:"contentDigest"`
+	ManifestListDigest digest.Digest                     `json:"listDigest"`
+	Config             *dockerv1client.DockerImageConfig `json:"config"`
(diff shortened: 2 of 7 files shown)


pkg/cli/admin/release/extract_tools.go
```diff
@@ -370,12 +370,17 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
 	releaseName := release.PreferredName()
 	refExact := release.ImageRef
-	refExact.Ref.Tag = ""
 	// if the release image is manifestlist image, we'll not change digest with
 	// arch based sha. Because we want that the extracted tool can be used for all archs.
 	if !release.IsManifestList {
 		refExact.Ref.ID = release.Digest.String()
+	} else {
+		// if the image is manifestlisted and referenced by tag, use the manifestlist digest.
+		if refExact.Ref.Tag != "" {
```
ardaguclu, line 379: I wonder if we can remove that conditional. What happened if we do this any case;

```
refExact.Ref.ID = release.ManifestListDigest
```

pkg/cli/admin/release/info.go
```diff
@@ -1214,7 +1216,12 @@ func describeReleaseInfo(out io.Writer, release *ReleaseInfo, showCommit, showCo
 	now := time.Now()
 	fmt.Fprintf(w, "Name:\t%s\n", release.PreferredName())
 	if release.IsManifestList {
-		fmt.Fprintf(w, "Digest:\t%s\n", release.ImageRef.Ref.ID)
+		// if the image is manifestlisted and referenced by tag, use the manifestlist digest.
+		if release.ImageRef.Ref.Tag != "" {
```
ardaguclu, line 1220: same with the above comment. Can't we just use;

```
fmt.Fprintf(w, "Digest:\t%s\n", release.ManifestListDigest)
```

ardaguclu (commented):
Thanks for the fix

pkg/cli/admin/release/extract_tools.go
Prashanth684: it shouldn't matter as it's the same digest..i'll make the change

pkg/cli/admin/release/info.go
```diff
@@ -692,11 +692,12 @@ type ReleaseManifestDiff struct {
...
-	ContentDigest  digest.Digest                   `json:"contentDigest"`
-	IsManifestList bool
+	Image              string                          `json:"image"`
+	ImageRef           imagesource.TypedImageReference `json:"-"`
+	Digest             digest.Digest                   `json:"digest"`
+	ContentDigest      digest.Digest                   `json:"contentDigest"`
+	IsManifestList     bool
+	ManifestListDigest string
 	// TODO: return the list digest in the future
 	// ListDigest    digest.Digest                       `json:"listDigest"`
```
aleskandro, line 702: @Prashanth684 is this TODO comment related to the same work you implemented here?

ardaguclu: This is a good point

pkg/cli/admin/release/extract_tools.go
```diff
@@ -375,6 +375,9 @@ func (o *ExtractOptions) extractCommand(command string) error {
 	// arch based sha. Because we want that the extracted tool can be used for all archs.
 	if !release.IsManifestList {
 		refExact.Ref.ID = release.Digest.String()
+	} else {
```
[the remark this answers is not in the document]
ardaguclu: That means we can remove `IsManifestList` field from `ReleaseInfo`. Because this information can be extracted from `ManifestListDigest`.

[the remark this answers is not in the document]
Prashanth684: agreed. will make this change.

pkg/cli/admin/release/info.go
Prashanth684: ooh...yeah it does look like that. @soltysh was this field meant to capture the manifestlist digest. if so i can just use it.

pkg/cli/admin/release/extract_tools.go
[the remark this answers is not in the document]
Prashanth684: fixed.

pkg/cli/admin/release/info.go
Prashanth684: removed TODO and made the manifestlistdigest a digest.Digest type

soltysh (approved):
/lgtm
/approve

PR #1177, opened 2022-06-22
Bug 2100138: Add json support for release info bug printing

ardaguclu:
Currently, bug source types can be Bugzilla or Jira and there is no
differentiator point for bugs in `oc adm release info`. By passing
`--output=json`, users will be able to see bugs belong to Jira or
Bugzilla.

This PR adds json output format support for release info bugs.

PR #1204, opened 2022-07-08
release info bug printing: enable `--skip-bug-check` for `--output=json` 

hoxhaeris:
The `--output=json` option is used on the `Openshift Release Controller` Jira and Bugzilla verifier. 
This option was added to be able to distinguish between Jira and Bugzilla bugs ([PR](https://github.com/openshift/oc/pull/1177)).

We use:
`oc adm release info --bugs=/tmp/git/ --output=json <from> <to>` to get the bug list for a given input.

When `oc` fails to retrieve a certain bug, it returns: 
`error: Bug XXXXX was not retrieved` 
This breaks our use case, and the `--skip-bug-check` option is required (at the moment this option is enabled for `--output=name` only)

This (`--output=json --skip-bug-check`) will return something like:
```json
[
  {
    "id": 2088606,
    "status": "",
    "priority": "",
    "summary": "",
    "source": 0
  }
]
```

### DIFF
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -416,8 +416,8 @@ func (o *InfoOptions) Validate() error {
 	if o.SkipBugCheck && len(o.BugsDir) == 0 {
 		return fmt.Errorf("--skip-bug-check requires --bugs")
 	}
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/approve

PR #1207, opened 2022-07-15
Bug 2097830: macOS: certificate is untrusted error

deejross:
On Mac, attempting to connect to an API server with a SHA1 signature (even if it also has a SHA256 signature) will cause `oc login` to respond with:

`certificate is not trusted`

This error is of type `*errors.errorString`, so there's not a great way to detect it, other than to check the error message string. Worse yet, the error must be created (not wrapped) as `x509.CertificateInvalidError` with reason as `x509.Expired` in order to include the original error message. While the user will likely not see this error, it may appear in rare conditions.

I'm open to suggestions on how to improve this.

/assign @soltysh


pkg/cli/login/loginoptions.go
```diff
@@ -136,6 +137,10 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 	// try to TCP connect to the server to make sure it's reachable, and discover
 	// about the need of certificates or insecure TLS
 	if err := dialToServer(*clientConfig); err != nil {
+		if strings.Contains(err.Error(), "certificate is not trusted") {
+			err = x509.CertificateInvalidError{Reason: x509.Expired, Detail: err.Error()}
```
ingvagabund, line 141: Is not this actually going to hide the error? What's the difference between the current behavior vs. the behavior after merging the changes? Is the idea here to provide a better formatted error string? Or, will this change lead to a different execution of the code?

deejross: Creating the `x509.CertificateInvalidError` causes the `switch case` to catch this like any other invalid certificate issue. The idea was to use the best descriptive `error` type, and unfortunately, `x509.CertificateInvalidError.Error()` method only outputs the `Detail` string when a certain `Reason` is set. In this case, `x509.Expired` was the closest reason I could find.

I haven't found anything in kubectl that handles this yet, but oc and kubectl have very different authentication mechanisms. I agree that reusing `x509.CertificateInvalidError` in such a way is not ideal. I was trying to keep the number of changes as small as possible, but I may have to create a custom error for this condition to better handle it and reduce the user confusion of getting both an "expired" and  "not trusted" message in the same error.

deejross:
I have submitted a PR upstream to resolve this issue, so hopefully this PR can get closed: https://github.com/golang/go/pull/53986

pkg/cli/login/loginoptions.go
deejross: I tested this on a Mac, and it does resolve the issue for logins, I'm not 100% sure about other commands though. It should be possible to replicate this fix globally if it is an issue outside of login.

ingvagabund: Based on https://go-review.googlesource.com/c/go/+/418835 the Go release cycle entered the code freeze phase (Nov-Jan). So I don't expect the fix getting reviews anytime soon. Though, it's still worth reaching the upstream Go community to get their attention so they are aware of the code change request. It seems noone is aware of the PR. @deejross worth checking the thread in https://groups.google.com/g/golang-dev/c/K7oGURi0wTM/m/HTyfcmY0BwAJ. Comment from Ian in https://groups.google.com/g/golang-dev/c/K7oGURi0wTM/m/HTyfcmY0BwAJ:
```
In general it should not be necessary to ask for an additional Googler
review. If your CL shows up on https://go.dev/s/needs-review, then
some Googler will get to it soon.

If your CL does not show up on that list, then most likely it has an
unresolved comment, or the trybots have not been run.

Please do ping if your CL is not on that list and you don't understand
why, or if it has been on the list for several days. But in general
pinging should not be required.
```

ingvagabund: When it comes to resolving this issue. @deejross if you are confident this change fixes the login issue without any negative side effects, let's proceed. @kasturinarra would you please perform pre-merge testing on all architectures for the `oc login`? Just to be sure we are not breaking anything.

ardaguclu: I think, we can put a TODO explaining why we did this and when we can remove this hacky solution with the links.

kasturinarra: @ingvagabund thanks, i will ask @zhouying7780 to take a look at it, thanks !!

pkg/cli/login/loginoptions.go
```diff
@@ -136,6 +137,10 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 	// try to TCP connect to the server to make sure it's reachable, and discover
 	// about the need of certificates or insecure TLS
 	if err := dialToServer(*clientConfig); err != nil {
+		if strings.Contains(err.Error(), "certificate is not trusted") {
```
ardaguclu, line 140: Can't we do this workaround only if the err is not type `x509.CertificateInvalidError`;

```go
if _, ok := err.(x509.CertificateInvalidError); !ok {
if strings.Contains(err.Error(), "certificate is not trusted") {
}
}
```

Otherwise, in the future when problem is fixed. We'll still be wrapping the correct error message to the same.

ingvagabund: The more the fix gets executed only for Mac the better.

cfergeau:
> I tested this on a Mac, and it does resolve the issue for logins, I'm not 100% sure about other commands though. It should be possible to replicate this fix globally if it is an issue outside of login.

Actually I think I'm hitting this problem in a different codepath.
https://github.com/crc-org/crc/blob/641ff18f521344a35df9e48f82fe31a49e2b20bd/pkg/crc/machine/kubeconfig.go#L130-L152
We call `tokencmd.RequestToken` with only one root CA, which should be fine, but looking at `RequestToken` implementation, it adds the root certs anyway:
https://github.com/openshift/oc/blob/142cb4431e96f510b4ed4c61a13d68cb13a76c3c/pkg/helpers/tokencmd/request_token.go#L186-L197
and `transportWithSystemRoots` has this code:
https://github.com/openshift/oc/blob/142cb4431e96f510b4ed4c61a13d68cb13a76c3c/pkg/helpers/tokencmd/request_token.go#L442-L480
which returns a `certificate is not trusted` error which is not currently caught.

deejross:
The upstream changes on Mac that lead to this were quite extensive, so I'm not surprised there would be issues with some use cases. The real question becomes, do we push upstream to resolve them, or do we try and fix oc-specific use cases ourselves?

cfergeau:
> The real question becomes, do we push upstream to resolve them, or do we try and fix oc-specific use cases ourselves?

This all depends on how fast a fix would land upstream, and be backported to 1.18/1.19 releases. My feeling is that this would take time (say more than a few weeks)? An alternative could be to get the a patch in the fedora/RHEL packages, but that might be a hard sell.
We have a lot more control over when it gets fixed if we add oc-specific changes to avoid the issue.

In short, I would pursue both. The oc changes would be a short-term fix and it would avoid annoying user-visible failures on macOS. The real fix belongs in go, but it may take some time to be fixed, so this is going to be a longer term fix. Once the issue is fixed in the go compiler(s) we use to build oc releases, the oc changes can be removed.

ardaguclu:
In my opinion, we need to fix that problem in oc. Because even we fix it in Go, that will not mean that some versions of oc(4.12) will include fix.

cfergeau:
This is also impacting odo, and users of odo https://github.com/redhat-developer/vscode-openshift-tools/issues/2693

cfergeau:
I rebuilt odo with the patch, it helps a bit, but is not enough.
- odo.git built with golang 1.19, no oc patches
```
% ./odo login -u developer https://api.crc.testing:6443   
Connecting to the OpenShift cluster

 �  x509: �kube-apiserver-lb-signer� certificate is not trusted
```

- odo.git built with golang 1.19 with the patch from this PR
```
% ./odo login  -u developer https://api.crc.testing:6443        
Connecting to the OpenShift cluster

The server is using an invalid certificate: x509: certificate has expired or is not yet valid: x509: �kube-apiserver-lb-signer� certificate is not trusted
You can bypass the certificate check, but any data you send to the server could be intercepted by others.
Use insecure connections? (y/n): y

 �  x509: �ingress-operator@1669088888� certificate is not trusted
```

- odo.git built with go 1.17, no patch
```
% ./odo login  -u developer https://api.crc.testing:6443   
Connecting to the OpenShift cluster

The server uses a certificate signed by an unknown authority.
You can bypass the certificate check, but any data you send to the server could be intercepted by others.
Use insecure connections? (y/n): y

Authentication required for https://api.crc.testing:6443 (openshift)
Username: developer
Password: 
```

cfergeau:
I did the same change as in this PR in a few more places in https://github.com/cfergeau/oc/commits/macos-cert-not-trusted , see commit https://github.com/cfergeau/oc/commit/d6ee3953e604c7e6dc424de66a289cb566cddaff

This fixes both crc and oc when I patch them to use this branch.

soltysh:
/hold
this should not  merge in the current form

gbraad:
Can you explain what the follow-up actions will be? As it is currently held, what does that mean for a possible solution?

cfergeau:
Newer go versions should no longer have this problem, see https://github.com/golang/go/issues/56891

ingvagabund:
https://github.com/golang/go/issues/57427 mentions 1.19.5

cfergeau:
> [golang/go#57427](https://github.com/golang/go/issues/57427) mentions 1.19.5

1.18.10, 1.19.5 and 1.20 should all have this fixed.

ingvagabund:
Fixed in golang 1.19.5
/close

PR #1213, opened 2022-07-25
Bug 2092731: Change error message to signal passing keep-manifest-list flag

ardaguclu:


pkg/cli/image/manifest/manifest.go
```diff
@@ -388,7 +388,7 @@ func ManifestToImageConfig(ctx context.Context, srcManifest distribution.Manifes
 		return base, layers, nil
 
 	default:
-		return nil, nil, fmt.Errorf("unknown image manifest of type %T from %s", srcManifest, location)
+		return nil, nil, fmt.Errorf("use --keep-manifest-list option for found image manifest type %T from %s", srcManifest, location)
```
soltysh, line 391: Not quite, I'd suggest adding a new case matching only ManifestList types (check to use all or an interface that will cover all) and only suggest `--keep-manifest-list` for those, and keep the default as it was before. This might be confusing as it stands.

soltysh (changes requested):

pkg/cli/image/manifest/manifest.go
ardaguclu: I think, `*manifestlist.DeserializedManifestList` is the thing you suggested

soltysh (approved):
/lgtm
/approve

PR #1215, opened 2022-07-29
bug 2095708: oc adm inspect: check a resource exists before its inspection

ingvagabund:
Collect all resources served by the server through the discovery client and check whether an about-to-be-inspected resource exists to avoid getting an error when it does not.

```
$ ./oc adm inspect co/console
Gathering data for ns/openshift-console-operator...
W0729 16:26:09.139371 1931005 util.go:119] the server doesn't have a resource type egressfirewalls, skipping the inspection
Gathering data for ns/openshift-console...
W0729 16:26:15.981392 1931005 util.go:119] the server doesn't have a resource type egressfirewalls, skipping the inspection
Wrote inspect data to inspect.local.8870216383603463254.
```

### DIFF
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -236,8 +236,28 @@ func (o *InspectOptions) Run() error {
 	}
 	defer o.logTimestamp()
 
+	// Collect all resources served by the server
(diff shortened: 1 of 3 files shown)


soltysh (approved):
/approve

PR #1217, opened 2022-08-05
Bug 2111537: (image info): Introduce show-multiarch flag

ardaguclu:
Currently if the passed image is multiarch image, we are throwing error
to force user to use `--filter-by-os` flag with the listed image manifests in
that error.

This PR introduces new `show-multiarch` flag. `oc image info --show-multiarch`
command will continue working exactly the same way as now for single arch images.
However, this command(with new flag) will also show informations of platform specific images
embedded in multiarch image.

This flag supports json and default printing and if the image is multiarch image,
result will be in json array format if output is json.


pkg/cli/image/info/info.go
```diff
@@ -143,7 +145,7 @@ func (o *InfoOptions) Run() error {
 				icspWarned = true
 			}
 
-			var image *Image
+			var image []*Image
```
ingvagabund, line 148: s/image/images

ingvagabund (changes requested):

pkg/cli/image/info/info.go
ardaguclu: @ingvagabund thanks. Done.

soltysh (approved):
/lgtm
/approve

PR #1232, opened 2022-09-06
OCPBUGS-843: Add limitranges into common namespaces for inspection

ardaguclu:
This PR adds `limitrages` resource into common namespaces for `oc adm inspect`.
`limitranges` is a useful resource as stated in referenced bug
and it takes a few kilobytes storage.


soltysh (changes requested):
/hold
we don't want to grow this list indefinitely, every resource on that list needs to have a purpose and be valuable to identifying problems

PR #1233, opened 2022-09-07
Remove version patching for mac arm64 archs

ardaguclu:
When binary versions are patched, mac os arm64 architectures does not allow binaries to be run by warning that binary is changed. To let `oc` and `openshift-install` working on mac os arm64, this PR removes version injection for these architectures.

Drawback of this PR is for that specific architectures, `oc version` and `openshift-install version` commands will not return correct data.

### DIFF
--- a/pkg/cli/admin/release/extract_tools.go
+++ b/pkg/cli/admin/release/extract_tools.go
@@ -189,10 +189,9 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			NewArch: true,
 			Mapping: extract.Mapping{Image: "cli-artifacts", From: "usr/share/openshift/mac_arm64/oc"},
 
-			LinkTo:               []string{"kubectl"},
-			Readme:               readmeCLIUnix,
-			InjectReleaseVersion: true,
-			ArchiveFormat:        "openshift-client-mac-arm64-%s.tar.gz",
+			LinkTo:        []string{"kubectl"},
+			Readme:        readmeCLIUnix,
+			ArchiveFormat: "openshift-client-mac-arm64-%s.tar.gz",
 		},
 		{
 			OS:      "linux",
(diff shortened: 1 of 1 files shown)


soltysh (changes requested):
/hold
we can't remove that functionality, at most I'm inclined to accept a warning which will explain users how to deal with these extracted binaries on mac

cfergeau:
> /hold we can't remove that functionality, at most I'm inclined to accept a warning which will explain users how to deal with these extracted binaries on mac

Where would that warning be?
The problem is that https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.11.1/openshift-client-mac-arm64-4.11.1.tar.gz has non-working binaries because the version patching invalidated the binary checksum contained in the M1 binaries.
The right way of fixing this is to readd a valid signature right after doing the binary patching, but this is not going to be possible short-term. In the mean time, having working binaries with wrong version information on https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.11.1/openshift-client-mac-arm64-4.11.1.tar.gz would be much better than something which crashes.

rgrunber:
@cfergeau 

Hi, I was looking at https://mirror.openshift.com/pub/openshift-v4/aarch64/clients/ocp/4.14.7/ and noticed :

```
openshift-client-mac-arm64-4.14.7.tar.gz: FAILED
openshift-install-mac-arm64-4.14.7.tar.gz: FAILED
sha256sum: WARNING: 2 computed checksums did NOT match
```

when comparing against the listed `sha256sum.txt` file.

This started happening in the [4.11.11](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.11.11/) release folder ([4.11.10](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.11.10/) seemed valid). Is the mismatch related to this issue? I wasn't able to find documentation to indicate this was expected.

cfergeau:
I don't know how/when the sha256sum.txt file is generated, and when re-signing is triggered, but the timeframe and the binaries modified by the commit match: https://github.com/openshift/oc/commit/1fe58b72ebd785892d2144f220efec03e559751a

@ardaguclu any idea about this?

ardaguclu:
> @cfergeau
> 
> Hi, I was looking at https://mirror.openshift.com/pub/openshift-v4/aarch64/clients/ocp/4.14.7/ and noticed :
> 
> ```
> openshift-client-mac-arm64-4.14.7.tar.gz: FAILED
> openshift-install-mac-arm64-4.14.7.tar.gz: FAILED
> sha256sum: WARNING: 2 computed checksums did NOT match
> ```
> 
> when comparing against the listed `sha256sum.txt` file.
> 
> This started happening in the [4.11.11](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.11.11/) release folder ([4.11.10](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.11.10/) seemed valid). Is the mismatch related to this issue? I wasn't able to find documentation to indicate this was expected.

@rgrunber thanks for commenting. As far as I can see, this issue is related to the commit @cfergeau mentioned above. I'm investigating and working on it to fix.

PR #1234, opened 2022-09-08
Add --overwrite flag into pod security violation error message

ardaguclu:
If user gets a pod security violation error, `oc` suggests a command how to overcome that. However, newly created namespaces already have `pod-security.kubernetes.io/enforce: restricted` this label and suggested command is failing by
`error: 'pod-security.kubernetes.io/enforce' already has a value (restricted), and --overwrite is false`.

This PR adds `--overwrite` flag into suggested commands in case new namespace might already contain such labels.


ardaguclu:
/assign @stlaz 

could you PTAL this trivial fix?. Thanks.

soltysh (approved):
/approve

ardaguclu:
How one very simple PR can create such noise :(;

/retest-required

PR #1235, opened 2022-09-08
(oc debug): Create temporary namespace for node, image-stream debugging

ardaguclu:
Currently `oc` returns pod security violation error when user tries to debug node or image-stream, if the current namespace is not privileged.

This hurts the usability of `oc debug` command by adding new step to use that command. This PR adds creating temporary namespace like we do in must-gather, when user wants to debug nodes or image streams.


ardaguclu:
cc: @stbenjam

ardaguclu:
There was an similar attempt for `oc debug` in the past and it was reverted https://github.com/openshift/oc/pull/668. For the same reasons, I'm closing this PR.

PR #1237, opened 2022-09-09
pkg/cli/admin/release/extract: Unify extraction cases

wking:
There's no reason we can't do things like "extract manifests to a directory" and "extract a single thing to a file", etc. simultaneously.  This pull request is the first portion of changes to consolidate shared logic, and reduce the amount of case-specific code.  It should mostly be a no-op refactor for users, except for the extension of verification coverage to cases that previously lacked verification.


soltysh (approved):
/approve

PR #1240, opened 2022-09-21
release: Add --command-arch parameter to 'release extract'

cfergeau:
It's currently possible to override the OS of the extracted command with
--command-os, but its arch is hardcoded to be amd64.
This is inconvenient if one wants to extract for example a darwin/arm64
oc binary.
This commit introduces a --command-arch parameter to allow to override
the arch of the binary which will be extracted.

This slightly changes `oc` behaviour, `oc adm release extract
--command=oc` will no longer extract an amd64 `oc` binary, but it will
choose instead a native one.


pkg/cli/admin/release/extract.go
```diff
@@ -129,7 +129,8 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 	flags.StringVar(&o.SigningKey, "signing-key", o.SigningKey, "Sign the sha256sum.txt generated by --tools with this GPG key. A sha256sum.txt.asc file signed by this key will be created. The key is assumed to be encrypted.")
 
 	flags.StringVar(&o.Command, "command", o.Command, "Specify 'oc' or 'openshift-install' to extract the client for your operating system.")
-	flags.StringVar(&o.CommandOperatingSystem, "command-os", o.CommandOperatingSystem, "Override which operating system command is extracted (mac, windows, linux). You map specify '*' to extract all tool archives.")
+	flags.StringVar(&o.CommandOperatingSystem, "command-os", o.CommandOperatingSystem, "Override which operating system command is extracted (mac, windows, linux). You may specify '*' to extract all tool archives.")
+	flags.StringVar(&o.CommandArch, "command-arch", o.CommandArch, "Override the arch of the command which will be extracted (amd64, arm64, ...). If --command-os is '*', this is ignored")
```
ardaguclu, line 133: I'd expect a default value for `commandArch` which is `amd64`?

cfergeau: Ah right, `amd64` as a default should avoid the behaviour change mentioned in the commit log, I'll try that!

cfergeau: Actually, this would change `oc adm release extract --command=oc` behaviour, at the moments this extracts oc for the native OS/arch, which seems right. Setting `commandArch` to `amd64` as the default would break this.

cfergeau: New iteration keeps `--command-os=darwin --command=oc` behaviour unchanged, it will always use amd64 unless `--command-arch` is used. `--command=oc` is also unchanged, it will use the native arch

ardaguclu:
This PR https://github.com/openshift/oc/pull/1305 was merged and fixes that problem. 

/close

PR #1241, opened 2022-09-21
OCPBUGS-1638: Increase default value of cache TTL to 6 hours

soltysh:
/assign @ardaguclu @ingvagabund


ingvagabund:
/retitle OCPBUGS-1638: Increase default value of cache TTL to 6 hours

PR #1242, opened 2022-09-23
Bug 2059125: release extract: Add binary re-signing for macos arm64

ardaguclu:
After injecting version numbers or images, extracted binaries need to be re-signed on macos arm64 platforms. Otherwise, these binaries should be manually signed via `codesign` tool. 

This PR mimics re-signing method of Go uses[1]. For standalone arm64 binaries, it re-signs as is. However, there is no straight forward way of re-signing binaries inside the archived files. Because, modules don't allow it.

That's why, this PR;
- extracts archived file to tmp directory
- re-signs binaries in there(skipping irrelevant ones such as README.md)
- overwrites previous archived file

1: https://go-review.googlesource.com/c/go/+/272257


ardaguclu:
/uncc @deads2k @mfojtik 
/hold

Until extensive manual testing is completed.

pkg/cli/admin/release/codesign/machoresign.go
```diff
@@ -0,0 +1,191 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func extractTarToTmpAndSign(path string) (string, error) {
+	tempDir, err := os.MkdirTemp(os.TempDir(), "oc-release-extract-*")
```
ingvagabund, line 124: What if there's no /tmp mounted? One can override the location by setting TMPDIR [1]. I wonder whether consumers of `oc adm release extract` assumes existence of /tmp and are given permissions to create dirs there in every case.

[1] https://pkg.go.dev/os#TempDir

pkg/cli/admin/release/codesign/machoresign.go
```diff
@@ -0,0 +1,191 @@
...
+			if _, err := wr.WriteAt(cs, int64(cmd.Dataoff)); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func signTarFile(path, command string, links []string) error {
```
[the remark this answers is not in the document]
ardaguclu: > Does the newly written tar file respects that same file permissions of the original tar? 

Yes, it is one of the important thing we need to consider and this PR preserves file permissions of original tar.

> Can we always expect the release image will always provide tar and not e.g. gzip archives?

That is a good point. In my initial draft, I had also implemented for `zip` archives. But after digging through the code, I noticed that `zip` is only supported for Windows(see [here](https://github.com/openshift/oc/blob/a1119e4cd75df284e74bfd731f22b65b2dac0c75/pkg/cli/admin/release/extract_tools.go#L228)) and we are only interested in macos arm64.

pkg/cli/admin/release/codesign/machoresign.go
ardaguclu: For users;
This would not be a catastrophic problem. Because this PR continues as is by [printing error message](https://github.com/openshift/oc/pull/1242/files#diff-2843ad56df617fe91ad742d36998e3292b81b86d85202ce5dc7d25ec05effd37R659) when there is any problem occurred(including temp directory permission error).

For our automation system;
This would create problem but I'm not sure we can use another safer directory than `/tmp/`

pkg/cli/admin/release/codesign/machoresign.go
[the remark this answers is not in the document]
ingvagabund: My understanding is the `oc adm release extract` command is executed on a linux machine/container. So zip, gzip, tar and other formats are supported. Not necessarily present.

pkg/cli/admin/release/codesign/machoresign.go
ingvagabund: @sosiouxme @thiagoalessio for awareness

pkg/cli/admin/release/codesign/machoresign.go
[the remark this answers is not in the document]
ardaguclu: Supported targets are listed in here https://github.com/openshift/oc/blob/a1119e4cd75df284e74bfd731f22b65b2dac0c75/pkg/cli/admin/release/extract_tools.go#L173. In reality, we support `zip` for Windows and `tar.gz` for the rest.

IMO, since we are re-signing archive files we produce as `tar.gz`, we can re-sign only `tar.gz` and ignore other formats.

ingvagabund:
This is only an improvement of the current case for MacOS arm64. In case the releasing pipeline needs to be updated to take this change into account it will be at the step which produces archives on https://mirror.openshift.com/pub/.

ardaguclu:
/assign @soltysh 

Could you please take a look at this patch?. Thanks.

pkg/cli/admin/release/codesign/codesign.go
```diff
@@ -0,0 +1,272 @@
+// Copyright 2020 The Go Authors. All rights reserved.
```
soltysh, line 1: Nit: can you move this under `pkg/cli/admin/internal/codesign` pacakge, this will ensure that nobody consuming oc will ever use it. Especially that we're copying golang code.

soltysh (changes requested):

pkg/cli/admin/release/extract_tools.go
```diff
@@ -649,6 +654,12 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			klog.V(2).Infof("Unable to set extracted file modification time: %v", err)
 		}
 
+		if (target.InjectReleaseVersion || target.InjectReleaseImage) && target.SignMachOBinary {
+			if err = codesign.ResignMacho(layer.Mapping.To, target.AsArchive, target.Command, target.LinkTo); err != nil {
+				klog.Infof("Unable to resign macho binaries:  %v", err)
```
soltysh, line 659: Nit: this could probably be warning.

soltysh (approved):
/approve

ardaguclu:
/hold

aws-serial looks bad.

PR #1243, opened 2022-09-26
add support for --next flag in oc adm release new

PratikMahajan:
Add support for `--next` flag like `--previous` while creating new releases. 
This will enable us to not have to repromote release payloads if release team bump a release

Checks done: 
- [x] cincinnati graph data interprets the next field correctly in payloads as expected


pkg/cli/admin/release/new.go
```diff
@@ -297,6 +299,7 @@ type CincinnatiMetadata struct {
 
 	Version  string   `json:"version"`
 	Previous []string `json:"previous"`
+	Next []string `json:"next"`
```
wking, line 302: `omitempty`?  It's a useful safety-valve, but there's enough risk in recommending vB update to an older vA this way, that I dunno if we want `"next": []` in metadata reminding folks that this is an option ;)

wking (approved):
/lgtm

PratikMahajan (commented):
/retest

PratikMahajan (commented):
/retest

PR #1248, opened 2022-09-28
Bug 2033167: oc extract: Create target directory if not exist

ardaguclu:
Currently, `oc extract` command returns error, if target directory is not existed. This PR creates all nested directories prior to extract data.

This will easen usage of `oc extract` command.

### DIFF
--- a/pkg/cli/extract/extract.go
+++ b/pkg/cli/extract/extract.go
@@ -4,9 +4,10 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
+
+	"github.com/pkg/errors"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


pkg/cli/extract/extract.go
```diff
@@ -105,19 +105,24 @@ func (o *ExtractOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args [
 		return err
 	}
 
+	if o.TargetDirectory != "-" {
```
ingvagabund, line 108: I'd prefer to move creation of the target directory as close to the line of code where it is needed. E.g. https://github.com/openshift/oc/blob/master/pkg/cli/extract/extract.go#L167

ingvagabund: When creating the target directory right before it is needed one might get "permission denied". Knowing this in advance would help to save the entire extraction step on the other hand. Which makes the current change more preferable on the other hand.

ardaguclu:
e2e-agnostic-cmd failure is real
/hold

soltysh (changes requested):

ardaguclu:
None of the failures is related to this patch;

/retest-required
/hold cancel

soltysh (approved):
/approve

PR #1263, opened 2022-10-12
Bug 2093046: oc debug: Add priorityClassName into node debugging pod template

ardaguclu:
When user tries to debug a pod, priorityClassName is automatically copied to this debug pod. However, there is no such information for node debugging.

This PR adds "openshift-user-critical" priorityClassName into node debugging pod which means that "it is fine node debugging pod to be preempted by user-workload and OOMKilled."[1]

[1] https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#priority-classes


ardaguclu:
cc: @stbenjam

soltysh (approved):
/lgtm
/approve

PR #1264, opened 2022-10-13
pkg/cli/admin/release/extract: Don't print image metadata verification messages when extracting to stdout

dcbw:
Otherwise it confuses other tools that take the JSON output from extraction.

Fixes: https://github.com/openshift/oc/pull/1237

@wking


soltysh (approved):
/lgtm
/approve

PR #1265, opened 2022-10-14
oc debug,version: minor todo improvements

ardaguclu:
This PR;

- uses `GetKustomizeModVersion()` function to get kustomize version instead using static one. Because https://github.com/kubernetes/kubernetes/pull/109430 patch was merged and landed on o/k
- adds check to not create temporary namespace if `dry-run` flag is passed to `oc debug node`. It is redundant operation for dry-run.
- relies on `printFlags` functions instead manual output calculations. `printFlags` already supports required functionality.


soltysh (approved):
/approve

PR #1270, opened 2022-10-19
Bug 2015119: bump(k8s.io/kubectl) to pick up k/k#110764

soltysh:
/assign @ardaguclu


soltysh:
Manually overwriting valid-bug since this is picking 4.11 change into 4.10. 
/remove-label bugzilla/invalid-bug
/label bugzilla/valid-bug

PR #1273, opened 2022-10-21
WRKLDS-593 - reordering the stack unwinding to fix how we execute the backup collection

sferich888:


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -311,6 +301,17 @@ func (o *MustGatherOptions) Run() error {
...
 
+	// Due to 'stack unwiding', this should happen after 'cleanupNamespace()' - WRKLDS-593
+	// Due to 'stack unwiding', this should happen after 'clusterState' printing, to ensure that we always
+	//  print our ClusterState information.
+	runBackCollection := true
+	defer func() {
+		if !runBackCollection {
+			return
+		}
+		o.BackupGathering(context.TODO(), errs)
```
wking, line 312: Shifting this after the `getNamespace` call means that failing `getNamespace()` call will bail out of `Run` without performing `BackupGathering`.  But we want `BackupGathering` even when we cannot create namespaces.  Something like this could cover both:

```go
var ns *corev1.Namespace
var cleanupNamespace func()

// ... ensure resource cleanup unless instructed otherwise ...
if !o.Keep {
  if cleanupNamespace != nil {
    defer cleanupNamespace()
  }
}

// Due to 'stack unwiding', this should happen after 'clusterState' printing, to ensure that we always
//  print our ClusterState information.
runBackCollection := true
defer func() {
  if !runBackCollection {
    return
  }
  o.BackupGathering(context.TODO(), errs, ns) // FIXME: also updating BackupGathering to take a possibly-nil *corev1.Namespace, which gets namespace gathering if it's not nil.
}()
	

// Get or create "working" namespace ...
var err error
ns, cleanupNamespace, err = o.getNamespace()
...
```

Then we'd have a flow tree like:

* Namespace creation fails
    1. Pop the deferred gather.  Gathers the default stuff, but no temporary namespace.
    2. Pop the deferred namespace removal.  Nothing to remove.
* Namespace creation succeeds
    * The main gather still fails, so `runBackCollection` is still `true`.
        1. Pop the deferred gather.  Gathers the default stuff and also (new with this PR) the temporary namespace.
        2. Pop the deferred namespace removal.  Removes the temporary namespace.
    * The main gather succeeds, so runBackCollection` is `false`.
        1. Pop the deferred gather.  `runBackCollection` is `false`, so it's a no-op.
        2. Pop the deferred namespace removal.  Removes the temporary namespace.

soltysh (approved):
/approve
/assign @ardaguclu

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: With respect to ordering, I agree with @wking.

We can still run backing up even if the `getNamespace` fails and the proposed code snippet above ^ sounds reasonable to me. 

However, backing up also must-gather temporary namespace might be handled in separate PR under the light of elaborated decision tree above.

PR #1276, opened 2022-10-24
release: extract linux/arm64 binaries from all payloads

yselkowitz:
A linux/arm64 installer binary will be created by https://github.com/openshift/installer/pull/6514, but this remains compatible with versions that don't have.  linux/arm64 oc has been available for a while.


yselkowitz:
Test failure is unrelated
/retest

yselkowitz:
clearly unrelated, please consider overriding
/retest-required

PR #1277, opened 2022-10-25
remove support for empty files and stdout in oc registry login 

atiratree:
to be in line with podman behaviour. Changes were originally introduced in https://github.com/openshift/oc/pull/1011

- workaround for empty files: delete the empty file before
- support for stdout output is removed


soltysh (approved):
/lgtm
/approve

PR #1284, opened 2022-11-07
OCPBUGS-1669: Use gokrb5 for GSSAPI auth

ardaguclu:
This PR uses `jcmturner/gokrb5/v8` as gssapi library instead the C dependent unmaintained repo. 

This PR is basically continuation of https://github.com/openshift/oc/pull/819

/cc @LorbusChris


ardaguclu:
/hold 
until the pre-merge tests

pkg/helpers/tokencmd/negotiator_gssapi.go
```diff
@@ -4,195 +4,91 @@
...
 	return &gssapiNegotiator{principalName: principalName}
 }
 
 func (g *gssapiNegotiator) Load() error {
-	_, err := g.loadLib()
-	return err
+	return nil
 }
 
 func (g *gssapiNegotiator) InitSecContext(requestURL string, challengeToken []byte) (tokenToSend []byte, err error) {
```
ardaguclu, line 43: @LorbusChris we are not using `challengeToken` now. I checked it and it seems that imported library handles that automatically, could you please confirm me?

pkg/helpers/tokencmd/negotiator_gssapi.go
```diff
@@ -4,195 +4,91 @@
...
+		}
+		keytab, err := krb5kt.Load(krb5KeytabLocation)
 		if err != nil {
-			return nil, convertGSSAPIError(err)
+			return nil, err
 		}
-		defer nameBuf.Release()
 
-		name, err := nameBuf.Name(lib.GSS_C_NT_HOSTBASED_SERVICE)
+		g.client = krb5client.NewWithKeytab(g.principalName, config.LibDefaults.DefaultRealm, keytab, config)
```
ardaguclu, line 64: @LorbusChris as far as I can see, realm is required. It seems to be safe to me to use `LibDefaults.DefaultRealm` in this case, what do you think?

pkg/helpers/tokencmd/negotiator_gssapi.go
```diff
@@ -4,195 +4,91 @@
...
-		klog.V(5).Infof("InitSecContext returned GSS_S_COMPLETE")
-		g.complete = true
-		return outgoingToken.Bytes(), nil
-	case gssapi.ErrContinueNeeded:
-		klog.V(5).Infof("InitSecContext returned GSS_S_CONTINUE_NEEDED")
+
+	_, key, err := g.client.GetServiceTicket(servicePrincipleName)
+	if err != nil {
+		g.client = nil
 		g.complete = false
```
ardaguclu, line 80: I said complete to false in case of errors which is another difference from the referenced PR

pkg/helpers/tokencmd/negotiator_gssapi.go
LorbusChris: agreed :+1:

pkg/helpers/tokencmd/negotiator_gssapi.go
LorbusChris: I'm no expert here, but that would be my impression tool. These changes will have to be tested against a real cluster that uses kerberos auth for verification.

zhouying7780:
pre-merge test done , no issue found

stlaz:
/hold
I'll need to review this thoroughly.

I wonder if @simo5 would care to look? :slightly_smiling_face:

simo5:
Ah @stlaz yeah, the only thing I care to say is that you should *not* use that gokrb5 package.
I can provide explanations if needed, but I did a long check on it a while ago and it is very far from being a security package that can be used in production, and feature wise is too incomplete to be useful.

ardaguclu:
> Ah @stlaz yeah, the only thing I care to say is that you should _not_ use that gokrb5 package. I can provide explanations if needed, but I did a long check on it a while ago and it is very far from being a security package that can be used in production, and feature wise is too incomplete to be useful.

Thanks @simo5  for your valuable comment. We are trying to remove the dependency to github.com/apcera/gssapi package in oc. Is there any go module you can suggest for our case?

simo5:
> The problem with binding a system C library is that requires cgo, which means the resulting binary is tied to the toolchain with which it was compiled, forcing us to keep build redistributable binaries with older RHEL versions.

Using FFI you can build in a way that can support multiple versions.
@derekparker can probably help you with the details.
We build golang itself to dynamically link to multiple versions of OpenSSL for the FIPS case for example, so it is definitely doable.

> IOW, even when the rest of OCP migrates to RHEL9, we would need to keep building oc on RHEL8 for a while longer, which is why we are looking to switch to a pure golang implementation. Certainly one decent one exists?

I am not aware of any implementation of the krb5 and GSSAPI protocols in Go that is reasonable to use at this time. The complexity of the protocols and implementations makes me doubt it. It is not trivial code, requires sustained maintenance effort and it is not exactly a sexy thing to work on these days.

I personally think it is better to invest in sane ways to deal with FFI and dynamic loading C code that can be used for multiple cases, then to try to pull in half-assed implementations "because they are native".

Even using IPC should be considered. For example the kernel uses IPC -> GSS-Proxy to avoid dealing with the whole krb5/gssapi thing. In that case only a thin XDR layer and handling of a unix socket plus a RPC are needed, and you leave all the protocol complexity to a separate process.

I maintain this:
https://github.com/gssapi/gssproxy
See x-files for the interface definition.
There is no Go-client for this, but writing one could be an option.
The only issue is if you need this to work on other operating systems like Windows, because gss-proxy is not build to work with SSPI, and on Windows you should really be using SSPI to deal with krb5/GSSAPI or you'd have an extremely poor experience.

PR #1285, opened 2022-11-08
OTA-818: pkg/cli/admin/upgrade: Support --to-multi-arch per OTA-818

jottofar:
(https://issues.redhat.com/browse/OTA-818) which allows the upgrade of a cluster from a homogeneous to a heterogeneous                        cluster architecture at the current version.


jottofar:
/retitle OTA-818: pkg/cli/admin/upgrade: Support --to-multi-arch per OTA-818

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -70,6 +70,9 @@ func New(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command
 			itself and report status that is available via "oc get clusterversion" and "oc describe
 			clusterversion".
 
+			Passing --to-multi-arch will upgrade the cluster from a single to a multi cluster
```
wking, line 73: nit: "multi cluster architecture" feels a bit weird.  Can we mirror [the docs][1] with "multi-architecture cluster"?  Something like:

> Passing `--to-multi-arch` will request a transition from a single-architecture cluster to a multi-architecture cluster.

[1]: https://docs.openshift.com/container-platform/4.11/post_installation_configuration/multi-architecture-configuration.html

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -202,6 +210,18 @@ func (o *Options) Run() error {
...
+	case o.ToMultiArch:
+		if cv.Spec.DesiredUpdate != nil && cv.Spec.DesiredUpdate.Architecture == configv1.ClusterVersionArchitectureMulti {
+			return fmt.Errorf("info: Update to multi cluster architecture has already been requested")
+		}
+		if err := patchDesiredUpdate(&configv1.Update{Architecture: configv1.ClusterVersionArchitectureMulti,
+			Version: cv.Status.Desired.Version}, o.Client, cv.Name); err != nil {
+
+			return err
+		}
+		fmt.Fprintln(o.Out, "Requesting update to multi cluster architecture")
```
wking, line 222: I realize this is an existing pattern, so feel free to stick with what you have, but I'm noticing that "Requesting ..." feels like it wants to be pre-patch logging, and I'd have expected "Requested ..." for post-patch logging.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -616,3 +631,17 @@ func targetMatch(target *configv1.Release, to string, toImage string) (bool, err
 
 	return false, nil
 }
+
+func patchDesiredUpdate(update *configv1.Update, client configv1client.Interface, clusterVersionName string) error {
```
wking, line 635: nit, to reduce the effort of eventually getting `Context` wired up for this `Patch` call, I'd prefer opening with a `ctx context.Context` argument here, and pushing the `TODO()` initialization as far up the stack as we conveniently can (but no need to restructure call signatures on existing functions).

pkg/cli/admin/upgrade/upgrade.go
jottofar: Agree. Changed everywhere.

soltysh (approved):
/approve

PR #1289, opened 2022-11-14
IR-260: Add CLI flag to set ImportMode when importing a tag

dorzel:
issues.redhat.com/browse/IR-260

### DIFF
--- a/pkg/cli/importimage/importimage.go
+++ b/pkg/cli/importimage/importimage.go
@@ -99,6 +99,7 @@ func NewImportImageOptions(streams genericclioptions.IOStreams) *ImportImageOpti
 		PrintFlags:      genericclioptions.NewPrintFlags("imported"),
 		IOStreams:       streams,
 		ReferencePolicy: tag.SourceReferencePolicy,
+		ImportMode:      string(imagev1.ImportModeLegacy),
 	}
 }
 
--- a/pkg/cli/importimage/importimage.go
+++ b/pkg/cli/importimage/importimage.go
@@ -212,6 +212,15 @@ func (o *ImportImageOptions) Validate() error {
 		return fmt.Errorf("valid reference policy values are source or local")
 	}
 
+	switch o.ImportMode {
+	case string(imagev1.ImportModeLegacy):
+	case string(imagev1.ImportModePreserveOriginal):
+	case "":
+		o.ImportMode = string(imagev1.ImportModeLegacy)
+	default:
+		return fmt.Errorf("valid ImportMode values are %s or %s", imagev1.ImportModeLegacy, imagev1.ImportModePreserveOriginal)
+	}
+
 	return nil
 }
 
--- a/pkg/cli/importimage/importimage.go
+++ b/pkg/cli/importimage/importimage.go
@@ -52,6 +52,9 @@ var (
 		# Update imported data for all tags in an existing image stream
 		oc import-image mystream --all
 
+		# Update imported data for a tag which points to a manifest list to include the full manifest list
+		oc import-image mystream --import-mode=PreserveOriginal
+
 		# Import all tags into a new image stream
 		oc import-image mystream --from=registry.io/repo/image --all --confirm
 
@@ -73,6 +76,7 @@ type ImportImageOptions struct {
 	Scheduled            bool
 	Insecure             bool
 	InsecureFlagProvided bool
+	ImportMode           string
 
 	DryRun bool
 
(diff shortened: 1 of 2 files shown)


flavianmissi (commented):
It would be nice to validate the user input against the known values for the import mode (before we hit the server). Something like what we do [here](https://github.com/openshift/oc/blob/master/pkg/cli/importimage/importimage.go?plain=1#L195-L211) for the `ReferencePolicy`.
WDYT?

dorzel:
> It would be nice to validate the user input against the known values for the import mode (before we hit the server). Something like what we do [here](https://github.com/openshift/oc/blob/master/pkg/cli/importimage/importimage.go?plain=1#L195-L211) for the `ReferencePolicy`. WDYT?

I think that makes sense to do, added a commit

pkg/cli/importimage/importimage_test.go
```diff
@@ -573,6 +574,44 @@ func TestCreateImageImport(t *testing.T) {
...
+				},
+			},
+			expectedImages: []imagev1.ImageImportSpec{{
+				From: corev1.ObjectReference{Kind: "DockerImage", Name: "repo.com/somens/someimage:latest"},
+				To:   &corev1.LocalObjectReference{Name: "latest"},
+				ImportPolicy: imagev1.TagImportPolicy{
+					ImportMode: imagev1.ImportModeLegacy,
+				},
+			}},
+		},
```
ardaguclu, line 614: Would you mind adding a test case also for `valid ImportMode values are %s or %s` error?

dorzel: I've added the test case in a new commit. I realized the test cases weren't running the validation step, so I've also added that step along with related error checking.

pkg/cli/importimage/importimage.go
```diff
@@ -207,6 +212,15 @@ func (o *ImportImageOptions) Validate() error {
 		return fmt.Errorf("valid reference policy values are source or local")
 	}
 
+	switch o.ImportMode {
+	case string(imagev1.ImportModeLegacy):
+	case string(imagev1.ImportModePreserveOriginal):
+	case "":
```
ardaguclu, line 218: Due to this assignment, there is no problem. But would it be better to set `ImportMode` to `Legacy` in this initialization https://github.com/openshift/oc/blob/d27e7c9efc6b2e1f6436d47eb4de2024e5a3d856/pkg/cli/importimage/importimage.go#L97 like we do for `ReferencePolicy`?

ardaguclu (commented):
I left a few non-blocker comments. After that I'll tag the PR. Thanks.

pkg/cli/importimage/importimage_test.go
```diff
@@ -597,11 +651,39 @@ func TestCreateImageImport(t *testing.T) {
 			o.InsecureFlagProvided = true
 		}
 
+		// set default for referencePolicy
+		if len(test.referencePolicy) == 0 {
+			for i := range test.expectedImages {
+				test.expectedImages[i].ReferencePolicy.Type = imagev1.SourceTagReferencePolicy
```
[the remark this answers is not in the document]
dorzel: I think the main reason for doing this here was to emulate the behavior of the user not needing to specify the default for it to be set, as it does when working with oc. These needed to be set because the Validate() step ends up mutating the ImportOptions object. But it would be nicer to not have to change the `expected`  fields ever.

pkg/cli/importimage/importimage.go
dorzel: That would seem to be the better place, but this is also done for `ReferencePolicy` just above. I'm wondering if this is needed for either of these.

dorzel: Ok, it seems this assignment is needed here regardless in the specific case that the user specifies `--reference-policy="" or --import-mode=""`.

pkg/cli/importimage/importimage_test.go
```diff
@@ -597,11 +651,39 @@ func TestCreateImageImport(t *testing.T) {
...
+				test.expectedRepository.ImportPolicy.ImportMode = imagev1.ImportModeLegacy
+			}
+		}
+
 		if err := o.parseImageReference(); err != nil {
 			t.Errorf("unexpected error: %v", err)
 			continue
 		}
 
+		if err := o.Validate(); err != nil {
```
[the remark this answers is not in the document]
dorzel: This makes sense. I've tweaked it a bit as retaining the behavior of `continue` is needed if there is a validation error so that the import object doesn't actually get created in the test. This makes it clearer what actually failed and emulates the actual behavior better. The general "unexpected error" case is already handled below.

PR #1291, opened 2022-11-21
pkg/cli/admin/upgrade: Drop "force" from "No updates available"

wking:
The outgoing wording dates back to the initial subcommand implementation in openshift/origin@65cce8c37d (openshift/origin#21605).  But "force" is easy to conflate with the `--force` option, and we occasionally hear of users who set `--force` under the mistaken impression that it is related to the recommended-ness of the target release image.  There's already a paragraph in the generic `--help` text discussing the risks of `--allow-explicit-upgrade`, and folks who use `--to-image` in the absence of recommended updates will hear about `--allow-explicit-upgrade` when a bare `--to-image` fails to turn up the requested pullspec in `availableUpdates`.

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -405,7 +405,7 @@ func (o *Options) Run() error {
 			if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.RetrievedUpdates); c != nil && c.Status == configv1.ConditionFalse {
 				fmt.Fprintf(o.ErrOut, "warning: Cannot display available updates:\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 			} else {
-				fmt.Fprintf(o.Out, "No updates available. You may force an upgrade to a specific release image, but doing so may not be supported and may result in downtime or data loss.\n")
+				fmt.Fprintf(o.Out, "No updates available. You may still upgrade to a specific release image with --to-image or wait for new updates to be available.\n")
 			}
 		}
 


pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -405,7 +405,7 @@ func (o *Options) Run() error {
 			if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.RetrievedUpdates); c != nil && c.Status == configv1.ConditionFalse {
 				fmt.Fprintf(o.ErrOut, "warning: Cannot display available updates:\n  Reason: %s\n  Message: %s\n\n", c.Reason, strings.ReplaceAll(c.Message, "\n", "\n  "))
 			} else {
-				fmt.Fprintf(o.Out, "No updates available. You may force an upgrade to a specific release image, but doing so may not be supported and may result in downtime or data loss.\n")
+				fmt.Fprintf(o.Out, "No updates available. You may still upgrade to a specific release image with --to-image, but doing so may not be supported and may result in downtime or data loss.\n")
```
LalatenduMohanty, line 408: We do not need to say `but doing so may not be supported and may result in downtime or data loss` anymore for --to-imgae, do we?

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/upgrade.go
wking: the dangerous flag is `--allow-explicit-upgrade`, which has [the `--help` paragraph][1] explaining exposure.  When `availableUpdates` is empty, the options are not updating (and waiting for new recommendations) or using `--allow-explicit-upgrade --to-image ...`.  Not updating seems fairly obvious, but I guess we could call that out in this message.  Using `--allow-explicit-upgrade --to-image ...` has risks, and the user going down that path will hear more about those risks in later commands if they go down the `oc adm upgrade` -> `oc adm upgrade --to-image ...` -> `oc adm upgrade --allow-explicit-upgrade --to-image ...` path.  But I think having some warning flags at this stage helps set the context, in case they are comfortable waiting, it avoid surprising them with warning flags after they sink time in walking more of the path towards `--allow-explicit-upgrade --to-image ...`.

[1]: https://github.com/openshift/oc/blob/3ac1b026860bbd7ccd51a8c7d1d6a14b83505b91/pkg/cli/admin/upgrade/upgrade.go#L73-L80

wking:
[Install error][1] [seems rare][2]:

```
level=error msg=Cluster operator network Degraded is True with ApplyOperatorConfig: Error while updating operator configuration: could not apply (/v1, Kind=ServiceAccount) openshift-multus/metrics-daemon-sa: failed to apply / update (/v1, Kind=ServiceAccount) openshift-multus/metrics-daemon-sa: Patch "https://api-int.ci-op-vkvihi4d-57f76.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/openshift-multus/serviceaccounts/metrics-daemon-sa?fieldManager=cluster-network-operator%2Foperconfig&force=true": read tcp 10.0.156.120:45978->10.0.156.134:6443: read: connection reset by peer
```

/test ci/prow/e2e-aws-ovn-upgrade

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1291/pull-ci-openshift-oc-master-e2e-aws-ovn-upgrade/1594798439705612288#1:build-log.txt%3A111
[2]: https://search.ci.openshift.org/chart?maxAge=24h&type=junit&search=Cluster+operator+network+Degraded+is+True+with+ApplyOperatorConfig

PR #1296, opened 2022-11-22
OCPBUGS-3526: Proceed archive if Lsetxattr gets unsupported error

ardaguclu:
Currently on darwin, archiving is failing by raising unsupported system error.
This PR handles that error and proceeds to archiving.

This is basically the same approach applied in here https://github.com/openshift/oc/blob/952066e54e7b9a487f321d3266161786ff8ab742/pkg/cli/image/archive/archive.go#L444

### DIFF
--- a/pkg/cli/image/archive/archive.go
+++ b/pkg/cli/image/archive/archive.go
@@ -392,7 +392,6 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		}
 	}
 
-	var errors []string
 	for key, value := range hdr.Xattrs {
 		// exclude security.capability unless you're root, Lsetxattr only works
 		// for linux based systems so that's fine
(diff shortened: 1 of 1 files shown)


zhouying7780:
without the fix pr , will hit error :
error: unable to extract layer sha256:7cff71538289f29a35fda6c24627b45b58f63c52b9727b061a0f8285fc372dc4 from quay.io/openshift-release-dev/ocp-release:4.11.18-x86_64: platform and architecture is not supported

with the pr the extract completed successfully.

ardaguclu:
@deejross could you PTAL?. Thanks.

pkg/cli/image/archive/archive.go
```diff
@@ -392,20 +392,23 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
...
 
-	var errors []string
 	for key, value := range hdr.Xattrs {
 		// exclude security.capability unless you're root, Lsetxattr only works
 		// for linux based systems so that's fine
 		if key == "security.capability" && currentUser != nil && currentUser.Username != "root" {
 			continue
 		}
 		if err := system.Lsetxattr(path, key, []byte(value), 0); err != nil {
+			if err == system.ErrNotSupportedPlatform {
```
soltysh, line 402: Maybe we should at least at some kind of warning or `klog.V(1)` here, in case someone wants to see it. Same for the error below.

soltysh (approved):
/lgtm

PR #1300, opened 2022-12-01
OCPBUGS-4280: oc import-image: reflect import image error

ardaguclu:
This PR returns import image error retrieved from server to return correct exit code.


ardaguclu:
/cc @ingvagabund 

could you PTAL?. Thanks.

PR #1301, opened 2022-12-05
handle admission webhook lists and CRD conversion webhooks

deads2k:
This adds support for admission webhook lists (that was missing before), so we collect their namespaces as well.  It also uses the same generic conversion/casting mechanism for routes and secrets.  And it adds CRD (and list) inspection capabilities for conversion webhook namespaces.

/assign @soltysh @ingvagabund


pkg/cli/admin/inspect/admission_webhooks.go
```diff
@@ -0,0 +1,118 @@
...
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/cli-runtime/pkg/resource"
+)
+
+type mutatingWebhookConfigList struct {
+	*admissionregistrationv1.MutatingWebhookConfigurationList
+}
+
+func (c *mutatingWebhookConfigList) addItem(obj interface{}) error {
```
deads2k, line 14: we appear to lack such a function in upstream kube.  I bet it's the typing problem

pkg/cli/admin/inspect/resource.go
```diff
@@ -94,30 +95,114 @@ func InspectResource(info *resource.Info, context *resourceContext, o *InspectOp
 		}
 		return nil
 
-	case schema.GroupResource{Group: "route.openshift.io", Resource: "routes"}:
+	case routev1.GroupVersion.WithResource("routes").GroupResource():
```
deads2k, line 98: matches the core secret one above.

pkg/cli/admin/inspect/resource.go
```diff
@@ -94,30 +95,114 @@ func InspectResource(info *resource.Info, context *resourceContext, o *InspectOp
...
+		}
+		return nil
+
 	default:
-		unstr, ok := info.Object.(*unstructured.Unstructured)
-		if ok {
-			// obtain associated objects for the current resource
-			if err := gatherRelatedObjects(context, unstr, o); err != nil {
-				return err
+		return gatherGenericObject(context, info, o)
```
deads2k, line 123: extracted this to a new method.

pkg/cli/admin/inspect/resource.go
```diff
@@ -94,30 +95,114 @@ func InspectResource(info *resource.Info, context *resourceContext, o *InspectOp
...
+	case *routev1.RouteList:
+		return &routeList{castObj}, nil
+
+	default:
+		return nil, fmt.Errorf("unhandledStructuredListType: %T", castObj)
+
+	}
+}
+
+func toStructuredObject[T any, TList any](obj runtime.Object) (runtime.Object, error) {
```
deads2k, line 151: my first experiment with generics.  Any better ideas for doing this?

pkg/cli/admin/inspect/resource.go
```diff
@@ -94,30 +95,114 @@ func InspectResource(info *resource.Info, context *resourceContext, o *InspectOp
...
+	if unstructureObj, ok := obj.(*unstructured.Unstructured); ok {
+		var t T
```
deads2k, line 153: I couldn't seem to do this with &T{}.  I get "invalid composite literal type T".

pkg/cli/admin/inspect/resource.go
```diff
@@ -126,8 +211,10 @@ func gatherRelatedObjects(context *resourceContext, unstr *unstructured.Unstruct
 		return err
 	}
 
-	relatedObjReferences = append(relatedObjReferences, obtainWebhookService(unstr)...)
```
deads2k, line 129: I could move some of the blocks from above to here.  it seemed a little more straightforward to add it to the main flow.

pkg/cli/admin/inspect/resource.go
```diff
@@ -186,42 +286,3 @@ func obtainRelatedObjects(obj *unstructured.Unstructured) ([]*configv1.ObjectRef
 
 	return relatedObjs, nil
 }
-
-// obtainWebhookService obtains namespaces related to ValidatingWebhookConfiguration or
-// MutatingWebhookConfiguration. These WebhookConfiguration resources do not contain status.relatedObject field,
-// instead service resources are stored in webhook.clientConfig.Service field and this function will
-// return namespace resources in which services related to webhook configurations are running.
-func obtainWebhookService(obj *unstructured.Unstructured) []*configv1.ObjectReference {
```
deads2k, line 194: This doesn't handle lists and we *do* sometimes get lists from the caller.

deads2k:
image problem

/retest

ardaguclu:
I think I should add e2e tests for inspect and must-gather commands to also verify such changes on CI

pkg/cli/admin/inspect/resource.go
soltysh: Theoretically this will be ok-ish, but `T` and `TList` do NOT have to be correlated, iow. I can pass `admissionregistrationv1.MutatingWebhookConfiguration` twice here and the compiler will happily accept that, which is what you're trying to prevent from happening, right?

soltysh (commented):
This looks mostly ok, except for that nit with generics and handling specific type vs list.

deads2k:
> Theoretically this will be ok-ish, but T and TList do NOT have to be correlated, iow. I can pass admissionregistrationv1.MutatingWebhookConfiguration twice here and the compiler will happily accept that, which is what you're trying to prevent from happening, right

I would like to restrict it, but the primitive we have now don't allow it and the type needs to be a concrete type you can instantiate in order to create the instances down below.

> I think I should add e2e tests for inspect and must-gather commands to also verify such changes on CI

our existing must-gather test cases appear to pass.

ardaguclu:
> > Theoretically this will be ok-ish, but T and TList do NOT have to be correlated, iow. I can pass admissionregistrationv1.MutatingWebhookConfiguration twice here and the compiler will happily accept that, which is what you're trying to prevent from happening, right
> 
> I would like to restrict it, but the primitive we have now don't allow it and the type needs to be a concrete type you can instantiate in order to create the instances down below.

I agree that we can not achieve this with current type structure(it might be possible if the upstream list types support `listaccessor` interface, etc.)

/lgtm
/retest-required

PR #1305, opened 2022-12-07
OCPBUGS-4517: oc adm release: Support extracting commands in specified architectures

ardaguclu:
Currently, user can extract tools in specific operating systems by setting `--command-os`. However, there is no way for user extracting specific architectures in these operating systems.

This PR adds an ability to `--command-os` flag to also specify architecture like `linux/arm64`, `mac/arm64`. Users are already used to use this syntax from `filter-by-os` flag and this can also be used for `--command-os` flag.


pkg/cli/admin/release/extract_tools.go
```diff
@@ -333,7 +333,19 @@ func (o *ExtractOptions) extractCommand(command string) error {
 		if currentOS == "*" {
 			currentArch = "*"
 		} else {
-			currentArch = "amd64"
+			parsed := strings.Split(currentOS, "/")
+			switch len(parsed) {
+			case 1:
+				// --command-os=linux
+				currentArch = "amd64"
```
cfergeau, line 340: I think this is missing `currentOS = parsed[0]` ?

cfergeau:
> @cfergeau you have a similar attempt for this issue #1240. But it seems that adding new flag for this will increase the dimension and maybe it is better to boil os and arch down to the `--command-os` flag.

No strong feeling either way :) `podman` has `--platform linux/amd64` plus separate `--os` and `--arch`. No idea how other tools deal with this.

pkg/cli/admin/release/extract_tools.go
ardaguclu: It is set 3 line above

PR #1306, opened 2022-12-09
FDN-112: Parse any jira reference from commit text, not just OCPBUGS

bparees:
for discussion on workloads call.

This allows the changelog to link to all the jira tickets(i.e. also stories/epics) referenced in commits, not just OCPBUGS tickets.

### DIFF
--- a/pkg/cli/admin/release/bug.go
+++ b/pkg/cli/admin/release/bug.go
@@ -9,61 +9,61 @@ import (
 	"net/url"
 	"regexp"
 	"sort"
-	"strconv"
 	"strings"
-
-	"k8s.io/klog/v2"
 )
 
-type BugType int
+type RefType int
 
 const (
-	Bugzilla BugType = iota
+	Bugzilla RefType = iota
 	Jira
 )
 
 const (
 	maxRetries = 2
 
-	bugzillaBrowseURL = "https://bugzilla.redhat.com/show_bug.cgi?id=%d"
+	bugzillaBrowseURL = "https://bugzilla.redhat.com/show_bug.cgi?id=%s"
 	bugzillaRestURL   = "https://bugzilla.redhat.com/rest/bug"
-	jiraBrowseURL     = "https://issues.redhat.com/browse/OCPBUGS-%d"
-	JiraRestURL       = "https://issues.redhat.com/rest/api/latest/issue/OCPBUGS-%d"
+	jiraBrowseURL     = "https://issues.redhat.com/browse/%s"
+	JiraRestURL       = "https://issues.redhat.com/rest/api/latest/issue/%s"
 )
 
 var (
+	// find strings like "Bug: 123,456 789: some commit description"
 	reBugzillaBug = regexp.MustCompile(`^[B|b]ug(?:s)?\s([\d\s,]+)(?:-|:)\s*`)
-	reJiraBug     = regexp.MustCompile(`^OCPBUGS-([\d\s,]+)(?:-|:)\s*`)
+
+	// find strings like "XYZ-123,ABC-456 DEF-789: some commit description"
+	reJiraRef = regexp.MustCompile(`^([A-Z]+-\d+[\s,]*)+(?:-|:)\s*`)
 )
 
-// BugList stores Bug list
-type BugList struct {
-	Bugs []Bug
+// RefList stores Bug list
+type RefList struct {
+	Refs []Ref
 }
 
-// Bug stores referenced bug ids and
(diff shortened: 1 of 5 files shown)


pkg/cli/admin/release/bug.go
```diff
@@ -255,18 +255,10 @@ func GetBugList(b map[string]Bug) []Bug {
 	return bs
 }
 
-func generateBugKey(bt BugType, id int) string {
```
bparees, line 258: not needed anymore because the jira ref IDs include the jira prefix now, so there's no risk of collision between a bugzilla bug "1234" and a jira bug "ocpbugs-1234"

pkg/cli/admin/release/bug.go
```diff
@@ -9,61 +9,61 @@ import (
...
+	JiraRestURL       = "https://issues.redhat.com/rest/api/latest/issue/%s"
 )
 
 var (
+	// find strings like "Bug: 123,456 789: some commit description"
 	reBugzillaBug = regexp.MustCompile(`^[B|b]ug(?:s)?\s([\d\s,]+)(?:-|:)\s*`)
-	reJiraBug     = regexp.MustCompile(`^OCPBUGS-([\d\s,]+)(?:-|:)\s*`)
+
+	// find strings like "XYZ-123,ABC-456 DEF-789: some commit description"
+	reJiraRef = regexp.MustCompile(`^([A-Z]+-\d+[\s,]*)+(?:-|:)\s*`)
```
nstielau, line 36: I'd like to support the existing `NO-ISSUE` semantic that some [teams are using](https://amd64.ocp.releases.ci.openshift.org/releasestream/4-dev-preview/release/4.13.0-ec.1).  If we do enforce, I think we might want a standard escape hatch.

nstielau (changes requested):
Would like the NO-ISSUE or other support escape hatch in the regex.

nstielau:
Thanks!  A few small behavior changes and a lot of s/bug/ref/ changes ;)

pkg/cli/admin/release/bug.go
bparees: this code doesn't enforce anything, this is just for adding links to the changelog so they can show up in the release controller payload diff report (and so we can potentially use those links to scrape more information about what went into the payload, based on the jira ticket that is linked).  

Since "no-issue" items by definition have nothing to link to, they should be ignored by this code (which they will be, since no-issue/no-jira won't match the regex)

the enforcement logic would be done in the prow bot + tide merge label requirements, which is unrelated to this codebase (other than in that both of them consume the same text: the pr title, effectively)

but yes, when we get to the prow bot logic, we'll want it to respect "no-issue/no-jira" as a valid "jira reference" for purposes of applying the "valid-jira" label to a PR.

bparees:
> Thanks! A few small behavior changes and a lot of s/bug/ref/ changes ;)

yes, pretty much. and i probably should have done it in separate commits to make reviewing it easier.

pkg/cli/admin/release/info.go
```diff
@@ -1619,27 +1619,29 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
 				continue
 			}
-			for _, v := range commit.Bugs.Bugs {
-				if _, ok := bugIDs[generateBugKey(v.Source, v.ID)]; !ok {
-					// We are using generated bug key according to the source type
-					// to prevent possible clashes like BUG 111 and OCPBUGS-111
-					bugIDs[generateBugKey(v.Source, v.ID)] = v
+			for _, v := range commit.Refs.Refs {
+				// ignore jira refs to non bugs
+				if v.Source == Jira && !(strings.HasPrefix(v.ID, "OCPBUGS-") || strings.HasPrefix(v.ID, "OCPBUGSM-")) {
```
ardaguclu, line 1627: Why we are ignoring non bug jira refs?

ardaguclu (commented):
Unit tests are covering this parsing mechanism and they are all passing.

I have only one questions. Other than that changes look good to me.

pkg/cli/admin/release/info.go
bparees: this code is in the function `describeBugs` which is only called when using the `oc adm release info --bugs` argument, which per the description:

```
  --bugs='':
	Generate bug listings from the changelogs in the git repositories
	extracted to this path.
```

so i made it so this function filters out anything that isn't a bug.  i'll update the comment to explain this better.

bparees: updated

PR #1310, opened 2022-12-13
IR-266: Add image manifests to 'describe image'

werled:
If the image to describe contains image manifests, this adds both os/arch and digest of each manifest to the output.

The output looks like this (this is the output of the stripped down test case):
```
Image Name:     test
Docker Image:   <none>
Name:           test
Layer Size:     0B
Image Created:  292 years ago
Author:         <none>
Arch:           <none>
Manifests:      linux/amd64     sha256:1234
                linux/s390x     sha256:5678
Command:        <none>
Working Dir:    <none>
User:           <none>
Exposes Ports:  <none>
Docker Labels:  <none>
```


dmage:
assigning to the root approver as mentioned by the ci bot
/assign @ardaguclu

PR #1311, opened 2022-12-13
OCPBUGS-1115: Use linux/arch when user's OS isn't in manifests 

bryan-cox:
Fixes [OCPBUGS-1115](https://issues.redhat.com/browse/OCPBUGS-1115).

When extracting 'oc' from a manifest listed image, if the --no-filter-by-os flag was not used and an image could not be found for the runtime OS/runtime arch combination, warn the user a manifest image could not be found for this combination and try searching for a linux/runtime arch combination again in the manifest listed image.

### DIFF
--- a/pkg/cli/image/extract/extract.go
+++ b/pkg/cli/image/extract/extract.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -375,17 +376,29 @@ func (o *ExtractOptions) Run() error {
 					return fmt.Errorf("unable to connect to image repository %s: %v", from.String(), err)
 				}
 
-				srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
+				srcManifest, location, err := retrieveSourceManifest(ctx, from, repo, o)
 				if err != nil {
-					if imagemanifest.IsImageForbidden(err) {
-						msg := fmt.Sprintf("image %q does not exist or you don't have permission to access the repository", from)
-						return imagemanifest.NewImageForbidden(msg, err)
+					if err != imagemanifest.AllImageFilteredErr {
+						return err
 					}
-					if imagemanifest.IsImageNotFound(err) {
-						msg := fmt.Sprintf("image %q not found: %s", from, err.Error())
-						return imagemanifest.NewImageNotFound(msg, err)
+
+					// Translate the current runtime OS to linux when looking through a manifest listed image since a manifest for the particular 'runtime OS'/'archType' may not exist.
+					// In cases where the runtime OS is windows or darwin (i.e. macOS), there is no expectation of a manifest image with those particular OS's.
+					if o.FilterOptions.DefaultOSFilter {
+						runtimeOS := runtime.GOOS
+						arch := runtime.GOARCH
+
+						klog.V(2).Infof("Warning: a manifest image could not be found for this platform: %s/%s. Converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image.\n", runtimeOS, arch, runtimeOS, arch)
+						o.FilterOptions.OSFilter, err = regexp.Compile("linux/" + arch)
+						if err != nil {
+							return fmt.Errorf("failed to compile OSFilter for linux/%s: %v", arch, err)
+						}
+
+						srcManifest, location, err = retrieveSourceManifest(ctx, from, repo, o)
+						if err != nil {
+							return err
+						}
 					}
-					return fmt.Errorf("unable to read image %s: %v", from, err)
 				}
 
 				contentDigest, err := registryclient.ContentDigestForManifest(srcManifest, location.Manifest.Algorithm())
@@ -527,6 +540,27 @@ func (o *ExtractOptions) Run() error {
 			})
 		}
 	})
+}
+
+// retrieveSourceManifest retrieves the first manifest at the request location that matches the filter function and handles any errors resulting from retrieving the manifest
+func retrieveSourceManifest(ctx context.Context, from imagesource.TypedImageReference, repo distribution.Repository, o *ExtractOptions) (distribution.Manifest, imagemanifest.ManifestLocation, error) {
+	srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
+	if err != nil {
+		emptyManifestLocation := imagemanifest.ManifestLocation{}
+		if imagemanifest.IsImageForbidden(err) {
+			msg := fmt.Sprintf("image %q does not exist or you don't have permission to access the repository", from)
+			return nil, emptyManifestLocation, imagemanifest.NewImageForbidden(msg, err)
+		}
+		if imagemanifest.IsImageNotFound(err) {
+			msg := fmt.Sprintf("image %q not found: %s", from, err.Error())
+			return nil, emptyManifestLocation, imagemanifest.NewImageNotFound(msg, err)
+		}
+		if err == imagemanifest.AllImageFilteredErr {
+			return nil, emptyManifestLocation, err
+		}
+		return nil, emptyManifestLocation, fmt.Errorf("unable to read image %s: %v", from, err)
+	}
+	return srcManifest, location, nil
 }
 
 func layerByEntry(r io.Reader, options *archive.TarOptions, layerInfo LayerInfo, fn TarEntryFunc, allLayers bool, alreadySeen map[string]struct{}) (bool, error) {
--- a/pkg/cli/image/manifest/errors.go
+++ b/pkg/cli/image/manifest/errors.go
@@ -1,9 +1,13 @@
 package manifest
 
 import (
+	"fmt"
+
 	"github.com/docker/distribution/registry/api/errcode"
 	registryapiv2 "github.com/docker/distribution/registry/api/v2"
 )
+
+var AllImageFilteredErr = fmt.Errorf("filtered all images from manifest list")
 
 type imageNotFound struct {
 	msg string
--- a/pkg/cli/image/manifest/manifest.go
+++ b/pkg/cli/image/manifest/manifest.go
@@ -296,7 +296,7 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
 		return nil, ManifestLocation{}, err
 	}
 	if len(srcManifests) == 0 {
-		return nil, ManifestLocation{}, fmt.Errorf("filtered all images from manifest list")
+		return nil, ManifestLocation{}, AllImageFilteredErr
 	}
 
 	if srcDigest != originalSrcDigest {


ardaguclu:
I'm not sure I understand the purpose of this PR. Wouldn't it be better to use `--command-os=linux/amd64` to get binary whatever platform we want?.

bryan-cox:
> 

I think that only works with single arch release images but doesn't work with multi-arch release images (at least it didn't when I just attempted to use it). The issue in the bug is when you try to run `oc adm release extract --command=oc quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5` on a Mac it will fail because it's looking for an image such as `darwin/arm64` in the manifest list. The solution we settled on was to translate `darwin` over to `linux` since there isn't  going to be a `darwin/(procArch)` in a manifest.

bryan-cox:
@ardaguclu - just following up, did my explanation help any?

ardaguclu:
But you should be able to get in whichever os you want via `--command-os` flag, even if it is multi arch. Did you try;

```sh
oc adm release extract --command=oc --command-os=darwin/amd64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5
```
Please correct me if I'm missing something but this PR prevents when user wants to extract image `--command-os=darwin/arm64`?

bryan-cox:
> But you should be able to get in whichever os you want via `--command-os` flag, even if it is multi arch. Did you try;
> 
> ```shell
> oc adm release extract --command=oc --command-os=darwin/amd64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5
> ```
> 
> Please correct me if I'm missing something but this PR prevents when user wants to extract image `--command-os=darwin/arm64`?

I tried `oc adm release extract --command=oc --command-os=darwin/amd64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5` and `oc adm release extract --command=oc --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5` and both fail. The issue is `darwin/arm64` combination is being searched for within the manifest list; as far as I understand, that os/proc combination is never going to be listed in an ocp multi-arch release image, so we opted to change `darwin` to `linux` in the code internally.

ardaguclu:
--command-os=darwin/arm64 feature was implemented in 4.13

Could you please check your `oc` version and tried below command in 4.13 or newer?
`oc adm release extract --command=oc --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5` 

If it fails in newer versions, would you mind pasting the error message?. Thanks.

bryan-cox:
> --command-os=darwin/arm64 feature was implemented in 4.13
> 
> Could you please check your `oc` version and tried below command in 4.13 or newer? `oc adm release extract --command=oc --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5`
> 
> If it fails in newer versions, would you mind pasting the error message?. Thanks.

It's still failing, most likely due to the the issue I noted in https://github.com/openshift/oc/pull/1311#issuecomment-1400231815. Here's the error message:
```
brcox@brcox-mac aws-oc-nightly % oc adm release extract --tools registry.ci.openshift.org/ocp-arm64/release-arm64:4.13.0-0.nightly-arm64-2023-02-27-192524 -a ~/all-the-pull-secrets.json
brcox@brcox-mac aws-oc-nightly % tar -xvf openshift-client-mac-arm64-4.13.0-0.nightly-arm64-2023-02-27-192524.tar.gz
brcox@brcox-mac aws-oc-nightly % ./oc adm release extract --command=oc --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi -v5
I0228 08:18:40.154661   97818 config.go:128] looking for config.json at /Users/brcox/.docker/config.json
I0228 08:18:40.154714   97818 config.go:136] found valid config.json at /Users/brcox/.docker/config.json
Warning: the default reading order of registry auth file will be changed from "${HOME}/.docker/config.json" to podman registry config locations in the future version of oc. "${HOME}/.docker/config.json" is deprecated, but can still be used for storing credentials as a fallback. See https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md for the order of podman registry config locations.
I0228 08:18:40.536746   97818 client_mirrored.go:174] Attempting to connect to quay.io/openshift-release-dev/ocp-release
I0228 08:18:40.606457   97818 client_mirrored.go:412] get manifest for sha256:53679d92dc0aea8ff6ea4b6f0351fa09ecc14ee9eda1b560deeb0923ca2290a1 served from registryclient.retryManifest{ManifestService:registryclient.manifestServiceVerifier{ManifestService:(*client.manifests)(0x1400059d560)}, repo:(*registryclient.retryRepository)(0x140007e0680)}: <nil>
I0228 08:18:40.606565   97818 manifest.go:407] Skipping image sha256:fcf4d95df9a189527453d8961a22a3906514f5ecbb05afbcd0b2cdd212aab1a2 for manifestlist.PlatformSpec{Architecture:"amd64", OS:"linux", OSVersion:"", OSFeatures:[]string(nil), Variant:"", Features:[]string(nil)} from quay.io/openshift-release-dev/ocp-release:4.11.4-multi
I0228 08:18:40.606593   97818 manifest.go:407] Skipping image sha256:1992a4713410b7363ae18b0557a7587eb9e0d734c5f0f21fb1879196f40233a3 for manifestlist.PlatformSpec{Architecture:"ppc64le", OS:"linux", OSVersion:"", OSFeatures:[]string(nil), Variant:"", Features:[]string(nil)} from quay.io/openshift-release-dev/ocp-release:4.11.4-multi
I0228 08:18:40.606607   97818 manifest.go:407] Skipping image sha256:3698082cd66e90d2b79b62d659b4e7399bfe0b86c05840a4c31d3197cdac4bfa for manifestlist.PlatformSpec{Architecture:"s390x", OS:"linux", OSVersion:"", OSFeatures:[]string(nil), Variant:"", Features:[]string(nil)} from quay.io/openshift-release-dev/ocp-release:4.11.4-multi
I0228 08:18:40.606620   97818 manifest.go:407] Skipping image sha256:15fc18c81f053cad15786e7a52dc8bff29e647ea642b3e1fabf2621953f727eb for manifestlist.PlatformSpec{Architecture:"arm64", OS:"linux", OSVersion:"", OSFeatures:[]string(nil), Variant:"", Features:[]string(nil)} from quay.io/openshift-release-dev/ocp-release:4.11.4-multi
I0228 08:18:40.606642   97818 workqueue.go:143] about to send work queue error: unable to read image quay.io/openshift-release-dev/ocp-release:4.11.4-multi: filtered all images from manifest list
I0228 08:18:40.606694   97818 workqueue.go:54] worker 0 stopping
error: unable to read image quay.io/openshift-release-dev/ocp-release:4.11.4-multi: filtered all images from manifest list
I0228 08:18:40.606738   97818 workqueue.go:60] work queue exiting
```

aleskandro:
Hello @ardaguclu, is there anything we need to do for this PR to land? 

The bug and PR have been open for a while, and, as stated by @bryan-cox , the `--command-os=darwin/amd64` acts in another part of the code.

Per my understanding, `--command-os` decides which binary to extract from a payload  (single-arch manifest) image.

When you use a multi-arch manifest-list image, before handling `--command-os`, `oc` will filter the manifest list using the GOARCH and GOOS values. When you do this on Darwin, this means that `oc` will try to extract the digest of the darwin/$arch manifest from the ManifestList payload image.

We don't produce Darwin payload images; therefore, it fails with the errors in the bug's and PR's descriptions.

Also, this bug reproduces on windows, too... what's the risk of setting the os to use for the ManifestList lookup always to be Linux. Are there cases where the changed function is used with Manifest Lists different than the payload (for example, windows containers)?

ardaguclu:
@aleskandro to be honest, this defaulting sounds to be hacky. Because when user explicitly passes `--filter-by-os=darwin/amd64`,  we are changing it to `--filter-by-os=linux/amd64`. What will happen in the future, when we have genuine darwing images in multi arch payload?.

The point I'm still trying to understand, in darwin machines(also in windows), why user does not pass `--filter-by-os=linux/amd64` and extracts linux/amd64 from multi arch payload?. Why we are trying to return linux image, although user deliberately requests `--filter-by-os=darwin` image. If we don't support `darwin` image, I'd at least prefer returning an error message notifying "darwin image is not supported currently, etc.".

If the problem is extracting an image when user does not pass `--filter-by-os`(using runtime OS);
* I doubt that this PR fixes this issue, since it modifies --filter-by-os
* Even if it fixes, shouldn't we notify user about the problem to pass `--filter-by-os`, instead directly defaulting to something

aleskandro:
@ardaguclu the problem is not about extracting the image, but about extracting the cli from the payload images. I'm not sure there will ever be a plan to support OCP payloads to run OCP clusters on Darwin hosts.

The problem is that adding --filter-by-os=linux/amd64, only for multi-arch payload, is IMHO, and not intuitive for a Darwin user.

What if the proposal becomes to fallback to linux/$GOARCH when unable to find the non-existing manifest?

ardaguclu:
> What if the proposal becomes to fallback to linux/$GOARCH when unable to find the non-existing manifest?

@aleskandro that looks more reasonable approach. But first I'd like to dig into the code. Just for clarification, could you please copy the exact non-working command?. 

Is this we are trying to fix?

```sh
oc adm release extract --command=oc --filter-by-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
```

aleskandro:
That won't work, as no existing Docker images are specifically built for darwin today. And it's expected imo.

What we want to do is to extract CLIs as we do from single-arch payloads, without forcing a specific os/arch via the `--filter-by-os`  argument. 

Extracting from a multi-arch payload should work seamlessly as extracting from a single-arch payload:

```
oc adm release extract --command=oc quay.io/openshift-release-dev/ocp-release:4.11.4-multi
```

ardaguclu:
Thanks. If I understand correctly, we are trying to behave as if we are passing `--filter-by-os=linux/amd64(or arm64)` on Darwin machines. What will be the expected outputs of below commands on Darwin(e.g. darwin/amd64)?;

```sh
$ oc adm release extract --command=oc --command-os=darwin/amd64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
```

```sh
$ oc adm release extract --command=oc --command-os=linux/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
```

```sh
$ oc adm release extract --command=oc --filter-by-os=linux/arm64 --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
```

ardaguclu:
* If user passes `--filter-by-os` flag explicitly, we will use it and if it is not found we return error without any defaulting because user deliberately wants that image but we don't have it(for example `--filter-by-os=darwin/amd64`)
* If user does not pass anything and we detect that image is not available for $OS/$ARCH(darwin/amd64), we can default to linux/$ARCH after warn user about what is happening?

@aleskandro what do you think?

aleskandro:
> Thanks. If I understand correctly, we are trying to behave as if we are passing `--filter-by-os=linux/amd64(or arm64)` on Darwin machines. What will be the expected outputs of below commands on Darwin(e.g. darwin/amd64)?;
> 
> ```shell
> $ oc adm release extract --command=oc --command-os=darwin/amd64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> ```
> 

the same as in the bug


> ```shell
> $ oc adm release extract --command=oc --command-os=linux/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> ```
> 
the same as in the bug


> ```shell
> $ oc adm release extract --command=oc --filter-by-os=linux/arm64 --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> ```

this should work.

ardaguclu:
> Thanks. If I understand correctly, we are trying to behave as if we are passing `--filter-by-os=linux/amd64(or arm64)` on Darwin machines. What will be the expected outputs of below commands on Darwin(e.g. darwin/amd64)?;
> 
> ```shell
> $ oc adm release extract --command=oc --command-os=darwin/amd64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> ```

I'd expect extracted image should be linux/amd64 and `oc` should be extracted darwin/amd64

> 
> ```shell
> $ oc adm release extract --command=oc --command-os=linux/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> ```

I'd expect extracted image should be linux/amd64 and `oc` should be extracted linux/arm64

> 
> ```shell
> $ oc adm release extract --command=oc --filter-by-os=linux/arm64 --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> ```

I'd expect extracted image should be linux/arm64 and `oc` should be extracted darwin/arm64

and for the example you gave;

```sh
oc adm release extract --command=oc quay.io/openshift-release-dev/ocp-release:4.11.4-multi
```

extracted image is linux/amd64 and oc should be extracted in darwin/amd64.

Are these ^ assumptions correct? @aleskandro @bryan-cox

aleskandro:
> > Thanks. If I understand correctly, we are trying to behave as if we are passing `--filter-by-os=linux/amd64(or arm64)` on Darwin machines. What will be the expected outputs of below commands on Darwin(e.g. darwin/amd64)?;
> > ```shell
> > $ oc adm release extract --command=oc --command-os=darwin/amd64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> > ```
> 
> I'd expect extracted image should be linux/amd64 and `oc` should be extracted darwin/amd64
> 

Yes. 

> > ```shell
> > $ oc adm release extract --command=oc --command-os=linux/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> > ```
> 
> I'd expect extracted image should be linux/amd64 and `oc` should be extracted linux/arm64
> 

Image should be `linux/$(uname -m)` (the arch of the client).

> > ```shell
> > $ oc adm release extract --command=oc --filter-by-os=linux/arm64 --command-os=darwin/arm64 quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> > ```
> 
> I'd expect extracted image should be linux/arm64 and `oc` should be extracted darwin/arm64
> 

yes.

> and for the example you gave;
> 
> ```shell
> oc adm release extract --command=oc quay.io/openshift-release-dev/ocp-release:4.11.4-multi
> ```
> 
> extracted image is linux/amd64 and oc should be extracted in darwin/amd64.
> 
> Are these ^ assumptions correct? @aleskandro @bryan-cox

extracted image is linux/$(uname -m), `oc` extracted is `$CLIENT_OS/$(uname -m)`

aleskandro:
the PR and the resutls lgtm. Waiting for the review from @ardaguclu and any other testing actions from @zhouying7780

aleskandro (approved):

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +386,33 @@ func (o *ExtractOptions) Run() error {
 					return fmt.Errorf("unable to connect to image repository %s: %v", from.String(), err)
 				}
 
-				srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
+				srcManifest, location, err := retrieveSourceManifest(ctx, from, repo, o)
```
ardaguclu, line 389: What is the reason of moving error handling into the `retrieveSourceManifest` function?. As far as I can see, we are still doing error handling below. Now, on the contrary, we are doing same error handling twice(in retrieveSourceManifest and also in below).

I think, it would be better to stay this as is.

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +386,33 @@ func (o *ExtractOptions) Run() error {
...
 				if err != nil {
-					if imagemanifest.IsImageForbidden(err) {
-						msg := fmt.Sprintf("image %q does not exist or you don't have permission to access the repository", from)
-						return imagemanifest.NewImageForbidden(msg, err)
-					}
-					if imagemanifest.IsImageNotFound(err) {
-						msg := fmt.Sprintf("image %q not found: %s", from, err.Error())
-						return imagemanifest.NewImageNotFound(msg, err)
+					if strings.Contains(err.Error(), "a manifest image could not be found for this platform") {
```
ardaguclu, line 391: Is `a manifest image could not be found for this platform` an error?. If not, we need to find a matching error to compare with `err`. If yes, `if err == ManifestNotFound` would be better instead calling `Contains` function.

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +386,33 @@ func (o *ExtractOptions) Run() error {
...
+						// Translate the current runtime OS to linux when looking through a manifest since there will not be a darwin/'archType' in a manifest image.
+						// In cases where the runtime OS is windows or darwin (i.e. macOS), there is no expectation of a manifest image with those particular OS's.
+						if o.FilterOptions.DefaultOSFilter {
+							runtimeOS := runtime.GOOS
+							arch := runtime.GOARCH
+							if linuxConversionWarning == "" {
```
ardaguclu, line 397: There is no need to this. Warn user and continue?

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +386,33 @@ func (o *ExtractOptions) Run() error {
...
+							if err != nil {
+								return fmt.Errorf("failed to compile OSFilter for linux/%s: %v", arch, err)
+							}
+
+							srcManifest, location, err = retrieveSourceManifest(ctx, from, repo, o)
+						}
+						if err != nil {
+							return err
+						}
+					} else {
```
ardaguclu, line 413: Return early, instead if - else

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +386,33 @@ func (o *ExtractOptions) Run() error {
...
+									fmt.Sprintf("Warning: converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image\n", runtimeOS, arch)
+							}
+							o.FilterOptions.OSFilter, err = regexp.Compile("linux/" + arch)
+							o.FilterOptions.FilterByOS = "linux/" + arch
+
+							if err != nil {
+								return fmt.Errorf("failed to compile OSFilter for linux/%s: %v", arch, err)
+							}
+
+							srcManifest, location, err = retrieveSourceManifest(ctx, from, repo, o)
```
ardaguclu, line 408: I see that the main reason of `retrieveSourceManifest` function is being called also in here. I still think, we might need this function

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +386,33 @@ func (o *ExtractOptions) Run() error {
...
+						// Translate the current runtime OS to linux when looking through a manifest since there will not be a darwin/'archType' in a manifest image.
+						// In cases where the runtime OS is windows or darwin (i.e. macOS), there is no expectation of a manifest image with those particular OS's.
+						if o.FilterOptions.DefaultOSFilter {
+							runtimeOS := runtime.GOOS
+							arch := runtime.GOARCH
+							if linuxConversionWarning == "" {
+								linuxConversionWarning = fmt.Sprintf("Warning: a manifest image could not be found for this platform: %s/%s\n", runtimeOS, arch) +
+									fmt.Sprintf("Warning: converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image\n", runtimeOS, arch)
+							}
+							o.FilterOptions.OSFilter, err = regexp.Compile("linux/" + arch)
```
ardaguclu, line 401: Why we are also setting `OSFilter` and also `FilterByOS`?

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +386,33 @@ func (o *ExtractOptions) Run() error {
...
+							o.FilterOptions.FilterByOS = "linux/" + arch
+
+							if err != nil {
```
ardaguclu, line 404: This err belongs to `OSFilter` assignment but it is not just after of it, is there any specific reason?

pkg/cli/image/manifest/errors.go
```diff
@@ -67,3 +67,10 @@ func IsImageNotFound(err error) bool {
 		return false
 	}
 }
+
+func IsFilteredAllImagesFromManifestList(err error) bool {
```
ardaguclu, line 71: This hurts readability. We don't need to wrap just 3 lines of code into a function.

ardaguclu (commented):
Thanks for the PR and the effort. But I'd expect simpler changes for this fix. If we need to refactor the code, these might be done in followup PRs.

pkg/cli/image/extract/extract.go
```diff
@@ -358,7 +361,15 @@ func (o *ExtractOptions) Run() error {
 	}
 
 	stopCh := make(chan struct{})
-	defer close(stopCh)
+	linuxConversionWarning := ""
+	defer func() {
+		close(stopCh)
+		setLinuxConversionWarningOnce.Do(func() {
```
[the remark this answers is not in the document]
bryan-cox: Yeah, I was just trying to prevent the message from appearing twice. If you are okay, with it appearing twice, I can remove this. As a user, I would not want the duplicated error messages.

I had the boolean flag prior to this, but it seems to add more complexity than the current solution. It would require a mutex since the flag could be accessed by both channels, right? This implementation seems simpler. I didn't add any additional defers; I just added the `Do.Once` function to the current defer.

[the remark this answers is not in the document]
ardaguclu: I agree, it makes sense to prevent duplicate messages from user point of view. I'd use boolean without managing it with mutexes.

pkg/cli/image/extract/extract.go
bryan-cox: It was moved into a function because we need to call the set of code twice, and I wanted to avoid duplicating the code. I see where it would be handled twice in `retrieveSourceManifest`; I'm not sure what you mean by 'also in below'. Would you be able to elaborate on that more please?

This was part of accomplishing when we can't find the current $OS/$ARCH, default to linux/$ARCH and warn the user about what is happening as suggested in https://github.com/openshift/oc/pull/1311#issuecomment-1471799668.

pkg/cli/image/extract/extract.go
bryan-cox: This is the implementation solution for
```
If user passes --filter-by-os flag explicitly, we will use it and if it is not found we return error without any defaulting because user deliberately wants that image but we don't have it(for example --filter-by-os=darwin/amd64)
If user does not pass anything
```
in https://github.com/openshift/oc/pull/1311#issuecomment-1471799668. This checks if `--filter-by-os` was passed in or not.

pkg/cli/image/extract/extract.go
bryan-cox: Yup, I should have done that. I'll update it.

pkg/cli/image/extract/extract.go
bryan-cox: I'm not sure exactly what you mean. It seems like an error to me, i.e., I was looking for this os/arch related manifest and it wasn't found. It's not an official error message in the Docker API as far as I know. 

`FirstManifest` returns an error message when there was no matching os/arch manifest [here](https://github.com/openshift/oc/blob/39dfced39bc2e2ce646c85837a5acf1ce081cb50/pkg/cli/image/manifest/manifest.go#L299).

That error message is processed in `IsFilteredAllImagesFromManifestList` [here](https://github.com/bryan-cox/oc/blob/0ea071e40fc4699cf87648a2e8aafc531994838a/pkg/cli/image/extract/extract.go#L571).

pkg/cli/image/extract/extract.go
bryan-cox: This is OBE by https://github.com/openshift/oc/pull/1311#discussion_r1142100859

pkg/cli/image/manifest/errors.go
bryan-cox: I will update this. I made the change due to some other feedback I had received.

bryan-cox:
@ardaguclu Thanks for the feedback! I believe I have everything address except https://github.com/openshift/oc/pull/1311#discussion_r1141742985. PTAL.

cc: @aleskandro

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +376,32 @@ func (o *ExtractOptions) Run() error {
 					return fmt.Errorf("unable to connect to image repository %s: %v", from.String(), err)
 				}
 
-				srcManifest, location, err := imagemanifest.FirstManifest(ctx, from.Ref, repo, o.FilterOptions.Include)
+				srcManifest, location, err := retrieveSourceManifest(ctx, from, repo, o)
 				if err != nil {
-					if imagemanifest.IsImageForbidden(err) {
-						msg := fmt.Sprintf("image %q does not exist or you don't have permission to access the repository", from)
-						return imagemanifest.NewImageForbidden(msg, err)
+					if !strings.Contains(err.Error(), "a manifest image could not be found for this platform") {
```
ardaguclu, line 381: Should this be;

```go
if err != AllIImagesFiltered... 
```

pkg/cli/image/extract/extract.go
```diff
@@ -4,17 +4,17 @@ import (
 	"archive/tar"
 	"context"
 	"fmt"
+	"github.com/spf13/cobra"
```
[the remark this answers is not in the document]
bryan-cox: Oh, my IDE did that I think. Fixed.

pkg/cli/image/extract/extract.go
```diff
@@ -103,6 +103,7 @@ var (
 		# Extract the last three layers of the image
 		oc image extract docker.io/library/centos:7[-3:]
 	`)
+	warnedAboutLinuxConversion bool
```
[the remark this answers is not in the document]
ardaguclu: It is the same principle with `icspWarned`. I doubt that warning message managed by `icspWarned` prints twice?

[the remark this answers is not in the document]
bryan-cox: Yeah it's still repeating twice.

FWIW - if I change `if !icspWarned && len(o.ICSPFile) > 0 && len(from.Ref.Tag) > 0 {` to `if !icspWarned {` that warning is repeated twice as well.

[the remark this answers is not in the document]
ardaguclu: Good catch, thanks. Does it make sense to move it in L:361?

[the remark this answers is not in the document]
bryan-cox: It repeats there too. It seems like `Run()` is executed twice.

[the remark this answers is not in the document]
bryan-cox: Hey @ardaguclu, just checking to see if you had any thoughts or suggestions on how to move forward on this comment.

I'd like to leave `warnedAboutLinuxConversion` where it is now.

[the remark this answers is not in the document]
ardaguclu: Could you please copy the command you tried?. I too would like to test it.

pkg/cli/image/manifest/errors.go
```diff
@@ -1,10 +1,13 @@
 package manifest
 
 import (
+	"fmt"
```
ardaguclu, line 4: nit: new line after this

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +377,31 @@ func (o *ExtractOptions) Run() error {
...
-						return imagemanifest.NewImageNotFound(msg, err)
+
+					// Translate the current runtime OS to linux when looking through a manifest listed image since a manifest for the particular 'runtime OS'/'archType' may not exist.
+					// In cases where the runtime OS is windows or darwin (i.e. macOS), there is no expectation of a manifest image with those particular OS's.
+					if o.FilterOptions.DefaultOSFilter {
+						runtimeOS := runtime.GOOS
+						arch := runtime.GOARCH
+						if !warnedAboutLinuxConversion {
+							warnedAboutLinuxConversion = true
+							klog.Warningf("Warning: a manifest image could not be found for this platform: %s/%s. Converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image.\n", runtimeOS, arch, runtimeOS, arch)
```
ardaguclu, line 393: After investigating this further;

It would be better to remove `warnedAboutLinuxConversion` flag altogether and logging this in any case with V(2) like this

```go
klog.V(2).Infof("a manifest image could not be found for this platform: %s/%s. Converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image.\n", runtimeOS, arch, runtimeOS, arch)
```

WDYT?

bryan-cox: I guess this would be okay since it's hidden in the verbose level. It still repeats there:
```
brcox@brcox-mac oc % ./oc adm release extract --command=oc quay.io/openshift-release-dev/ocp-release:4.12.8-multi --to=/tmp/ -a ~/all-the-pull-secrets.json -v 2
I0329 06:50:01.736430   59066 extract.go:393] Warning: a manifest image could not be found for this platform: %s/%s. Converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image.
darwinamd64darwinamd64
I0329 06:50:02.013744   59066 manifest.go:451] Chose linux/amd64 manifest from the manifest list.
I0329 06:50:03.377163   59066 extract_tools.go:479] Will extract usr/share/openshift/mac/oc from quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:8653f3d359970b5126f6ada6618898a61da9ae7ff88dbcbec59eed249dc85765
I0329 06:50:03.377555   59066 extract_tools.go:466] Skipping openshift-client-mac-arm64-%s.tar.gz, does not match current architecture arm64
I0329 06:50:03.377917   59066 extract_tools.go:461] Skipping openshift-client-linux-%s.tar.gz, does not match current OS linux
I0329 06:50:03.377936   59066 extract_tools.go:461] Skipping openshift-client-linux-amd64-%s.tar.gz, does not match current OS linux
I0329 06:50:03.377950   59066 extract_tools.go:461] Skipping openshift-client-linux-arm64-%s.tar.gz, does not match current OS linux
I0329 06:50:03.377962   59066 extract_tools.go:461] Skipping openshift-client-windows-%s.zip, does not match current OS windows
I0329 06:50:03.651705   59066 extract.go:393] Warning: a manifest image could not be found for this platform: %s/%s. Converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image.
darwinamd64darwinamd64
I0329 06:50:03.821987   59066 manifest.go:451] Chose linux/amd64 manifest from the manifest list.
I0329 06:53:02.564941   59066 extract_tools.go:853] Found match for release version at 3832 (len=4096, offset=93, n=4003)
```

bryan-cox: If you are good with that, I will push the changes up.

pkg/cli/image/extract/extract.go
```diff
@@ -375,17 +376,29 @@ func (o *ExtractOptions) Run() error {
...
-						msg := fmt.Sprintf("image %q not found: %s", from, err.Error())
-						return imagemanifest.NewImageNotFound(msg, err)
+
+					// Translate the current runtime OS to linux when looking through a manifest listed image since a manifest for the particular 'runtime OS'/'archType' may not exist.
+					// In cases where the runtime OS is windows or darwin (i.e. macOS), there is no expectation of a manifest image with those particular OS's.
+					if o.FilterOptions.DefaultOSFilter {
+						runtimeOS := runtime.GOOS
+						arch := runtime.GOARCH
+
+						klog.V(2).Info(fmt.Printf("Warning: a manifest image could not be found for this platform: %s/%s. Converting runtime OS, %s, to 'linux' to pull the linux/%s equivalent image.\n", runtimeOS, arch, runtimeOS, arch))
```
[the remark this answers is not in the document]
bryan-cox: Done.

ardaguclu:
Thanks @bryan-cox.

/lgtm
/hold
until pre-merge test is completed.

@zhouying7780 could you PTAL for the pre-merge verification?. Thanks.

zhouying7780:
pre-merge tested , no issue found

PR #1312, opened 2022-12-14
IR-261: Add --import-mode to 'tag'

werled:
Adds the given value (one of "PreserveOriginal", "Legacy", "") to the resulting ImageStreamTag. Invalid values result in an error by the API server.

https://issues.redhat.com/browse/IR-261
See also: https://github.com/openshift/oc/pull/1289/

### DIFF
--- a/pkg/cli/tag/tag.go
+++ b/pkg/cli/tag/tag.go
@@ -37,6 +37,7 @@ type TagOptions struct {
 	namespace    string
 
 	referencePolicy string
+	importMode      string
 
 	ref            imagev1.DockerImageReference
 	sourceKind     string
@@ -44,8 +45,6 @@ type TagOptions struct {
 	destNameAndTag []string
 
 	genericclioptions.IOStreams
-
-	importMode string
 }
 
 var (
--- a/pkg/cli/tag/tag.go
+++ b/pkg/cli/tag/tag.go
@@ -44,6 +44,8 @@ type TagOptions struct {
 	destNameAndTag []string
 
 	genericclioptions.IOStreams
+
+	importMode string
 }
 
 var (
@@ -111,6 +113,7 @@ func NewCmdTag(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.C
 	cmd.Flags().BoolVar(&o.scheduleTag, "scheduled", o.scheduleTag, "Set a container image to be periodically imported from a remote repository. Defaults to false.")
 	cmd.Flags().BoolVar(&o.insecureTag, "insecure", o.insecureTag, "Set to true if importing the specified container image requires HTTP or has a self-signed certificate. Defaults to false.")
 	cmd.Flags().StringVar(&o.referencePolicy, "reference-policy", SourceReferencePolicy, "Allow to request pullthrough for external image when set to 'local'. Defaults to 'source'.")
+	cmd.Flags().StringVar(&o.importMode, "import-mode", o.importMode, "Imports the full manifest list of a tag when set to 'PreserveOriginal'. Defaults to 'Legacy'.")
 
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


flavianmissi:
good stuff
/lgtm

/assign @ardaguclu 
@ardaguclu could you help us with approval here too?

ardaguclu:
This PR conflicts with https://github.com/openshift/oc/pull/1289. Which one we should presume?

flavianmissi:
Hi @ardaguclu, https://github.com/openshift/oc/pull/1289 adds the `--import-mode` flag to `oc import-image` (I see now it's title is a bit misleading though). This PR adds the same flag to `oc tag`. They're very similar, but I'm not sure how they are conflicting. Could you expand?

ardaguclu:
> Hi @ardaguclu, #1289 adds the `--import-mode` flag to `oc import-image` (I see now it's title is a bit misleading though). This PR adds the same flag to `oc tag`. They're very similar, but I'm not sure how they are conflicting. Could you expand?

ah sorry, I missed that part :). Thanks.

pkg/cli/tag/tag.go
```diff
@@ -111,6 +113,7 @@ func NewCmdTag(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.C
 	cmd.Flags().BoolVar(&o.scheduleTag, "scheduled", o.scheduleTag, "Set a container image to be periodically imported from a remote repository. Defaults to false.")
 	cmd.Flags().BoolVar(&o.insecureTag, "insecure", o.insecureTag, "Set to true if importing the specified container image requires HTTP or has a self-signed certificate. Defaults to false.")
 	cmd.Flags().StringVar(&o.referencePolicy, "reference-policy", SourceReferencePolicy, "Allow to request pullthrough for external image when set to 'local'. Defaults to 'source'.")
+	cmd.Flags().StringVar(&o.importMode, "import-mode", o.importMode, "Imports the full manifest list of a tag when set to 'PreserveOriginal'. Defaults to 'Legacy'.")
```
ardaguclu, line 116: It is better to add an explanation like in here https://github.com/openshift/oc/pull/1289/files#diff-f0e0ab5ccc1ab1ecac81c724a4c086b8e9d16f525c0d23a91b37d8b2b10b365fR55

pkg/cli/tag/tag.go
```diff
@@ -44,6 +44,8 @@ type TagOptions struct {
 	destNameAndTag []string
 
 	genericclioptions.IOStreams
+
+	importMode string
```
ardaguclu, line 48: Could you please move it to under `referencePolicy string`?

pkg/cli/tag/tag.go
```diff
@@ -302,7 +305,7 @@ func isCrossImageStream(namespace string, srcRef imagev1.DockerImageReference, d
 }
 
 // Validate validates all the required options for the tag command.
-func (o TagOptions) Validate() error {
+func (o *TagOptions) Validate() error {
```
ardaguclu, line 308: What is the reason of this change with respect to the new flag?. If just for standardization, could you please also change this one https://github.com/openshift/oc/blob/f372c323adfdac963c09af0dda2745401c3b1c96/pkg/cli/tag/tag.go#L376?

ardaguclu (commented):
Thanks. This looks good to me. I left a few comments.

wewang58:
Finished testing, according test case
/label qe-approved

pkg/cli/tag/tag.go
werled: I changed this, because the Validate method now actively sets the option to "Legacy" in case it is missing (as opposed to leaving the fallback to the API server after the resource is created): https://github.com/openshift/oc/blob/f372c323adfdac963c09af0dda2745401c3b1c96/pkg/cli/tag/tag.go#L367

This is the same as in the respective change in https://github.com/openshift/oc/pull/1289/files

From what I've seen, the way option structs are passed to these method is not really standardized in the oc code base. So my intuition was to only do this if the method actually changes the struct.

pkg/cli/tag/tag.go
werled: Thanks for the feedback. I added an explanation. I wasn't sure whether `openshift/origin-control-plane:latest` is a good example here, because it doesn't have a manifest list. However, IMHO it's easier to parse the explanation as a user because it's the same as in the other examples (and they probably wouldn't tag this specific image anyways).

pkg/cli/tag/tag.go
ardaguclu: Good point, you are right

PR #1313, opened 2022-12-15
WRKLDS-629: oc adm must-gather: unhidden --run-namespace flag

ingvagabund:
Provide a visible flag to run must-gather in its own name space. Originally an experimental flag.


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -100,7 +100,6 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericclioptions.IOStream
 	cmd.Flags().StringVar(&o.SourceDir, "source-dir", o.SourceDir, "Set the specific directory on the pod copy the gathered data from.")
 	cmd.Flags().StringVar(&o.timeoutStr, "timeout", "10m", "The length of time to gather data, like 5s, 2m, or 3h, higher than zero. Defaults to 10 minutes.")
 	cmd.Flags().StringVar(&o.RunNamespace, "run-namespace", o.RunNamespace, "An existing namespace where must-gather pods should run. If not specified a temporary namespace will be generated.")
```
ardaguclu, line 102: I think, it is better to mention in this description about "namespace should be privileged". Otherwise, when users get pod security admission violation error, they might complain to us.

ingvagabund: Good point, thx

PR #1314, opened 2022-12-21
Update errors when debug pod fails

soltysh:
/assign @sanchezl @ardaguclu 

This will unblock the rebase PR, but I think @ardaguclu we need to revisit this bit of code to be a bit smarter when dealing with failures from creating debug pods. This change is coming from k8s 1.26, so I'd suggest to re-visit it once 1.26 lands. 

1.26 PR: https://github.com/openshift/kubernetes/pull/1432

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -595,8 +595,19 @@ func (o *DebugOptions) RunDebug() error {
 			// switch to logging output
 		case err == krun.ErrPodCompleted, err == conditions.ErrContainerTerminated:
 			resultPod, ok := containerRunningEvent.Object.(*corev1.Pod)
-			if ok && resultPod.Status.Reason == "NodeAffinity" && len(resultPod.Spec.NodeSelector) != 0 {
-				return fmt.Errorf("debug pod could not be scheduled: %v. To fix this you may want to create a new namespace with empty node selector and run the debug there. For example: oc adm new-project --node-selector=\"\" debug", resultPod.Status.Message)
+			if ok {
+				if resultPod.Status.Reason == "NodeAffinity" && len(resultPod.Spec.NodeSelector) != 0 {
+					return fmt.Errorf("debug pod could not be scheduled: %v. To fix this you may want to create a new namespace with empty node selector and run the debug there. For example: oc adm new-project --node-selector=\"\" debug", resultPod.Status.Message)
+				}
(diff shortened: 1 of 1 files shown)


pkg/cli/debug/debug.go
```diff
@@ -595,8 +595,19 @@ func (o *DebugOptions) RunDebug() error {
...
 		case err == krun.ErrPodCompleted, err == conditions.ErrContainerTerminated:
 			resultPod, ok := containerRunningEvent.Object.(*corev1.Pod)
-			if ok && resultPod.Status.Reason == "NodeAffinity" && len(resultPod.Spec.NodeSelector) != 0 {
-				return fmt.Errorf("debug pod could not be scheduled: %v. To fix this you may want to create a new namespace with empty node selector and run the debug there. For example: oc adm new-project --node-selector=\"\" debug", resultPod.Status.Message)
+			if ok {
+				if resultPod.Status.Reason == "NodeAffinity" && len(resultPod.Spec.NodeSelector) != 0 {
+					return fmt.Errorf("debug pod could not be scheduled: %v. To fix this you may want to create a new namespace with empty node selector and run the debug there. For example: oc adm new-project --node-selector=\"\" debug", resultPod.Status.Message)
+				}
+				for _, c := range resultPod.Status.Conditions {
+					if c.Type == corev1.AlphaNoCompatGuaranteeDisruptionTarget {
```
ardaguclu, line 603: This is a very specific conditional. What does `AlphaNoCompatGuaranteeDisruptionTarget` mean?, why we can not gather logs when this exists.

pkg/cli/debug/debug.go
```diff
@@ -595,8 +595,19 @@ func (o *DebugOptions) RunDebug() error {
 			// switch to logging output
 		case err == krun.ErrPodCompleted, err == conditions.ErrContainerTerminated:
 			resultPod, ok := containerRunningEvent.Object.(*corev1.Pod)
```
ardaguclu, line 597: Do you have any solid idea how we can make it smarter;

* not only relying on this casting to gather pod logs(because it might include exceptions like in below)?.
* or this casting is mandatory and `getLogs` would be smart enough for such cases and we'll not have these exceptions below?

ardaguclu:
After the discussion on Slack with @soltysh, in order to unblock rebase PR, we can merge this one.

It already looks good to me;

/lgtm

PR #1317, opened 2022-12-22
OTA-824: hack/verify-commands: New harness for end-to-end oc testing

wking:
Ensuring we don't break compatibility with:

```code
$ oc adm release extract --file=image-references quay.io/openshift-release-dev/ocp-release:4.12.0-rc.0-x86_64
{
  "kind": "ImageStream",
  "apiVersion": "image.openshift.io/v1",
  ...
    "dockerImageRepository": ""
  }
}
```

Compare [this output][1] from a different release image with the broken `oc`:

```
{
  "kind": "ImageStream",
  "apiVersion": "image.openshift.io/v1",
  ...
    "dockerImageRepository": ""
  }
}Extracted release payload created at 2022-10-13T12:24:43Z
```

I've tried to set things up to make it easy for other subcommands to add their own tests, without getting so complicated that we need to use an off-the-shelf test harness, and pick up the dependency cost that sort of thing brings along.  Once we land this, or something like it,  I'll reroll openshift/release#34989 to run it in a presubmit, so we protect ourselves from backsliding.

[1]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-multiarch-master-nightly-4.12-ocp-e2e-sdn-serial-aws-arm64/1580538617384669184/artifacts/release/artifacts/release-images-arm64-latest


wking:
[CI][1] [looks good][2]:

```
...
OC=./oc hack/verify-commands.sh testdata/commands
testing testdata/commands/admin/upgrade/extract/image-references ...
...
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1317/pull-ci-openshift-oc-master-verify/1606057590729478144
[2]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_oc/1317/pull-ci-openshift-oc-master-verify/1606057590729478144/artifacts/test/build-log.txt

testdata/commands/admin/upgrade/extract/image-references/command.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+"${OC:-oc}" adm release extract --file=image-references quay.io/openshift-release-dev/ocp-release:4.11.0-x86_64
```
ardaguclu, line 2: Why we don't write unit tests to test this command's output is correct or not?

wking: it would be possible to inline all the stdout/stderr/exit-status capture and comparison in this script.  But presumably folks will have other things that also want that logic, so I've centralized it in `verify-commands.sh`.  We could convert to a `verify-commands.go` library, if that makes it more unit-test-y, but we still want to be executing an external `oc` command to get that end-to-end testing, and we had existing verification like `verify-generated-completions.sh` in shell, so :shrug:, it seemed easier to write the centralized code in shell.  But I don't feel too strongly about it, and I can port to the Go-library-calling-an-external-`oc`, if that's what `oc`-maintainers prefer.

ardaguclu: I checked the code and it would be difficult to add unit test case for `adm release extract` command without some refactorings. 

In my opinion, instead of adding a whole brand new step just for testing the output of this command, best thing is adding a e2e test case in here https://github.com/openshift/origin/blob/master/test/extended/cli/admin.go by skipping in disconnected. That will ensure that command works as expected?

wking:
Replacing with openshift/origin#27822

PR #1318, opened 2023-01-04
OCPBUGS-4906: oc process: Set original namespace if it differs

ardaguclu:
When template passed to `oc process` has namespace different from the `--namespace`. It returns an error;

```
the namespace of the provided object does not match the namespace sent on the request
``` 

This PR uses the same fix applied in https://github.com/openshift/oc/pull/1272


ardaguclu:
/hold
until manually verification is done. I opened this PR to complete CI jobs prior to that.

soltysh (approved):
/lgtm

PR #1320, opened 2023-01-10
OCPBUGS-5010: Remove must-gather pod after it completes when explicit namespace is used

ardaguclu:
Currently, when user passes `--run-namespace` flag, must-gather pod becomes stale and is not removed after it completes.

The reason is that `copy` container in the pod is set sleeping infinite, thus pod never falls into `completed` status to be collected by garbage collector.

One option is setting to a precise duration for sleeping and wait until garbage collector removes pods in `completed` status. But that also means that we will delay pod removal to a certain time which is presumably long(2 hours maybe?).

Other option is removing the must-gather pod manually.

This PR applies second option because we know when task is completed and there might be no need to delay this removal.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -379,6 +379,19 @@ func (o *MustGatherOptions) Run() error {
 		go func(pod *corev1.Pod) {
 			defer wg.Done()
 
+			if len(o.RunNamespace) > 0 && !o.Keep {
+				defer func() {
+					// must-gather runs in its own separate namespace as default , so after it is completed
+					// it deletes this namespace and all the pods are removed by garbage collector.
+					// However, if user specifies namespace via `run-namespace`, these pods need to
+					// be deleted manually.
+					err = o.Client.CoreV1().Pods(o.RunNamespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+					if err != nil {
+						klog.V(4).Infof("pod deletion error %v", err)
+					}
+				}()
+			}
+
 			log := newPodOutLogger(o.Out, pod.Name)
 
 			// wait for gather container to be running (gather is running)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -379,6 +379,19 @@ func (o *MustGatherOptions) Run() error {
 		go func(pod *corev1.Pod) {
 			defer wg.Done()
 
+			if len(o.RunNamespace) > 0 {
+				defer func() {
```
ingvagabund, line 383: In case the `oc adm must-gather` command is interrupted by a user/hanging in the background, the must-gather pod will get removed prematurely/not get removed at all.

ardaguclu:
/hold

After @ingvagabund comment, I noticed that we should not delete must-gather pod when user passes `--keep` flag.

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: > In case the `oc adm must-gather` command is interrupted by a user/hanging in the background, the must-gather pod will get removed prematurely/not get removed at all.
> 
> There's also a request to implement `oc adm must-gather [resume]` which will "re-join" the already running must-gather pod in case the command was stopped/interrupted.

Isn't this problem also valid for default must-gather?

ingvagabund: It is. The internet connection from the client to the server may be interrupted. So the pod will not get deleted. Also, when a temporary namespace is created it will not get removed either.

Even with checking the keep option, the must-gather pod can be still deleted prematurely. Worth checking whether aborting the command will still call the defer here.

ardaguclu: > It is. The internet connection from the client to the server may be interrupted. So the pod will not get deleted. Also, when a temporary namespace is created it will not get removed either.
> 
> Even with checking the keep option, the must-gather pod can be still deleted prematurely. Worth checking whether aborting the command will still call the defer here.

I checked from the code and if user passes `--keep` flag, I could not find any code portion that can delete namespace or the pod. But for the `--keep=false` case, you are right, namespace or pod become stale after manual interruption. But maybe we can propose a general fix for all cases when user interrupts the connection in a followup PR, since this PR only focuses on removal of pod when user passes `--run-namespace`. What do you think?

ardaguclu: @ingvagabund what do you think about this https://github.com/openshift/oc/pull/1320#discussion_r1065667621?.

ingvagabund: Using --keep make sense. As a user I was not aware of this option. So I presume the oc adm must-gather does not remove a pod when the command is accidentally interrupted by a user. Seen an unwritten rule by observation. Nevertheless, given this is a "new" option, we can state the interruption case in the release notes and advise users to consider usage of the keep option.

ardaguclu: Thanks, sounds good to me

PR #1321, opened 2023-01-11
Enable Changelog to output in JSON

bradmwilliams:
A request has been made to enable the release-controller to display feature-orientated differences, specifically, for 4.Next releases, but it will also be available for all releases.  To accomplish this, we'd like the enable `oc` to be able to generate it's changelog information in a format other than markdown, for easier post-processing.  This PR has an initial attempt at making as close to a 1-1 mapping from the markdown output to a JSON payload.

For example, the release-controller's output for a difference between `4.12.0-rc.7` and `4.12.0-rc.8` can be observed [here](https://amd64.ocp.releases.ci.openshift.org/dashboards/compare?from=4.12.0-rc.7&to=4.12.0-rc.8).  The release-controller does a minimal amount of processing to the output of the following command:
```
oc adm release info --changelog /tmp/git quay.io/openshift-release-dev/ocp-release:4.12.0-rc.7-x86_64 quay.io/openshift-release-dev/ocp-release:4.12.0-rc.8-x86_64
```

This PR will enable the same command to output a JSON payload by specifying the `--output=json` option, like:
```
oc adm release info --changelog /tmp/git quay.io/openshift-release-dev/ocp-release:4.12.0-rc.7-x86_64 quay.io/openshift-release-dev/ocp-release:4.12.0-rc.8-x86_64 --output=json
```
The output can be seen [here](https://github.com/openshift/oc/files/10396417/changelog.json.txt)

### DIFF
--- a/pkg/cli/admin/release/bug.go
+++ b/pkg/cli/admin/release/bug.go
@@ -240,6 +240,22 @@ func (b RefList) PrintRefs(out io.Writer) {
 	}
 }
 
+// GetRefsForSource returns a map of URLs, sorted by ID, for the specified RefType
+func (b RefList) GetRefsForSource(source RefType) map[string]string {
+	refs := make(map[string]string)
+	for _, ref := range b.Refs {
+		if ref.Source == source {
+			switch ref.Source {
+			case Bugzilla:
+				refs[ref.ID] = fmt.Sprintf(bugzillaBrowseURL, ref.ID)
+			case Jira:
+				refs[ref.ID] = fmt.Sprintf(jiraBrowseURL, ref.ID)
+			}
+		}
+	}
+	return refs
+}
+
 // GetRefList converts Ref map to Ref list after sorting
 // according to the ID field.
 func GetRefList(b map[string]Ref) []Ref {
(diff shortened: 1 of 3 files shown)


nstielau:
The output looks good to me.

pkg/cli/admin/release/bug.go
```diff
@@ -278,6 +294,7 @@ func extractRefs(msg string) (RefList, string) {
 	if bt == Jira {
 		refs = strings.TrimSpace(parsedMsg[0])
 		refs = strings.TrimSuffix(refs, ":")
+		refs = strings.TrimSuffix(refs, "-")
```
ardaguclu, line 297: Since you control these naming patterns, I assume this change is safe

ardaguclu (commented):
Changes look good to me

/approve

pkg/cli/admin/release/bug.go
bradmwilliams: AFAIK, the Jira automation, that we have in place, only expects the `:`.  The regex used in `oc`, for Bugzilla and Jira, allows for both `:` and `-`.  
In validating these changes, I found a bunch of occurrences where the changelog had been incorrectly parsing Jira tickets that used the `-`, as seen here:
JiraParsingIssue
Instead of ignoring these tickets all together (by changing the regex), I chose to simply trim off the `-` if when it appears.

PR #1324, opened 2023-01-19
added show-groups option to display group membership

christianh814:
Added an option `-g` or `--show-groups` that shows the group of the user that is logged in.

Signed-off-by: Christian Hernandez <christian@chernand.io>


vrutkovs:
/retitle NO-JIRA: Display whoami information as JSON or YAML
/refresh

@ardaguclu PTAL

kaovilai:
@vrutkovs fyi jira exists here: https://issues.redhat.com/browse/RFE-3594 should we use that?

christianh814:
> /retitle NO-JIRA: Display whoami information as JSON or YAML /refresh
> 
> @ardaguclu PTAL

@vrutkovs JIRA was created 2 years ago https://issues.redhat.com/browse/RFE-3594

vrutkovs:
This RFE seems deferred, so we'll need a bug or a task ticket for this. In any case, JIRA ticket here is required if we want to backport it - and it can be done later

pkg/cli/whoami/whoami.go
```diff
@@ -153,14 +164,34 @@ func (o *WhoAmIOptions) Run() error {
 		}
 		fmt.Fprintf(o.Out, "%s\n", consoleUrl)
 		return nil
-	}
+	case o.Output == "yaml":
+		u, err := o.UserInterface.Users().Get(context.TODO(), "~", metav1.GetOptions{})
```
ardaguclu, line 168: SelfSubjectReview needs to take precedence from UserInterface (logic https://github.com/openshift/oc/blob/093ad8719d283e1a9f79584a6c92aa9b687d7f5b/pkg/cli/whoami/whoami.go#L88-L107). Actually, there is no need to duplicate the effort. Why can't we print what we get in `WhoAmI` function.

pkg/cli/whoami/whoami.go
```diff
@@ -153,14 +164,34 @@ func (o *WhoAmIOptions) Run() error {
...
-	o.UserInterface, err = userv1typedclient.NewForConfig(o.ClientConfig)
-	if err != nil {
-		return err
+		y, err := yaml.Marshal(u)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintf(o.Out, "%s\n", string(y))
+		return nil
+	case o.Output == "json":
```
ardaguclu, line 180: We can use built-in printing flags to print yaml and json files. Example, https://github.com/kubernetes/kubectl/blob/5ff92a69e3ecb06912d5f468186d458bf610244e/pkg/cmd/auth/whoami.go#L64

pkg/cli/whoami/whoami.go
```diff
@@ -47,6 +49,8 @@ type WhoAmIOptions struct {
 	ShowContext    bool
 	ShowServer     bool
 	ShowConsoleUrl bool
+	ShowGroups     bool
```
ardaguclu, line 52: Unused and not related to this PR?

ardaguclu (commented):
Thank you for this PR. However, SelfSubjectReview takes precedence over User object in OpenShift. So that this PR needs to be modified accordingly.

christianh814:
@ardaguclu  Updated/resolved comments. PTAL cc @kaovilai

kaovilai:
Lol

pkg/cli/whoami/whoami.go
```diff
@@ -9,7 +9,8 @@ import (
 	v1 "k8s.io/api/authentication/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/cli-runtime/pkg/genericiooptions"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
```
ardaguclu, line 12: This is deprecated. `genericiooptions` needs to be used.

pkg/cli/whoami/whoami.go
```diff
@@ -171,19 +205,18 @@ func (o *WhoAmIOptions) Run() error {
 		}
 		fmt.Fprintf(o.Out, "%s\n", consoleUrl)
 		return nil
+	case o.resourcePrinterFunc != nil:
+		// If output format is specified, get the user info and print it in the requested format
+		me, err := o.WhoAmI()
```
ardaguclu, line 210: `WhoAmI` function still prints the username into the terminal. So the output of this function is not an valid json/yaml. I think we can pass a boolean to print username or not.

ardaguclu (commented):
Overall idea looks good to me. Dropped a few comments. Also it is better to add unit tests to verify the behavior in whoami_test.go

Thank you

pkg/cli/whoami/whoami.go
christianh814: But isn't `genericclioptions` used for PrintFlags? I see I can use `genericiooptions` for IOStream. Both are used right? Because `rollback` and `startBuild` also use this approach

ardaguclu: `genericclioptions` has been deprecated in favor of `genericiooptions`. So theoretically you can use `genericclioptions`. However, it is always preferable to use the recommended one (i.e. `genericiooptions`), since the other one will be removed entirely at some point.

ardaguclu:
I saw merge commits in your PR history. This shouldn't happen. Could you please rebase onto the first commit?.

kaovilai:
if you use merge method squash it shouldn't matter? can do via label for tide to follow.

kaovilai:
[ardaguclu](https://github.com/ardaguclu) we good?

ardaguclu:
My comments have not been resolved yet. Besides, this PR contains merge commits which are not allowed.

It is better to close this inactive PR
/close

ardaguclu:
@kaovilai feel free to open a new PR by applying my comments above. I can gladly review it.

PR #1334, opened 2023-02-06
Allow keeping sparse manifest lists

Jamstah:
Using this configuration will require that the target registry supports sparse manifest lists.


Jamstah:
Sparse manifest support in registries:
- https://github.com/distribution/distribution/pull/3536
- https://issues.redhat.com/browse/PROJQUAY-3114

Covers this RFE:
- https://issues.redhat.com/browse/RFE-3333

dmesser:
@Jamstah That's awesome, thank you. 

I think the `--keep-manifest-list` option is now somewhat debatable. The scenario where you don't rely on that behavior is effectively going to break all digest-based references. 

IMHO you will always want to output a manifest list as a result of a mirroring manifest list from a source registry. Then the only question that remains to me is: do you want to filter that list down to the architectures you are interested in? So it remains a manifest list but with fewer arch-specific child manifest. So the `--keep-manifest-list` is not at odds with the `--filter-by-os` switch.

I would probably even make keeping the manifest list the default, if it is not already, and then deprecate the `--keep-manifest-list` switch and replace it with an option that only ever mirrors a certain architecture and the result would not be a manifest list but a regular manifest with that arch-variant of the image.

Jamstah:
I think the scenarios (assuming a manifest list with multiple architectures):

- **You want to mirror only one architecture and you are using tags**
oc image mirror can mirror the arch specific image and tag it using that tag for the manifest list in the source. It does this today.

- **You want to mirror only one architecture and you are using digests**
oc image mirror must mirror the individual architecture and the unchanged manifest list. Registry must support sparse indexes.

- **You want to mirror a subset of architectures and you are using tags**
oc image mirror can mirror the specific architectures and generate a new manifest list, and tag that new list using the tag for the manifest list in the source. It does this today (although I think its buggy...).

- **You want to mirror a subset of architectures and you are using digests**
oc image mirror must mirror the specific architectures and the unchanged manifest list. Registry must support sparse indexes.

- **You want to mirror all architectures and you are using tags**
oc image mirror can mirror the manifest list in its entirety. It does this today.

- **You want to mirror all architectures and you are using digests**
oc image mirror can mirror the manifest list in its entirety. It does this today.

Because some of those options require sparse manifest support in the registry, I don't think we can change the default behaviour to always keep the manifest list yet, although I agree that long term, using sparse manifests is the sensible default for all filtering cases to maintain digest stability.

Are you suggesting changes to this PR?

Jamstah:
Can I get a review on this one? I'd like to get this one reviewed before doing a similar change for `oc adm catalog mirror`.

ardaguclu:
Thanks for the PR @Jamstah. I'm planning to review this next week;

/ok-to-test

ardaguclu:
AFAIK, unit test will be fixed with this https://github.com/openshift/oc/pull/1377

Jamstah:
> AFAIK, unit test will be fixed with this #1377

Ah, thanks. I'll wait for that to merge and rebase then, rather than merging it into this branch early. Hopefully that will be in time for your review next week :)

pkg/cli/image/manifest/manifest.go
```diff
@@ -434,28 +435,29 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
...
-			return srcManifests, srcManifests[0], manifestDigest, nil
+			return nil, childManifests[0], manifestDigest, nil
 		default:
-			return append(srcManifests, manifestList), manifestList, manifestDigest, nil
+			return childManifests, manifestList, manifestDigest, nil
 		}
 
 	default:
-		return []distribution.Manifest{srcManifest}, srcManifest, srcDigest, nil
+		return nil, srcManifest, srcDigest, nil
```
ardaguclu, line 460: Aren't we changing the behavior for cases where `keepManifestList` is false?

pkg/cli/image/manifest/manifest.go
```diff
@@ -434,28 +435,29 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
...
-			manifestDigest, err := registryclient.ContentDigestForManifest(srcManifests[0], srcDigest.Algorithm())
+		case len(childManifests) == 1 && !keepManifestList:
+			// Just return the single platform specific image
+			manifestDigest, err := registryclient.ContentDigestForManifest(childManifests[0], srcDigest.Algorithm())
 			if err != nil {
 				return nil, nil, "", err
 			}
 			klog.V(2).Infof("Chose %s/%s manifest from the manifest list.", t.Manifests[0].Platform.OS, t.Manifests[0].Platform.Architecture)
-			return srcManifests, srcManifests[0], manifestDigest, nil
+			return nil, childManifests[0], manifestDigest, nil
```
ardaguclu, line 454: Same, what is the reason of not returning `srcManifests`?

pkg/cli/image/mirror/mirror.go
```diff
@@ -538,12 +543,12 @@ func (o *MirrorImageOptions) plan() (*plan, error) {
...
 						// filter or load manifest list as appropriate
 						originalSrcDigest := srcDigest
-						srcManifests, srcManifest, srcDigest, err := imagemanifest.ProcessManifestList(ctx, srcDigest, srcManifest, manifests, src.ref.Ref, o.FilterOptions.IncludeAll, o.KeepManifestList)
+						srcChildren, srcManifest, srcDigest, err := imagemanifest.ProcessManifestList(ctx, srcDigest, srcManifest, manifests, src.ref.Ref, o.FilterOptions.IncludeAll, o.KeepManifestList)
 						if err != nil {
 							plan.AddError(retrieverError{src: src.ref, err: err})
 							return
 						}
-						if len(srcManifests) == 0 {
+						if srcManifest == nil {
```
ardaguclu, line 551: What is the reason of changing from `len(srcChildren) == 0` to this?

pkg/cli/image/mirror/mirror.go
```diff
@@ -601,18 +606,17 @@ func (o *MirrorImageOptions) plan() (*plan, error) {
 								// upload all the blobs
 								srcBlobs := srcRepo.Blobs(ctx)
 
-								// upload each manifest
-								for _, srcManifest := range srcManifests {
+								addBlobsForManifest := func(srcManifest distribution.Manifest) {
```
ardaguclu, line 609: Is the reason of this change for `addBlobsForManifest(srcManifest)`?

pkg/cli/image/mirror/plan.go
```diff
@@ -615,7 +614,7 @@ type repositoryManifestPlan struct {
 	}
 }
 
-func (p *repositoryManifestPlan) Copy(srcDigest godigest.Digest, srcManifest distribution.Manifest, tags []string, to distribution.ManifestService, toBlobs distribution.BlobService) {
+func (p *repositoryManifestPlan) Copy(srcDigest godigest.Digest, srcManifest distribution.Manifest, prerequisites []godigest.Digest, tags []string, to distribution.ManifestService, toBlobs distribution.BlobService) {
```
ardaguclu, line 617: Now manifests not in `manifestlist.DeserializedManifestList` type should pass `prerequisites` as nil?

ardaguclu (commented):
Thanks for the PR @Jamstah. My first iteration of this PR mostly about figuring out the reasons of the changes. In addition to that CI failures might be related to this change, worth to have a look at.

pkg/cli/image/manifest/manifest.go
Jamstah: Yes, but this PR also modifies the callers to handle that change.

pkg/cli/image/mirror/mirror.go
Jamstah: Yes, so we can reuse the blob code for the parent and the children. The other option was to make a new slice with the parent and the children combined, but that would require a new allocation and I thought this was easier.

Jamstah:
> Thanks for the PR @Jamstah. My first iteration of this PR mostly about figuring out the reasons of the changes. In addition to that CI failures might be related to this change, worth to have a look at.

Thanks for the comprehensive review. Don't be shy, I'm open to any and all feedback!

I did look at putting some tests in here, but its not an easy thing to mock up. I feel like the right approach would probably be to have a mock plan object using something like gomock so that we could validate the calls to plan based on the args to `oc image mirror`. gomock isn't already a test dependency of the project though, so I didn't want to try that without a discussion.

Otherwise, there is no easy way to compare two plans, and writing that was too big a job for the time I had available :)

pkg/cli/image/manifest/manifest.go
```diff
@@ -434,28 +435,29 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
 			klog.V(5).Infof("Filtered manifest list to new digest %s:\n%s", manifestDigest, body)
 		}
 
-		for i, manifest := range t.Manifests {
+		for i, manifest := range filtered {
```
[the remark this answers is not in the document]
ardaguclu: It makes sense but I wonder how it was working as expected before. Because we were also appending the unfiltered manifests

ardaguclu (commented):
Code looks good to me, left a few comments. 

But as far as I can see this PR also tries to do same changes on `append` command https://github.com/openshift/oc/pull/1361, it is worth to check it.

pkg/cli/image/manifest/manifest.go
```diff
@@ -291,17 +291,28 @@ func FirstManifest(ctx context.Context, from imagereference.DockerImageReference
...
 
 	originalSrcDigest := srcDigest
-	srcManifests, srcManifest, srcDigest, err := ProcessManifestList(ctx, srcDigest, srcManifest, manifests, from, filterFn, false)
+	srcChildren, srcManifest, srcDigest, err := ProcessManifestList(ctx, srcDigest, srcManifest, manifests, from, filterFn, false)
 	if err != nil {
 		return nil, ManifestLocation{}, err
 	}
-	if len(srcManifests) == 0 {
+	if srcManifest == nil {
+		// Everything got filtered from the list so no image returned at all
```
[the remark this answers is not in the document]
Jamstah: The call is explicitly passing keepmanifestlists=false here because we're trying to filter for the first manifest and its the only safe way to do it.

pkg/cli/image/manifest/manifest.go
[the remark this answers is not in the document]
Jamstah: Before, if you were keeping the list you needed all the images, and if you weren't keeping the list it made you a new list and put it in t with only the filtered images.

Now, we want to keep the list, but not necessarily copy everything underneath it, and filtered matches that use case.

pkg/cli/image/manifest/manifest.go
```diff
@@ -411,12 +422,13 @@ func ProcessManifestList(ctx context.Context, srcDigest digest.Digest, srcManife
 			filtered = append(filtered, manifest)
 		}
 
-		if len(filtered) == 0 {
+		if len(filtered) == 0 && !keepManifestList {
```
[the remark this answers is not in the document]
Jamstah: I don't think we return nil for child manifests, I think we return an empty slice in this case. It just follows the exact same logic as if it had selected children, but with an empty slice.

pkg/cli/image/mirror/mirror.go
```diff
@@ -621,20 +625,28 @@ func (o *MirrorImageOptions) plan() (*plan, error) {
...
+								for _, childManifest := range srcChildren {
+									addBlobsForManifest(childManifest)
+								}
 							}
 
-							if len(srcManifests) > 1 {
-								for _, srcManifest := range srcManifests {
+							prerequisites := make([]godigest.Digest, 0, len(srcChildren))
+							if len(srcChildren) > 0 {
+								for _, srcManifest := range srcChildren {
```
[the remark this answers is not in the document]
Jamstah: Yes it would :)

ardaguclu:
Thanks @Jamstah code looks good to me. I'd prefer this PR https://github.com/openshift/oc/pull/1361 merged first to assure that no merge conflicts and we can iterate reviewing one last time. After that @zhouying7780 can do the pre-merge testing.

Jamstah:
> Thanks @Jamstah code looks good to me. I'd prefer this PR #1361 merged first to assure that no merge conflicts and we can iterate reviewing one last time. After that @zhouying7780 can do the pre-merge testing.

Yep, happy to rebase after that one goes in, although at the moment, I really don't understand what that one is doing :)

Jamstah:
Code is good but I won't try and rerun these for now because I'm expecting to rebase on #1361 (although I checked and it doesn't conflict).

What is the policy on squashing, does the project prefer to squash or leave history?

ardaguclu:
e2e-aws-ovn-serial is perm failing currently and no need to retest

tidy can squash automatically;
/tide merge-method-squash

ardaguclu:
/lgtm
/retest-required
/hold
until pre-merge testing is completed.

zhouying7780:
Do some pre-merge test , we will hit error when use --keep-manifest-list and --filter-by-os at the same time:
oc image mirror --keep-manifest-list=true  --filter-by-os linux/s390x quay.io/openshift-release-dev/ocp-release-nightly@sha256:e12b4260d63b0491715b6c4623dcfa33fb3ac19f69b48ca84e82ff7fc667adcc   localhost:5000/openshift-release-nightly:414390  --insecure  
Warning: the default reading order of registry auth file will be changed from "${HOME}/.docker/config.json" to podman registry config locations in the future version of oc. "${HOME}/.docker/config.json" is deprecated, but can still be used for storing credentials as a fallback. See https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md for the order of podman registry config locations.
localhost:5000/
  openshift-release-nightly
    blobs:
      quay.io/openshift-release-dev/ocp-release-nightly sha256:54fd50e23884489e441ac80517d8edd91d8d1ed9e63af72a05c56c9de5300e25 1.412KiB
      quay.io/openshift-release-dev/ocp-release-nightly sha256:221040e36e1e22356932aafde770932ca1ffa30b66328e791fe25b55ab6f00d5 1.713KiB
      quay.io/openshift-release-dev/ocp-release-nightly sha256:affd0d6ced07dd6f8e7688c46f79c38a4f01eecab2ced33eaccbd344dc8e23c2 969.6KiB
      quay.io/openshift-release-dev/ocp-release-nightly sha256:830ee24ebb9c0b0b4512b74fd757b361dbbb997158b7ae855f03016add85d647 11.78MiB
      quay.io/openshift-release-dev/ocp-release-nightly sha256:2c3b4b484b7988a9cac2a0582dc71972ce6f217e2e6f6ba8e854fb83110fbdb0 24.43MiB
      quay.io/openshift-release-dev/ocp-release-nightly sha256:b80727e221a204d972bb0ad36f21184f254d9ee5b3e79d35102ca35cbed1099f 27.75MiB
      quay.io/openshift-release-dev/ocp-release-nightly sha256:9ed68b94598760816e142fcf7f8e9572af00b5567df5cc818ddb0810919627a6 73.08MiB
    manifests:
      sha256:f5c36ae697bbfb7f6aa33181657ae06a440ef6ac0531158eaa0462ecf9ef18e6
      sha256:e12b4260d63b0491715b6c4623dcfa33fb3ac19f69b48ca84e82ff7fc667adcc -> 414390
  stats: shared=0 unique=7 size=138MiB ratio=1.00

phase 0:
  localhost:5000 openshift-release-nightly blobs=7 mounts=0 manifests=2 shared=0

info: Planning completed in 2.91s
uploading: localhost:5000/openshift-release-nightly sha256:affd0d6ced07dd6f8e7688c46f79c38a4f01eecab2ced33eaccbd344dc8e23c2 969.6KiB
uploading: localhost:5000/openshift-release-nightly sha256:b80727e221a204d972bb0ad36f21184f254d9ee5b3e79d35102ca35cbed1099f 27.75MiB
uploading: localhost:5000/openshift-release-nightly sha256:2c3b4b484b7988a9cac2a0582dc71972ce6f217e2e6f6ba8e854fb83110fbdb0 24.43MiB
uploading: localhost:5000/openshift-release-nightly sha256:830ee24ebb9c0b0b4512b74fd757b361dbbb997158b7ae855f03016add85d647 11.78MiB
uploading: localhost:5000/openshift-release-nightly sha256:9ed68b94598760816e142fcf7f8e9572af00b5567df5cc818ddb0810919627a6 73.08MiB
sha256:f5c36ae697bbfb7f6aa33181657ae06a440ef6ac0531158eaa0462ecf9ef18e6 localhost:5000/openshift-release-nightly
error: unable to push manifest to localhost:5000/openshift-release-nightly:414390: errors:
manifest blob unknown: blob unknown to registry
manifest blob unknown: blob unknown to registry
manifest blob unknown: blob unknown to registry

info: Mirroring completed in 8.3s (17.43MB/s)
error: one or more errors occurred while uploading images

Jamstah:
Yes, it will fail unless you have a registry that supports sparse manifest lists. For that you could use this PR to distribution: https://github.com/distribution/distribution/pull/3536

Jamstah:
What are the next steps here?

ardaguclu:
Due to the current CI issues, only bug fixes are allowed to be merged. We just need to wait;

/lgtm
/hold cancel

PR #1336, opened 2023-02-07
OCPBUGS-7190: Reuse LDAP connection when performing group sync

stlaz:
The code was written more like a kube-client but it makes more sense to reuse the connection to the LDAP server for successive queries to prevent LDAP conn/bind spamming.

/hold
fake bumps

/assign @ibihim


pkg/cli/admin/groups/sync/ad.go
```diff
@@ -4,7 +4,7 @@ import (
 	legacyconfigv1 "github.com/openshift/api/legacyconfig/v1"
 	"github.com/openshift/library-go/pkg/security/ldapclient"
 	ldapquery "github.com/openshift/library-go/pkg/security/ldapquery"
-	"github.com/openshift/oc/pkg/helpers/groupsync"
+	syncgroups "github.com/openshift/oc/pkg/helpers/groupsync"
```
ibihim, line 7: How did it work before? I see that `syncgroups` is referenced in the file, but it was not defined before.

stlaz: I have no idea, this was an automatic change

ibihim (approved):
/lgtm

ibihim (changes requested):
We need to close the connection at some point...

pkg/cli/admin/groups/sync/ad.go
ibihim: witchcraft

ardaguclu:
Idea behind the changes look good to me. I think we need a pre-merge testing for this after PR is ready to be merged(after applying real bump instead fake bump)

stlaz:
Thanks! There is in fact group sync testing as a part of the e2e serial test suite. We should go ahead with the library-go changes, then :+1:

zhouying7780:
Done pre-merge test . no issue found . see details from the bug .

PR #1337, opened 2023-02-13
oc debug: Use own fields instead directly depending on Attach's

ardaguclu:
Currently, `Quiet` and `ContainerName` fields are managed by setting
directly to `Attach` command's fields. This creates tight coupling,
that's why, this PR adds these fields into `oc debug` command.

Main motivation behind this change is better aligning with `kubectl debug`
in terms of flag management to ease the future integration.

Moreover, this PR does a few refactorings like renaming and moving
adding flags to it's own function.


soltysh (approved):
/lgtm

PR #1343, opened 2023-02-15
Use registry.k8s.io and update image version

soltysh:
/assign @ardaguclu 
ref https://groups.google.com/a/kubernetes.io/g/dev/c/kgWu5q6h8fs/m/ANmnNAatAQAJ


ardaguclu:
`aws` jobs are failing in deprovisioning step and according to the Slack, it is known currently. I'm not sure it is fixed or not.

PR #1353, opened 2023-02-22
IR-300,IR-301: generates ImageStreamTags with import-mode when using oc new-build and oc-new-app

prb112:
[IR-300](https://issues.redhat.com//browse/IR-300)
- add new flag --import-mode=Legacy or --import-mode=PreserveOriginal
- the default is Legacy

[IR-301](https://issues.redhat.com//browse/IR-301)
- support Legacy and PreserveOriginal import-mode with oc new-build
- add a guard to protect a panic when no DockerConfig exists
- add a PreserveOriginal guard for IsBuilderImage

The scope of this issue is limited to the imported images. The use of BuildConfig is not addressed in this PR, and is addressed in a separate backlog issue with the BuildConfig team.

### DIFF
--- a/pkg/cli/newapp/newapp.go
+++ b/pkg/cli/newapp/newapp.go
@@ -99,6 +99,9 @@ var (
 		# Use a MySQL image in a private registry to create an app and override application artifacts' names
 		oc new-app --image=myregistry.com/mycompany/mysql --name=private
 
+		# Use an image with the full manifest list to create an app and override application artifacts' names
+		oc new-app --image=myregistry.com/mycompany/image --name=private --import-mode=PreserveOriginal
+
 		# Create an application from a remote repository using its beta4 branch
 		oc new-app https://github.com/openshift/ruby-hello-world#beta4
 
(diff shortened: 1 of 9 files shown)


prb112:
Hi @phoracek 

https://github.com/sclorg/s2i-ruby-container/commit/c42405dfd04be6ebac652047d7e8395e54ae8a7f
Breaks the `oc cli` build I'm making the change to point to 3.1. 

Is there another version I should point to? 

Thanks

Paul

prb112:
/cc @coreydaley 

HI Corey are you the best reviewer for this one? Thanks, Paul

coreydaley:
@prb112 I am thinking that it would be better to have a bool flag called `--import-mode-preserve-original` that defaults to false. I see no reason to have a flag that takes additional options if there are only two basic options (true and false). Thoughts?

flavianmissi:
Hi @coreydaley, although you make an interesting point we already implemented this flag in the way that is done here, so I think at this point we have no option but to use it as is.

prb112:
Thank you Florian, and yes - I implemented it to be consistent with the flags use in other spots.

PR #1355, opened 2023-02-27
OCPBUGS-3393: Always copy the blobs if the target isn't a registry

Jamstah:
The assuption that the manifest existing means that the blobs exist only works when the target is a registry that will validate that the blobs have been uploaded before the manifest. On other targets, we should check for layer existence each time because the user may have used "--continue-on-error" and end up with missing layers.

https://issues.redhat.com/browse/OCPBUGS-3393

### DIFF
--- a/pkg/cli/image/mirror/mirror.go
+++ b/pkg/cli/image/mirror/mirror.go
@@ -586,6 +586,9 @@ func (o *MirrorImageOptions) plan() (*plan, error) {
 								mustCopyLayers = true
 							case src.ref.EqualRegistry(dst.ref) && canonicalFrom.String() == canonicalTo.String():
 								// if the source and destination repos are the same, we don't need to copy layers unless forced
+							case dst.ref.Type != imagesource.DestinationRegistry:
+								// If we're not copying to a registry, the destination doesn't guarantee the blobs exist for the manifest
+								mustCopyLayers = true
 							default:
 								if _, err := toManifests.Get(ctx, srcDigest); err != nil {
 									mustCopyLayers = true


pkg/cli/image/mirror/mirror.go
```diff
@@ -586,6 +586,9 @@ func (o *MirrorImageOptions) plan() (*plan, error) {
 								mustCopyLayers = true
 							case src.ref.EqualRegistry(dst.ref) && canonicalFrom.String() == canonicalTo.String():
 								// if the source and destination repos are the same, we don't need to copy layers unless forced
+							case dst.ref.Type != imagesource.DestinationRegistry:
```
ingvagabund, line 589: Does the same hold for case where the destination is s3?

Jamstah: I haven't tested with S3, but I think so, which is why I put the test in this way.

My logic: S3 will not check if the layers for the manifest have already been uploaded, so you can end up in a situation where there is a manifest missing layers, so checking manifest existence doesn't obviate the need to check layer existence.

Jamstah:
Also, I see I have an invalid bug label. I don't know how the RH jira/pr's link up so I'm guessing that the jira release needs updating, but I don't have that access :)

From an IBM cloud paks perspective, I think we would want this backported to at least 4.10 (and that's my opinion, I guess someone raised at 4.8 for a reason?). Would that then be a second jira ticket to backport the fix? Happy to create tickets/prs as required but don't want to spam anything if its not required!

ardaguclu:
> Also, I see I have an invalid bug label. I don't know how the RH jira/pr's link up so I'm guessing that the jira release needs updating, but I don't have that access :)
> 
> From an IBM cloud paks perspective, I think we would want this backported to at least 4.10 (and that's my opinion, I guess someone raised at 4.8 for a reason?). Would that then be a second jira ticket to backport the fix? Happy to create tickets/prs as required but don't want to spam anything if its not required!

@Jamstah thanks for the fix and it looks reasonable to me. 

But for the backporting part, I'm a bit hesitant to backport to older releases. Because as you said in Jira, `force` flag can be used as a workaround to always copying blobs. Another reason is that although there seems to be no side effects of this patch, we haven't proved that yet.

/lgtm
/hold
for pre-merge testing. @zhouying7780 could you PTAL this patch for pre-merge testing?. Thanks.

ingvagabund:
@Jamstah we do not backport fixes to older branches unless there's a strong reason for it. We can backport the fix down to 4.11 if the impact is significant. 4.10 is in the maintenance phase so only critical fixes are backported. Given there's an existing workaround we can currently accept the fix into 4.13.

Jamstah:
> @Jamstah we do not backport fixes to older branches unless there's a strong reason for it. We can backport the fix down to 4.11 if the impact is significant. 4.10 is in the maintenance phase so only critical fixes are backported. Given there's an existing workaround we can currently accept the fix into 4.13.

No complaints here.

zhouying7780:
pre merge test done , no issue found.

ardaguclu:
Considering that aws-ovn-upgrade job is permanently failing, we need to wait until it starts passing again.
/hold

Jamstah:
ovn-upgrade build passed elsewhere, so
/retest-required

PR #1357, opened 2023-02-28
Microshift command docs

soltysh:
/assign @ardaguclu @ShaunaDiaz

### DIFF
--- a/tools/gendocs/gen_openshift_docs.go
+++ b/tools/gendocs/gen_openshift_docs.go
@@ -7,12 +7,14 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/openshift/oc/pkg/cli"
 	"github.com/openshift/oc/tools/gendocs/gendocs"
 )
 
 var admin = flag.Bool("admin", false, "Generate admin commands docs")
+var microshift = flag.Bool("microshift", false, "Generate MicroShift commands docs")
 
 func OutDir(path string) (string, error) {
 	outDir, err := filepath.Abs(path)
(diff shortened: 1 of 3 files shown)


tools/gendocs/gen_openshift_docs.go
```diff
@@ -51,12 +53,16 @@ func main() {
 	out := os.Stdout
 	cmd := cli.NewOcCommand(&bytes.Buffer{}, out, ioutil.Discard)
 
-	outFile := filepath.Join(outDir, "oc-by-example-content.adoc")
+	fileName := "oc-by-example-content.adoc"
```
ShaunaDiaz, line 56: Should this be "microshift-oc-by-example.adoc"?

tools/gendocs/gendocs/microshift.go
```diff
@@ -0,0 +1,38 @@
+package gendocs
+
+import (
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// a set of commands excluded from microshift
```
ShaunaDiaz, line 7: Are excluding these, or adding just the ones in the Google doc?

ShaunaDiaz (changes requested):

tools/gendocs/gen_openshift_docs.go
dhellmann: The prefix is added down on lines 60-62, so I think this is OK as a base filename.

tools/gendocs/gendocs/gendocs.go
```diff
@@ -70,18 +70,21 @@ func GenDocs(cmd *cobra.Command, filename string, admin bool) error {
...
 }
 
-func extractExamples(cmd *cobra.Command, admin bool) Examples {
+func extractExamples(cmd *cobra.Command, admin, microshift bool) Examples {
 	objs := Examples{}
 	for _, c := range cmd.Commands() {
 		if len(c.Deprecated) > 0 {
 			continue
 		}
+		if microshift && microshiftCommands.Has(c.CommandPath()) {
```
dhellmann, line 79: nit: It's a little odd to call it `microshiftCommands` when it is the list of exclusions.

tools/gendocs/gendocs/microshift.go
dhellmann: The logic in gendocs.go on lines 79-81 says that if it sees a command in this list it skips producing output.

dhellmann (commented):
The list of commands looks good to me. I had one nit about the variable name, but it's not worth holding things up. I'll leave a hold in case you want other contributors to look at it before merging, but you can remove that if you're ready to merge it.

/lgtm
/hold

soltysh:
The problem in e2e-aws-ovn is not related with this PR, which is only modifying the docs generation helper, but it's being tracked separately in https://issues.redhat.com/browse/OCPBUGS-8092

/override ci/prow/e2e-aws-ovn

PR #1359, opened 2023-02-28
OCPBUGS-8048: pkg/cli/admin/upgrade: Client-side checks for --to-multi-arch

wking:
`Invalid=True`, `Failing=True`, and `Progressing=True` are all reasons we might want to delay a transition to multi-arch.  Copy over the block from the version-bumping switch case, so we get the usual checks for `--to-multi-arch` too.  The user may opt to waive these guards with `--allow-upgrade-with-warnings` as usual, but at least they'll be aware that there are issues they may want to look into.


evakhoni:
pre merge verified in https://issues.redhat.com/browse/OCPBUGS-8048?focusedCommentId=21834768&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21834768

/label qe-approved

LalatenduMohanty (approved):
/lgtm

PR #1361, opened 2023-03-02
append: expose keep-manifest-list option and preserve manifestlist when appending to specific arch images

Prashanth684:
The keep-manifest-list option for the append command was not exposed externally and only used in the case of creating the release image. This PR aims to expose this option while also keeping functionality intact. So, this means:

  -  keep-manifest-list will always preserve the manifestlist when appending images. when used with --filter-by-os, the 
      append will only happen for filtered images.
  -  if keep-manifest-list is not specified, functionality will be same as before.

### DIFF
--- a/pkg/cli/image/append/append.go
+++ b/pkg/cli/image/append/append.go
@@ -57,9 +57,12 @@ var (
 		add the --drop-history flag to remove information from the image about the system that
 		built the base image.
 
-		Images in manifest list format will automatically select an image that matches the current
-		operating system and architecture unless you use --filter-by-os to select a different image.
-		This flag has no effect on regular images.
+		Images in manifest list format with keep-manifest-list specified will automatically append layers
+		to all sub manifests in the list unless filter-by-os is specified in which case the append will
+		only happen for the filtered manifests while preserving the manifestlist. If keep-manifest-list is
+		not specified, automatically select an image that matches the current operating system and architecture
+		unless --filter-by-os is used to select a different image.
+		These flags have no effect on regular images.
 	`)
 
 	example = templates.Examples(`
@@ -84,8 +87,14 @@ var (
 		oc image append --from-dir v2 --to myregistry.com/myimage:latest layer.tar.gz
 
 		# Add a new layer to a multi-architecture image for an os/arch that is different from the system's os/arch
-		# Note: Wildcard filter is not supported with append. Pass a single os/arch to append
+		# Note: Wildcard filter is not supported with append. Pass a single os/arch to append (when --keep-manifest-list is not specified)
 		oc image append --from docker.io/library/busybox:latest --filter-by-os=linux/s390x --to myregistry.com/myimage:latest layer.tar.gz
+
+		# Add a new layer to a multi-architecture image for all the os/arch manifests when keep-manifest-list is specified
+		oc image append --from docker.io/library/busybox:latest --keep-manifest-list --to myregistry.com/myimage:latest layer.tar.gz
+
+		# Add a new layer to a multi-architecture image for all the os/arch manifests that is specified by the filter, while preserving the manifestlist
+		oc image append --from docker.io/library/busybox:latest --filter-by-os=linux/s390x --keep-manifest-list --to myregistry.com/myimage:latest layer.tar.gz
 
 	`)
 )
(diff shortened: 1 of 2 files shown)


pkg/cli/image/append/append.go
```diff
@@ -57,8 +57,10 @@ var (
 		add the --drop-history flag to remove information from the image about the system that
 		built the base image.
 
-		Images in manifest list format will automatically select an image that matches the current
-		operating system and architecture unless you use --filter-by-os to select a different image.
+		Images in manifest list format with keep-manifest-list specified will automatically append layers
+		to all sub manifests in the list unless filter-by-os is specified in which case the append will
+		only happen for the filtered manifests while preserving the manifestlist. If keep-manifest-list is
+		not specified, a single manifest is filtered and appended to depending on the filter.
```
flavianmissi, line 63: Do we want to keep the information about the automatic selection of an image when keep-manifest-list is not used? i.e, this part from the previous doc: 
```
automatically select an image that matches the operating system and architecture
```

pkg/cli/image/append/append.go
```diff
@@ -84,9 +86,14 @@ var (
...
 
 		# Add a new layer to a multi-architecture image for an os/arch that is different from the system's os/arch
-		# Note: Wildcard filter is not supported with append. Pass a single os/arch to append
 		oc image append --from docker.io/library/busybox:latest --filter-by-os=linux/s390x --to myregistry.com/myimage:latest layer.tar.gz
 
+		# Add a new layer to a multi-architecture image for all the os/arch manifests when keep-manifest-list is specified
+		oc image append --from docker.io/library/busybox:latest --keep-manifest-list --to myregistry.com/myimage:latest layer.tar.gz
+
+		# Add a new layer to a multi-architecture image for all the os/arch manifests that is specified by the filter, while preserving the manifestlist
+		oc image append --from docker.io/library/busybox:latest --filter-by-os=linux/s390x --keep-manifest-list --to myregistry.com/myimage:latest layer.tar.gz
```
flavianmissi, line 95: does `--filter-by-os` accepts multiple os/arch to be specified? The previous comment was very clear about it only accepting one, as well as not allowing wild cards. I think the new comment should be too.

flavianmissi (commented):
looks good overall

pkg/cli/image/append/append.go
Prashanth684: ah..yes..added that statement back.

soltysh (approved):
/approge

ardaguclu:
This PR has already been in merging phase, but just for the awareness there is an another PR https://github.com/openshift/oc/pull/1334 working on `--keep-manifest-list` flag and might have impact on `append` command.

pkg/cli/image/append/append.go
```diff
@@ -84,9 +87,15 @@ var (
 		oc image append --from-dir v2 --to myregistry.com/myimage:latest layer.tar.gz
 
 		# Add a new layer to a multi-architecture image for an os/arch that is different from the system's os/arch
-		# Note: Wildcard filter is not supported with append. Pass a single os/arch to append
+		# Note: Wildcard filter is not supported with append. Pass a single os/arch to append (when --keep-manifest-list is not specified)
```
Jamstah, line 90: Isn't the wildcard filter the default? Why do we say its not supported?

Jamstah:
I'm confiused about this PR. The append command seems to be designed to modify an image. If you modify an image, the digest of the manifest changes. If the digest of the manifest changes, the digest of the manifest list changes. If the digest of the manifest list changes, you cannot `keep-manifest-list`, it will always change.

What are we trying to achieve with this option that wasn't possible with just the filter option?

Jamstah:
Ah, ok, I see we have a --from and a --to, we're not updating a tag in place and we're not expecting to maintain the digest, just expecting to maintain the fact that it's a list. That's a little different to the definition in `oc image mirror` where `--keep-manifest-list` implies that we want that exact manifest list and it cannot be altered.

So, without keep manifest list, it could pick a single architecture image from the list, append to that, and push just that image to the --to. With keep manifest list, the --to will always be a list containing all the images in --from, but the append will only be applied to the image selected by the filter.

My final question then, is why does the doc say it only supports a single filtered image? Why wouldn't we support the '.*' filter? The code as it is written looks like it is capable of selecting more than one image to append to based on the filter.

Prashanth684:
> Ah, ok, I see we have a `--from` and a `--to`, we're not updating a tag in place and we're not expecting to maintain the digest, just expecting to maintain the fact that it's a list. That's a little different to the definition in #1334 where `--keep-manifest-list` implies that we want that exact manifest list and it cannot be altered. That might need reconciling, will have to think about it.
> 
> From `oc image mirror` help text, which implies that keep-manifest-list should maintain the digest:
> 
> > `flag.BoolVar(&o.KeepManifestList, "keep-manifest-list", o.KeepManifestList, "If an image is part of a manifest list, always mirror the list even if only one image is found. The default is to mirror the specific image unless unless --filter-by-os is passed. This flag is equivalent to setting --filter-by-os to '.*' since you cannot preserve the manifest list digest while filtering out any of the manifests included in the list.")`
> 
> But for this PR: Without `--keep-manifest-list`, it could pick a single architecture image from the list, append to that, and push just that image to the `--to`. With `--keep-manifest-list`, the `--to` will always be a list containing all the images in `--from`, and the append will only be applied to the image selected by the filter.
> 
> My final question then, is why does the doc say it only supports a single filtered image? Why wouldn't we support the `.*` filter? The code as it is written looks like it is capable of selecting more than one image to append to based on the filter.

The wildcard filter does not work in the case where `--keep-manifest-list` is not specified. This is the default behavior where it filters and appends to a single manifest.

Jamstah:
Got it, thanks :)

pkg/cli/image/append/append.go
Jamstah:
```suggestion
		# Note: The first image in the manifest list that matches the filter will be returned when --keep-manifest-list is not specified
```

Jamstah (commented):
I think this is more clear and accurate. The code will actually pick the first list if the filter matches more than one image.

PR #1362, opened 2023-03-07
Support OCI manifest lists in image mirror

Jamstah:
We only need to tell the registry we accept them, the registry client code already handles OCI manifests transparently for us.

### DIFF
--- a/pkg/cli/image/manifest/manifest.go
+++ b/pkg/cli/image/manifest/manifest.go
@@ -214,6 +214,7 @@ type FilterFunc func(*manifestlist.ManifestDescriptor, bool) bool
 // PreferManifestList specifically requests a manifest list first
 var PreferManifestList = distribution.WithManifestMediaTypes([]string{
 	manifestlist.MediaTypeManifestList,
+	imagespecv1.MediaTypeImageIndex,
 	schema2.MediaTypeManifest,
 	imagespecv1.MediaTypeImageManifest,
 })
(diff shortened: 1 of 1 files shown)


pkg/cli/image/manifest/manifest.go
```diff
@@ -214,6 +214,7 @@ type FilterFunc func(*manifestlist.ManifestDescriptor, bool) bool
 // PreferManifestList specifically requests a manifest list first
 var PreferManifestList = distribution.WithManifestMediaTypes([]string{
 	manifestlist.MediaTypeManifestList,
+	imagespecv1.MediaTypeImageIndex,
```
ardaguclu, line 217: AFAICS, `imagespecv1.MediaTypeImageManifest` was added to support OCI manifests in [here](https://github.com/openshift/oc/pull/697/files#diff-88a19aa652d0dad2300f5b2080f14494546c5036b5b357e07ae150e24d67c072R216), what is the reason of adding this one?

ardaguclu (commented):
Thanks for this patch. I have just one question, other than that;

/lgtm

pkg/cli/image/manifest/manifest.go
ardaguclu: ah, I missed that sorry. You're right.

Jamstah: I should have titled it right :)

PR #1390, opened 2023-03-29
OCPBUGS-10879: Fix deprecated oc command suggestion

ardaguclu:
Currently when user runs `oc status --suggest` if misconfigured registry is detected, `oc` suggests to run `oc adm registry -h` command. However, this command is deprecated and removed, this PR instead suggests correct command `oc registry -h`.


ardaguclu:
@ingvagabund could you PTAL this trivial fix when you have a chance?. Thanks.

PR #1393, opened 2023-03-30
OCPBUGS-10843: oc debug unique pod name

ardaguclu:
`oc debug` command uses same deterministic pod name in every execution(if matching pod name exists, it deletes this pod and re-creates it). 
Currently, some CI jobs rely on executing `oc debug` multiple times in a short period
of time. When it happens, sometimes informers might return previous container's
terminated state(although genuine container hasn't started yet). This causes debug command fails intermittently.

This might be considered an issue in informers but wouldn't it be also better to add suffix(just how [kubectl debug does](https://github.com/kubernetes/kubernetes/blob/d89d5ab2680bc74fe4487ad71e514f4e0812d9ce/staging/src/k8s.io/kubectl/pkg/cmd/debug/debug.go#L584)) to set apart debug pods from each other. If user still wants to attach same container, he/she can pass `--preserve-pod` flag and attach it to via `attach` command.

In addition to that this PR prevents deletion of temporary namespace if the `--preserve-pod` flag is
passed to keep debug pod in cluster.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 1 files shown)


ardaguclu:
e2e-aws-ovn failure is real and this test https://github.com/openshift/origin/blob/98d9c70b162c863f3dd6856ded4d214cf9aca977/test/extended/cli/debug.go#L51 needs to be changed, if we decide to move forward with this patch.

/hold

soltysh (approved):
/lgtm

PR #1400, opened 2023-04-10
replace trimLeft with trimPrefix

airycanon:
replaces trimLeft function with the more appropriate trimPrefix to avoid potential bugs.
for example:
```golang
// will output ello.test
fmt.Println(strings.TrimLeft("hello.test", "http://"))
```


airycanon:
@dmage I have recently rebased the code. Could you please take a look and add the label if appropriate? Thanks.

PR #1402, opened 2023-04-11
AUTH-355: Add OAuth2 Authorization Code Grant Flow for login

liouk:
This PR reopens #1031 to add OAuth2 Authorization Code Grant Flow for `oc login`, adding minor improvements to the original PR. This can be invoked with `oc login --web`.

Requires
- https://github.com/openshift/cluster-authentication-operator/pull/606
- https://github.com/openshift/oauth-server/pull/121
- https://github.com/openshift/oauth-server/pull/128


liouk:
@stlaz: would we need a timeout while waiting for the authentication/authorization to happen via the browser?

stlaz:
> oc login --web the user logs in the openshift console on the browser and thus authorizes the code grant; this also means that as long as the user is logged in the openshift console, there won't be a prompt for login, so subsequent invocations of oc login --web will result in automatic grants. If one needs to login with a different user, they would first need to logout from the openshift console -- is this desired/acceptable?

I believe this should be fine - two users would not generally share the same cookie cache.

> would we need a timeout while waiting for the authentication/authorization to happen via the browser?

The wait should be on the client side so I think we can just keep the localhost server running as long as necessary.

> would it be a good idea to print the browser URL on the console when oc login --web is invoked, in case the user needs to manually open it? (e.g. automatically opening it didn't work, user accidentally closed tab)

definitely

liouk:
/hold

PR commits need squashing before merging

liouk:
Changed to WIP in order to incorporate https://issues.redhat.com/browse/AUTH-376

liouk:
Rebase requires:
- https://github.com/openshift/library-go/pull/1531

liouk:
Marked as WIP due to https://github.com/openshift/oc/pull/1402#issuecomment-1551144640

liouk:
@stlaz, I've added the changes of https://github.com/openshift/library-go/pull/1531 manually as a proof of functionality until the rebase is possible.

vendor/github.com/pkg/browser/LICENSE
```diff
@@ -0,0 +1,23 @@
+Copyright (c) 2014, Dave Cheney <dave@cheney.net>
```
stlaz, line 1: remove vendor changes from this commit

pkg/cli/login/loginoptions.go
```diff
@@ -265,7 +265,7 @@ func (o *LoginOptions) gatherAuthInfo() error {
 	clientConfig.CertFile = o.CertFile
 	clientConfig.KeyFile = o.KeyFile
 
-	token, err := tokenrequest.RequestToken(o.Config, o.getAuthChallengeHandler())
+	token, err := tokenrequest.RequestTokenWithChallengeHandlers(o.Config, o.getAuthChallengeHandler())
```
stlaz: move these changes outside the vendor commit

pkg/helpers/tokencmd/oauth_callback.go
```diff
@@ -0,0 +1,61 @@
+package tokencmd
+
+import (
+	"net/http"
+
+	"github.com/RangelReale/osincli"
```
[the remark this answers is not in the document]
liouk: @stlaz is this maybe relevant for [library-go#1531](https://github.com/openshift/library-go/pull/1531/files#diff-f30259f7a13841a226555f2801678f096da8bb80456cf414201b8cfd647da728R16)?

pkg/cli/admin/catalog/mirror.go
```diff
@@ -27,8 +27,8 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"github.com/alicebob/sqlittle"
-	"github.com/docker/distribution"
-	"github.com/docker/distribution/manifest/manifestlist"
+	"github.com/distribution/distribution/v3"
```
stlaz, line 30: Is this change necessary? I have a feeling this may break things

stlaz: nvm, this apparently comes from https://github.com/openshift/library-go/pull/1529/commits

liouk: This was introduced recently in library-go: https://github.com/openshift/library-go/pull/1529

And unfortunately it's necessary wherever relevant library-go code is used :confused:

pkg/cli/login/loginoptions.go
```diff
@@ -265,10 +269,21 @@ func (o *LoginOptions) gatherAuthInfo() error {
 	clientConfig.CertFile = o.CertFile
 	clientConfig.KeyFile = o.KeyFile
 
-	token, err := tokenrequest.RequestToken(o.Config, o.getAuthChallengeHandler())
+	var token string
+	if o.WebLogin {
+		urlHandler := func(u *url.URL) error {
+			strUrl := u.String()
```
stlaz, line 275: `loginURL` ?

liouk: Fixed here 01dc4ce9ee16c1da8aa6caef6134d5ced4733fbc

liouk:
/retest-required
infrastructure issues

pkg/cli/login/login.go
```diff
@@ -83,6 +87,8 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra
 	cmds.Flags().StringVarP(&o.Username, "username", "u", o.Username, "Username for server")
 	cmds.Flags().StringVarP(&o.Password, "password", "p", o.Password, "Password for server")
 
+	cmds.Flags().BoolVarP(&o.WebLogin, "web", "w", o.WebLogin, "Login with web browser")
```
ardaguclu, line 90: As being asked by @atiratree, what is the expected behavior of this flag against 4.13 server?

liouk: @ardaguclu @atiratree this flag needs the server to have a specific default OAuth client (`openshift-cli-client`) and also the oauth server to have the latest OSIN library; i.e., it needs this code, which will only exist on 4.14 onwards:
- https://github.com/openshift/cluster-authentication-operator/pull/606
- https://github.com/openshift/oauth-server/pull/121
- https://github.com/openshift/oauth-server/pull/128

Any attempt to invoke `oc login --web` against a pre 4.14 server will result on an error on the browser, as the server won't have the necessary oauth client with which this flow proceeds:
```json
{"error":"invalid_request","error_description":"The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed."}
```

atiratree:
@liouk opened a rebase of the dependent PR to pull in the library-go changes: https://github.com/openshift/oc/pull/1495

pkg/cli/login/login.go
atiratree: I would not block on this, but I think it would be better from UX perspective if the user got a preemptive error in the CLI that the web authentication is not possible. It could use the same mechanism as we use in 

https://github.com/openshift/oc/blob/master/pkg/cli/login/loginoptions.go#L292

ardaguclu:
Nice feature!
/lgtm

PR #1403, opened 2023-04-14
OCPBUGS-11652: Extend adm node-logs to new API

aravindhp:
- Map --tail internally to tailLines
- Map --unit internally to query
- Use service instead of journal in the help to normalize across Linux  and Windows

### DIFF
--- a/pkg/cli/admin/node/logs.go
+++ b/pkg/cli/admin/node/logs.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
-
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
@@ -72,8 +71,8 @@ type LogsOptions struct {
 	Boot              int
 	BootChanaged      bool
 	Units             []string
-	SinceTime         string
-	UntilTime         string
+	Since             string
+	Until             string
 	Tail              int
 	Output            string
 
@@ -112,21 +111,19 @@ func NewCmdLogs(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 		},
 	}
 
-	cmd.Flags().StringVar(&o.Path, "path", o.Path, "Retrieve the specified path within the node's /var/logs/ folder. The 'journal' value will allow querying the journal on supported operating systems.")
-
-	cmd.Flags().StringSliceVarP(&o.Units, "unit", "u", o.Units, "Return log entries from the specified unit(s). Only applies to node journal logs.")
-	cmd.Flags().StringVarP(&o.Grep, "grep", "g", o.Grep, "Filter log entries by the provided regex pattern. Only applies to node journal logs.")
+	cmd.Flags().StringVar(&o.Path, "path", o.Path, "Retrieve the specified path within the node's /var/logs/ folder. The 'journal' value will allow querying the services on supported operating systems.")
+	cmd.Flags().StringSliceVarP(&o.Units, "unit", "u", o.Units, "Return log entries from the specified services(s) Only applies to node service logs.")
+	cmd.Flags().StringVarP(&o.Grep, "grep", "g", o.Grep, "Filter log entries by the provided regex pattern. Only applies to node service logs.")
 	cmd.Flags().BoolVar(&o.GrepCaseSensitive, "case-sensitive", o.GrepCaseSensitive, "Filters are case sensitive by default. Pass --case-sensitive=false to do a case insensitive filter.")
-	cmd.Flags().StringVar(&o.SinceTime, "since", o.SinceTime, "Return logs after a specific ISO timestamp or relative date. Only applies to node journal logs.")
-	cmd.Flags().StringVar(&o.UntilTime, "until", o.UntilTime, "Return logs before a specific ISO timestamp or relative date. Only applies to node journal logs.")
-	cmd.Flags().IntVar(&o.Boot, "boot", o.Boot, " Show messages from a specific boot. Use negative numbers, allowed [-100, 0], passing invalid boot offset will fail retrieving logs. Only applies to node journal logs.")
-	cmd.Flags().StringVarP(&o.Output, "output", "o", o.Output, "Display journal logs in an alternate format (short, cat, json, short-unix). Only applies to node journal logs.")
-	cmd.Flags().IntVar(&o.Tail, "tail", o.Tail, "Return up to this many lines (not more than 100k) from the end of the log. Only applies to node journal logs.")
-
+	cmd.Flags().StringVar(&o.Since, "since", o.Since, "Return logs after a specific ISO timestamp or relative date. Only applies to node service logs.")
+	cmd.Flags().StringVar(&o.Until, "until", o.Until, "Return logs before a specific ISO timestamp or relative date. Only applies to node service logs.")
+	cmd.Flags().IntVar(&o.Boot, "boot", o.Boot, " Show messages from a specific boot. Use negative numbers, allowed [-100, 0], passing invalid boot offset will fail retrieving logs. Only applies to node service logs.")
+	cmd.Flags().StringVarP(&o.Output, "output", "o", o.Output, "Display service logs in an alternate format (short, cat, json, short-unix). Only applies to node service logs.")
+	cmd.Flags().IntVar(&o.Tail, "tail", o.Tail, "Return up to this many lines (not more than 100k) from the end of the log. Only applies to node service logs.")
 	cmd.Flags().StringVar(&o.Role, "role", o.Role, "Set a label selector by node role.")
 	cmd.Flags().StringVarP(&o.Selector, "selector", "l", o.Selector, "Selector (label query) to filter on.")
 	cmd.Flags().BoolVar(&o.Raw, "raw", o.Raw, "Perform no transformation of the returned data.")
-	cmd.Flags().BoolVar(&o.Unify, "unify", o.Unify, "Interleave logs by sorting the output. Defaults on when viewing node journal logs.")
+	cmd.Flags().BoolVar(&o.Unify, "unify", o.Unify, "Interleave logs by sorting the output. Defaults on when viewing node service logs.")
 
 	return cmd
 }
@@ -192,7 +189,7 @@ type logRequest struct {
 	skipPrefix bool
 }
 
-// WriteTo prefixes the error message with the current node if necessary
+// WriteRequest prefixes the error message with the current node if necessary
 func (req *logRequest) WriteRequest(out io.Writer) error {
 	if req.err != nil {
 		return req.err
@@ -259,11 +256,11 @@ func (o LogsOptions) RunLogs() error {
 			SetHeader("Accept", "text/plain, */*").
 			SetHeader("Accept-Encoding", "gzip")
 		if o.Path == "journal" {
-			if len(o.UntilTime) > 0 {
-				req.Param("until", o.UntilTime)
+			if len(o.Until) > 0 {
+				req.Param("until", o.Until)
 			}
-			if len(o.SinceTime) > 0 {
-				req.Param("since", o.SinceTime)
+			if len(o.Since) > 0 {
+				req.Param("since", o.Since)
 			}
 			if len(o.Output) > 0 {
 				req.Param("output", o.Output)
@@ -273,15 +270,21 @@ func (o LogsOptions) RunLogs() error {
 			}
 			if len(o.Units) > 0 {
 				for _, unit := range o.Units {
+					// Needed to allow working with kubelet that does not support query
 					req.Param("unit", unit)
+					req.Param("query", unit)
 				}
 			}
 			if len(o.Grep) > 0 {
+				// Needed to allow working with kubelet that does not support query
 				req.Param("grep", o.Grep)
+				req.Param("pattern", o.Grep)
 				req.Param("case-sensitive", fmt.Sprintf("%t", o.GrepCaseSensitive))
 			}
 			if o.Tail > 0 {
+				// Needed to allow working with kubelet that does not support query
 				req.Param("tail", strconv.Itoa(o.Tail))
+				req.Param("tailLines", strconv.Itoa(o.Tail))
 			}
 		}
 


pkg/cli/admin/node/logs.go
```diff
@@ -258,37 +287,44 @@ func (o LogsOptions) RunLogs() error {
...
+			} else if len(o.Units) > 0 {
 				for _, unit := range o.Units {
-					req.Param("unit", unit)
+					req.Param("query", unit)
 				}
 			}
 			if len(o.Grep) > 0 {
-				req.Param("grep", o.Grep)
+				req.Param("pattern", o.Grep)
 				req.Param("case-sensitive", fmt.Sprintf("%t", o.GrepCaseSensitive))
```
ardaguclu, line 314: Do we use `case-sensitive` flag in new API?. For example, when we deprecate `grep` and starts using `pattern` in next release, can we also deprecate `case-sensitive`?

ardaguclu (commented):
@aravindhp thanks for the PR. I left a few comments.

pkg/cli/admin/node/logs.go
```diff
@@ -112,28 +122,35 @@ func NewCmdLogs(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
...
 
 func (o *LogsOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
 	if !cmd.Flags().Lookup("unify").Changed {
-		o.Unify = o.Path == "journal"
+		o.Unify = o.Path == "journal" || len(o.Query) > 0
+	}
+
+	if len(o.Query) > 0 {
+		// We don't need to add the special "journal" path if in query mode
+		o.Path = ""
```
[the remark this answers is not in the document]
aravindhp: I did not want to modify struct items in `Validate()`. Moreover `if len(o.Query) > 0 && (len(o.Path) > 0 || len(o.Units) > 0) ` will always catch if I don't reset it here. So I can't do it after 193.

pkg/cli/admin/node/logs.go
```diff
@@ -258,37 +287,44 @@ func (o LogsOptions) RunLogs() error {
 		req := client.Get().RequestURI(path).
 			SetHeader("Accept", "text/plain, */*").
 			SetHeader("Accept-Encoding", "gzip")
-		if o.Path == "journal" {
-			if len(o.UntilTime) > 0 {
-				req.Param("until", o.UntilTime)
+		if o.Path == "journal" || len(o.Query) > 0 {
+			if len(o.Until) > 0 {
```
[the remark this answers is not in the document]
aravindhp: It is a bit more difficult to convert `Since` to `SinceTime` and `Until` to `UntilTime` as `Since` and `Until` could be a ISO date or a relative time. i.e. `YYYY-MM-DD HH:MM:SS` or `(+|-)[0-9]+(s|m|h|d)`. So the relative time to ISO 3339 format is especially tricky and I wanted to avoid that. So my suggestion we keep them separate and mark `Since` and `Until` for deprecation in the future.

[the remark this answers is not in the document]
ardaguclu: Yes, type incompatibility between `Since` and `SinceTime` sounds good to me, but what will happen if any CI script relies on `Since`;

```sh
$ oc adm node-logs --since
```
will it continue working?

pkg/cli/admin/node/logs.go
ardaguclu: Thanks

pkg/cli/admin/node/logs.go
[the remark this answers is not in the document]
aravindhp: Yes that will continue. Please see line 294.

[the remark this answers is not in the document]
aravindhp: Fixed.

pkg/cli/admin/node/logs.go
```diff
@@ -173,6 +190,18 @@ func (o LogsOptions) Validate() error {
...
 		return fmt.Errorf("--boot accepts values [-100, 0]")
 	}
+	if len(o.Query) > 0 && (len(o.Path) > 0 || len(o.Units) > 0) {
+		return fmt.Errorf("--path and --unit flags are mutually exclusive with --query")
+	}
+	if len(o.SinceTime) > 0 && (len(o.Since) > 0 || len(o.Until) > 0) {
+		return fmt.Errorf("cannot combine --since-time with --since or --until")
+	}
+	if len(o.SinceTime) > 0 && (len(o.Since) > 0 || len(o.Until) > 0) {
+		return fmt.Errorf("cannot combine --since-time with --since or --until")
```
ardaguclu, line 200: Can we also change this in the format of https://github.com/openshift/oc/pull/1403/files#diff-fe5df7f76dde8b46f44cbcfa239451ec304c914cfc3d6a17087debb317079ee0R194

pkg/cli/admin/node/logs.go
```diff
@@ -173,6 +190,18 @@ func (o LogsOptions) Validate() error {
 	if o.BootChanaged && (o.Boot < -100 || o.Boot > 0) {
 		return fmt.Errorf("--boot accepts values [-100, 0]")
 	}
+	if len(o.Query) > 0 && (len(o.Path) > 0 || len(o.Units) > 0) {
+		return fmt.Errorf("--path and --unit flags are mutually exclusive with --query")
+	}
+	if len(o.SinceTime) > 0 && (len(o.Since) > 0 || len(o.Until) > 0) {
+		return fmt.Errorf("cannot combine --since-time with --since or --until")
```
ardaguclu, line 197: Can we also change this in the format of https://github.com/openshift/oc/pull/1403/files#diff-fe5df7f76dde8b46f44cbcfa239451ec304c914cfc3d6a17087debb317079ee0R194?

pkg/cli/admin/node/logs.go
```diff
@@ -173,6 +190,18 @@ func (o LogsOptions) Validate() error {
...
+	}
+	if len(o.SinceTime) > 0 && (len(o.Since) > 0 || len(o.Until) > 0) {
```
ardaguclu, line 199: I think this should be `len(o.UntilTime)`?

ardaguclu (commented):
If the user sets `query`, we are setting `path` to empty string. But we are generating Path by using `path` flag in here https://github.com/openshift/oc/blob/a65a29a57e4d213df00e2ee1fa3641dc19912e3b/pkg/cli/admin/node/logs.go#L253, will it work for the new API?

pkg/cli/admin/node/logs.go
aravindhp: Done

pkg/cli/admin/node/logs.go
aravindhp: Sorry. Fixed.

pkg/cli/admin/node/logs.go
aravindhp: Done.

aravindhp (commented):
> If the user sets `query`, we are setting `path` to empty string. But we are generating Path by using `path` flag in here
> 
> https://github.com/openshift/oc/blob/a65a29a57e4d213df00e2ee1fa3641dc19912e3b/pkg/cli/admin/node/logs.go#L253
> , will it work for the new API?

Yes that will still work. Setting `path` to `journal` means the API call will be `proxy/logs/journal?options=`. With `query` it will be `proxy/logs/?query=...` and that works.

aravindhp:
/test e2e-agnostic-ovn-cmd

The [failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1403/pull-ci-openshift-oc-master-e2e-agnostic-ovn-cmd/1650568692833980416) is in `[sig-instrumentation] Prometheus [apigroup:image.openshift.io] when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured` and has nothing to do with the changes in this PR.

aravindhp:
/retest-required

e2e-aws-ovn-serial [failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1403/pull-ci-openshift-oc-master-e2e-aws-ovn-serial/1650568693001752576) in `[sig-arch] events should not repeat pathologically` which has nothing to do with this PR

ardaguclu:
According to the pre-merging tests, this PR is compatible with older server versions. I think, we can merge it especially considering that current `oc adm node-logs` is not working already.

/lgtm
/hold
feel free to cancel hold tag

pkg/cli/admin/node/logs.go
```diff
@@ -112,28 +121,35 @@ func NewCmdLogs(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 		},
 	}
 
-	cmd.Flags().StringVar(&o.Path, "path", o.Path, "Retrieve the specified path within the node's /var/logs/ folder. The 'journal' value will allow querying the journal on supported operating systems.")
-
-	cmd.Flags().StringSliceVarP(&o.Units, "unit", "u", o.Units, "Return log entries from the specified unit(s). Only applies to node journal logs.")
-	cmd.Flags().StringVarP(&o.Grep, "grep", "g", o.Grep, "Filter log entries by the provided regex pattern. Only applies to node journal logs.")
+	cmd.Flags().StringVar(&o.Path, "path", o.Path, "Retrieve the specified path within the node's /var/logs/ folder. The 'journal' value will allow querying the services on supported operating systems.")
```
soltysh, line 124: We'll most likely need to translate in the future `--path` value to `query` with `/var/logs` prepended to the value. But you'll need to support that in upstream first (see my slack comment).

pkg/cli/admin/node/logs.go
```diff
@@ -112,28 +121,35 @@ func NewCmdLogs(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
...
+	cmd.Flags().StringSliceVarP(&o.Query, "query", "q", o.Units, "Specifies services(s) or files from which to return logs")
```
soltysh, line 125: Drop as mentioned above.

pkg/cli/admin/node/logs.go
```diff
@@ -112,28 +121,35 @@ func NewCmdLogs(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
...
-	cmd.Flags().IntVar(&o.Boot, "boot", o.Boot, " Show messages from a specific boot. Use negative numbers, allowed [-100, 0], passing invalid boot offset will fail retrieving logs. Only applies to node journal logs.")
-	cmd.Flags().StringVarP(&o.Output, "output", "o", o.Output, "Display journal logs in an alternate format (short, cat, json, short-unix). Only applies to node journal logs.")
-	cmd.Flags().IntVar(&o.Tail, "tail", o.Tail, "Return up to this many lines (not more than 100k) from the end of the log. Only applies to node journal logs.")
-
+	cmd.Flags().StringVar(&o.Since, "since", o.Since, "Return logs after a specific ISO timestamp or relative date. Only applies to node service logs.")
+	cmd.Flags().StringVar(&o.SinceTime, "since-time", o.Since, "Return logs after a specific RFC3339 timestamp. Only applies to node service logs.")
+	cmd.Flags().StringVar(&o.Until, "until", o.Until, "Return logs before a specific ISO timestamp or relative date. Only applies to node service logs.")
+	cmd.Flags().StringVar(&o.UntilTime, "until-time", o.Since, "Return logs until a specific RFC3339 timestamp. Only applies to node service logs.")
```
soltysh, line 132: Drop `--since-time` and `--until-time` they are not needed, we should handle the translation in oc to use either of the two.

pkg/cli/admin/node/logs.go
aravindhp: I don't follow this. `query` does support taking in a path in upstream.

pkg/cli/admin/node/logs.go
aravindhp: Will do.

pkg/cli/admin/node/logs.go
aravindhp: Like I mentioned in Slack, converting `since` to `sinceTime` and `until` to `untilTime` is a bit tough. since and until can be both a relative time or ISO time stamp. `sinceTime` and `untilTime` only supports RFC3339 timestamps. I will drop `--since-time` and `--until-time` for now.

aravindhp (commented):
Thanks for the review, @soltysh. I will address your requests shortly.

aravindhp:
@soltysh I have addressed your comments. Please review.

aravindhp:
/retest-required

Cluster failed to come up [e2e-aws-ovn-upgrade](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1403/pull-ci-openshift-oc-master-e2e-aws-ovn-upgrade/1656056990503276544)

pkg/cli/admin/node/logs.go
```diff
@@ -273,15 +273,21 @@ func (o LogsOptions) RunLogs() error {
 			}
 			if len(o.Units) > 0 {
 				for _, unit := range o.Units {
+					// Needed to allow working with kubelet that does not support query
 					req.Param("unit", unit)
+					req.Param("query", unit)
```
ardaguclu, line 278: Shouldn't we have a some kind of transformation instead directly assigning?

pkg/cli/admin/node/logs.go
```diff
@@ -112,21 +114,19 @@ func NewCmdLogs(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
...
-	cmd.Flags().StringVar(&o.SinceTime, "since", o.SinceTime, "Return logs after a specific ISO timestamp or relative date. Only applies to node journal logs.")
-	cmd.Flags().StringVar(&o.UntilTime, "until", o.UntilTime, "Return logs before a specific ISO timestamp or relative date. Only applies to node journal logs.")
-	cmd.Flags().IntVar(&o.Boot, "boot", o.Boot, " Show messages from a specific boot. Use negative numbers, allowed [-100, 0], passing invalid boot offset will fail retrieving logs. Only applies to node journal logs.")
-	cmd.Flags().StringVarP(&o.Output, "output", "o", o.Output, "Display journal logs in an alternate format (short, cat, json, short-unix). Only applies to node journal logs.")
-	cmd.Flags().IntVar(&o.Tail, "tail", o.Tail, "Return up to this many lines (not more than 100k) from the end of the log. Only applies to node journal logs.")
-
+	cmd.Flags().StringVar(&o.Since, "since", o.Since, "Return logs after a specific ISO timestamp or relative date. Only applies to node service logs.")
```
ardaguclu, line 121: `SinceTime` and `UntilTime` can not be removed in any case, because we haven't deprecated them yet

aravindhp: Ah you are right. I keep mixing up `--since` and `--since-time`. Sorry about that.

pkg/cli/admin/node/logs.go
aravindhp: No transformation is required as such.

pkg/cli/admin/node/logs.go
aravindhp: I looked more closely at this. The cli API is staying at `--since` and `--until`. It is only the internal struct variables that are being renamed to `Since` and `Until` to match the cli. Isn't okay given we will be introducing `--since-time` and `--until-time` in a future release and will have to support both for a while until we deprecate `--since` and `--until`?

ardaguclu: ah, sorry you are right

pkg/cli/admin/node/logs.go
```diff
@@ -54,6 +53,9 @@ var (
 
 		# Display cron log file from all masters
 		oc adm node-logs --role master --path=cron
+
+		# Display kubelet log from worker nodes using the new query sub-command
+		oc adm node-logs --role worker --query=kubelet
```
zhouying7780, line 58: Now , we can't support query, please remove this example.

aravindhp: Sorry about that. I have now removed it.

ardaguclu:
@aravindhp  What are we waiting for this PR?, is it requiring a new round of review?

ardaguclu:
I think, we are waiting for the last review from @soltysh;
/test e2e-metal-ipi-ovn-ipv6

ardaguclu:
@aravindhp I have one last question just for clarification;
will this https://github.com/openshift/origin/blob/6aeee601dea8ecb548c251f6a457ad66f0e44e77/pkg/monitor/nodedetails/node_log.go#L37 continue working after this PR?

aravindhp:
Yes the relative time options for `--since` and `--until` will continue to work. QE and I have manually have verified this.

PR #1406, opened 2023-04-18
OCPBUGS-12143: oc login: unwrap tls.CertificateVerificationError to use x509 errors

ardaguclu:
Currently, `oc login` has some logic when user gets x509 certificate errors.
However in Go 1.20(with [this commit](https://go-review.googlesource.com/c/go/+/449336)) and upward versions, these certificate errors are wrapped in
`tls.CertificateVerificationError`(https://pkg.go.dev/crypto/tls#CertificateVerificationError). In order to keep the same functionality,
`oc login` needs to detect this type and extract correct error.
Furthermore, since this new type has been only added into Go 1.20,
we also need to bump `oc` to 1.20. 

This PR is also prerequisite for the openshift/kubernetes
1.27 vendoring of `oc`.

### DIFF
--- a/pkg/cli/login/loginoptions.go
+++ b/pkg/cli/login/loginoptions.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -22,7 +23,7 @@ import (
 	kterm "k8s.io/kubectl/pkg/util/term"
 
 	projectv1typedclient "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
-	"github.com/openshift/oc/pkg/helpers/errors"
+	ocerrors "github.com/openshift/oc/pkg/helpers/errors"
 	cliconfig "github.com/openshift/oc/pkg/helpers/kubeconfig"
 	"github.com/openshift/oc/pkg/helpers/motd"
 	"github.com/openshift/oc/pkg/helpers/project"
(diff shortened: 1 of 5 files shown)


ardaguclu:
This PR also covers this automatic bump https://github.com/openshift/oc/pull/1405 as well as fixing a change in 1.20.

ardaguclu:
Even if this PR is approved, we need to wait unblock signal from TRT to merge this. That's why;

/hold

pkg/cli/login/loginoptions.go
```diff
@@ -139,6 +139,12 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 	// try to TCP connect to the server to make sure it's reachable, and discover
 	// about the need of certificates or insecure TLS
 	if err := dialToServer(*clientConfig); err != nil {
+		// In go 1.20 and upwards versions, x509 errors in the switch statement
+		// are wrapped in tls.CertificateVerificationError.
+		if cerr, ok := err.(*tls.CertificateVerificationError); ok {
```
stlaz, line 144: errors.As()

pkg/helpers/tokencmd/request_token.go
```diff
@@ -461,7 +461,7 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 		}
 		return systemRootsRT, nil
 	case x509.UnknownAuthorityError, x509.HostnameError, x509.CertificateInvalidError, x509.SystemRootsError,
-		tls.RecordHeaderError, *net.OpError:
```
stlaz, line 464: this switch case is checking for specific verification errors. For now, we could just unwrap the error before the switch I guess

pkg/cli/login/loginoptions.go
```diff
@@ -22,7 +23,7 @@ import (
 	kterm "k8s.io/kubectl/pkg/util/term"
 
 	projectv1typedclient "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
-	"github.com/openshift/oc/pkg/helpers/errors"
+	errs "github.com/openshift/oc/pkg/helpers/errors"
```
stlaz, line 26: nit: `ocerrors` would've been clearer

ardaguclu: Done

soltysh (approved):
/lgtm

PR #1414, opened 2023-04-24
Clarify the use of the filter without keep-manifest-list

Jamstah:
There is no way to guarantee a single arch has been specified because the filter is a regex. This clarification explains the actual behaviour.


ardaguclu:
Thanks for the clarification

/ok-to-test
/lgtm

Jamstah:
Rebased for merge conflicts.

PR #1415, opened 2023-04-26
OCPBUGS-12793: adds a mapping for exposed ports to DockerConfig when manifest listed

prb112:
Discussed with @Prashanth684 and @pravin-dsilva 

Prashanth identified the code is unable to see the manifest listed config. 
We pick off the first entry in the manifest, and pick the exposed port. 

Added mappings per discussion with @Prashanth684

### DIFF
--- a/pkg/helpers/newapp/app/dockerimagelookup.go
+++ b/pkg/helpers/newapp/app/dockerimagelookup.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"sort"
@@ -331,6 +332,19 @@ func (s ImageImportSearcher) Search(precise bool, terms ...string) (ComponentMat
 		}
 
 		dockerImage, ok := image.Image.DockerImageMetadata.Object.(*dockerv10.DockerImage)
+
+		// When a ManifestListed Image is used the nested DockerConfig objects in a DockerImageManifest are not propagated
+		// The first entry of the Manifest List entry's config is assigned to the top-level DockerImage
+		if ok && image.Manifests != nil && len(image.Manifests) != 0 && dockerImage.Config == nil {
+			conf := image.Manifests[0].DockerImageConfig
+			dc := &docker10.DockerImage{}
+			if err := json.Unmarshal([]byte(conf), &dc); err != nil {
+				// Errors do not cause issues in unmarshalling, and it is intentional, and logged out.
+				klog.V(4).Infof("unmarshalling failed for manifest listed image's dockerConfig: %#v %#v", conf, err)
+			} else {
+				dockerImage.Config = dc.Config
+			}
+		}
 		if !ok {
 			continue
 		}


coreydaley:
/approve
/assign @nalind 
for lgtm

pkg/helpers/newapp/app/dockerimagelookup.go
```diff
@@ -265,6 +267,46 @@ func (s ImageImportSearcher) Type() string {
...
+		return arx, nil
+	}
+	return nil, errors.New("value does not exist")
+}
+
+func selectFromMapStruct(dc interface{}, field string) (map[string]struct{}, error) {
+	ep, ok := dc.(map[string]interface{})[field]
+	if ok {
+		exs := map[string]struct{}{}
+		for ex := range ep.(map[string]interface{}) {
```
nalind, line 302: When `ep` is `nil`, this panics.  For example, when I used crc 2.17.0+44e15711 and ran `oc new-app --import-mode=PreserveOriginal --search php`.

nalind:
This appears to be looking in the right places for the information, but the unchecked type assertions can cause panics in unexpected places, so I'd like to see more of the type assertions being checked.

pkg/helpers/newapp/app/dockerimagelookup.go
```diff
@@ -331,6 +373,65 @@ func (s ImageImportSearcher) Search(precise bool, terms ...string) (ComponentMat
...
+			if len(conf) != 0 {
+				var dockerConfig map[string]interface{}
+				if err := json.Unmarshal([]byte(conf), &dockerConfig); err != nil {
+					errs = append(errs, err)
+				}
+
+				dc, ok := dockerConfig["config"]
+				if ok {
+					// Fields match the v1 spec https://github.com/moby/moby/blob/master/image/spec/v1.md
+					if in, err := selectFromInt(dc, "Memory"); err == nil {
```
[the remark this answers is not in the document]
nalind: Zero values might as well not be there, true.  If at some point there are defaults we don't want to zero out when the config blob doesn't include a specific field, they can be brought back.

The config struct we're digging into appears to be a field of the `github.com/docker/docker/api/types/container.Config` type in a `github.com/docker/docker/image.V1Image` embedded in a `github.com/docker/docker/image.Image`, and in contrast to the spec, some of the fields we're looking for aren't in the struct definition.  Of course, it's always possible that the JSON we get doesn't exactly comply with the specifications.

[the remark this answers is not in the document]
Prashanth684: Indeed . It looks like  the internal representation of the docker config has the additional fields: https://github.com/openshift/api/blob/master/image/docker10/types_docker.go#L32 and so does the dockerclient: https://github.com/fsouza/go-dockerclient/blob/main/container.go#L250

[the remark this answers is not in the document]
prb112: Right. I opted not to go for everything, and use a smaller mapping based on common usage. Are you suggesting we map all of these?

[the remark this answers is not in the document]
nalind: Those two are structs that get passed to APIs which create containers at run-time.  The presence of both a `VolumesFrom` field and an `OnBuild` field in both of those definitions is kind of weird.

pkg/helpers/newapp/app/dockerimagelookup.go
prb112: Added TypeOf Guards

pkg/helpers/newapp/app/dockerimagelookup.go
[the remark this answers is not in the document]
prb112: Used the struct from docker10/types_docker.go

prb112:
@nalind @Prashanth684 updated per your comments

pkg/helpers/newapp/app/dockerimagelookup.go
nalind: Not a blocker, since the above example no longer crashes, but using the `reflect` package to avoid panicking seems like a lot.  It would be enough to check the type assertions (per the last paragraph in [the section on type assertions in the language reference](https://go.dev/ref/spec#Type_assertions)).

prb112: Sure. I just went heavy weight. If you'd like, I can change away from reflect.

prb112:
Changed per comment on type assertions.

prb112:
Hi @nalind @Prashanth684 if you can review when you have a moment. Thank you, Paul

pkg/helpers/newapp/app/dockerimagelookup.go
```diff
@@ -331,6 +445,15 @@ func (s ImageImportSearcher) Search(precise bool, terms ...string) (ComponentMat
 		}
 
 		dockerImage, ok := image.Image.DockerImageMetadata.Object.(*dockerv10.DockerImage)
+
+		// Exposed Ports nested in a DockerImageManifest are not propogated
```
nalind, line 449:
```suggestion
		// Exposed Ports nested in a DockerImageManifest are not propagated
```

nalind:
At this point I'm not clear on why `processDockerConfig()` isn't unmarshalling directly into the `docker10.DockerConfig` object that's passed to it, but this should work.

prb112:
> At this point I'm not clear on why `processDockerConfig()` isn't unmarshalling directly into the `docker10.DockerConfig` object that's passed to it, but this should work.

Taken into account, and now unmarshals directly to docker10.DockerImage and takes the Config object.

pkg/helpers/newapp/app/dockerimagelookup.go
```diff
@@ -331,6 +332,18 @@ func (s ImageImportSearcher) Search(precise bool, terms ...string) (ComponentMat
...
+		// The code marshalls the first entry's config, and assigns it to the top-level DockerImage
+		if len(image.Image.DockerImageManifests) != 0 && dockerImage.Config == nil {
+			conf := image.Manifests[0].DockerImageConfig
+			dc := &docker10.DockerImage{}
+			if err := json.Unmarshal([]byte(conf), &dc); err != nil {
+				// Errors do not cause issues in unmarshalling, and it is intentional, and logged out.
+				klog.V(4).Infof("unmarshalling failed for manifest listed image's dockerConfig: %#v %#v", conf, err)
+			}
+			dockerImage.Config = dc.Config
+		}
```
nalind, line 346: I think this block should go after the `if !ok` check that's currently after it - in the `ok == false` case I think `dockerImage.Config` will be dereferencing a `nil` pointer.  Otherwise LGTM.

prb112: oh good catch - I will take care of it.

nalind: This looks like it's still going to be accessing fields of `dockerImage` before testing that `ok` is true.

prb112: Hi Nalin, Hopefully I addressed your concerns above. thanks, Paul

nalind: Yes, that should prevent attempts to access fields in a `nil` `dockerImage` value when `ok` is false.

prb112: Thanks for the diligent review

prb112:
Will retry when CI has a good signal

prb112:
Hey @nalind and @coreydaley 

I hope all is well. 

I think I have addressed the questions/comments.

Any next steps I should take? 

Thank you 

Paul

nalind:
Assuming one image in the list is as good as any other (and if we only care about the `ExposedPorts` field in the config blob, it is very likely to be), LGTM.

PR #1416, opened 2023-04-28
OCPBUGS-12901: preserve explicit release image in ClusterVersion

bverschueren:
This preserves the specified release image (using `--to-image=''`) in the ClusterVersion object.


petr-muller (approved):
The change looks good to me and works as intended. It would be great if we could get these things tested to capture desired behavior better but that's hard to do in the current form of `Run` without further refactoring :(

Thanks!

PR #1420, opened 2023-05-09
WRKLDS-700: bump(k8s) to v1.27.2

atiratree:

PR #1424, opened 2023-05-11
introduce plugin and plugin shadowing tests for oc

atiratree:
It is a loose copy from upstream tests with following distinctions
- run shim before executing the tests to initialize ValidPluginFilenamePrefixes
- test resolution of oc plugins
- test preference for oc plugins before kubectl plugins

Build upon and depends on the kube rebase PR and the discussion upstream about these tests:

TODO:
- [ ]: https://github.com/openshift/oc/pull/1420
- [ ]: https://github.com/kubernetes/kubernetes/pull/117952


pkg/cli/cli_test.go
```diff
@@ -0,0 +1,344 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/spf13/cobra"
+	"k8s.io/kubectl/pkg/cmd"
+	"k8s.io/kubectl/pkg/cmd/plugin"
+	cmdtesting "k8s.io/kubectl/pkg/cmd/testing"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+func TestOCSubcommandShadowPlugin(t *testing.T) {
```
ardaguclu, line 33: I assume we will not merge this PR until the upstream PR is merged in order to sync them to the latest version?

ardaguclu (commented):
This is indeed missing for feature parity between oc and kubectl and thanks @atiratree for adding these.

pkg/cli/cli_test.go
ardaguclu: I think, I missed an important point. We are making changes on upstream 1.28. But our oc rebase is basically picking 1.27.1. That means we'll have to wait next rebase to move these changes. 
In our oc rebase PR, we should live with this patch for this release.

pkg/cli/cli.go
```diff
@@ -97,19 +96,27 @@ var (
     To see the full list of commands supported, run 'oc --help'.`)
 )
 
-func NewDefaultOcCommand(in io.Reader, out, errout io.Writer) *cobra.Command {
-	cmd := NewOcCommand(in, out, errout)
+var defaultConfigFlags = genericclioptions.NewConfigFlags(true).WithDiscoveryBurst(350).WithDiscoveryQPS(50.0)
```
ardaguclu, line 99: see more about why we shouldn't have this https://github.com/kubernetes/kubernetes/pull/117668#discussion_r1194742874

pkg/cli/cli_test.go
atiratree: I suppose, if we decide to sync and base it on upstream. If not, we could introduce it sooner.

ardaguclu:
@atiratree nowadays, I'm a good follower of your PRs :). Do you think we need this PR in 4.14?.

atiratree:
@ardaguclu :) I think it would be nice to have as it makes sure the shadowing works properly in oc (+ some refactoring). I can finalize this PR in 2 weeks when I am back. Also there is still a feature gap that has to be resolved https://github.com/openshift/oc/pull/1420#issuecomment-1544777587

pkg/cli/cli.go
atiratree: thanks, I have moved this to a function - also added `WithDeprecatedPasswordFlag` which is missing as compared to the kubectl

pkg/cli/cli.go
```diff
@@ -97,15 +96,22 @@ var (
     To see the full list of commands supported, run 'oc --help'.`)
 )
 
-func NewDefaultOcCommand(in io.Reader, out, errout io.Writer) *cobra.Command {
-	cmd := NewOcCommand(in, out, errout)
+func defaultConfigFlags() *genericclioptions.ConfigFlags {
+	return genericclioptions.NewConfigFlags(true).WithDeprecatedPasswordFlag().WithDiscoveryBurst(350).WithDiscoveryQPS(50.0)
+}
+
+func NewDefaultOcCommand(o kubecmd.KubectlOptions) *cobra.Command {
```
ardaguclu, line 103: This is breaking interface change. Thereby,  I searched external usage of this function and there is none found which means that this change is safe.

pkg/cli/kubectl_compat_test.go
```diff
@@ -2,11 +2,11 @@ package cli
 
 import (
 	"bytes"
-	"io/ioutil"
+	"io"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
```
ardaguclu, line 6: nit: move into k8s.io group

ardaguclu (commented):
/lgtm
/hold
until ipv6 test is also passed

pkg/cli/cli.go
```diff
@@ -166,8 +172,8 @@ func NewDefaultOcCommand(in io.Reader, out, errout io.Writer) *cobra.Command {
 	return cmd
 }
 
-func NewOcCommand(in io.Reader, out, err io.Writer) *cobra.Command {
-	warningHandler := rest.NewWarningWriter(err, rest.WarningWriterOptions{Deduplicate: true, Color: kterm.AllowsColorOutput(err)})
+func NewOcCommand(o kubecmd.KubectlOptions) *cobra.Command {
```
ardaguclu, line 175: This is breaking interface change. Thereby, I searched external usage of this function and there is none found which means that this change is safe.

pkg/cli/kubectl_compat_test.go
atiratree: fixed the imports

PR #1425, opened 2023-05-16
OCPBUGS-11123: oc adm groups sync: print a warning when two or more groups are mapped to the same ldap uid

ingvagabund:
The original design limited the ldap to openshift group mapping to a single group: https://github.com/openshift/api/blob/4f3c947a2271264efd787eebeec05deb141e0333/legacyconfig/v1/types.go#L1306

The current implementation does not allow to cleanly error without disrupting already existing automation which is either not aware of this restriction or willingly ignores it. We can only print a warning.

### DIFF
--- a/pkg/helpers/groupsync/grouplister.go
+++ b/pkg/helpers/groupsync/grouplister.go
@@ -8,6 +8,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 
 	userv1 "github.com/openshift/api/user/v1"
 	userv1client "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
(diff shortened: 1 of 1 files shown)


pkg/helpers/groupsync/grouplister.go
```diff
@@ -62,6 +63,9 @@ func (l *allOpenShiftGroupLister) ListGroups() ([]string, error) {
 
 		ldapGroupUID := group.Annotations[LDAPUIDAnnotation]
 		l.ldapGroupUIDToOpenShiftGroupName[ldapGroupUID] = group.Name
+		if _, exists := l.ldapGroupUIDToOpenShiftGroupName[ldapGroupUID]; exists {
```
ardaguclu, line 66: nit:
```suggestion
		if existedGroupUID, exists := l.ldapGroupUIDToOpenShiftGroupName[ldapGroupUID]; exists {
```
`existedGroupUID` can be used instead long `l.ldapGroupUIDToOpenShiftGroupName[ldapGroupUID]`

ardaguclu (commented):
What about the other objects that implement `LDAPGroupLister` interface such as https://github.com/openshift/oc/blob/39679d53195c77ff541aeea2717b25bc2a03b3bc/pkg/helpers/groupsync/rfc2307/ldapinterface.go#L141?

pkg/helpers/groupsync/grouplister.go
ingvagabund: Done.

ingvagabund:
> What about the other objects that implement LDAPGroupLister interface such as

The remaining do not list the openshift groups. The scope of the changes are the openshift groups only.

ardaguclu:
Thanks @ingvagabund 
/lgtm

PR #1429, opened 2023-05-25
OCPBUGS-11632: Skip invalid events yamls and continue

ardaguclu:
Although it is rare, if `events.yaml` file is invalid, `oc adm must-gather` and `oc adm inspect` commands return non zero exit code because they fail when generating `event-filter.html`. Since we don't have control on events, we can warn user and continue process instead returning error.

### DIFF
--- a/pkg/cli/admin/inspect/util.go
+++ b/pkg/cli/admin/inspect/util.go
@@ -191,7 +191,8 @@ func getAllEventsRecursive(rootDir string) (*corev1.EventList, error) {
 			}
 			events, err := readEvents(eventBytes)
 			if err != nil {
-				return fmt.Errorf("failed to read event err: %v", err)
+				klog.Warningf("skipping %s, failed to read event err: %v", string(eventBytes), err)
+				return nil
 			}
 			eventLists.Items = append(eventLists.Items, events.Items...)
 			return nil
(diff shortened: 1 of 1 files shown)


ardaguclu:
@ingvagabund could you PTAL?. Thanks.

pkg/cli/admin/inspect/util.go
```diff
@@ -191,7 +191,8 @@ func getAllEventsRecursive(rootDir string) (*corev1.EventList, error) {
 			}
 			events, err := readEvents(eventBytes)
 			if err != nil {
-				return fmt.Errorf("failed to read event err: %v", err)
+				klog.Warningf("skipping %s, failed to read event err: %v", string(eventBytes), err)
```
ingvagabund, line 194: `getAllEventsRecursive` should be renamed then to reflect the fact the function will do best-effort instead. E.g. `getEventsRecursive` + mentioning the best-effort intention in the function description.

ardaguclu: Thanks for review @ingvagabund. You are right, I renamed function.

ardaguclu:
Unit test failures should be fixed by https://github.com/openshift/oc/pull/1430

ingvagabund:
/lgtm

With the error there's no webpage generated so we are not making the situation worst. Actually, the webpage gets generated with at least a subset of the original events.

ardaguclu:
Known flaky test(`cmd/images.sh`, I'll fix it soon);
/test e2e-agnostic-ovn-cmd

ardaguclu:
/hold
until unit tests are stable

PR #1431, opened 2023-05-26
Add package subcommand in release command

kevydotvinu:
- It shows the contents of `pkglist.txt` in the machine-os-content image.
```
$ oc adm release package 4.13.0 | grep NetworkManager
NetworkManager-1:1.42.2-1.el9.x86_64
NetworkManager-cloud-setup-1:1.42.2-1.el9.x86_64
NetworkManager-libnm-1:1.42.2-1.el9.x86_64
NetworkManager-ovs-1:1.42.2-1.el9.x86_64
NetworkManager-team-1:1.42.2-1.el9.x86_64
NetworkManager-tui-1:1.42.2-1.el9.x86_64
```

### DIFF
--- a/pkg/cli/admin/release/package.go
+++ b/pkg/cli/admin/release/package.go
@@ -1,0 +1,156 @@
+package release
+
+import (
+	"fmt"
+	"github.com/openshift/oc/pkg/cli/image/extract"
+	"github.com/openshift/oc/pkg/cli/image/imagesource"
+	imagemanifest "github.com/openshift/oc/pkg/cli/image/manifest"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+	"os"
+	"strings"
+)
+
+func NewPackageOptions(streams genericclioptions.IOStreams) *PackageOptions {
+	return &PackageOptions{
+		IOStreams:              streams,
+		KubeTemplatePrintFlags: *genericclioptions.NewKubeTemplatePrintFlags(),
+		ParallelOptions:        imagemanifest.ParallelOptions{MaxPerRegistry: 4},
+	}
+}
+
+func NewPackage(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
+	o := NewPackageOptions(streams)
+	cmd := &cobra.Command{
+		Use:   "package IMAGE",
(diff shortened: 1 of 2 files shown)


riginoommen (commented):
LGTM ❤️ this looks nice

riginoommen (commented):
LGTM ❤️ this looks nice

pkg/cli/admin/release/package.go
```diff
@@ -0,0 +1,156 @@
...
+	return o.FilterOptions.Validate()
+}
+
+func (o *PackageOptions) Run() error {
+	var exitErr error
+	image := strings.Join(o.Images, " ")
+	infoopts := NewInfoOptions(o.IOStreams)
+	infoopts.SecurityOptions = o.SecurityOptions
+	infoopts.FilterOptions = o.FilterOptions
+	infoopts.ImageFor = "machine-os-content"
```
sinnykumari, line 119: machine-os-content is deprecated and will anytime go away from release image, see https://issues.redhat.com/browse/MCO-392 . Instead, you should use rhel-coreos (contains base RHCOS content) and rhel-coreos-extensions (contains OCP supported extensions and RT kernel packages) which is available from 4.13.

sinnykumari: in 4.12, they are referenced as rhel-coreos-8 and rhel-coreos-8-extensions. So, based on which all releases this impact, changes can be handled accordingly.

riginoommen:
Unfreeze please.... om Kreem kuttichatha

pkg/cli/admin/release/package.go
kevydotvinu: Thanks @sinnykumari, for the suggestion. I will add that changes.

kevydotvinu:
> I also think `package` is too generic. In OCP 4, the operating system is _just one component_ (albeit, an important one) of the platform.
> 
> So this could be something like `oc adm release operating-system-info` perhaps?
> 
> Is this command really worth it over just referring to e.g. https://github.com/openshift/os/blob/master/docs/faq.md#q-how-do-i-determine-what-version-of-an-rpm-is-included-in-an-rhcos-release ?

Thanks, @cgwalters, `operating-system-info` would be neat. Even if the `pkglist.txt`file is not present it the image, achieving the same using the `podman` like the below would be an attempt to simply the task. What do you think?

https://github.com/openshift/oc/blob/7e74d82e13fae0a514805cec1e2c9164cda06325/pkg/cli/admin/release/git.go#L52

riginoommen (approved):

PR #1442, opened 2023-06-07
pkg/cli/admin/upgrade: Surface Recommended=Unknown more prominently

wking:
Make a distinction between "all of these are not recommended for your cluster" and "we're having trouble figuring out if these risks apply to your cluster" more obvious to customers who haven't set `--include-not-recommended`.  For example, if the monitoring stack backing PromQL risk evaluation is unavailable, customers may want to assess their exposure manually, instead of waiting for an update recommendation that is not impacted by PromQL risks.


petr-muller (approved):
LGTM

petr-muller:
/lgtm

This repo does not care about GH reviews, boo
/meow

petr-muller:
yeah, Trevor and I are ones :) I just complained about Prow not treating an approving GH review as `/lgtm` which is configurable in repo config

PR #1439, opened 2023-06-06
add oc adm ocp-certificates regenerate-top-level

deads2k:
replace https://github.com/openshift/oc/pull/1438

Adds a command that will cause operators to regenerate the self-signed signing certificates at the top of their chains.

[deads@fedora ~]$ oc adm ocp-certificates regenerate-signers -A secrets --all --dry-run 
secret/aggregator-client-signer regeneration set
secret/kube-apiserver-to-kubelet-signer regeneration set
secret/kube-control-plane-signer regeneration set
secret/loadbalancer-serving-signer regeneration set
secret/localhost-recovery-serving-signer regeneration set
secret/localhost-serving-signer regeneration set
secret/node-system-admin-signer regeneration set
secret/service-network-serving-signer regeneration set
secret/csr-signer-signer regeneration set
secret/ovn-ca regeneration set
secret/signer-ca regeneration set
Needs unit tests.

/assign @stlaz

PR #1443, opened 2023-06-08
add command for regenerating OCP leaf certs

stlaz:
builds on top of https://github.com/openshift/oc/pull/1439

/assign @deads2k


stlaz:
we may want to extend this for the Prometheus/OpenShift-authenticator client certs

pkg/cli/admin/ocpcertificates/certregen/command.go
```diff
@@ -61,8 +61,8 @@ func NewRegenerateTopLevelOptions(restClientGetter genericclioptions.RESTClientG
 	}
 }
 
-func NewCmdRegenerateTopLevel(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
-	o := NewRegenerateTopLevelOptions(restClientGetter, streams)
+func NewCmdcertregen(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
```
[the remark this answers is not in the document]
deads2k: Is this an accidental rename?

pkg/cli/admin/ocpcertificates/certregen/leavesregen.go
```diff
@@ -0,0 +1,94 @@
...
+
+type LeavesRegen struct {
+	ValidBefore *time.Time
+}
+
+// split here for convenience of unit testing
+func (o *LeavesRegen) forceRegenerationOnSecret(objPrinter *objectPrinter, kubeClient kubernetes.Interface, secret *corev1.Secret, dryRun bool) error {
+	if len(secret.Annotations[certrotation.CertificateIssuer]) == 0 ||
+		len(secret.Annotations[certrotation.CertificateNotBeforeAnnotation]) == 0 ||
+		isRevisioned(secret.ObjectMeta) {
```
deads2k, line 27: should add this to the target one.

deads2k: err the signer one.

pkg/cli/admin/ocpcertificates/certregen/leavesregen.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		if notBefore.After(*o.ValidBefore) {
+			// not for us
+			return nil
+		}
+	}
+
+	applyOptions := metav1.ApplyOptions{
+		Force:        true,
+		FieldManager: RegenerateSignersFieldManager,
```
deads2k, line 65: This field manager needs to be different.  If you use the same field manager and set different fields on an object, it will clear fields you no longer care about.  While probably ok here, we don't want to risk cross talk.

deads2k:
https://github.com/openshift/oc/pull/1443/files#r1224178638 is blocking. lgtm after

pkg/cli/admin/ocpcertificates/certregen/command.go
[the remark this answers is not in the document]
stlaz: yes, probably a result of mutliple strings replacement

pkg/cli/admin/ocpcertificates/certregen/leavesregen.go
stlaz: good catch

PR #1444, opened 2023-06-08
add a new command to wait for all clusteroperators to go stable

stlaz:
/assign @deads2k


pkg/cli/clusteroperator/waitforstable/command_test.go
```diff
@@ -0,0 +1,105 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/printers"
+
+	fakeconfig "github.com/openshift/client-go/config/clientset/versioned/fake"
+)
+
+func TestWaitForStableOptions_Run(t *testing.T) {
```
stlaz, line 18: still need more unit tests that wait for the cluster to stay stable

pkg/cli/clusteroperator/waitforstable/command.go
```diff
@@ -0,0 +1,157 @@
...
+var (
+	waitForStableLong = templates.LongDesc(`TODO:`)
+
+	waitForStableExample = templates.Examples(`
+		TODO: 
+	`)
+)
+
+func NewWaitForStableOptions(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *WaitForStableOptions {
+	return &WaitForStableOptions{
```
deads2k, line 45: Default timeout?  Maybe 60 minutes?

pkg/cli/clusteroperator/waitforstable/command.go
```diff
@@ -0,0 +1,157 @@
...
+
+		return true, nil
+	})
+
+	return waitErr
+}
+
+func isOperatorStable(operator *configv1.ClusterOperator) bool {
+	if v1helpers.IsStatusConditionFalse(operator.Status.Conditions, configv1.OperatorAvailable) ||
+		v1helpers.IsStatusConditionTrue(operator.Status.Conditions, configv1.OperatorProgressing) ||
```
deads2k, line 152: not false isn't the same as true.

pkg/cli/clusteroperator/waitforstable/command.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		return true, nil
+	})
+
+	return waitErr
+}
+
+func isOperatorStable(operator *configv1.ClusterOperator) bool {
+	if v1helpers.IsStatusConditionFalse(operator.Status.Conditions, configv1.OperatorAvailable) ||
```
deads2k, line 151: not true isn't the same as false.

pkg/cli/clusteroperator/waitforstable/command.go
```diff
@@ -0,0 +1,157 @@
...
+		IOStreams: streams,
+	}
+}
+
+// NewCmdLogout implements a command to wait for OpenShift platform operators to become stable
+func NewCmdWaitForStableClusterOperators(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
+	o := NewWaitForStableOptions(restClientGetter, streams)
+
+	cmd := &cobra.Command{
+		Use:     "waitforstable",
```
deads2k, line 60: `wait-for-stable-cluster` I think.

pkg/cli/clusteroperator/waitforstable/command.go
```diff
@@ -0,0 +1,157 @@
...
+		Short:   "wait for the platform operators to become stable",
```
deads2k, line 61: We may choose to make this wait for machineconfigpools to steady too.

deads2k:
post some sample output for me?

deads2k:
Let's get the conditions and name fixed up and then I think we're ready.

stlaz:
the output changed slightly with the latest changes:
```
$ oc adm clusteroperator wait-for-stable-cluster --timeout=5m2s
operator "authentication" is not yet stable: degraded != false
operator "authentication" is not yet stable: degraded != false
operator "authentication" is not yet stable: degraded != false
```

pkg/cli/clusteroperator/waitforstable/command.go
```diff
@@ -0,0 +1,148 @@
...
+
+		return true, nil
+	})
+
+	return waitErr
+}
+
+func unstableOperatorReason(operator *configv1.ClusterOperator) string {
+	if !v1helpers.IsStatusConditionTrue(operator.Status.Conditions, configv1.OperatorAvailable) {
+		return "available != true"
```
deads2k, line 139: we may twiddle these in the future, but the conditions look correct now.

cgwalters: I had to read this twice since we're returning something that *looks like (Go) code* but is actually for humans?  Why not something more like
`fmt.Sprintf("operator %s is not available", operator.name)` etc?

PR #1445, opened 2023-06-08
add printing for new revisions

deads2k:
Adds printing of static pod operator revisions so that we can watch the rollout happen and get reports of why a particular revision was created.  This is helpful for watching the manner in which new certificates trigger rollouts.

/assign @stlaz


stlaz:
the new commits look ok, for the base from #1441 the related review applies

PR #1446, opened 2023-06-08
add command to produce an updated CA bundle for trusting the kube-apiserver

deads2k:
Realistically the starting point will only be useable by cluster-admins.  The point of this is to use existing trust to get the updated/rotated trust during a rotation.

Convenient way to see the ca bundle in your kubeconfig

`oc config view --raw -o jsonpath='{.clusters[?(@.name == "the-name")].cluster.certificate-authority-data}' | base64 -d`

I got this compatible with general kubeconfigs.  Still needs good testing.  

/assign @stlaz

### DIFF
--- a/pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
+++ b/pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
@@ -63,7 +63,10 @@ func NewCmdConfigRefreshCABundle(restClientGetter genericclioptions.RESTClientGe
 		Example:               setClusterExample,
 		Run: func(cmd *cobra.Command, args []string) {
 			cmdutil.CheckErr(o.Complete(cmd))
-			cmdutil.CheckErr(o.Run(context.TODO()))
+			cmdutil.CheckErr(o.Validate())
+			r, err := o.ToRuntime()
+			cmdutil.CheckErr(err)
+			cmdutil.CheckErr(r.Run(context.TODO()))
 
 		},
 	}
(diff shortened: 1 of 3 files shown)


pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
```diff
@@ -0,0 +1,181 @@
...
+	cmd.Flags().BoolVar(&o.DryRun, "dry-run", o.DryRun, "display the CA bundle, but don't make any changes to the kubeconfig")
+
+	return cmd
+}
+
+func (o RefreshCABundleOptions) Run(ctx context.Context) error {
+	err := o.Validate()
+	if err != nil {
+		return err
+	}
```
stlaz, line 80: I'd expect this to come right after `cmdutil.CheckErr(o.Complete(cmd))` in `NewCmdConfigRefreshCABundle()`

pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
```diff
@@ -0,0 +1,181 @@
...
+
+	clientConfig, err := o.RESTClientGetter.ToRESTConfig()
+	if err != nil {
+		return err
+	}
+	kubeClient, err := kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
```
stlaz, line 89: the client setup should probably be a part of `Complete()` as fake client is easier to mock than the whole `genericclioptions.RESTClientGetter` for unit tests

pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
```diff
@@ -0,0 +1,181 @@
...
+
+	kubeAPIServerCABundleConfigMap, err := kubeClient.CoreV1().ConfigMaps("openshift-kube-controller-manager").Get(ctx, "serviceaccount-ca", metav1.GetOptions{})
```
stlaz, line 91: if you wanted to do this as an unprivileged user, the `kube-root-ca.crt` that's distributed in all namespace appears to share the same content

pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
```diff
@@ -0,0 +1,181 @@
...
+
+		},
+	}
+
+	cmd.Flags().BoolVar(&o.DryRun, "dry-run", o.DryRun, "display the CA bundle, but don't make any changes to the kubeconfig")
+
+	return cmd
+}
+
+func (o RefreshCABundleOptions) Run(ctx context.Context) error {
```
stlaz, line 76: will this command break trust for you if you rely on system-trust and you're keeping your kubeconfig CA empty?

If so, we may want to add that in the description of the command or add a prompt to confirm in that case.

pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
deads2k: > if you wanted to do this as an unprivileged user, the kube-root-ca.crt that's distributed in all namespace appears to share the same content

that configmap isn't guaranteed to be accurate.  You could race with evil-namespace-admin.

pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to read the CA bundle from the cluster: %w", err)
+	}
+	caBundle := kubeAPIServerCABundleConfigMap.Data["ca-bundle.crt"]
+	if len(caBundle) == 0 {
+		return fmt.Errorf("cluster somehow missing the CA bundle: not an OCP cluster?")
+	}
+
+	if r.DryRun {
+		fmt.Fprintf(r.Out, caBundle)
```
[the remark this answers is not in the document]
stlaz: either that or it might mention the rewrite of the CA file (if that would happen) in case it's more generic trust bundle and we'd wipe it

stlaz:
/lgtm
No blocking comments

pkg/cli/config/refreshcabundle/refresh_ca_bundle.go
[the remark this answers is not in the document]
deads2k: I'm open to a `-o` option for kubeconfig or ca-bundle.  I'd like to pin this far for merge over the weekend and refine next week or later as a followup.

cgwalters (commented):
I gave this a superficial skim and it looks sane to me FWIW

PR #1447, opened 2023-06-09
Add a command to remove older trust

stlaz:
Adds the `oc adm ocp-certificates remove-old-trust` command.

TODO:
- [x] add an easy way to exclude certain CMs
- [x] add unit tests
- [ ] test to see if it works

/assign @deads2k 

built on top of https://github.com/openshift/oc/pull/1443

### DIFF
--- a/pkg/cli/admin/ocpcertificates/certregen/rootregen_test.go
+++ b/pkg/cli/admin/ocpcertificates/certregen/rootregen_test.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
 	"io"
 	"reflect"
 	"testing"
@@ -29,11 +30,12 @@ import (
 
 func TestRootsRegen_forceRegenerationOnSecret(t *testing.T) {
 	tests := []struct {
-		name           string
-		validBefore    *time.Time
-		inputSecret    *corev1.Secret
-		expectedUpdate bool
-		wantErr        string
+		name             string
+		validBefore      *time.Time
+		inputSecret      *corev1.Secret
+		expectedUpdate   bool
+		injectApplyError bool
+		wantErr          string
 	}{
 		{
 			name:        "no annotations",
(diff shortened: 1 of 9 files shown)


pkg/cli/admin/ocpcertificates/trustpurge/trustpurge.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if r.dryRun {
+		applyOptions.DryRun = []string{metav1.DryRunAll}
+	}
+
+	applyCM := applycorev1.ConfigMap(cm.Name, cm.Namespace)
+	applyCM = applyCM.WithData(newCMData)
+
+	finalObj, err := r.KubeClient.CoreV1().ConfigMaps(cm.Namespace).Apply(context.TODO(), applyCM, applyOptions)
```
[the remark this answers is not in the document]
stlaz: Looks like resourceversion constraint is only available for `Get`, `List` and `Delete` actions

deads2k:
followup we should attempt.  read all the secrets that hold the targets and see if the CA bundle we would write can validate them.  That would protect us in case someone runs this command inappropriately.

pkg/cli/admin/ocpcertificates/certregen/regenerate.go
```diff
@@ -61,10 +61,6 @@ func (o *RegenerateCertsRuntime) forceRegenerationOnResourceInfo(info *resource.
 		return err
 	}
 
-	if secretKind != info.Object.GetObjectKind().GroupVersionKind() {
```
deads2k, line 64: please leave in place.

pkg/cli/admin/ocpcertificates/certregen/rootregen_test.go
```diff
@@ -114,7 +128,7 @@ func testPrinter(runtime.Object, io.Writer) error {
 }
 
 func testSecret(t *testing.T) *corev1.Secret {
-	key, err := rsa.GenerateKey(rand.Reader, 4096)
+	key, err := rsa.GenerateKey(rand.Reader, 1024)
```
deads2k, line 131: 2048 or fips gets angry.  Yes, even in tests.

pkg/cli/admin/ocpcertificates/trustpurge/trustpurge.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (r *RemoveOldTrustRuntime) purgeTrustFromConfigMap(cm *corev1.ConfigMap) error {
+	var err error
+	var finalObj *corev1.ConfigMap
+
+	if excludedConfigMaps := r.excludeBundles[cm.Namespace]; excludedConfigMaps.Has(cm.Name) {
+		return nil
+	}
+
+	for retriesLeft := 2; retriesLeft > 0; retriesLeft -= 1 {
```
deads2k, line 86: Use the retryonconflict method

deads2k:
solid starting point.  We should write a precondition check that the target cert involved can be validated using the keys that will remain.  That will likely end up *requiring* the date.

/lgtm
/approve

PR #1448, opened 2023-06-09
Restart kubelet

deads2k:
Add command to restart kubelets on nodes using a pod placed on each one.  Timeouts need testing.

`oc adm clusteroperator restart-kubelet nodes -l node-role.kubernetes.io/worker --parallelism=2`

/assign @rphillips


pkg/cli/clusteroperator/restartkubelet/restart-pod-template.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: restart-kubelet-pod-
+spec:
+  restartPolicy: Never
+  containers:
+    - command:
+        - sleep
```
cgwalters, line 9: Presumably something will replace this?  What?

pkg/cli/clusteroperator/restartkubelet/restart-pod-template.yaml
```diff
@@ -0,0 +1,13 @@
...
+        - "10"
+      image: registry.redhat.io/openshift4/ose-must-gather:latest
```
cgwalters, line 11: Why this image?

pkg/cli/clusteroperator/restartkubelet/restart_kubelet.go
```diff
@@ -0,0 +1,229 @@
...
+		numberOfNodesInParallel = o.NumberOfNodesInParallel
+	case o.PercentOfNodesInParallel > 0:
+		numberOfNodesInParallel = len(interestingNodes) / o.PercentOfNodesInParallel
+	}
+	if numberOfNodesInParallel < 1 {
+		numberOfNodesInParallel = 1
+	}
+
+	// create a namespace to work in
+	nsName := "!!-dry-run"
```
cgwalters, line 71: Hmm...do we need a full new namespace for this?  Why not `kube-system` or something?

pkg/cli/clusteroperator/restartkubelet/restart_kubelet.go
```diff
@@ -0,0 +1,229 @@
...
+	if !o.DryRun {
+		ns, err := o.KubeClient.CoreV1().Namespaces().Create(ctx, namespace, metav1.CreateOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to create namespace: %w", err)
+		}
+		nsName = ns.Name
+		fmt.Fprintf(o.Out, "Created namespace/%v\n", ns.Name)
+		defer func() {
+			// using new context so the cancel doesn't stop the cleanup.
```
cgwalters, line 80: Yeah but we don't hook up `oc` to SIGINT, so if the user Ctrl-C's the cli while waiting for kubelet restart, we will leak the namespace and also error out on later attempts rerun this command because of the unconditional `Create` above right?

pkg/cli/clusteroperator/restartkubelet/restart-pod-template.yaml
deads2k: > Presumably something will replace this? What?

Whatever Ryan ends up developing.  :)

pkg/cli/clusteroperator/restartkubelet/restart-pod-template.yaml
deads2k: > Why this image?

Same default as our debugging today.

pkg/cli/clusteroperator/restartkubelet/restart_kubelet.go
deads2k: kube-system is naughty to use.  Shared namespaces are fraught.  Notice the cleanup on this script is a clean namespace delete to avoid cruft.

pkg/cli/clusteroperator/restartkubelet/restart_kubelet.go
deads2k: > Yeah but we don't hook up oc to SIGINT, so if the user Ctrl-C's the cli while waiting for kubelet restart, we will leak the namespace and also error out on later attempts rerun this command because of the unconditional Create above right?

Which signals should I catch?  I'm happy to update.

cgwalters: Trying to properly handle SIGINT is hard.  And in the truly general case there's also Go panics (between the creation and `defer` setup), network flakes, operating system crashes - all of which can result in the cleanup code not being run.

Kubernetes wants us to do eventual consistency/reconcilation, so let's do that instead of trying to crash recovery?

i.e. idempotently create namespace, don't error if it exists, be sure we handle any pods created before etc

rphillips:
@deads2k Here is a patch... the pod exits cleanly, kubelet is restarted, I am not sure why the oc tool is hanging.

https://github.com/rphillips/oc/commit/81c4b104e3c1267d39b3bc42786d8d51e3fbd828

pkg/cli/clusteroperator/restartkubelet/restart_kubelet.go
deads2k: > Kubernetes wants us to do eventual consistency/reconcilation, so let's do that instead of trying to crash recovery?

how would you do this with a client command as opposed to a long running process?

deads2k:
> @deads2k Here is a patch... the pod exits cleanly, kubelet is restarted, I am not sure why the oc tool is hanging.

Output from your hang?  Are the pods stuck running?

pkg/cli/clusteroperator/restartkubelet/restart_kubelet.go
cgwalters: I'm not blocking this PR to be clear.  If the process gets Ctrl-C'd, I guess we can leave it up to the user to deal with.  It looks like what would happen here is:

```
$ oc adm ocp-certificates restart-kubelet
Created namespace/openshift-restart-kubelet-xyz
# Ctrl-c or pod/machine/laptop dies or whatever
$ oc adm ocp-certificates  restart-kubelet
error: namespace openshift-restart-kubelet-xyz already exists
```
Then we'd just tell them to do

```
$ oc delete namespace/openshift-restart-kubelet-xyz
```
and restart?

pkg/cli/clusteroperator/restartkubelet/restart-pod-template.yaml
stlaz: isn't that `"openshift/tools:latest"` as in https://github.com/openshift/oc/blob/641d6f024a7a8f1a66ed311f3b8070c8a4157570/pkg/cli/debug/debug.go#L388 ?

deads2k:
superceded by https://github.com/openshift/oc/pull/1454

pkg/cli/clusteroperator/restartkubelet/restart_kubelet.go
deads2k: it's a generated namespace name.  We'd just get a new namespace when the command runs.

PR #1449, opened 2023-06-09
ocpcertificates: Add regenerate-mco

cgwalters:
This certificate chain is special as it may need external-to-cluster action to rotate in the general case.  Let's make this a separate CLI entrypoint.


cgwalters:
cc @deads2k @yuqi-zhang

yuqi-zhang:
Give it a test spin on 4.13 AWS (4.13 gcp failed to scale altogether, for some reason):

1. run the command above
2. get the new rotated ca crt bundle (encoded) at `secret/machine-config-server-ca -n openshift-machine-config-operator`'s tls.cert
3. get the worker-user-data secret from openshift-machine-api namespace
4. base64decode the contents, swap in the new ca cert under security/tls/certificateAuthorities
5. base64encode the new contents, update worker-user-data secret
6. scale a machine

The machine doesn't come up, and looking at the AWS console, I see `GET error: Get "https://api-int.jerzhang-230609-test.devcluster.openshift.com:22623/config/worker": x509: certificate signed by unknown authority`

Uh hmm, the MCS deployment is certainly referencing the cert:
```
      - name: certs
        secret:
          defaultMode: 420
          secretName: machine-config-server-tls
```

Oh, I see we load the cert from `/etc/ssl/mcs/tls.*` and then passes it to serve TLS here: https://github.com/openshift/machine-config-operator/blob/master/pkg/server/api.go#L71 presumably that isn't dynamically updated.

Restarting all 3 MCS pods and trying a new node scaleup gets me...

`GET error: Get "https://api-int.jerzhang-230609-test.devcluster.openshift.com:22623/config/worker": x509: certificate is valid for https://api-int.jerzhang-230609-test.devcluster.openshift.com:6443, not api-int.jerzhang-230609-test.devcluster.openshift` <- the logs are actually truncated here on the web console and I can't seem to otherwise connect to it, but that's where it's getting stuck now.

Looks like, we're missing a `https://` ? Did I mess up a step somewhere?

cgwalters:
Closing in favor of targeting the combined https://github.com/openshift/oc/pull/1450

PR #1450, opened 2023-06-11
ocpcertificates: add ability to rotate MCS CA/cert

yuqi-zhang:
New oc command to rotate the Machine-Config-Server ca + cert. Also see: https://github.com/openshift/oc/pull/1449

To test this, take any cluster and run the command. Afterwards, try scaling a node into the cluster (e.g. on AWS, patch a worker machineset to add 1 more replica)

### DIFF
--- a/pkg/cli/admin/ocpcertificates/regeneratemco/command.go
+++ b/pkg/cli/admin/ocpcertificates/regeneratemco/command.go
@@ -2,18 +2,11 @@ package regeneratemco
 
 import (
 	"context"
-	"fmt"
-	"net/url"
 	"time"
 
-	configclient "github.com/openshift/client-go/config/clientset/versioned"
-	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/certrotation"
-	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/spf13/cobra"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
+
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
@@ -22,52 +15,57 @@ import (
 )
 
 const (
-	OneYear    = 365 * 24 * time.Hour
-	caExpiry   = 10 * OneYear
-	caRefresh  = 9 * OneYear
+	oneYear    = 365 * 24 * time.Hour
+	caExpiry   = 10 * oneYear
+	caRefresh  = 9 * oneYear
 	keyExpiry  = caExpiry
 	keyRefresh = caRefresh
 
 	mcoNamespace   = "openshift-machine-config-operator"
-	controllerName = "OcMachineConfigServerRotator"
+	mapiNamespace  = "openshift-machine-api"
+	controllerName = "OCMachineConfigServerRotator"
 	mcsName        = "machine-config-server"
 
 	// mcsTlsSecretName is created by the installer and is not owned by default
 	mcsTlsSecretName = mcsName + "-tls"
-	// signerName is the name injected into the certificate as the signer
-	signerName = "oc-for-machine-config-server"
-	// openshiftOrg is the openshift organizational unit
-	openshiftOrg = "openshift"
 
-	// legacyRootCANamespace is the namespace of the legacy root CA
-	legacyRootCANamespace = "kube-system"
-	// legacyRootCA is the name of the configmap holding the MCS CA created by openshift-install
-	legacyRootCA = "root-ca"
+	// newMCSCASecret is the location of the CA after rotation
+	newMCSCASecret = "machine-config-server-ca"
+	userDataKey    = "userData"
 
-	generatedAnnotationKey   = "openshift.io/generated-by"
-	generatedAnnotationValue = "oc"
+	// mcoManagedWorkerSecret is the unused, MCO-managed stub ignition for workers
+	mcoManagedWorkerSecret = "worker-user-data-managed"
+	// mcoManagedMasterSecret is the unused, MCO-managed stub ignition for masters
+	mcoManagedMasterSecret = "master-user-data-managed"
+	// mcsLabelSelector is used to select the MCS pods
+	mcsLabelSelector = "k8s-app=machine-config-server"
+
+	// ign* are for the user-data ignition fields
+	ignFieldIgnition = "ignition"
+	ignFieldSource   = "source"
 )
 
 var (
 	regenerateMCOLong = templates.LongDesc(`
 		Regenerate the Machine Config Operator certificates for an OCP v4 cluster.
-
-		More information about these certificates are in the product documentation;
-		visit [docs.openshift.com](https://docs.openshift.com)
-		and select the version of OpenShift you are using.
+		This is the certificate used to verify the MCS contents when a new nodes attempts to join the cluster.
 
 		Experimental: This command is under active development and may change without notice.
 	`)
 
 	regenerateMCOExample = templates.Examples(`
-		oc adm certificates regenerate-mco
+	    # Regenerate the MCO certs without modifying user-data secrets
+		oc adm certificates regenerate-machine-config-server-serving-cert --update-ignition=false
+
+		# Update the user-data secrets to use new MCS certs
+		oc adm certificates update-ignition-ca-bundle-for-machine-config-server
 	`)
 )
 
 type RegenerateMCOOptions struct {
 	RESTClientGetter genericclioptions.RESTClientGetter
 
-	mcoNamespace string
+	ModifyUserData bool
 
 	genericclioptions.IOStreams
 }
@@ -76,16 +74,17 @@ func NewCmdRegenerateTopLevel(restClientGetter genericclioptions.RESTClientGette
 	o := &RegenerateMCOOptions{
 		RESTClientGetter: restClientGetter,
 		IOStreams:        streams,
+		ModifyUserData:   true,
 	}
 
 	cmd := &cobra.Command{
-		Use:                   "regenerate-mco",
+		Use:                   "regenerate-machine-config-server-serving-cert",
 		DisableFlagsInUseLine: true,
 		Short:                 i18n.T("Regenerate the machine config operator certificates in an OpenShift cluster"),
 		Long:                  regenerateMCOLong,
 		Example:               regenerateMCOExample,
 		Run: func(cmd *cobra.Command, args []string) {
-			cmdutil.CheckErr(run(o, args))
+			cmdutil.CheckErr(o.Run(context.WithValue(context.Background(), certrotation.RunOnceContextKey, true)))
 		},
 	}
 
@@ -94,92 +93,28 @@ func NewCmdRegenerateTopLevel(restClientGetter genericclioptions.RESTClientGette
 	return cmd
 }
 
-// AddFlags registers flags for a cli
-func (o *RegenerateMCOOptions) AddFlags(cmd *cobra.Command) {
-	cmd.Flags().StringVar(&o.mcoNamespace, "mco-namespace", mcoNamespace, "Operate instead on this namespace for the MCO (useful for testing)")
+func NewCmdUpdateUserData(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
+	o := &RegenerateMCOOptions{
+		RESTClientGetter: restClientGetter,
+		IOStreams:        streams,
+		ModifyUserData:   true,
+	}
+
+	cmd := &cobra.Command{
+		Use:                   "update-ignition-ca-bundle-for-machine-config-server",
+		DisableFlagsInUseLine: true,
+		Short:                 i18n.T("Update user-data secrets in an OpenShift cluster to use updated MCO certfs"),
+		Long:                  regenerateMCOLong,
+		Example:               regenerateMCOExample,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(o.RunUserDataUpdate(context.Background()))
+		},
+	}
+
+	return cmd
 }
 
-func run(o *RegenerateMCOOptions, args []string) error {
-	recorder := events.NewLoggingEventRecorder("oc")
-
-	clientConfig, err := o.RESTClientGetter.ToRESTConfig()
-	if err != nil {
-		return err
-	}
-	clientset, err := kubernetes.NewForConfig(clientConfig)
-	if err != nil {
-		return err
-	}
-
-	oconfig, err := configclient.NewForConfig(clientConfig)
-	if err != nil {
-		return err
-	}
-
-	host, err := oconfig.ConfigV1().Infrastructures().Get(context.TODO(), "cluster", metav1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("unable to get cluster infrastructure resource: %w", err)
-	}
-	if host.Status.APIServerInternalURL == "" {
-		return fmt.Errorf("no APIServerInternalURL")
-	}
-	apiserverIntURL, err := url.Parse(host.Status.APIServerInternalURL)
-	if err != nil {
-		return fmt.Errorf("failed to parse %s: %w", apiserverIntURL, err)
-	}
-
-	inf := informers.NewSharedInformerFactoryWithOptions(
-		clientset,
-		24*time.Hour,
-		informers.WithNamespace(o.mcoNamespace))
-
-	caName := mcsName + "-ca"
-	cont := certrotation.NewCertRotationController(
-		controllerName,
-		certrotation.RotatedSigningCASecret{
-			Namespace:     o.mcoNamespace,
-			Name:          caName,
-			Validity:      caExpiry,
-			Refresh:       caRefresh,
-			Informer:      inf.Core().V1().Secrets(),
-			Lister:        inf.Core().V1().Secrets().Lister(),
-			Client:        clientset.CoreV1(),
-			EventRecorder: recorder,
-		},
-		certrotation.CABundleConfigMap{
-			Namespace:     o.mcoNamespace,
-			Name:          caName,
-			Lister:        inf.Core().V1().ConfigMaps().Lister(),
-			Informer:      inf.Core().V1().ConfigMaps(),
-			Client:        clientset.CoreV1(),
-			EventRecorder: recorder,
-		},
-		certrotation.RotatedSelfSignedCertKeySecret{
-			Namespace: o.mcoNamespace,
-			Name:      mcsTlsSecretName,
-			Validity:  keyExpiry,
-			Refresh:   keyRefresh,
-			CertCreator: &certrotation.ServingRotation{
-				Hostnames: func() []string { return []string{apiserverIntURL.Hostname()} },
-			},
-			Lister:        inf.Core().V1().Secrets().Lister(),
-			Informer:      inf.Core().V1().Secrets(),
-			Client:        clientset.CoreV1(),
-			EventRecorder: recorder,
-		},
-		nil, // no operatorclient needed
-		recorder,
-	)
-
-	ch := make(chan struct{})
-	inf.Start(ch)
-	inf.WaitForCacheSync(ch)
-
-	ctx := context.WithValue(context.Background(), certrotation.RunOnceContextKey, true)
-	syncCtx := factory.NewSyncContext(mcsName, recorder)
-	if err := cont.Sync(ctx, syncCtx); err != nil {
-		return err
-	}
-
-	return nil
+// AddFlags registers flags for a cli
+func (o *RegenerateMCOOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&o.ModifyUserData, "update-ignition", o.ModifyUserData, "If true, automatically update user-data secrets (ignition) in machine-api namespace. Not useful if node scaling not backed by MachineSet.")
 }
--- a/pkg/cli/admin/ocpcertificates/regeneratemco/command.go
+++ b/pkg/cli/admin/ocpcertificates/regeneratemco/command.go
@@ -1,0 +1,185 @@
+package regeneratemco
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"time"
+
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/certrotation"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/kubectl/pkg/util/i18n"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+const (
+	OneYear    = 365 * 24 * time.Hour
+	caExpiry   = 10 * OneYear
+	caRefresh  = 9 * OneYear
+	keyExpiry  = caExpiry
+	keyRefresh = caRefresh
+
+	mcoNamespace   = "openshift-machine-config-operator"
+	controllerName = "OcMachineConfigServerRotator"
+	mcsName        = "machine-config-server"
+
+	// mcsTlsSecretName is created by the installer and is not owned by default
+	mcsTlsSecretName = mcsName + "-tls"
+	// signerName is the name injected into the certificate as the signer
+	signerName = "oc-for-machine-config-server"
+	// openshiftOrg is the openshift organizational unit
+	openshiftOrg = "openshift"
+
+	// legacyRootCANamespace is the namespace of the legacy root CA
+	legacyRootCANamespace = "kube-system"
+	// legacyRootCA is the name of the configmap holding the MCS CA created by openshift-install
+	legacyRootCA = "root-ca"
+
+	generatedAnnotationKey   = "openshift.io/generated-by"
+	generatedAnnotationValue = "oc"
+)
+
+var (
+	regenerateMCOLong = templates.LongDesc(`
+		Regenerate the Machine Config Operator certificates for an OCP v4 cluster.
+
+		More information about these certificates are in the product documentation;
+		visit [docs.openshift.com](https://docs.openshift.com)
+		and select the version of OpenShift you are using.
+
+		Experimental: This command is under active development and may change without notice.
+	`)
+
+	regenerateMCOExample = templates.Examples(`
+		oc adm certificates regenerate-mco
+	`)
+)
+
+type RegenerateMCOOptions struct {
+	RESTClientGetter genericclioptions.RESTClientGetter
+
+	mcoNamespace string
+
+	genericclioptions.IOStreams
+}
+
+func NewCmdRegenerateTopLevel(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
+	o := &RegenerateMCOOptions{
+		RESTClientGetter: restClientGetter,
+		IOStreams:        streams,
+	}
+
+	cmd := &cobra.Command{
+		Use:                   "regenerate-mco",
+		DisableFlagsInUseLine: true,
+		Short:                 i18n.T("Regenerate the machine config operator certificates in an OpenShift cluster"),
+		Long:                  regenerateMCOLong,
+		Example:               regenerateMCOExample,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(run(o, args))
+		},
+	}
+
+	o.AddFlags(cmd)
+
+	return cmd
+}
+
+// AddFlags registers flags for a cli
+func (o *RegenerateMCOOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.mcoNamespace, "mco-namespace", mcoNamespace, "Operate instead on this namespace for the MCO (useful for testing)")
+}
+
+func run(o *RegenerateMCOOptions, args []string) error {
+	recorder := events.NewLoggingEventRecorder("oc")
+
+	clientConfig, err := o.RESTClientGetter.ToRESTConfig()
+	if err != nil {
+		return err
+	}
+	clientset, err := kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
+	oconfig, err := configclient.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
+	host, err := oconfig.ConfigV1().Infrastructures().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to get cluster infrastructure resource: %w", err)
+	}
+	if host.Status.APIServerInternalURL == "" {
+		return fmt.Errorf("no APIServerInternalURL")
+	}
+	apiserverIntURL, err := url.Parse(host.Status.APIServerInternalURL)
+	if err != nil {
+		return fmt.Errorf("failed to parse %s: %w", apiserverIntURL, err)
+	}
+
+	inf := informers.NewSharedInformerFactoryWithOptions(
+		clientset,
+		24*time.Hour,
+		informers.WithNamespace(o.mcoNamespace))
+
+	caName := mcsName + "-ca"
+	cont := certrotation.NewCertRotationController(
+		controllerName,
+		certrotation.RotatedSigningCASecret{
+			Namespace:     o.mcoNamespace,
+			Name:          caName,
+			Validity:      caExpiry,
+			Refresh:       caRefresh,
+			Informer:      inf.Core().V1().Secrets(),
+			Lister:        inf.Core().V1().Secrets().Lister(),
+			Client:        clientset.CoreV1(),
+			EventRecorder: recorder,
+		},
+		certrotation.CABundleConfigMap{
+			Namespace:     o.mcoNamespace,
+			Name:          caName,
+			Lister:        inf.Core().V1().ConfigMaps().Lister(),
+			Informer:      inf.Core().V1().ConfigMaps(),
+			Client:        clientset.CoreV1(),
+			EventRecorder: recorder,
+		},
+		certrotation.RotatedSelfSignedCertKeySecret{
+			Namespace: o.mcoNamespace,
+			Name:      mcsTlsSecretName,
+			Validity:  keyExpiry,
+			Refresh:   keyRefresh,
+			CertCreator: &certrotation.ServingRotation{
+				Hostnames: func() []string { return []string{apiserverIntURL.Hostname()} },
+			},
+			Lister:        inf.Core().V1().Secrets().Lister(),
+			Informer:      inf.Core().V1().Secrets(),
+			Client:        clientset.CoreV1(),
+			EventRecorder: recorder,
+		},
+		nil, // no operatorclient needed
+		recorder,
+	)
+
+	ch := make(chan struct{})
+	inf.Start(ch)
+	inf.WaitForCacheSync(ch)
+
+	ctx := context.WithValue(context.Background(), certrotation.RunOnceContextKey, true)
+	syncCtx := factory.NewSyncContext(mcsName, recorder)
+	if err := cont.Sync(ctx, syncCtx); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
+++ b/pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
@@ -1,0 +1,122 @@
+package regeneratemco
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"time"
+
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/certrotation"
+	"github.com/openshift/library-go/pkg/operator/events"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+)
+
+func (o *RegenerateMCOOptions) Run(ctx context.Context) error {
+	recorder := events.NewLoggingEventRecorder("oc")
+
+	clientConfig, err := o.RESTClientGetter.ToRESTConfig()
+	if err != nil {
+		return err
+	}
+	clientset, err := kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
+	oconfig, err := configclient.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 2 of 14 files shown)


pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -177,5 +187,78 @@ func run(o *RegenerateMCOOptions, args []string) error {
...
+
+	caCert, exists := mcsSecret.Data[CASecretCertName]
+	if !exists {
+		return fmt.Errorf("could not find MCS CA cert at %s after rotation", CASecretCertName)
+	}
+	encodedCACert := base64.StdEncoding.EncodeToString(caCert)
+
+	// Get user-data-secret
+	mapiSecrets := clientset.CoreV1().Secrets(mapiNamespace)
+	secretList, err := mapiSecrets.List(ctx, metav1.ListOptions{})
```
cgwalters, line 209: So I think this is fine, but we will be trying to edit the `user-data-managed` secrets generated/owned by the MCO today that are unused.  I think the MCO will edit them back?

In the general case as far as I understand it, probably what we should be doing is iterating over the *machineset* objects, and then we can find references to their user data from there.

yuqi-zhang: Right, to rotate those, we'd have to change the MCO code I think. Let's check if we would need to do so or not

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -177,5 +187,78 @@ func run(o *RegenerateMCOOptions, args []string) error {
...
+			ignMap := userDataIgn.(map[string]interface{})
+			ign, isIgn := ignMap["ignition"].(map[string]interface{})
+			if !isIgn {
+				// Didn't find "ignition in user-data, warn but continue"
+				fmt.Printf("Unable to find ignition in user-data, skipping secret %s\n", secret.Name)
+				continue
+			}
+
+			// This is uh, quite unsavory, but arguably beats doing nested ignition version parsing?
+			if len(ign["security"].(map[string]interface{})["tls"].(map[string]interface{})["certificateAuthorities"].([]interface{})) == 1 {
```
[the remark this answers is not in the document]
yuqi-zhang: ack, will consider improving this, thanks!

cgwalters:
In my testing, I did have to bounce the MCS pods via `oc delete pods -l k8s-app=machine-config-server`.

cgwalters:
Another thing we could do here is make this more idempotent by adding an annotation to the secret with the last generation version of the input CA secret.  Right now we edit the userdata every run even when it'd be a no-op.

yuqi-zhang:
> In my testing, I did have to bounce the MCS pods via oc delete pods -l k8s-app=machine-config-server.

Ah hmm, I think I may have tested twice or something, I'll add that to the script

> Another thing we could do here is make this more idempotent by adding an annotation to the secret with the last generation version of the input CA secret. Right now we edit the userdata every run even when it'd be a no-op.

That is an option, although I doubt most people who would be using this script would run it that regularly (and really, it might be helpful to refresh the user-data regardless, just in case the user has made other manual modifications they would like to have the script auto-rectify)

cgwalters:
https://opensource.com/article/22/9/dynamically-update-tls-certificates-golang-server-no-downtime looks relevant; the default MCS logic here just loads the key once at startup. So we can and should change it to watch for rotation, but for now bouncing the pods seems OK.

cgwalters:
Ah even better, look at https://github.com/openshift/hypershift/blob/7cfb7f32a4b72c03a3a1e1ed552fe8b9a731dfee/ignition-server/cmd/start.go#L264 :arrow_right:  https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/certwatcher#CertWatcher

cgwalters:
@yuqi-zhang minor cleanup, how about de-indenting things here?

```
diff --git a/pkg/cli/admin/ocpcertificates/regeneratemco/command.go b/pkg/cli/admin/ocpcertificates/regeneratemco/command.go
index 5e20375cd..af28890b4 100644
--- a/pkg/cli/admin/ocpcertificates/regeneratemco/command.go
+++ b/pkg/cli/admin/ocpcertificates/regeneratemco/command.go
@@ -219,42 +219,42 @@ func run(o *RegenerateMCOOptions, args []string) error {
 		// 1. have a userData field
 		// 2. is an ignition config
 		userData, exists := secret.Data[userDataKey]
-		if exists {
-			// userData is an ignition config. To save the effort of multiple-version parsing, just parse it as a json
-			var userDataIgn interface{}
-			err = json.Unmarshal(userData, &userDataIgn)
-			if err != nil {
-				return fmt.Errorf("failed to unmarshal decoded user-data to json (secret %s): %w", secret.Name, err)
-			}
-			ignMap := userDataIgn.(map[string]interface{})
-			ign, isIgn := ignMap["ignition"].(map[string]interface{})
-			if !isIgn {
-				// Didn't find "ignition in user-data, warn but continue"
-				fmt.Printf("Unable to find ignition in user-data, skipping secret %s\n", secret.Name)
-				continue
-			}
-
-			// This is uh, quite unsavory, but arguably beats doing nested ignition version parsing?
-			if len(ign["security"].(map[string]interface{})["tls"].(map[string]interface{})["certificateAuthorities"].([]interface{})) == 1 {
-				ca := userDataIgn.(map[string]interface{})["ignition"].(map[string]interface{})["security"].(map[string]interface{})["tls"].(map[string]interface{})["certificateAuthorities"].([]interface{})[0]
-				ca.(map[string]interface{})["source"] = "data:text/plain;charset=utf-8;base64," + encodedCACert
-				fmt.Printf("Successfully updated for secret %s\n", secret.Name)
-			} else {
-				return fmt.Errorf("Additional CAs detected, cannot modify automatically. Aborting")
-			}
-
-			// Now marshal back to json
-			updatedIgnition, err := json.Marshal(userDataIgn)
-			if err != nil {
-				return fmt.Errorf("failed to marshal updated ignition back to json (secret %s): %w", secret.Name, err)
-			}
-
-			secret.Data[userDataKey] = updatedIgnition
-			_, err = mapiSecrets.Update(ctx, &secret, metav1.UpdateOptions{})
-			if err != nil {
-				return fmt.Errorf("could not update secret %s: %w", secret.Name, err)
-			}
+		if !exists {
+			continue
+		}
+		// userData is an ignition config. To save the effort of multiple-version parsing, just parse it as a json
+		var userDataIgn interface{}
+		err = json.Unmarshal(userData, &userDataIgn)
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal decoded user-data to json (secret %s): %w", secret.Name, err)
+		}
+		ignMap := userDataIgn.(map[string]interface{})
+		ign, isIgn := ignMap["ignition"].(map[string]interface{})
+		if !isIgn {
+			// Didn't find "ignition in user-data, warn but continue"
+			fmt.Printf("Unable to find ignition in user-data, skipping secret %s\n", secret.Name)
+			continue
+		}
+
+		// This is uh, quite unsavory, but arguably beats doing nested ignition version parsing?
+		if len(ign["security"].(map[string]interface{})["tls"].(map[string]interface{})["certificateAuthorities"].([]interface{})) == 1 {
+			ca := userDataIgn.(map[string]interface{})["ignition"].(map[string]interface{})["security"].(map[string]interface{})["tls"].(map[string]interface{})["certificateAuthorities"].([]interface{})[0]
+			ca.(map[string]interface{})["source"] = "data:text/plain;charset=utf-8;base64," + encodedCACert
+			fmt.Printf("Successfully updated for secret %s\n", secret.Name)
+		} else {
+			return fmt.Errorf("Additional CAs detected, cannot modify automatically. Aborting")
+		}
+
+		// Now marshal back to json
+		updatedIgnition, err := json.Marshal(userDataIgn)
+		if err != nil {
+			return fmt.Errorf("failed to marshal updated ignition back to json (secret %s): %w", secret.Name, err)
+		}
 
+		secret.Data[userDataKey] = updatedIgnition
+		_, err = mapiSecrets.Update(ctx, &secret, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("could not update secret %s: %w", secret.Name, err)
 		}
 	}
 
```

yuqi-zhang:
Rebased + updated based on suggestions above. Removing draft for further review

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// mcsTlsSecretName is created by the installer and is not owned by default
+	mcsTlsSecretName = mcsName + "-tls"
+	// signerName is the name injected into the certificate as the signer
+	signerName = "oc-for-machine-config-server"
+	// openshiftOrg is the openshift organizational unit
+	openshiftOrg = "openshift"
+
+	// legacyRootCANamespace is the namespace of the legacy root CA
+	legacyRootCANamespace = "kube-system"
```
deads2k, line 46: we placed this in kube-system before?  How naughty of us.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	o.AddFlags(cmd)
+
+	return cmd
+}
+
+// AddFlags registers flags for a cli
+func (o *RegenerateMCOOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.mcoNamespace, "mco-namespace", mcoNamespace, "Operate instead on this namespace for the MCO (useful for testing)")
```
deads2k, line 116: hide this flag.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+
+	mcoNamespace   string
+	modifyUserData bool
+	outputNewCert  bool
+
+	genericclioptions.IOStreams
+}
+
+func NewCmdRegenerateTopLevel(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
+	o := &RegenerateMCOOptions{
```
deads2k, line 93: set the `mcoNamespace` here.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+	mcsSecret, err := mcoSecrets.Get(ctx, newMCSCASecret, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("cannot read MCS secret: %w", err)
+	}
+
+	caCert, exists := mcsSecret.Data[CASecretCertName]
+	if !exists {
+		return fmt.Errorf("could not find MCS CA cert at %s after rotation", CASecretCertName)
+	}
+	encodedCACert := base64.StdEncoding.EncodeToString(caCert)
```
deads2k, line 233: Get this from the configmap in the rotator above, not the secret itself.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+		var userDataIgn interface{}
+		err = json.Unmarshal(userData, &userDataIgn)
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal decoded user-data to json (secret %s): %w", secret.Name, err)
+		}
+		ignMap := userDataIgn.(map[string]interface{})
+		ign, isIgn := ignMap["ignition"].(map[string]interface{})
+		if !isIgn {
+			// Didn't find "ignition in user-data, warn but continue"
+			fmt.Printf("Unable to find ignition in user-data, skipping secret %s\n", secret.Name)
```
deads2k, line 272: Fprintf to the iostream.  Doing that allows composition and unit testing for output if desired at some point.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+	"k8s.io/client-go/kubernetes"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/kubectl/pkg/util/i18n"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+const (
+	OneYear    = 365 * 24 * time.Hour
```
stlaz, line 27: unexport

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		ignMap := userDataIgn.(map[string]interface{})
+		ign, isIgn := ignMap["ignition"].(map[string]interface{})
+		if !isIgn {
+			// Didn't find "ignition in user-data, warn but continue"
+			fmt.Printf("Unable to find ignition in user-data, skipping secret %s\n", secret.Name)
+			continue
+		}
+
+		if len(ign["security"].(map[string]interface{})["tls"].(map[string]interface{})["certificateAuthorities"].([]interface{})) == 1 {
```
deads2k, line 276: all the other constants and this one is where you draw the line? ;)

Can you use `unstructured.NestedString` to make this slightly more readable?

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+	mapiNamespace  = "openshift-machine-api"
+	controllerName = "OcMachineConfigServerRotator"
+	mcsName        = "machine-config-server"
+
+	// mcsTlsSecretName is created by the installer and is not owned by default
+	mcsTlsSecretName = mcsName + "-tls"
+	// signerName is the name injected into the certificate as the signer
+	signerName = "oc-for-machine-config-server"
+	// openshiftOrg is the openshift organizational unit
+	openshiftOrg = "openshift"
```
stlaz, line 43: unused

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+
+	mcoNamespace   = "openshift-machine-config-operator"
+	mapiNamespace  = "openshift-machine-api"
+	controllerName = "OcMachineConfigServerRotator"
+	mcsName        = "machine-config-server"
+
+	// mcsTlsSecretName is created by the installer and is not owned by default
+	mcsTlsSecretName = mcsName + "-tls"
+	// signerName is the name injected into the certificate as the signer
+	signerName = "oc-for-machine-config-server"
```
stlaz, line 41: unused

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
stlaz: Unused. Go through all the consts and remove unused references.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+	if o.outputNewCert {
+		fmt.Printf("New Certificate:\n%s\n", encodedCACert)
+	}
+
+	if !o.modifyUserData {
+		return nil
+	}
+
+	// Get user-data-secret
+	mapiSecrets := clientset.CoreV1().Secrets(mapiNamespace)
```
deads2k, line 244: Based on usage and potential failures, I think everything from here down is logically a different command.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+
+	regenerateMCOExample = templates.Examples(`
+		oc adm certificates regenerate-mco
+	`)
+)
+
+type RegenerateMCOOptions struct {
+	RESTClientGetter genericclioptions.RESTClientGetter
+
+	mcoNamespace   string
```
stlaz, line 85: why is this parametrized?

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use:                   "regenerate-mco",
+		DisableFlagsInUseLine: true,
+		Short:                 i18n.T("Regenerate the machine config operator certificates in an OpenShift cluster"),
+		Long:                  regenerateMCOLong,
+		Example:               regenerateMCOExample,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(run(o, args))
```
stlaz, line 105: typically, `Run()` is a method of the options, that get `Validate()`-ed before use

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return err
+	}
+
+	oconfig, err := configclient.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
+	host, err := oconfig.ConfigV1().Infrastructures().Get(context.TODO(), "cluster", metav1.GetOptions{})
```
stlaz, line 138: get a real context here, should be passed to `run()`

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+		},
+		nil, // no operatorclient needed
+		recorder,
+	)
+
+	ch := make(chan struct{})
+	inf.Start(ch)
+	inf.WaitForCacheSync(ch)
+
+	ctx := context.WithValue(context.Background(), certrotation.RunOnceContextKey, true)
```
stlaz, line 197: wire in a proper context from `run()`

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
cgwalters: Hmm.  Yes, I see your point.  On the other hand:

- these secret objects are admin owned and not expected to change by anything else
- The cert code is already a "run once" controller and hence is idempotent and can be run multiple times reliably
- We really do want to update both of these things in "one go" as our default stance; doing one but not the other will result in a broken state

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+		// 1. have a userData field
+		// 2. is an ignition config
+		userData, exists := secret.Data[userDataKey]
+		if !exists {
+			continue
+		}
+		// userData is an ignition config. To save the effort of multiple-version parsing, just parse it as a json
+		var userDataIgn interface{}
+		err = json.Unmarshal(userData, &userDataIgn)
+		if err != nil {
```
stlaz, line 265: nit: could be one line

stlaz (changes requested):
/lgtm cancel

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+			Refresh:   keyRefresh,
+			CertCreator: &certrotation.ServingRotation{
+				Hostnames: func() []string { return []string{apiserverIntURL.Hostname()} },
+			},
+			Lister:        inf.Core().V1().Secrets().Lister(),
+			Informer:      inf.Core().V1().Secrets(),
+			Client:        clientset.CoreV1(),
+			EventRecorder: recorder,
+		},
+		nil, // no operatorclient needed
```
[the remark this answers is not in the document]
cgwalters: No, that's what the `certrotation.RunOnceContextKey` does.  (We've tested this code)

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+			Client:        clientset.CoreV1(),
+			EventRecorder: recorder,
+		},
+		certrotation.RotatedSelfSignedCertKeySecret{
+			Namespace: o.mcoNamespace,
+			Name:      mcsTlsSecretName,
+			Validity:  keyExpiry,
+			Refresh:   keyRefresh,
+			CertCreator: &certrotation.ServingRotation{
+				Hostnames: func() []string { return []string{apiserverIntURL.Hostname()} },
```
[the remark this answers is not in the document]
cgwalters: Yes

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
yuqi-zhang: We can probably remove this flag, since the command is a bit more stable, and the testing should not otherwise break cluster functionality

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
yuqi-zhang: I had a comment about how unsavoury this is but then I removed it :D I'll make it a constant at least

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -97,6 +114,8 @@ func NewCmdRegenerateTopLevel(restClientGetter genericclioptions.RESTClientGette
 // AddFlags registers flags for a cli
 func (o *RegenerateMCOOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.mcoNamespace, "mco-namespace", mcoNamespace, "Operate instead on this namespace for the MCO (useful for testing)")
+	cmd.Flags().BoolVar(&o.modifyUserData, "update-user-data", true, "If true, automatically update user-data secrets in machine-api namespace. Not useful if node scaling not backed by MachineSet.")
+	cmd.Flags().BoolVar(&o.outputNewCert, "output-new-cert", false, "If true, outputs the new CA cert for ignition use. Useful if need to manually update UPI clusters.")
```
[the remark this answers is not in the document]
cgwalters: Hmm, probably somewhere under

https://docs.openshift.com/container-platform/4.13/machine_management/index.html
perhaps
https://docs.openshift.com/container-platform/4.13/machine_management/user_infra/adding-compute-user-infra-general.html

But I don't think we need to block on this, we can just remove this code for now.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
deads2k: @yuqi-zhang I'm ok with that approach.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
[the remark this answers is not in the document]
stlaz: I see, interesting

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+
+		secret.Data[userDataKey] = updatedIgnition
+		_, err = mapiSecrets.Update(ctx, &secret, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("could not update secret %s: %w", secret.Name, err)
+		}
+	}
+
+	fmt.Println("Successfully modified user-data secrets.")
```
stlaz, line 295: this should print to `o.Out`, too

yuqi-zhang:
Ok, updated it to have two entrypoints, hopefully I've addressed all the comments

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,128 @@
...
+
+func NewCmdUpdateUserData(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
+	o := &RegenerateMCOOptions{
+		RESTClientGetter: restClientGetter,
+		IOStreams:        streams,
+		modifyUserData:   true,
+	}
+
+	cmd := &cobra.Command{
+		Use:                   "update-mco-user-data",
```
yuqi-zhang, line 112: Any ideas for a better name here?

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,128 @@
...
+	// mcsLabelSelector is used to select the MCS pods
+	mcsLabelSelector = "k8s-app=machine-config-server"
+
+	// ign* are for the user-data ignition fields
+	ignFieldIgnition = "ignition"
+	ignFieldSecurity = "security"
+	ignFieldTLS      = "tls"
+	ignFieldCA       = "certificateAuthorities"
+	ignFieldSource   = "source"
+	ignDataPrefix    = "data:text/plain;charset=utf-8;base64,"
```
cgwalters, line 49: OK as is, but it would be cleaner to vendor github.com/vincent-petithory/dataurl  in the future I think as we do in the MCO and butane.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,128 @@
...
+	o := &RegenerateMCOOptions{
+		RESTClientGetter: restClientGetter,
+		IOStreams:        streams,
+		modifyUserData:   true,
+	}
+
+	cmd := &cobra.Command{
+		Use:                   "regenerate-mco",
+		DisableFlagsInUseLine: true,
+		Short:                 i18n.T("Regenerate the machine config operator certificates in an OpenShift cluster"),
```
cgwalters, line 91: I wonder if we should clarify by s/operator/server/ here?  But eh.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
cgwalters, line 112: Minor bikeshed; maybe `update-machineset-user-data`?  Technically it's not fully owned by the MCO today.  Or maybe `update-ignition-pointer-config`?  

Eh, fine as is :smile:

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -0,0 +1,99 @@
...
+		if len(caSlice.([]interface{})) == 1 {
+			caSlice.([]interface{})[0].(map[string]interface{})[ignFieldSource] = ignDataPrefix + encodedCACert
+		} else {
+			return fmt.Errorf("additional CAs detected, cannot modify automatically. Aborting")
+		}
+
+		updatedIgnition, err := json.Marshal(userDataIgn)
+		if err != nil {
+			return fmt.Errorf("failed to marshal updated ignition back to json (secret %s): %w", secret.Name, err)
+		}
```
cgwalters, line 88: Might be nice to do:
```
if bytes.Equal(userData, updatedIgnition) {
  fmt.Fprintf(o.IOStreams.Out, "Nothing to do for %s\n", secret.Name)
  continue
}
```
or so

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -0,0 +1,99 @@
...
+
+		secret.Data[userDataKey] = updatedIgnition
+		_, err = mapiSecrets.Update(ctx, &secret, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("could not update secret %s: %w", secret.Name, err)
+		}
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "Successfully modified user-data secrets\n")
```
cgwalters, line 97: I think it'd be a bit clearer to move this print after we call `Update` to print each time we change a secret; would pair well with the above suggestion.

cgwalters (approved):
Overall LGTM, just super minor nits.  Will leave open for a bit in case anyone else has anything, if not then I'll drop lgtm later today or tomorrow.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
yuqi-zhang: I think for most users, the higher level "mco certificate" is hand-wavy enough to point them to the right path :)

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
yuqi-zhang: Yeah, I mused about that a bit, I think it's better to keep the whole chain referencing the MCO, so it's easier for a non-in-depth user to understand the connection

But yeah, probably not the best naming

yuqi-zhang:
Updated to match Colin's suggestions above

cgwalters (approved):

cgwalters:
OK I've looked through and it looks like you've adjusted for all the review comments so far.  Awesome!  I think we can do any other fixues as followups.
/lgtm

yuqi-zhang:
/retest-required

seems like flakes

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,128 @@
...
+	ignDataPrefix    = "data:text/plain;charset=utf-8;base64,"
+)
+
+var (
+	regenerateMCOLong = templates.LongDesc(`
+		Regenerate the Machine Config Operator certificates for an OCP v4 cluster.
+
+		More information about these certificates are in the product documentation;
+		visit [docs.openshift.com](https://docs.openshift.com)
+		and select the version of OpenShift you are using.
```
stlaz, line 58: Remove this part please. From reading this, I do not know where to look, still. Reading the docs if something's unclear is what we'd expect from our users even if we don't mention it here.

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
stlaz: To get the answer to that question, I 'd need to answer this other one: what kind of data are we talking about here?

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return err
+	}
+
+	host, err := oconfig.ConfigV1().Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to get cluster infrastructure resource: %w", err)
+	}
+	if host.Status.APIServerInternalURL == "" {
+		return fmt.Errorf("no APIServerInternalURL found in cluster infrastructure resource")
```
stlaz, line 41: much better now :+1:

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
```diff
@@ -0,0 +1,123 @@
...
+			Client:        clientset.CoreV1(),
+			EventRecorder: recorder,
+		},
+		nil, // no operatorclient needed
+		recorder,
+	)
+
+	ch := make(chan struct{})
+	inf.Start(ch)
+	inf.WaitForCacheSync(ch)
```
stlaz, line 93: use `ctx.Done()`

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
```diff
@@ -0,0 +1,123 @@
...
+
+	syncCtx := factory.NewSyncContext(mcsName, recorder)
+	if err := cont.Sync(ctx, syncCtx); err != nil {
+		return err
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "Successfully rotated MCS CA + certs\n")
```
stlaz, line 100: Explain what's happening next so that the user is not confused as to why the command hasn't finished yet

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (o *RegenerateMCOOptions) updateSecrets(ctx context.Context) error {
+	clientConfig, err := o.RESTClientGetter.ToRESTConfig()
+	if err != nil {
+		return err
+	}
+	clientset, err := kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
```
stlaz, line 29: nit: if you had a `Complete()` method, you could store these in your `RegenerateMCOOptions` and would not have to reinitialize here in case of a call from `func (o *RegenerateMCOOptions) Run(ctx context.Context) error {`

It would also help you with unit testing where you'd be able to supply your fake clients to the method.

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -0,0 +1,105 @@
...
+	// Get the updated CA cert
+	mcoSecrets := clientset.CoreV1().Secrets(mcoNamespace)
+	mcsSecret, err := mcoSecrets.Get(ctx, newMCSCASecret, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("cannot read MCS secret: %w", err)
+	}
+
+	caCert, exists := mcsSecret.Data[corev1.TLSCertKey]
+	if !exists {
+		return fmt.Errorf("could not find MCS CA cert at %s after rotation", corev1.TLSCertKey)
```
stlaz, line 40:
```suggestion
		return fmt.Errorf("could not find MCS CA cert at %s", corev1.TLSCertKey)
```
Maybe the user did not perform the rotation yet.

Also, the `%s` is not helpful here, you'll want to call out the location of the secret, too.

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		_, isIgn, err := unstructured.NestedMap(userDataIgn.(map[string]interface{}), ignFieldIgnition)
+		if !isIgn {
+			// Didn't find ignition in user-data, warn but continue
+			fmt.Fprintf(o.IOStreams.Out, "Unable to find ignition in user-data, skipping secret %s\n", secret.Name)
+			continue
+		}
+
+		caSlice, isSlice, err := unstructured.NestedFieldNoCopy(userDataIgn.(map[string]interface{}), ignFieldIgnition, ignFieldSecurity, ignFieldTLS, ignFieldCA)
```
stlaz, line 75:
```suggestion
		caSlice, isSlice, err := unstructured.NestedFieldNoCopy(userDataIgn.(map[string]interface{}), ignCAPath...)
```
where
```
ignCAPath = []string{ignFieldIgnition, "security", "tls",  "certificateAuthorities"}
```
defined above instead of the constants

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -0,0 +1,105 @@
...
+		if !isSlice || err != nil {
+			return fmt.Errorf("failed to find certificateauthorities field in ignition (secret %s): %w", secret.Name, err)
+		}
+		if len(caSlice.([]interface{})) == 1 {
+			caSlice.([]interface{})[0].(map[string]interface{})[ignFieldSource] = dataurl.EncodeBytes(caCert)
```
stlaz, line 80: `unstructured.SetNestedStringSlice()` should work here I think

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -0,0 +1,105 @@
...
+			// Didn't find ignition in user-data, warn but continue
+			fmt.Fprintf(o.IOStreams.Out, "Unable to find ignition in user-data, skipping secret %s\n", secret.Name)
+			continue
+		}
+
+		caSlice, isSlice, err := unstructured.NestedFieldNoCopy(userDataIgn.(map[string]interface{}), ignFieldIgnition, ignFieldSecurity, ignFieldTLS, ignFieldCA)
+		if !isSlice || err != nil {
+			return fmt.Errorf("failed to find certificateauthorities field in ignition (secret %s): %w", secret.Name, err)
+		}
+		if len(caSlice.([]interface{})) == 1 {
```
stlaz, line 79: nit: invert the condition to reduce nesting:
```
if len(caSlice.([]interface{})) > 1 {
    return fmt.Errorf("additional CAs detected, cannot modify automatically. Aborting")
}
<set the string slice here>
```

vendor/github.com/vincent-petithory/dataurl/LICENSE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
```
stlaz, line 1: please move vendoring into a separate commit

stlaz (changes requested):
/lgtm cancel

Could we also add unit tests, please? I did not realize that yesterday, sorry :neutral_face:

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,128 @@
...
+
+		# Update the user-data secrets to use new MCS certs
+		oc adm certificates update-mco-user-data
+	`)
+)
+
+type RegenerateMCOOptions struct {
+	RESTClientGetter genericclioptions.RESTClientGetter
+
+	modifyUserData bool
```
deads2k, line 75: followup: CLIs expose all provided fields for external composition

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
stlaz: nvm, types are probably correct

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
```diff
@@ -0,0 +1,128 @@
...
+			cmdutil.CheckErr(o.RunUserDataUpdate(context.Background()))
+		},
+	}
+
+	return cmd
+}
+
+// AddFlags registers flags for a cli
+func (o *RegenerateMCOOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&o.modifyUserData, "update-user-data", o.modifyUserData, "If true, automatically update user-data secrets in machine-api namespace. Not useful if node scaling not backed by MachineSet.")
```
deads2k, line 127:
```suggestion
	cmd.Flags().BoolVar(&o.modifyUserData, "update-ignition", o.modifyUserData, "If true, automatically update user-data secrets in machine-api namespace. Not useful if node scaling not backed by MachineSet.")
```

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
```diff
@@ -0,0 +1,123 @@
...
+		err := mcoPods.Delete(ctx, pod.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return fmt.Errorf("cannot delete MCS pod %s: %w", pod.Name, err)
+		}
+	}
+
+	// TODO maybe add a watcher to make sure the MCS daemonset is ready here
+
+	if o.modifyUserData {
+		return o.updateSecrets(ctx)
```
deads2k, line 120:
```suggestion
		return o.RunUserDataUpdate(ctx)
```

?

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
yuqi-zhang: Yes, sorry, so the data looks something like this:

```
"security":{"tls":{"certificateAuthorities":[{"source":"data:text/plain;charset=utf-8;base64,...","verification":{}}]}}
```

i.e. certificateAuthorities is a [] of ca objects as defined by ignition. Short of vendoring ignition, and doing multiple versions of parsing (ignition is very strict) this was the simplest way I could come up with. I opted not to use `SetNestedStringSlice` since it's not a string slice in here

yuqi-zhang:
Rebased, renamed the commands, updated to use ctx.Done, hopefully addressed the other issues, and re-tested to be still working on a fresh cluster.

Thanks for all the review!

pkg/cli/admin/ocpcertificates/regeneratemco/command.go
cgwalters: It's both...it's ignition in user data.  The actual API objects we are modifying are called e.g. `worker-user-data`.

PR #1451, opened 2023-06-12
require an explicit output format when using --dry-run

deads2k:
per @stlaz 

I cannot do a -okubeconfig because its possible that multiple kubeconfigs need to be written due to merging rules.  This at least makes the format explicit for people.

/assigni @stlaz


stlaz:
As far as I can tell, the PR only adds complexity as to how the --dry-run option is handled.

The use-case I had in mind was protecting my trust bundle files from being overriden which might lead to broken/lost trusts since we override all of the files.

Dry-run should be capable to at least report which files will be written/overriden by the command.

PR #1453, opened 2023-06-12
Minor updates to CLI help text

bergerhoffer:
A few help text updates for grammar/consistency, mainly targeted at content that also shows up in the OCP CLI reference docs.


soltysh:
Also manually adding the valid-bug jira, due to the same reason. 
/label jira/valid-bug

PR #1454, opened 2023-06-12
add a command to copy content to every node

deads2k:
`oc adm clusteroperator copy-to-node nodes --all --copy=../marker.txt=/etc/kubernetes/proof.txt`

it needs some refinement, but the last commit shows how this can be done.  I don't think the refinement is structural.

/cc @rphillips  @cgwalters 
/assign @yuqi-zhang


pkg/cli/clusteroperator/copytonode/command.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+func NewCmdCopyToNode(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
+	o := NewRestartKubelet(restClientGetter, streams)
+
+	cmd := &cobra.Command{
+		Use:                   "copy-to-node",
+		DisableFlagsInUseLine: true,
+		Short:                 i18n.T("Copies specified files to the node."),
```
cgwalters, line 55: There's actually `oc rsync` today, it just doesn't support `oc rsync foo node:`.

(And a bunch of backing infrastructure there that maybe we could share some with)

pkg/cli/clusteroperator/copytonode/copy_to_node.go
```diff
@@ -0,0 +1,166 @@
...
+		return cleanupFn, nil
+	}
+
+	copyCommands := []string{}
+	for source, destination := range secretDataKeyToFilename {
+		sourcePath := filepath.Join("/source-data", source)
+		destPath := filepath.Join("/host-root", destination)
+		parentDir := filepath.Dir(destPath)
+		copyCommands = append(copyCommands, fmt.Sprintf("mkdir -p %q", parentDir))
+		copyCommands = append(copyCommands, fmt.Sprintf("cp --dereference -fr %q %q", sourcePath, destPath))
```
cgwalters, line 66: Not fatal, but the problem with generating shell script here is that if any of the source/dest have shell metacharacters we won't be properly handling escaping.  There's e.g. https://pkg.go.dev/github.com/alessio/shellescape

We could also generate a pod per file to copy and just directly inject `cp` into the commands, although that's a bit messier with the semantic to make parent directories here.

pkg/cli/clusteroperator/copytonode/copy_to_node.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	createPodFn := func(ctx context.Context, namespaceName, nodeName, imagePullSpec string) (*corev1.Pod, error) {
+		restartObj := pod.DeepCopy()
+		restartObj.Namespace = namespaceName
+		restartObj.Spec.NodeName = nodeName
+		restartObj.Spec.Containers[0].Image = imagePullSpec
+		restartObj.Spec.Containers[0].Command[2] = strings.ReplaceAll(
+			restartObj.Spec.Containers[0].Command[2],
+			"### COPY_COMMANDS_HERE",
```
cgwalters, line 76: It seems simpler here to just append to the argument list, i.e. the default command is 

```
- command:
        - /bin/bash
        - -c
```

and we just do:

```
restartObj.Spec.COntainers[0].Command = append(Command, `#!/bin/bash ...`)
```
or so

pkg/cli/clusteroperator/pernodepod/runtime.go
```diff
@@ -0,0 +1,242 @@
...
+	if watchErr != nil {
+		// TODO inspect and report the pod state.
+		retErr := fmt.Errorf("node/%v failed waiting for restart using --namespace=%v pod/%v: %w", node.Name, namespaceName, createdPod.Name, watchErr)
+		fmt.Fprintln(r.ErrOut, retErr.Error())
+		return retErr
+	}
+	finalPodState := lastWatchEvent.Object.(*corev1.Pod)
+
+	switch {
+	case finalPodState.Status.Phase == corev1.PodSucceeded:
```
cgwalters, line 182: The semantics here get a bit weird, because this pod is rebooting the node...so the pod can't enter "completed" status really, unless we do the reboot *asynchronously* of the pod exiting successfully, which is what I was arguing with `systemd-run`.

In the general case this problem of "reporting reboot success" is so messy that in other places what we've done is just to fire off a "reboot" command, and then poll the node's `/proc/sys/kernel/boot-id`.  In kube that's actually part of the node status object, see `bootID` in  https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#nodesysteminfo-v1-core

pkg/cli/clusteroperator/pernodepod/command.go
```diff
@@ -0,0 +1,141 @@
...
+	return ctx, cancel
+}
+
+// AddFlags registers flags for a cli
+func (o *PerNodePodOptions) AddFlags(cmd *cobra.Command) {
+	o.PrintFlags.AddFlags(cmd)
+	o.ResourceBuilderFlags.AddFlags(cmd.Flags())
+
+	cmd.Flags().BoolVar(&o.DryRun, "dry-run", o.DryRun, "Set to true to use server-side dry run.")
+	cmd.Flags().StringVar(&o.Parallelism, "parallelism", o.Parallelism, "parallelism is a raw number or a percentage of the nodes we need to restart at the same time.")
```
cgwalters, line 74: In this context we're not restarting...just "applying" or so right?

pkg/cli/clusteroperator/restartkubelet/command.go
```diff
@@ -0,0 +1,92 @@
...
+		oc adm clusteroperators restart-kubelet nodes --all
+
+		# Restart all the nodes,  20 nodes at a time
+		oc adm clusteroperators restart-kubelet nodes --all --parallelism=20
+
+		# Restart all the nodes,  15% at a time
+		oc adm clusteroperators restart-kubelet nodes --all --parallelism=15%
+
+		# Restart all the masters at the same time
+		oc adm clusteroperators restart-kubelet nodes -l node-role.kubernetes.io/master --parallelism=100%`)
```
cgwalters, line 34: Hmm...ISTM we should be a bit more opinionated here.  First, the way the MCO works is to roll one controlplane and one worker simultaneously.  Instead of operating on nodes, we could actually operate on pools...

But even that aside I think we shouldn't document trying to roll the entire control plane simultaneously, right?  I mean, let's document doing that for the *workers* perhaps.

cgwalters (commented):
Just a skim

pkg/cli/clusteroperator/copytonode/command.go
deads2k: > There's actually oc rsync today, it just doesn't support oc rsync foo node:.
>
> (And a bunch of backing infrastructure there that maybe we could share some with)

Doing a single pod versus doing large numbers of nodes presents a different problem.  Rather than setting up oc client to pod connections for every single node, this creates an intermediate secret which is then mounted, so the network connection remains internal to the cluster.

pkg/cli/clusteroperator/pernodepod/runtime.go
deads2k: > The semantics here get a bit weird, because this pod is rebooting the node...so the pod can't enter "completed"

The output suggests that we do indeed enter a succeeded state. 

```yaml
    state:
      terminated:
        containerID: cri-o://eed849055daccd50d2e730a8f5b8e694b0751c84c8597151ce1460f5620da49d
        exitCode: 0
        finishedAt: "2023-06-13T11:12:39Z"
        reason: Completed
        startedAt: "2023-06-13T11:12:33Z"
  hostIP: 10.0.128.4
  phase: Succeeded
 ```

pkg/cli/clusteroperator/restartkubelet/command.go
deads2k: > But even that aside I think we shouldn't document trying to roll the entire control plane simultaneously, right? I mean, let's document doing that for the workers perhaps.

A kubelet restart is non-disruptive.  It doesn't impact pod lifecycles. I think restarting all of them concurrently is acceptable.

pkg/cli/clusteroperator/copytonode/copy_to_node.go
deads2k: fixed

pkg/cli/clusteroperator/pernodepod/runtime.go
cgwalters: OK.  Just saying I've seen historical issues in this area.

cgwalters:
My biggest question on this is - do we really need this powerful of a CLI primitive?  Or even if we do, I think we should have a much higher level `oc adm ocp-certificates rotate-kubelet-bootstrap` or something that synthesizes the desired contents for `/etc/kubernetes/kubeconfig` in the correct way.

pkg/cli/clusteroperator/pernodepod/runtime.go
deads2k: > OK. Just saying I've seen historical issues in this area.

I know we've had issues rebooting nodes, I think restarting kubelets will be ok for this  usage.

deads2k:
> Or even if we do, I think we should have a much higher level oc adm ocp-certificates rotate-kubelet-bootstrap or something that synthesizes the desired contents for /etc/kubernetes/kubeconfig in the correct way.

Agreed.  I'd like to merge pieces as they're ready.

deads2k:
/override ci/prow/e2e-aws-ovn 

hpa CRD failure noted a few weeks back

/override ci/prow/e2e-aws-ovn-upgrade

deads2k:
This is next major piece in line for a couple pieces.  will make other changes post-merge if desired.

PR #1456, opened 2023-06-13
Add tls-server-name when property exists in kubeconfig

kwoodson:
Ran into #1457 today when using tsh, kubectl, and oc. I was able to track this down and this patch was able to fix the missing field issue.


kwoodson:
@deads2k @mfojtik Could someone please review this PR? This is a problem for us in our environment when using teleport.

lalan7:
@maszulik Can you help review this PR? Thanks

soltysh (approved):
/lgtm
/approve

PR #1452, opened 2023-06-12
create command to create new adminkubeconfig

deads2k:
Add `oc config new-admin-kubeconfig` to create a new admin.kubeconfig using an existing connection.

/assign @stlaz

PR #1458, opened 2023-06-13
add command to create new bootstrap kubeconfig for kubelet

deads2k:
builds on https://github.com/openshift/oc/pull/1452

`oc config new-kubelet-bootstrap-kubeconfig > ~/Downloads/bootstrap.kubeconfig`

/assign @rphillips

### DIFF
--- a/pkg/cli/config/adminkubeconfig/new_admin_kubeconfig.go
+++ b/pkg/cli/config/adminkubeconfig/new_admin_kubeconfig.go
@@ -26,7 +26,7 @@ import (
 	"k8s.io/kubectl/pkg/util/templates"
 )
 
-type NewAdminKubeConfigOptions struct {
+type NewAdminKubeconfigOptions struct {
 	RESTClientGetter genericclioptions.RESTClientGetter
 
 	genericclioptions.IOStreams
@@ -45,20 +45,21 @@ var (
 )
 
 const (
+	// adminKubeconfigClientCAConfigMap is described in https://github.com/openshift/api/blob/master/tls/docs/kube-apiserver%20Client%20Certificates/README.md#kube-apiserver-admin-kubeconfig-client-ca
 	adminKubeconfigClientCAConfigMap = "admin-kubeconfig-client-ca"
 
 	tenYears = 24 * time.Hour * 365 * 10
 )
 
-func NewAdminKubeconfigOptions(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *NewAdminKubeConfigOptions {
-	return &NewAdminKubeConfigOptions{
+func NewNewAdminKubeconfigOptions(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *NewAdminKubeconfigOptions {
+	return &NewAdminKubeconfigOptions{
 		RESTClientGetter: restClientGetter,
 		IOStreams:        streams,
 	}
 }
 
-func NewCmdConfigRefreshCABundle(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
-	o := NewAdminKubeconfigOptions(restClientGetter, streams)
+func NewCmdNewAdminKubeconfigOptions(restClientGetter genericclioptions.RESTClientGetter, streams genericclioptions.IOStreams) *cobra.Command {
+	o := NewNewAdminKubeconfigOptions(restClientGetter, streams)
 
 	cmd := &cobra.Command{
 		Use:                   "new-admin-kubeconfig",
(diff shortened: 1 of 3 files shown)


pkg/cli/config/kubeletbootstrapkubeconfig/new_kubelet_bootstrap_kubeconfig.go
```diff
@@ -0,0 +1,158 @@
...
+	if len(token) == 0 {
+		return fmt.Errorf("token is missing from the secret: %w", err)
+	}
+
+	serverCABundle, err := refreshcabundle.GetCABundleToTrustKubeAPIServer(ctx, r.KubeClient)
+	if err != nil {
+		return fmt.Errorf("unable to get the CA bundle from the cluster: %w", err)
+	}
+
+	newConfig := clientcmdapi.Config{
```
cgwalters, line 130: Probably worth adding a comment here like

```
// keep in sync with https://github.com/openshift/machine-config-operator/blob/3d84f653e08d760d446442ddc80c3da21d8d7e59/pkg/server/cluster_server.go#L167
```

pkg/cli/config/kubeletbootstrapkubeconfig/new_kubelet_bootstrap_kubeconfig.go
```diff
@@ -0,0 +1,158 @@
...
+			},
+		},
+		CurrentContext: "bootstrap",
+	}
+	newKubeletBootstrapConfig, err := clientcmd.Write(newConfig)
+	if err != nil {
+		return fmt.Errorf("unable to serialize new kubeconfig: %w", err)
+	}
+
+	fmt.Fprintln(r.Out, string(newKubeletBootstrapConfig))
```
cgwalters, line 155: Under what scenario does the admin need to see this directly instead of having us directly rotate the file on the node?

I'm OK creating two separate primitives here to be clear, I just don't quite understand why we are instead of having a more opinionated wired-together flow.

pkg/cli/config/adminkubeconfig/new_admin_kubeconfig.go
```diff
@@ -0,0 +1,267 @@
...
+		The key is produced locally and is not persisted to disk.  The public half is pushed to the cluster
+		for the kube-apiserver to trust the locally created admin.kubeconfig.`))
+
+	newAdminKubeconfigExample = templates.Examples(`
+		# Generate a new admin kubeconfig
+		oc config new-admin-kubeconfig`)
+)
+
+const (
+	adminKubeconfigClientCAConfigMap = "admin-kubeconfig-client-ca"
```
cgwalters, line 48: Worth a comment like
```
// adminKubeconfigClientCAConfigMap is described in https://github.com/openshift/api/blob/master/tls/docs/kube-apiserver%20Client%20Certificates/README.md#kube-apiserver-admin-kubeconfig-client-ca
```
perhaps?

pkg/cli/config/adminkubeconfig/new_admin_kubeconfig.go
```diff
@@ -0,0 +1,267 @@
...
+
+	tenYears = 24 * time.Hour * 365 * 10
```
cgwalters, line 50: At some point we'll want to centralize our many "tenYears" constants...

pkg/cli/config/adminkubeconfig/new_admin_kubeconfig.go
```diff
@@ -0,0 +1,267 @@
...
+				Name:      adminKubeconfigClientCAConfigMap,
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			return fmt.Errorf("unable to create starting --namespace=openshift-config configmap/%s : %w", adminKubeconfigClientCAConfigMap, err)
+		}
+	} else if err != nil {
+		return fmt.Errorf("unable to read configmap %w", err)
+	}
+	caBundle, err := combineCABundles(existingConfigMap.Data["ca-bundle.crt"], string(signerCertBytes))
```
cgwalters, line 152: This is ensuring we continue trust the old keys for now?  Worth a comment if so.

cgwalters:
None of my comments are blocking to be clear.

pkg/cli/config/kubeletbootstrapkubeconfig/new_kubelet_bootstrap_kubeconfig.go
deads2k: > Under what scenario does the admin need to see this directly instead of having us directly rotate the file on the node?
>
> I'm OK creating two separate primitives here to be clear, I just don't quite understand why we are instead of having a more opinionated wired-together flow.

if our rotation command fails, having this available allows us to do the hard part (content of the file), while allowing any distribution method.  Perhaps our "put it on the nodes" fails and we need to re-run it.  Or perhaps its doomed in some scenario and they need to SSH it. This gives us those options.

pkg/cli/config/adminkubeconfig/new_admin_kubeconfig.go
deads2k: > At some point we'll want to centralize our many "tenYears" constants...

Just in case 10 years changes in duration some day? ;)

deads2k:
Comments fixed, though I decline to be worried about 10y changing durations on us over time :)

pkg/cli/config/adminkubeconfig/new_admin_kubeconfig.go
cgwalters: I mean more that I can imagine us wanting to have a more consistent policy for certs, e.g. something much closer to Let's Encrypt expiry at e.g. 3 months to ensure that rotation really works.  To do that we'd want to centralize these currently disparate constants.

Just a side comment, not saying we need to change anything now.

deads2k:
bringing in for a complete set to match directions.

PR #1459, opened 2023-06-13
allow running a command while the kubelet is off

deads2k:
`oc adm clusteroperator restart-kubelet nodes --all --parallelism=2 --command='rm -f /host-root/var/lib/kubelet/kubeconfig'`

this should allow us to quickly snip some files.

/assign @rphillips

### DIFF
--- a/pkg/cli/clusteroperator/pernodepod/runtime.go
+++ b/pkg/cli/clusteroperator/pernodepod/runtime.go
@@ -124,7 +124,7 @@ func (r *PerNodePodRuntime) Run(ctx context.Context, prePodHookFn PrePodHookFunc
 					if !stillReady {
 						return
 					}
-					if restartErr := r.RestartNode(ctx, createPodFn, nsName, node); restartErr != nil {
+					if restartErr := r.HandleNode(ctx, createPodFn, nsName, node); restartErr != nil {
 						errCh <- restartErr
 					}
 				case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


pkg/cli/clusteroperator/pernodepod/runtime.go
```diff
@@ -143,7 +143,24 @@ func (r *PerNodePodRuntime) Run(ctx context.Context, prePodHookFn PrePodHookFunc
...
+		}
+		if condition.Status == corev1.ConditionTrue {
+			break
+		}
+		notReadyMessage = fmt.Sprintf("%s since %v - %s: %s", condition.Status, condition.LastTransitionTime.Format(time.RFC3339), condition.Reason, condition.Message)
+		break
+	}
+	if len(notReadyMessage) > 0 {
+		retErr := fmt.Errorf("nodes/%v is not ready, %v", node.Name, notReadyMessage)
+		fmt.Fprintln(r.ErrOut, retErr.Error())
```
cgwalters, line 160: Is there a reason to both print an return the error here?  That usually results in double logging unnecessarily.

cgwalters (commented):
One nonblocking question

pkg/cli/clusteroperator/pernodepod/runtime.go
deads2k: > Is there a reason to both print an return the error here? That usually results in double logging unnecessarily.

It does, it's not perfect.  However if you don't report until the command finishes, there is a long feedback delay.  Given enough nodes, it's probably worthwhile to collect just the errors at the end anyway.  With only 6, it does look weird.

pkg/cli/clusteroperator/restartkubelet/command.go
```diff
@@ -46,7 +46,7 @@ func NewRestartKubelet(restClientGetter genericclioptions.RESTClientGetter, stre
 	return &RestartKubeletOptions{
 		PerNodePodOptions: pernodepod.NewPerNodePodOptions(
 			"openshift-restart-kubelet-",
-			"copied to node",
+			"restarted kubelet",
```
stlaz, line 53: original copy-pasta error I assume

stlaz (approved):

PR #1426, opened 2023-05-17
OCPNODE-1656: oc image extract: Introduce IDMS as alternative source

ardaguclu:
This PR;

* adds foundational work to support IDMS(aka ImageDigestMirrorSet) as an alternative source
in addition to the ICSP. 
* introduces `--idms-file` in `oc image extract` depending on the foundational work above ^
* deprecates `--icsp-file` flag in `oc image extract` command to suggest `--idms-file` instead

PR #1464, opened 2023-06-15
OCPBUGS-15012: oc image extract: idms-file flag map to IDMSFile field

ardaguclu:
`--idms-file` flag was introduced in oc image extract command with this PR https://github.com/openshift/oc/pull/1426/, however `-idms-file` is incorrectly mapping to `ICSPFile` field instead `IDMSFile`.

This PR fixes it.

/assign @QiWang19


ardaguclu:
@QiWang19 would you mind tagging this simple fix?. Thanks.

QiWang19 (commented):
/lgtm

PR #1468, opened 2023-06-15
add reboot-nodes

deads2k:
`oc adm reboot-machine-config-pool mcp --all`

or 

`oc adm reboot-machine-config-pool mcp/master`

/assign @yuqi-zhang


pkg/cli/admin/rebootmachineconfigpool/reboot_machine_config_pool.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+	return metav1.UpdateOptions{
+		FieldManager: RebootMachineConfigPoolFieldManager,
+	}
+}
+
+func (r *RebootMachineConfigPoolRuntime) rebootMachineConfigPool(ctx context.Context, machineConfigPool *MachineConfigPool) error {
+	// do something better.
+	if machineConfigPool.Spec.MachineConfigSelector == nil {
```
deads2k, line 123: check these

pkg/cli/admin/rebootmachineconfigpool/reboot_machine_config_pool.go
```diff
@@ -0,0 +1,200 @@
...
+	case err != nil:
+		return err
+
+	default:
+		currMachineConfig := &MachineConfig{}
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(currMachineConfigUnstructured.Object, currMachineConfig); err != nil {
+			return err
+		}
+		machineConfig.ResourceVersion = currMachineConfig.ResourceVersion
+		// TODO figure out how to read the existing int and set a bigger one.
```
deads2k, line 170: this TODO is quite important

yuqi-zhang: I think probably having an annotation on the MC object to match the contents would be easiest, and better to track instead of reading the machineconfig directly. Those can be updated in tandem. WDYT?

pkg/cli/admin/rebootmachineconfigpool/reboot_machine_config_pool.go
yuqi-zhang: So there's two things to this:

1. the `master` and `worker` pools will always be required be in a cluster (even something like SNO, it would simply have an empty worker pool), strictly named (and labelled) as such
2. all custom pools must inherit from `worker` pool, so by updating worker pool, you also roll out any custom pool e.g. `infra` you may have

So to simply this for us, I wonder if it may be easier to only allow specifying `worker` and `master` and strictly match it as such here? A user rebooting `infra` pool directly isn't very valuable, since if they reboot `worker` they will just reboot it again.

pkg/cli/admin/rebootmachineconfigpool/reboot_machine_config_pool.go
deads2k: > I think probably having an annotation on the MC object to match the contents would be easiest, and better to track instead of reading the machineconfig directly. Those can be updated in tandem. WDYT?

no opinion.  I don't know how to change the int either way.

pkg/cli/admin/rebootmachineconfigpool/reboot_machine_config_pool.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	if len(machineConfigPool.Spec.MachineConfigSelector.MatchLabels) == 0 {
+		return fmt.Errorf("machineconfigpools/%v cannot be rebooted with this command due to missing .spec.machineConfigSelector.matchLabels", machineConfigPool.Name)
+	}
+
+	machineConfig := restartTemplate.DeepCopy()
+	if machineConfig.Labels == nil {
+		machineConfig.Labels = map[string]string{}
+	}
+	machineConfig.Name = fmt.Sprintf("95-certrotation-rebooter-%s", machineConfigPool.Name)
```
cgwalters, line 138: This isn't strictly scoped to cert rotation, so we could probably s/certrotation/oc-initiated-reboot/ maybe?

deads2k: Updated

pkg/cli/admin/rebootmachineconfigpool/command.go
```diff
@@ -0,0 +1,112 @@
...
+
+		Does not wait for the reboot to complete, only initiates it.  This command will honor paused pools.
+		Degraded, failed, or otherwise not healthy nodes will not restart.
+
+		Experimental: This command is under active development and may change without notice.
+	`)
+
+	rebootMachineConfigPoolExample = templates.Examples(`
+		# Reboot all MachineConfigPools
+		oc adm clusteroperator reboot-machine-config-pool machineconfigpools --all
```
yuqi-zhang, line 27: nit: this is just `oc adm reboot-machine-config-pool`, no clusteroperator subcommand

Also, `--all` isn't a flag

deads2k: I promise to fix it later.

pkg/cli/admin/rebootmachineconfigpool/command.go
```diff
@@ -0,0 +1,112 @@
...
+
+		# Reboot masters
+		oc adm clusteroperator reboot-machine-config-pool mcp/master`)
```
yuqi-zhang, line 30: same here

deads2k:
going with this so that we have something for QE to test with tonight.  Will fix other comments in a followup.

pkg/cli/admin/rebootmachineconfigpool/reboot_machine_config_pool.go
```diff
@@ -0,0 +1,242 @@
...
+	if !ok || err != nil {
+		return ""
+	}
+	existingContent, ok, err := unstructured.NestedString(fileList[0].(map[string]interface{}), "contents", "source")
+	if !ok || err != nil {
+		return ""
+	}
+
+	// we write like `data:,1%0A`
+	if len(existingContent) < len(`data:,1%0A`) {
```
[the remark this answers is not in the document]
deads2k: once we get to 10, the length will be greater than the static string, right?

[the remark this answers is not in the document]
deads2k: just tested, went to 12 just fine.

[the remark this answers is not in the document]
yuqi-zhang: Hmm, I could have sworn I saw that as `if len(existingContent) == len(`data:,1%0A`) {` :facepalm: ignore me

PR #1341, opened 2023-02-15
OCPNODE-1580: Add --print-mirror-instructions to oc adm release mirror to allow idms instructions

QiWang19:
story: https://issues.redhat.com/browse/OCPNODE-1580

Add `--print-mirror-instructions` flag to command oc adm release.
The printout instructions of the command will keep it as is if the flag is unset or set to `icsp`.
If set to `idms`, update the instruction output of `oc adm release mirror` to ImageDigestMirrorSet example.
The new ImageDigestMirrorSet examples are:
- use `imageDigestSources` field in install-config.yaml[1] that will creates ImageDigestMirrorSet manifests,
- manifest example of ImageDigestMirrorSet CRD

[1] Add `imageDigestSources` to install-config: https://github.com/openshift/installer/pull/6235

result of this patch:
```
LOCAL_REGISTRY=localhost:5000
LOCAL_REPOSITORY=ocp4/openshift4
LOCAL_SECRET_JSON=/home/qiwan/cluster/mysecret
sudo podman run -d -p 5000:5000 docker.io/registry

$ oc adm release mirror -a ${LOCAL_SECRET_JSON} --from=registry.ci.openshift.org/ocp/release:4.14.0-0.ci-2023-04-04-023533 --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}

phase 0:
  localhost:5000 ocp4/openshift4 blobs=597 mounts=0 manifests=189 shared=6

info: Planning completed in 36.98s
info: Dry run complete

Success
Update image:  localhost:5000/ocp4/openshift4:4.14.0-0.ci-2023-04-04-023533-x86_64
Mirror prefix: localhost:5000/ocp4/openshift4

To use the new mirrored repository to install, add the following section to the install-config.yaml:

imageDigestSources:
- mirrors:
  - localhost:5000/ocp4/openshift4
  source: registry.ci.openshift.org/ocp/4.14-2023-04-04-023533
- mirrors:
  - localhost:5000/ocp4/openshift4
  source: registry.ci.openshift.org/ocp/release


To use the new mirrored repository for upgrades, use the following to create an ImageDigestMirrorSet:

apiVersion: config.openshift.io/v1
kind: ImageDigestMirrorSet
metadata:
  name: example
spec:
  imageDigestMirrors:
  - mirrors:
    - localhost:5000/ocp4/openshift4
    source: registry.ci.openshift.org/ocp/4.14-2023-04-04-023533
  - mirrors:
[...]

PR #1330, opened 2023-01-26
pkg/cli/admin/upgrade: Add post-period to space to some error messages

wking:
Cleaning up some nits from 6c8d9352d3 (#986).

PR #1472, opened 2023-06-15
OTA-932: pkg/cli/admin/update: Rename "upgrade" to "update" and similar

wking:
The outgoing `upgrade` phrasing still works, but now logs deprecation warnings.  We may never remove the outgoing `upgrade` phrasing, because it seems unlikely to be worth the compatibility churn.  But moving the canonical phrasing to `update` aligns with openshift/openshift-docs@a68b607325 (openshift/openshift-docs#40525).  And logging the deprecation warnings makes the new canonical phrasing more obvious, which should help gradually transition existing consumers, because the more consistently folks use the canonical phrasing, the easier it is to find it with grep-searches when searching scripts and similar.


wking:
Testing with a local build:

```console
$ go build ./cmd/oc
$ ./oc adm update  # new status approach
Cluster version is 4.13.2
...
$ ./oc adm upgrade  # old status approach, confirming a warning is logged
oc adm update is deprecated.  Use 'oc adm update ...' instead
Cluster version is 4.13.2
...
$ ./oc adm update --allow-upgrade-with-warnings --to 4.99.99  # two command line flags also moved, and using them gets a warning logged, but isn't fatal (the non-existed release used for testing is what fails this call)
Flag --allow-upgrade-with-warnings has been deprecated, please use --allow-update-with-warnings instead
error: the update is not one of the possible targets: 4.13.3. specify --to-image to continue with the update.
$ ./oc adm upgrade --help | grep -1 upgrade  # the deprecated options are hidden
Aliases:
update, upgrade

```

So that looks good to me :)

wking:
Rebased around #1341 to resolve diff-context-overlap with 2098a8515 -> 7b9597252.

cgwalters:
What's the motivation behind logging deprecation warnings?  I mean really:
> we may never remove the outgoing upgrade phrasing, because it seems unlikely to be worth the compatibility churn.

Exactly. Many tools (e.g. `dnf`) supports both `dnf upgrade` and `dnf update` as aliases, we did the same change for rpm-ostree a while ago.

If we're never going to remove it, let's not nag users to change pointlessly.

cgwalters:
Confusingly, `apt update` and `apt upgrade` are totally different things.  It looks like `zypper update` exists, but not `zypper upgrade`.

pkg/cli/admin/update/update.go
```diff
@@ -52,64 +52,77 @@ func NewOptions(streams genericclioptions.IOStreams) *Options {
...
+			if cmd.CalledAs() != "update" {
+				fmt.Fprintf(o.ErrOut, "%s is deprecated.  Use '%s ...' instead\n", cmd.CommandPath(), cmd.CommandPath())
+			}
+			return nil
+		},
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Complete(f, cmd, args))
 			kcmdutil.CheckErr(o.Run())
 		},
+		Aliases: []string{"upgrade"},
```
LalatenduMohanty, line 107: :+1:

LalatenduMohanty: Do we need to add some code to warn people saying `upgrade` is deprecated and they should start using `update` even if we will never remove `upgrade` alias.

LalatenduMohanty: Deleted my previous comment about the warning message about `upgraded` because I saw that we already printing the message.

LalatenduMohanty:
> If we're never going to remove it, let's not nag users to change pointlessly.

The change from `upgrade` to `update` is not single sub-command change as we have used upgrade string in other places e.g. flags etc.  So we wanted to push folks to use the updated names without breaking older scripts.  

But I can live without the warning message as well. If you think strongly about it. 

```
$ oc adm upgrade --help | grep -i upgrade
Check on upgrade status or upgrade the cluster to a newer version
 This command assists with cluster upgrades. If no arguments are passed the command will retrieve the current version info and display whether an upgrade is in progress or whether any errors might prevent an upgrade, as well as show the suggested updates available to the cluster. Information about compatible updates is periodically retrieved from the update server and cached on the cluster - these are updates that are known to be supported as upgrades from the current version.
 Passing --to=VERSION or --to-image=IMAGE will upgrade the cluster to one of the available updates or report an error if no such version exists. The cluster will then upgrade itself and report status that is available via "oc get clusterversion" and "oc describe clusterversion".
 Passing --to-multi-arch will upgrade the cluster from a single-architecture to a multi-architecture cluster at the current version.
 If there are no versions available, or a bug in the cluster version operator prevents updates from being retrieved, --to-image may be combined with the more powerful and dangerous --allow-explicit-upgrade. This instructs the cluster to upgrade to the contents of the specified release image, regardless of whether that upgrade is known to be recommended for the current version. While rolling back to a previous patch (z stream) version (4.1.2 -> 4.1.1) may be safe, upgrading more than one minor version ahead (4.1 -> 4.3) or downgrading one minor version (4.2 -> 4.1) is likely to cause data corruption or to completely break a cluster.
 If the cluster is already being upgraded, or if the cluster is reporting a failure or other error, the update will not be triggered.  It is usually best to give these conditions time to resolve, or to actively work to resolve them.  But if you decide to trigger the update regardless of these concerns, use --allow-upgrade-with-warnings.
 The cluster may report that the upgrade should not be performed due to a content verification error or update precondition failures such as operators blocking upgrades. Do not upgrade to images that are not appropriately signed without understanding the risks of upgrading your cluster to untrusted code. If you must override this protection use the --force flag.
  oc adm upgrade
  oc adm upgrade --to-latest=true
    --allow-explicit-upgrade=false:
	Upgrade even if the upgrade target is not listed in the available versions list.
	Allows upgrade to a version when it is supported but not recommended for updates.
    --allow-upgrade-with-warnings=false:
	Upgrade even if an upgrade is in process or a cluster error is blocking the update.
	If an upgrade has been requested but not yet downloaded, cancel the update. This has no effect once the update has started.
	Forcefully upgrade the cluster even when upgrade release image validation fails and the cluster is reporting errors.
	Specify the version to upgrade to. The version must be on the list of available updates.
	Provide a release image to upgrade to.
	Upgrade current version to multi architecture.
  oc adm upgrade --to=VERSION [flags] [options]
```

LalatenduMohanty (approved):
/lgtm

wking:
> If we're never going to remove it, let's not nag users to change pointlessly.

I agree that the update is pointless as far as "does the invocation work" goes.  But as I point out briefly in [my initial post][1], there are ecosystem benefits to consistency.  The more widespread the canonical phrasing is, the less mental overhead there is among folks sharing scripts, documenting behavior, and so on.  The logged stderr messages will do a little bit to gradually shift existing call-sites over to the new canonical phrasing, and make it less likely that new callers chose the outdated format, and gradually the bulk of users will not even need to be aware that  the outdated (but still functional) phrasing ever existed.

[1]: https://github.com/openshift/oc/pull/1472#issue-1759377848

pkg/cli/admin/update/update.go
```diff
@@ -52,64 +52,77 @@ func NewOptions(streams genericclioptions.IOStreams) *Options {
...
+	flags.BoolVar(&o.AllowExplicitUpdate, "allow-explicit-update", o.AllowExplicitUpdate, "Update even if the update target is not listed in the available versions list.")
+	flags.BoolVar(&o.AllowUpdateWithWarnings, "allow-update-with-warnings", o.AllowUpdateWithWarnings, "Update regardless of client-side guard failures, such as updates in progress or failing clusters.")
 	flags.BoolVar(&o.IncludeNotRecommended, "include-not-recommended", o.IncludeNotRecommended, "Display additional updates which are not recommended based on your cluster configuration.")
-	flags.BoolVar(&o.AllowNotRecommended, "allow-not-recommended", o.AllowNotRecommended, "Allows upgrade to a version when it is supported but not recommended for updates.")
+	flags.BoolVar(&o.AllowNotRecommended, "allow-not-recommended", o.AllowNotRecommended, "Allows update to a version when it is supported but not recommended for updates.")
+
+	flags.BoolVar(&o.AllowExplicitUpdate, "allow-explicit-upgrade", o.AllowExplicitUpdate, "Update even if the update target is not listed in the available versions list.")
```
evakhoni, line 121: so both the new "update" a well as the old "upgrade" flags variants are expected to work identically, while the later just won't be mentioned anymore?

evakhoni: flags.MarkDeprecated is just writing to stderr, or there's other differences expected?

evakhoni: also what about autocomplete? will it be able to pick the former and not the latter?

wking: [`MarkDeprecated`][1] sets up stderr logging and also hides the flag from help text.

[1]: https://pkg.go.dev/github.com/spf13/pflag#FlagSet.MarkDeprecated

pkg/cli/admin/update/update.go
```diff
@@ -52,64 +52,77 @@ func NewOptions(streams genericclioptions.IOStreams) *Options {
...
 
-			Cluster-side guards include checks for release verification and upgradeable conditions.
+			Cluster-side guards include checks for release verification and updateable conditions.
 			Again, it is usually best to give these conditions time to resolve, or to actively work to
 			resolve them.  But if you decide to trigger the update regardless of these concerns,
 			use --force, which is passed through to ClusterVersion's spec.desiredUpdate.force.
 		`),
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if cmd.CalledAs() != "update" {
+				fmt.Fprintf(o.ErrOut, "%s is deprecated.  Use '%s ...' instead\n", cmd.CommandPath(), cmd.CommandPath())
```
evakhoni, line 99:
```
❯ oc adm upgrade
oc adm update is deprecated.  Use 'oc adm update ...' instead
```
`..., cmd.CommandPath(), cmd.CommandPath())` seems wrong. we're using the same string twice

evakhoni:
/hold
seems like an issue to solve before merging

pkg/cli/admin/update/update.go
```diff
@@ -52,64 +52,77 @@ func NewOptions(streams genericclioptions.IOStreams) *Options {
...
+			regardless of these concerns, use --allow-update-with-warnings.
 
-			Cluster-side guards include checks for release verification and upgradeable conditions.
+			Cluster-side guards include checks for release verification and updateable conditions.
 			Again, it is usually best to give these conditions time to resolve, or to actively work to
 			resolve them.  But if you decide to trigger the update regardless of these concerns,
 			use --force, which is passed through to ClusterVersion's spec.desiredUpdate.force.
 		`),
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if cmd.CalledAs() != "update" {
```
evakhoni, line 98:
```
❯ oc adm update channel 
oc adm update channel is deprecated.  Use 'oc adm update channel ...' instead
info: Cluster channel is already clear (no change)
```
hmm.. we're not deprecating the channel subcommand right?

pkg/cli/admin/update/update.go
wking: I've added a `Replace` to replace `update` with `upgrade` in 003a50de3 -> 220d10e50, which also rebases around #1330:

```console
$ ./oc adm upgrade
oc adm upgrade is deprecated.  Use 'oc adm update ...' instead
Cluster version is 4.13.2
....
```

pkg/cli/admin/update/update.go
wking: Thanks!  Fixed in 003a50de3 -> 5f78beb41:

```console
$ ./oc adm upgrade channel
oc adm upgrade channel is deprecated.  Use 'oc adm update channel ...' instead
error: You are requesting to clear the update channel. The current channel "candidate-4.13" is one of the available channels, you must pass --allow-explicit-channel to continue
```

wking:
[build05 registry 504s][1].

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1472/pull-ci-openshift-oc-master-verify/1671196105590706176

ardaguclu:
I dropped my comments on Slack but wanted to write also in here;

I'd highly suggest deprecating `upgrade` command instead directly removing it. There is a version compatibility skew(and this PR's current status breaks this, `oc adm upgrade` 4.14 won't work on 4.13).

Instead we can deprecate and wait at least 1 releases for users to update their scripts to use new `update` command.

pkg/cli/admin/update/update.go
evakhoni: retested - `update channel` still triggering a false positive!
```
❯ oc adm update channel
oc adm upgrade channel is deprecated.  Use 'oc adm update channel ...' instead
info: Cluster channel is already clear (no change)
```

evakhoni: btw, there's a double space in the middle - intentional?

pkg/cli/admin/release/info.go
```diff
@@ -75,7 +75,7 @@ func NewInfo(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Com
 			If no arguments are specified the release of the currently connected cluster is displayed.
 			Specify one or more images via pull spec to see details of each release image. You may also
 			pass a semantic version (4.11.2) as an argument, and if cluster version object has seen such a
-			version in the upgrades channel it will find the release info for that version.
+			version in the update channel it will find the release info for that version.
```
ardaguclu, line 78: nit:
```suggestion
			version in the updates channel it will find the release info for that version.
```

ardaguclu (commented):
/approve

ardaguclu:
I'd prefer to see the e2e-metal-ipi-ovn-ipv6 is also green. But since it is not required job;
/hold
/retest

sdodson:
/hold
I know this is already held, but just noting here that there may be a desire to offer considerably different UX around `oc adm update` versus the deprecated `oc adm upgrade` such as forcing specification of intent to update to latest z-stream or minor version, etc.

Will leave a note on the card as well.

LalatenduMohanty:
We are not planning to rename upgrade to update. Rather we would re-imagine a new user experience for update sub-command, hence closing this PR

PR #1473, opened 2023-06-15
[wip] Wait for reboot

deads2k:
bugged, but gives a view of where I'm headed


pkg/cli/admin/rebootmachineconfigpool/reboot_machine_config_pool.go
```diff
@@ -216,23 +221,51 @@ func (r *RebootMachineConfigPoolRuntime) rebootMachineConfigPool(ctx context.Con
...
+	}
+
+	existingContent := ""
+	for _, file := range fileList {
+		filename, ok, err := unstructured.NestedString(file.(map[string]interface{}), "path")
+		if !ok || err != nil {
+			continue
+		}
+		// this is the file we use to indicate reboot numbers
+		if filename != "/etc/machine-config-operator-oc-initiated-reboot-number" {
```
stlaz, line 240: should we use the `/etc/kubernetes` folder?

cgwalters:
I think what we want in the future is a clean way to reboot machines without going through MCs.  Something like a flow:

- client reads node object's current linux boot ID
- client adds annotation `io.openshift.request-reboot-for-bootid=$bootid` to node
- MCC reacts to annotation, and chooses which nodes to reboot using the same "upgrade" logic it has today, including the drain
- MCC writes an annotation like `io.openshift.do-reboot-for-bootid=$bootid` which tells the MCD to reboot

Mmm, and doing this cleanly would also need reporting in the pools...perhaps everywhere we compare "current == desired" for config we can check the reboot request annoation and treat it as an upgrade?

OK yeah, not a simple patch.

I guess a simpler way would be to just enshrine a magical file path (or...hmm, have a reboot-timestamp in a machineconfig object).

yuqi-zhang:
I think the longer term path would be a more centralized API to drain/reboot nodes as requested by all operators, handled by some other operator.

For now though I think this approach is good enough :)

pkg/cli/admin/waitfornoderestart/wait_for_node_restart.go
```diff
@@ -0,0 +1,281 @@
...
+		r.workerRestartNumber = r.RebootNumber
+		return nil
+	}
+
+	masterMachineConfig, err := r.DynamicClient.Resource(rebootmachineconfigpool.MachineConfigResource).Get(ctx, rebootmachineconfigpool.MasterRebootingMachineConfigName, metav1.GetOptions{})
+	switch {
+	case apierrors.IsNotFound(err):
+	// do nothing.  If we need to use it, we'll fail
+	case err != nil:
+		return fmt.Errorf("unable to read existing state: %w", err)
```
yuqi-zhang, line 58: It's possible that the user hasn't actually started a reboot (or was unsuccessful in the MachineConfig application. Maybe be more descriptive here to say what we're looking for, and no reboots might be queued?

Similarly, the user may have initiated just a worker reboot for some reason, maybe we don't fail here and just let both pools describe their own state

deads2k:
> I think the longer term path would be a more centralized API to drain/reboot nodes as requested by all operators, handled by some other operator.

We already have an operator that manages nodes, I think I'd expect this to be an MCO concern. Rebooting by pools seems like a reasonable path. I think an eventual improvement to selectively choose either nodes (as imperative from the CLI) or pools (as declarative via the API) and a "wait for this" that allows either would be a great improvement.

pkg/cli/admin/waitfornoderestart/wait_for_node_restart.go
```diff
@@ -0,0 +1,281 @@
...
+			return fmt.Errorf("not unstructured: %w", err)
+		}
+		node := &corev1.Node{}
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(uncastObj.Object, node); err != nil {
+			return fmt.Errorf("not a node: %w", err)
+		}
+		if _, ok := node.Labels["node-role.kubernetes.io/master"]; ok {
+			hasMaster = true
+		} else {
+			hasWorker = true
```
yuqi-zhang, line 110: So I think the confusing part here might be that, technically, we are lumping all other roles into "worker". Like we discussed on the other PR, all custom pools will react to worker changes, so by initiating a worker pool reboot all custom pools will reboot in parallel.

Instead of describing the case as worker, maybe we just describe it as "worker + all custom pools" somehow. Same as to the reboot command, the user should be notified that the worker change will apply to all custom pools as well.

pkg/cli/admin/waitfornoderestart/wait_for_node_restart.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func (r *WaitForNodeRestartRuntime) Run(ctx context.Context) error {
+	if err := r.setRestartNumber(ctx); err != nil {
+		return err
+	}
+
+	hasMaster := false
+	hasWorker := false
+	remainingNodes := []*corev1.Node{}
```
yuqi-zhang, line 89: remainingNodes end up being all nodes right? Or did you want this to be a subset of nodes?

deads2k:
Seeing @cgwalters proposal for declarative node restart, that looks nicer than imperative via the CLI.  But still a future concern with clear semantics established for how pool lifecycle impacts the node reboot.

pkg/cli/admin/waitfornoderestart/wait_for_node_restart.go
```diff
@@ -0,0 +1,281 @@
...
+	RebootNumber int
+
+	masterRestartNumber int
+	workerRestartNumber int
+
+	genericclioptions.IOStreams
+}
+
+func (r *WaitForNodeRestartRuntime) setRestartNumber(ctx context.Context) error {
+	if r.RebootNumber > 0 {
```
yuqi-zhang, line 47: Hmm, not sure how useful this is as a general case, but I guess the expectation is that all nodes will be rebooting in tandem?

As in, we aren't really expecting users to run this command 10x for worker pools just because they feel like it for other use cases.

yuqi-zhang (commented):
Some general comments:

pkg/cli/admin/waitfornoderestart/wait_for_node_restart.go
deads2k: If I were a cluster-admin trying to restart a particular pool, and knew about this, I would create a machineconfig that overrides the file with my own value and then run this command with a specific, "expect this one" against the nodes.

pkg/cli/admin/waitfornoderestart/wait_for_node_restart.go
yuqi-zhang: Mostly naming. The main scenario I don't want to confuse people is, let's say they have `master, worker, infra` pools. They can't initiate a reboot just for infra, and worker reboots will also reboot infra.

Basically, make sure the user is aware of what's happening if there are non-worker pools

yuqi-zhang (approved):
functionality works fine, we can follow up on the details if needed. Let's get this into a testing build

/lgtm

PR #1474, opened 2023-06-16
bring some cert rotation helpers back into 4.13

deads2k:


stlaz:
These changes appear to match master/4.14.

PR #1475, opened 2023-06-16
bring some cert rotation helpers back into 4.12

deads2k:


stlaz:
The changes and the final state of the significant files appears to match the 4.13.

However, it appears that the "Patch" type is not supported by the fake client in leaf and root certificate regeneration :sob:

PR #1477, opened 2023-06-16
bring some cert rotation helpers back into 4.10

deads2k:


stlaz:
The patching output seems to match the 4.11 fixes

PR #1496, opened 2023-07-05
Wire built-in dry-run flag functionality instead custom one

ardaguclu:
This PR wires builtin dry-run functionality into the newly added certificate related commands instead relying on the custom non-standard dry-run flag as stated in the TODO.

### DIFF
--- a/pkg/cli/admin/ocpcertificates/certregen/command.go
+++ b/pkg/cli/admin/ocpcertificates/certregen/command.go
@@ -2,6 +2,7 @@ package certregen
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


pkg/cli/admin/ocpcertificates/certregen/command.go
```diff
@@ -152,11 +153,20 @@ func (o *RegenerateCertificatesOptions) ToRuntime(args []string) (*RegenerateCer
 		return nil, err
 	}
 
+	dryRunStrategy, err := cmdutil.GetDryRunStrategy(cmd)
```
ardaguclu, line 156: standalone `--dry-run` is also deprecated. I hope that there is no script relying on passing `--dry-run` instead `--dry-run=server`.

deads2k:
Rather than degrade the usage of dry-run for a command that *always* used server-side dry-run, the upstream cli-runtime should have an option that reflects "dry-run *always* means server-side dry run"..  The only reason it is a string is due to legacy debt.

ardaguclu:
I still believe that this looks non-standard from the user point of view, since upstream supports `--dry-run=server/client/none` and we'll support `--dry-run=true`. But I agree that in the long term, there is a movement to drop client dry-run totally and `--dry-run=server` will be default. 

I'm closing this PR in favor of this.
/close

PR #1497, opened 2023-07-07
OCPBUGS-15776: mcs cert: account for environments that use IP directly

yuqi-zhang:
In some cases, the MCS endpoint refers to the apiserver IP directly instead of the api-int url. Update the cert generation to account for this, otherwise we fail for those platforms.

Also update the kube-system root-ca cert for BM IPI, since BM IPI is the only platform that uses the MCO-generated and managed secret

### DIFF
--- a/pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
+++ b/pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
@@ -131,15 +131,30 @@ func getServerIPsFromInfra(cfg *configv1.Infrastructure) []string {
 	}
 	switch cfg.Status.PlatformStatus.Type {
 	case configv1.BareMetalPlatformType:
-		return cfg.Status.PlatformStatus.BareMetal.APIServerInternalIPs
+		if cfg.Status.PlatformStatus.BareMetal.APIServerInternalIPs != nil {
+			return cfg.Status.PlatformStatus.BareMetal.APIServerInternalIPs
+		}
+		return []string{cfg.Status.PlatformStatus.BareMetal.APIServerInternalIP}
 	case configv1.OvirtPlatformType:
-		return cfg.Status.PlatformStatus.Ovirt.APIServerInternalIPs
+		if cfg.Status.PlatformStatus.Ovirt.APIServerInternalIPs != nil {
+			return cfg.Status.PlatformStatus.Ovirt.APIServerInternalIPs
+		}
+		return []string{cfg.Status.PlatformStatus.Ovirt.APIServerInternalIP}
 	case configv1.OpenStackPlatformType:
-		return cfg.Status.PlatformStatus.OpenStack.APIServerInternalIPs
+		if cfg.Status.PlatformStatus.OpenStack.APIServerInternalIPs != nil {
+			return cfg.Status.PlatformStatus.OpenStack.APIServerInternalIPs
+		}
+		return []string{cfg.Status.PlatformStatus.OpenStack.APIServerInternalIP}
 	case configv1.VSpherePlatformType:
-		return cfg.Status.PlatformStatus.VSphere.APIServerInternalIPs
+		if cfg.Status.PlatformStatus.VSphere.APIServerInternalIPs != nil {
+			return cfg.Status.PlatformStatus.VSphere.APIServerInternalIPs
+		}
+		return []string{cfg.Status.PlatformStatus.VSphere.APIServerInternalIP}
 	case configv1.NutanixPlatformType:
-		return cfg.Status.PlatformStatus.Nutanix.APIServerInternalIPs
+		if cfg.Status.PlatformStatus.Nutanix.APIServerInternalIPs != nil {
+			return cfg.Status.PlatformStatus.Nutanix.APIServerInternalIPs
+		}
+		return []string{cfg.Status.PlatformStatus.Nutanix.APIServerInternalIP}
 	default:
 		return []string{}
 	}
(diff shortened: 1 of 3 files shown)


yuqi-zhang:
Ok, removing draft status since I tested this to work on vsphere IPI. Bare metal is still a big question, see latest comment in https://issues.redhat.com/browse/OCPBUGS-15776

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
```diff
@@ -120,3 +124,26 @@ func (o *RegenerateMCOOptions) Run(ctx context.Context) error {
...
+	}
+	switch cfg.Status.PlatformStatus.Type {
+	case configv1.BareMetalPlatformType:
+		return cfg.Status.PlatformStatus.BareMetal.APIServerInternalIPs
+	case configv1.OvirtPlatformType:
+		return cfg.Status.PlatformStatus.Ovirt.APIServerInternalIPs
+	case configv1.OpenStackPlatformType:
+		return cfg.Status.PlatformStatus.OpenStack.APIServerInternalIPs
+	case configv1.VSpherePlatformType:
+		if cfg.Status.PlatformStatus.VSphere.APIServerInternalIPs != nil {
```
cgwalters, line 140: I know we copied this from the relevant installer code, but I don't think we need it because the nil slice will behave the same as the empty slice for us.  I think they did this conditional because they behave differently when serialized, but that's not relevant to us.

pkg/cli/admin/ocpcertificates/regeneratemco/updatesecrets.go
```diff
@@ -47,8 +47,9 @@ func (o *RegenerateMCOOptions) updateSecrets(ctx context.Context) error {
 		return fmt.Errorf("cannot list MAO secrets: %w", err)
 	}
 	for _, secret := range secretList.Items {
-		// These two are managed by the MCO but unused. Skip them since the MCO will write them back.
-		if secret.Name == mcoManagedWorkerSecret || secret.Name == mcoManagedMasterSecret {
+		// Managed secrets should be created by the MCO but are unused. Normally they would be written back,
+		// but baremetal IPI uses the managed worker secret. Skip only the Master secret for now.
```
cgwalters, line 51: But won't the MCO just rewrite the `worker-user-data-managed` secret back next time `syncRenderConfig()` runs?  That's kind of the idea, it's managed state...so to fix this properly we need to sync the `-managed` logic in the MCO right?

yuqi-zhang: Yes, I was wondering if BM did something different, but otherwise I am updating this locally.

Essentially to do this via oc we would need to update the root-ca configmap which unfortunately will trigger another reboot. We can also patch this via the MCO but that would require another payload to update to which I am not sure is feasible for older versions

yuqi-zhang:
Updated per comments + also update the root-ca selectively on BM platforms

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
```diff
@@ -120,3 +124,38 @@ func (o *RegenerateMCOOptions) Run(ctx context.Context) error {
...
+func getServerIPsFromInfra(cfg *configv1.Infrastructure) []string {
+	if cfg.Status.PlatformStatus == nil {
+		return []string{}
+	}
+	switch cfg.Status.PlatformStatus.Type {
+	case configv1.BareMetalPlatformType:
+		if cfg.Status.PlatformStatus.BareMetal.APIServerInternalIPs != nil {
+			return cfg.Status.PlatformStatus.BareMetal.APIServerInternalIPs
+		}
+		return []string{cfg.Status.PlatformStatus.BareMetal.APIServerInternalIP}
```
yuqi-zhang, line 137: As it turns out our BM IPI QE installs only have this (legacy?) field. Just to make triple sure, all platforms will fall back to IP and not IPs

ingvagabund:
@yuqi-zhang would you please pull the installer team for the review? All the changes here are quite specific. The oc roofs many commands owned by many teams. @sinnykumari for the MCO review.

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
```diff
@@ -120,3 +124,38 @@ func (o *RegenerateMCOOptions) Run(ctx context.Context) error {
 	}
 	return nil
 }
+
+func getServerIPsFromInfra(cfg *configv1.Infrastructure) []string {
```
cgwalters, line 128: It looks like we have [multiple copies of this logic](https://github.com/search?q=org%3Aopenshift%20%22PlatformStatus.BareMetal.APIServerInternalIP%22&type=code) around now; this would be the 3rd (the baremetal one is a different variant, but related and is a 4th thing).

Not a blocker but we should definitely get a helper function for this into openshift/api at least.

cgwalters (approved):

yuqi-zhang:
/assign @deads2k 

Since David helped with the initial commands for reviews. Is there a good contact for installer team? The logic here is pretty specific to how the MCO uses these configurations after the fact

Based on the latest comments in https://issues.redhat.com/browse/OCPBUGS-15776, this should be working for all platforms, and I have verified manually on BM/vsphere as well

pkg/cli/admin/ocpcertificates/regeneratemco/rotatecerts.go
deads2k: > Not a blocker but we should definitely get a helper function for this into openshift/api at least.

openshift/api has no code.  But openshift/library-go, ok.

deads2k:
/assign @cgwalters 

for lgtm at his discretion.

PR #1499, opened 2023-07-11
OCPBUGS-16009: reboot: set ignition version to 3.1

yuqi-zhang:
The oldest version we want to run the cert rotation command on is 4.6, which has support only up to ignition spec 3.1 (3.2+ was introduced in 4.7)

Update the template to use 3.1 since it should be supported for all versions we want to run the procedure on.


deads2k:
/approve

/assign @cgwalters 

for lgtm

cgwalters (approved):

PR #1504, opened 2023-07-11
[release-4.11] OCPBUGS-16059: mcs cert: account for environments that use IP directly

yuqi-zhang:
Doing a manual pick here since the bot-PR doesn't work for 4.11/4.10


yuqi-zhang:
/test e2e-agnostic-cmd

has passed before

PR #1513, opened 2023-07-17
login: improve usage message for --web option

liouk:
This PR adds more informative text in the usage doc of `oc login --web` flag.

### DIFF
--- a/pkg/cli/login/login.go
+++ b/pkg/cli/login/login.go
@@ -46,7 +46,7 @@ var (
 		oc login localhost:8443 --username=myuser --password=mypass
 
 		# Log in to the given server through a browser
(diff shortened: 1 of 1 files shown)


pkg/cli/login/login.go
```diff
@@ -46,7 +46,7 @@ var (
 		oc login localhost:8443 --username=myuser --password=mypass
 
 		# Log in to the given server through a browser
-		oc login --web --callback-port 8280 localhost:8443
+		oc login localhost:8443 --web --callback-port 8280
```
stlaz, line 49: funny, this is probably from the times when `oc cluster up` was still a thing

PR #1516, opened 2023-07-20
Add profiling functionality and flags in oc

ardaguclu:
To preserve the feature parity between oc and kubectl, this commit takes the profiling functionality in kubectl into the oc
as detected by @atiratree in https://github.com/openshift/oc/pull/1420#issuecomment-1544777587

### DIFF
--- a/pkg/cli/profiling.go
+++ b/pkg/cli/profiling.go
@@ -1,0 +1,90 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"runtime"
+	"runtime/pprof"
+
+	"github.com/spf13/pflag"
+)
+
+var (
+	profileName   string
+	profileOutput string
+)
+
+func addProfilingFlags(flags *pflag.FlagSet) {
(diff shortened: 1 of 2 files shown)


ardaguclu:
/hold
It would be better to do pre-merge testing of this and hold until it is done.

ardaguclu:
After thinking on this, I think there is no need for a pre-merge testing since the feature is straightforward. What do you think @atiratree ?

pkg/cli/profiling.go
```diff
@@ -0,0 +1,90 @@
...
+	profileName   string
+	profileOutput string
+)
+
+func addProfilingFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&profileName, "profile", "none", "Name of profile to capture. One of (none|cpu|heap|goroutine|threadcreate|block|mutex)")
+	flags.StringVar(&profileOutput, "profile-output", "profile.pprof", "Name of the file to write the profile to")
+}
+
+func initProfiling() error {
```
atiratree, line 23: Would be great if we could import these from kubectl. I suppose we can copy for now, as it does not seem there are planned changes to these in the future

atiratree:
I agree, the feature looks straightforward
/lgtm

PR #1520, opened 2023-07-25
OCPBUGS-16735: Truncate existing files when writing from inspect

MarSik:
oc adm inspect generated files sometime have the leading "---" and some time do not. This depends on the order of objects collected. This by itself is not an issue.

However this becomes an issue when combined with multiple invocations of oc adm inspect and collecting data to the same directory like must-gather does.

If an object is collected multiple times then the second time oc might overwrite the original file improperly and leave 4 bytes of the original content behind.

This is happening when not writing the "---\n" in the second invocation as this makes the content 4B shorter and the original tailing 4B are left in the file intact.

This garbage confuses YAML parsers.

### DIFF
--- a/pkg/cli/admin/inspect/writer.go
+++ b/pkg/cli/admin/inspect/writer.go
@@ -51,7 +51,7 @@ func (r *resourceWriterReadCloser) Close() error {
 type simpleFileWriter struct{}
 
 func (f *simpleFileWriter) Write(filepath string, src fileWriterSource) error {
-	dest, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE, 0755)
+	dest, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
 	if err != nil {
 		return err
 	}


pkg/cli/admin/inspect/writer.go
```diff
@@ -51,7 +51,7 @@ func (r *resourceWriterReadCloser) Close() error {
 type simpleFileWriter struct{}
 
 func (f *simpleFileWriter) Write(filepath string, src fileWriterSource) error {
-	dest, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE, 0755)
+	dest, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
```
ardaguclu, line 54: In terms of yaml files generated by oc adm inspect command, this is indeed a problem and I'm supportive of this change. 

On the other hand, this writer is also used for saving the collected logs of pods, operators, etc. and I think, we should ensure that this change will not cause truncating the logs.

MarSik: Well this always starts writing from the beginning, so it will either overwrite from the beginning (causing the same bug) or drop the "old" data. The only way to behave differently would be to pass O_APPEND.

Or is there something I am missing?

ardaguclu: I agree with you @MarSik. Since we are not passing O_APPEND, we are already overwriting on the old data and it would be good to truncate it.

Thanks for fixing this.

ardaguclu:
/lgtm
/hold
until checking the must-gather of CI jobs

MarSik:
e2e-aws-ovn looks good to me, the bug is not present in the generated must-gather and logs are there.

PR #1521, opened 2023-07-27
OTA-559: pkg/cli/admin/release/extract: Add --included and --install-config

wking:
There are many manifests in release images.  CredentialsRequests, Deployments, ClusterOperators, etc., etc.  There are increasingly complicated conditions on whether the cluster-version operator reconciles or ignores a manifest, including [cluster profiles, cluster capabilities, and feature sets][1].  To make it easier for users to see the manifests that actually matter for their cluster, this commit adds `--included`, which uses [the same `Manifest.Include` logic used by the cluster-version operator][2].  As long as we keep the vendored copies of library-go close-ish between `oc` and the CVO, this will allow users to just get the manifests that apply to their cluster.  For example:

* You have enabled `TechPreviewNoUpgrade`, so here's [a CredentialsRequest for the `openshift-cluster-api` namespace][3], or
* You have not enabled `TechPreviewNoUpgrade`, so I'll filter out this CredentialsRequest for the `openshift-cluster-api namespace`.

This should work fairly well as long as `oc` can reach the cluster to retrieve the same configuration data that the cluster-version operator is using to make its decisions, although there is a risk of drift between the separate code bases.  It gets trickier when we need to run before the cluster exists.  For those cases, we have `--install-config`, and `oc` can get some information reliably from there (enabled capability (sets), cloud provider, etc.) but not others (known capabilities, cluster profile, etc.)  We may need to circle back later with additional flags for these properties, so the user can explicitly pass in the information that we cannot figure out from the install-config alone.

[1]: https://github.com/openshift/enhancements/blob/fb6021978e149e4cd12a223134a148172eff51ad/dev-guide/cluster-version-operator/dev/operators.md#manifest-inclusion-annotations
[2]: https://github.com/openshift/cluster-version-operator/blob/4eb9433f345cc551c7d31996ee441d649f677b4e/pkg/cvo/sync_worker.go#L963
[3]: https://github.com/openshift/cluster-capi-operator/blob/1593e40d6422e46473beee2f4419514a0418da1c/manifests/0000_30_cluster-api_00_credentials-request.yaml#L9-L15

### DIFF
--- a/pkg/cli/admin/release/extract.go
+++ b/pkg/cli/admin/release/extract.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/rest"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/templates"
 
@@ -75,7 +76,7 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 			must be installed on the cluster for a given version.
 
 			The --tools and --command flags allow you to extract the appropriate client binaries
-			for	your operating system to disk. --tools will create archive files containing the
+			for your operating system to disk. --tools will create archive files containing the
 			current OS tools (or, if --command-os is set to '*', all OS versions). Specifying
 			--command for either 'oc' or 'openshift-install' will extract the binaries directly.
 			You may pass a PGP private key file with --signing-key which will create an ASCII
@@ -86,6 +87,14 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 			The --credentials-requests flag filters extracted manifests to only cloud credential
 			requests. The --cloud flag further filters credential requests to a specific cloud.
 			Valid values for --cloud include alibabacloud, aws, azure, gcp, ibmcloud, nutanix, openstack, ovirt, powervs, and vsphere.
+
+			The --included flag filters extracted manifests to those that are expected to be included
+			with the cluster.  Filters are cumulative, so '--credentials-requests --included' will
+			only include cloud credential requests which are expected to be included with the cluster.
+			If --install-config is set, it will be used to determine the expected cluster configuration,
+			otherwise the command will interrogate your current cluster to determine its configuration.
+			This command is most accurate when the version of the extracting client matches the version
+			of the cluster under consideration.
 
 			Instead of extracting the manifests, you can specify --git=DIR to perform a Git
 			checkout of the source code that comprises the release. A warning will be printed
@@ -111,7 +120,7 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Complete(f, cmd, args))
 			kcmdutil.CheckErr(o.Validate())
-			kcmdutil.CheckErr(o.Run())
+			kcmdutil.CheckErr(o.Run(cmd.Context()))
 		},
 	}
 	flags := cmd.Flags()
@@ -133,8 +142,11 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 	flags.StringVar(&o.CommandOperatingSystem, "command-os", o.CommandOperatingSystem, "Override which operating system command is extracted (mac, windows, linux) or can be specified with arch(linux/arm64, mac/amd64). You map specify '*' to extract all tool archives.")
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be copied under.")
 
-	flags.BoolVar(&o.CredentialsRequests, "credentials-requests", o.CredentialsRequests, "Extract credential request manifests only")
-	flags.StringVar(&o.Cloud, "cloud", o.Cloud, "Specify the cloud for which credential request manifests should be extracted. Works only in combination with --credentials-requests.")
+	flags.BoolVar(&o.Included, "included", o.Included, "Exclude manifests that are not expected to be included in the cluster.")
+	flags.StringVar(&o.InstallConfig, "install-config", o.InstallConfig, "Path to an install-config file, as consumed by the openshift-install command.  Works only in combination with --included.")
+
+	flags.BoolVar(&o.CredentialsRequests, "credentials-requests", o.CredentialsRequests, "Exclude manifests which are not credential requests.")
+	flags.StringVar(&o.Cloud, "cloud", o.Cloud, "Exclude credential requests which are not relevant to the given cloud provider.  Works only in combination with --credentials-requests.")
 
 	flags.StringVarP(&o.Output, "output", "o", o.Output, "Output format. Supports 'commit' when used with '--git'.")
 	return cmd
@@ -146,6 +158,9 @@ type ExtractOptions struct {
 	SecurityOptions imagemanifest.SecurityOptions
 	FilterOptions   imagemanifest.FilterOptions
 	ParallelOptions imagemanifest.ParallelOptions
+
+	// RESTConfig is a REST client configuration for connecting to a cluster if neccessary.
+	RESTConfig *rest.Config
 
 	ICSPFile string
 
@@ -159,7 +174,16 @@ type ExtractOptions struct {
 	CommandOperatingSystem string
 	SigningKey             string
 
-	// CredentialsRequests if true, results in only credential request manifests getting extracted.
+	// Included, if true, results in only included manifests getting extracted.
+	// For example, manifests associated with optional capabilities will be excluded unless
+	// the cluster configuration enables that capability.
+	Included bool
+
+	// InstallConfig is the path to an install-config file, as
+	// consumed by the openshift-install command.
+	InstallConfig string
+
+	// CredentialsRequests, if true, results in only credential request manifests getting extracted.
 	// If Cloud is specified, then only the credential requests for that cloud are extracted.
 	CredentialsRequests bool
 	Cloud               string
@@ -196,6 +220,9 @@ func (o *ExtractOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args [
 		return fmt.Errorf("you may only specify a single image via --from or argument")
 	}
 	o.From = args[0]
+	if o.RESTConfig, err = f.ToRESTConfig(); err != nil {
+		return err
+	}
 
 	return o.FilterOptions.Complete(cmd.Flags())
 }
@@ -204,7 +231,7 @@ func (o *ExtractOptions) Validate() error {
 	return o.FilterOptions.Validate()
 }
 
-func (o *ExtractOptions) Run() error {
+func (o *ExtractOptions) Run(ctx context.Context) error {
 	sources := 0
 	if o.Tools {
 		sources++
@@ -226,8 +253,12 @@ func (o *ExtractOptions) Run() error {
 		return fmt.Errorf("--output is only supported with --git")
 	}
 
-	if !o.CredentialsRequests && len(o.Cloud) > 0 {
-		return fmt.Errorf("--cloud is only supported with --credentials-requests")
+	if len(o.InstallConfig) > 0 && !o.Included {
+		return fmt.Errorf("--install-config is only supported with --included")
+	}
+
+	if len(o.Cloud) > 0 && !o.CredentialsRequests && !o.Included {
+		return fmt.Errorf("--cloud is only supported with --credentials-requests or --included")
 	}
 	if len(o.Cloud) > 0 {
 		if _, ok := credRequestCloudProviderSpecKindMapping[o.Cloud]; !ok {
@@ -310,8 +341,33 @@ func (o *ExtractOptions) Run() error {
 	if o.ExtractManifests {
 		expectedProviderSpecKind := credRequestCloudProviderSpecKindMapping[o.Cloud]
 
+		include := func(m *manifest.Manifest) error { return nil } // default to including everything
+		if o.Included {
+			context := "connected cluster"
+			inclusionConfig := manifestInclusionConfiguration{}
+			if o.InstallConfig == "" {
+				inclusionConfig, err = findClusterIncludeConfig(ctx, o.RESTConfig)
+			} else {
+				inclusionConfig, err = findClusterIncludeConfigFromInstallConfig(ctx, o.InstallConfig)
+				context = o.InstallConfig
+			}
+			if err != nil {
+				return err
+			}
+			if inclusionConfig.Platform != nil {
+				if o.Cloud != "" && *inclusionConfig.Platform != o.Cloud {
+					return fmt.Errorf("--cloud %q set, but %s has %q", o.Cloud, context, *inclusionConfig.Platform)
+				}
+				var ok bool
+				if expectedProviderSpecKind, ok = credRequestCloudProviderSpecKindMapping[*inclusionConfig.Platform]; !ok {
+					return fmt.Errorf("unrecognized platform for CredentialsRequests: %q", *inclusionConfig.Platform)
+				}
+			}
+			include = newIncluder(inclusionConfig)
+		}
+
 		tarEntryCallbacks = append(tarEntryCallbacks, func(hdr *tar.Header, _ extract.LayerInfo, r io.Reader) (bool, error) {
-			if hdr.Name == "image-references" {
+			if hdr.Name == "image-references" && !o.CredentialsRequests {
 				buf := &bytes.Buffer{}
 				if _, err := io.Copy(buf, r); err != nil {
 					return false, fmt.Errorf("unable to load image-references from release payload: %w", err)
@@ -339,7 +395,7 @@ func (o *ExtractOptions) Run() error {
 					return true, err
 				}
 				return true, nil
-			} else if hdr.Name == "release-metadata" {
+			} else if hdr.Name == "release-metadata" && !o.CredentialsRequests {
 				out := o.Out
 				if o.Directory != "" {
 					out, err = os.Create(filepath.Join(o.Directory, hdr.Name))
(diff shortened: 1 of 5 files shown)


wking:
Example local testing:

```console
$ cat install-config.yaml 
apiVersion: v1
platform:
  aws:
    dummy: data
capabilities:
  baselineCapabilitySet: None
  XadditionalEnabledCapabilities:
  - openshift-samples
featureSet: TechPreviewNoUpgrade
$ go build ./cmd/oc
$ rm -rf manifests
$ ./oc --v=4 adm release extract --to manifests --credentials-requests --included --install-config install-config.yaml quay.io/openshift-release-dev/ocp-release:4.14.0-ec.3-x86_64 2>&1 | grep -v 'Found manifest'
I0727 12:19:41.117364    3535 extract_tools.go:972] If the eventual cluster will not be the same minor version as this unknown 'oc', the known capability sets may differ.    <-- 'unknown' would be a version like "4.14.0-ec.3" in an official oc release
...
0727 12:19:42.381014    3535 extract.go:419] Excluding Group: "apiextensions.k8s.io" Kind: "CustomResourceDefinition" Name: "preprovisioningimages.metal3.io": disabled capabilities: baremeta
...
I0727 12:19:43.155522    3535 extract.go:419] Excluding Group: "apps" Kind: "Deployment" Namespace: "openshift-console-operator" Name: "console-operator": include.release.openshift.io/self-managed-high-availability unset
I0727 12:19:43.159270    3535 extract.go:419] Excluding Group: "apps" Kind: "Deployment" Namespace: "openshift-console-operator" Name: "console-operator": disabled capabilities: Console
...
$ ls manifests
0000_30_cluster-api_00_credentials-request.yaml
0000_50_cloud-credential-operator_05-iam-ro-credentialsrequest.yaml
0000_50_cluster-image-registry-operator_01-registry-credentials-request.yaml
0000_50_cluster-ingress-operator_00-ingress-credentials-request.yaml
0000_50_cluster-network-operator_02-cncc-credentials.yaml
image-references
release-metadata
```

Hmm, I guess I don't want `image-references` or `release-metadata` in `--credentials-requests` mode; I'll update to filter those out...

wking:
Should have `image-references` and `release-metadata` excluded for `--credentials-requests` now with 37326e7 -> ae80d31.

/retest

pkg/cli/admin/release/extract.go
```diff
@@ -310,8 +334,32 @@ func (o *ExtractOptions) Run() error {
 	if o.ExtractManifests {
 		expectedProviderSpecKind := credRequestCloudProviderSpecKindMapping[o.Cloud]
 
+		include := func(m *manifest.Manifest) error { return nil } // default to including everything
```
LalatenduMohanty, line 337: I am not sure how this will include everything. Or it is some code which will get changed in the final code?

jiajliu:
> You have not enabled TechPreviewNoUpgrade, so I'll filter out this CredentialsRequest for the openshift-cluster-api namespace.
@wking from my understanding(correct me if wrong), all manifests of disabled  feature set and cluster capability on a live cluster should not be extracted when specifying --included var, right? But it doesn't seem like that from pre-merge test result, more details in https://issues.redhat.com/browse/OTA-968

pkg/cli/admin/release/extract.go
wking: If later code calls this `include` function, it will always return a non-error `nil` for "include this manifest".  If the caller set `--included`, we replace the value of this `include` variable with the more restrictive implementation that errors for manifests that the CVO is not expected to include (e.g. "this manifest is tech-preview, but the current cluster is not, so I doubt it would be included").  I don't expect changes to that pattern before merging.  But I can add a more elaborate comment here or in the commit message if you think  that would help clarify for future developers?

pkg/cli/admin/release/extract_tools.go
```diff
@@ -908,3 +927,90 @@ func copyAndReplace(errorOutput io.Writer, w io.Writer, r io.Reader, bufferSize
...
 }
+
+func findClusterIncludeConfig(ctx context.Context, installConfigPath string, restConfig *rest.Config) (requiredFeatureSet *string, profile *string, platform *string, capabilities *configv1.ClusterVersionCapabilitiesStatus, overrides []configv1.ComponentOverride, err error) {
+	clientVersion, reportedVersion, err := version.ExtractVersion()
+	if err != nil {
+		return nil, nil, nil, nil, nil, err
+	}
+	if reportedVersion == "" {
+		reportedVersion = clientVersion.String()
+	}
```
petr-muller, line 938: These seem to be used only in the `installConfigPath` branch

pkg/cli/admin/release/extract_tools.go
```diff
@@ -908,3 +927,90 @@ func copyAndReplace(errorOutput io.Writer, w io.Writer, r io.Reader, bufferSize
...
+				}
+				capabilities.EnabledCapabilities = append(capabilities.EnabledCapabilities, enabled...)
+			}
+			capabilities.EnabledCapabilities = append(capabilities.EnabledCapabilities, data.Capabilities.AdditionalEnabledCapabilities...)
+
+			klog.Infof("If the eventual cluster will not be the same minor version as this %s 'oc', the known capability sets may differ.", reportedVersion)
+			capabilities.KnownCapabilities = configv1.KnownClusterVersionCapabilities
+		}
+
+		return requiredFeatureSet, profile, platform, capabilities, overrides, nil
```
petr-muller, line 978: I think we can extract all this to a `func includesFromInstallConfig(installConfigPath string) (*string, *string, *string, *configv1.ClusterVersionCapabilitiesStatus, []configv1.ComponentOverride, error)` method to reduce complexity... And with an even simpler signature if we end up collecting these bits in to a struct! :)

pkg/cli/admin/release/extract.go
```diff
@@ -133,6 +142,9 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 	flags.StringVar(&o.CommandOperatingSystem, "command-os", o.CommandOperatingSystem, "Override which operating system command is extracted (mac, windows, linux) or can be specified with arch(linux/arm64, mac/amd64). You map specify '*' to extract all tool archives.")
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be copied under.")
 
+	flags.BoolVar(&o.Included, "included", o.Included, "Extract credential request manifests only")
```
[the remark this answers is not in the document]
wking: Thanks :)  I've pushed 6a5869577 -> ae4ddeddc to shuffle doc-strings for the various filter parameters, because it's easier (at least for me) to understand how multiple exclusions fit together than it is to reason about how multiple inclusion properties fit together.

pkg/cli/admin/release/info.go
```diff
@@ -301,22 +300,22 @@ func replaceStableSemanticArgs(args []string, semanticArgs map[string]semver.Ver
 	return nil
 }
 
-func replaceClusterSemanticArgs(f kcmdutil.Factory, args []string, semanticArgs map[string]semver.Version) ([]string, error) {
+func getClusterVersion(ctx context.Context, f kcmdutil.Factory) (*configv1.ClusterVersion, error) {
```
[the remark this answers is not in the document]
wking: I'd done this when I expected to use it in `findClusterIncludeConfig`.  But I ended up going a different way there, because I need more than ClusterVersion.  I've dropped my changes to this file in ae4ddeddc -> 0a9024bb1.

pkg/cli/admin/release/extract.go
wking: I like using an `error` because it's "I want to include this manifest.  Helper function, if you disagree tell me why."  I guess it doesn't allow the caller to conveniently fail-open ("Hmm, my helper function failed to form an opinion, so I guess I'll assume it had not concerns" is not possible), and instead "helper function failed to form an opinion" is just one of several possible reasons we might decide to exclude a manifest.  But that's somewhat academic, because there are no external calls in this "accept everything" lambda or in the `Manifest.Include` function.

pkg/cli/admin/release/extract.go
```diff
@@ -310,8 +341,33 @@ func (o *ExtractOptions) Run() error {
...
 
+		include := func(m *manifest.Manifest) error { return nil } // default to including everything
+		if o.Included {
+			var excludeIdentifier *string // not worth filtering on this, it is just for exclude.release.openshift.io/internal-openshift-hosted ROKS
+
+			context := "connected cluster"
+			if o.InstallConfig != "" {
+				context = o.InstallConfig
+			}
+			requiredFeatureSet, profile, platform, capabilities, overrides, err := findClusterIncludeConfig(ctx, o.InstallConfig, o.RESTConfig)
```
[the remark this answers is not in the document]
wking: I guess I can add the helper struct here now, and copy it into library-go later...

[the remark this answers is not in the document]
wking: Struct added in 0a9024bb1 -> e82184c92

pkg/cli/admin/release/extract_tools.go
wking: Refactored into an install-config function and a REST-config function with e82184c92 -> ee48493c0.

pkg/cli/admin/release/extract_tools.go
wking: I'll use them there in a bit, but they're two functions now, so linters will catch any unused variables ;)

wking:
I've pushed ee48493c0 -> ac6a9032d delivering cluster-profile extraction from the `cluster-version-operator` Deployment, which was the final missing piece, so I've removed the WIP tag.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -908,3 +954,121 @@ func copyAndReplace(errorOutput io.Writer, w io.Writer, r io.Reader, bufferSize
...
+		config.Overrides = clusterVersion.Spec.Overrides
+		config.Capabilities = &clusterVersion.Status.Capabilities
+
+		// FIXME: eventually pull in GetImplicitlyEnabledCapabilities from https://github.com/openshift/cluster-version-operator/blob/86e24d66119a73f50282b66a8d6f2e3518aa0e15/pkg/payload/payload.go#L237-L240 for cases where a minor update would implicitly enable some additional capabilities.  For now, 4.13 to 4.14 will always enable MachineAPI.
+		currentVersion := clusterVersion.Status.Desired.Version
+		matches := regexp.MustCompile(`^(\d+[.]\d+)[.].*`).FindStringSubmatch(currentVersion)
+		if len(matches) < 2 {
+			return config, fmt.Errorf("failed to parse major.minor version from ClusterVersion status.desired.version %q", currentVersion)
+		} else if matches[1] == "4.13" {
+			machineAPI := configv1.ClusterVersionCapability("MachineAPI")
```
wking, line 1035: This bit (from the earlier FIXME coment through this whole code block) is a ugly hack to work around 4.13-to-4.14 updates implicitly enabling `MachineAPI`.  I hope to drop it soon, and replace it with a [`GetImplicitlyEnabledCapabilities`][1] shifted to library-go.  But we've been through a few rounds of library-go bumps already, and this temporary hack seems reliable enough (to me) for 4.14, with something more robust landing shortly afterwards for 4.15+.

[1]: https://github.com/openshift/cluster-version-operator/blob/86e24d66119a73f50282b66a8d6f2e3518aa0e15/pkg/payload/payload.go#L237-L240

wking: Oh, and the reason we can't just ignore `MachineAPI` and assume that 4.13 clusters will already have 4.14-compatible credentials in-cluster is:

```console
$ oc adm release extract --to 4.13 --credentials-requests quay.io/openshift-release-dev/ocp-release:4.13.0-x86_64
$ oc adm release extract --to 4.14 --credentials-requests quay.io/openshift-release-dev/ocp-release:4.14.0-ec.4-x86_64
$ diff -u3 <(yaml2json <4.13/0000_30_machine-api-operator_00_credentials-request.yaml | jq '.[] | select(.metadata.name == "openshift-machine-api-powervs").spec.providerSpec.policies[]') <(yaml2json <4.14/0000_30_machine-api-operator_00_credentials-request.yaml | jq '.[] | select(.metadata.name == "openshift-machine-api-powervs").spec.providerSpec.policies[]')
--- /dev/fd/63  2023-08-09 16:41:04.479985075 -0700
+++ /dev/fd/62  2023-08-09 16:41:04.482985075 -0700
@@ -22,3 +22,16 @@
     "crn:v1:bluemix:public:iam::::role:Viewer"
   ]
 }
+{
+  "attributes": [
+    {
+      "name": "serviceName",
+      "value": "is"
+    }
+  ],
+  "roles": [
+    "crn:v1:bluemix:public:iam::::role:Editor",
+    "crn:v1:bluemix:public:iam::::role:Operator",
+    "crn:v1:bluemix:public:iam::::role:Viewer"
+  ]
+}
```

4.13 PowerVS clusters will need that additional policy before updating to 4.14, and the extraction logic needs to know that after the update to 4.14, `MachineAPI`  will be enabled so it can include that CredentialsRequest.  The only other `MachineAPI` change from 4.13 to 4.14 is `openshift-machine-api-azure` adding `serviceAccountNames`, and I'm not sure if that matters for `ccoctl` or not.

pkg/cli/admin/release/extract.go
petr-muller: Yeah I see the usefulness, it's basically an annotated boolean. My concern is mostly that using `error` in this way is surprising for someone who's expecting idiomatic Go.

Just engaging in academic discussion here, I don't think you need to change the code :) It follows existing code which has a reasonable Godoc, so I'm good with it.

pkg/cli/admin/release/extract.go
```diff
@@ -133,8 +142,11 @@ func NewExtract(f kcmdutil.Factory, streams genericclioptions.IOStreams) *cobra.
 	flags.StringVar(&o.CommandOperatingSystem, "command-os", o.CommandOperatingSystem, "Override which operating system command is extracted (mac, windows, linux) or can be specified with arch(linux/arm64, mac/amd64). You map specify '*' to extract all tool archives.")
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be copied under.")
 
-	flags.BoolVar(&o.CredentialsRequests, "credentials-requests", o.CredentialsRequests, "Extract credential request manifests only")
-	flags.StringVar(&o.Cloud, "cloud", o.Cloud, "Specify the cloud for which credential request manifests should be extracted. Works only in combination with --credentials-requests.")
+	flags.BoolVar(&o.Included, "included", o.Included, "Exclude manifests that are not expected to be included in cluster reconciliation.")
+	flags.StringVar(&o.InstallConfig, "install-config", o.InstallConfig, "Path to an install-config file, as consumed by the openshift-install command.  Works only in combination with --included.")
```
LalatenduMohanty, line 146: The assumption here is that install-config only makes sense for manifests which will be reconciled by CVO which is part of `included`. Can you clarify? I am wondering  why we can not call install-config and included separately even if the result might be same for installed clusters?

pkg/cli/admin/release/extract_tools.go
```diff
@@ -908,3 +954,121 @@ func copyAndReplace(errorOutput io.Writer, w io.Writer, r io.Reader, bufferSize
...
+		config.RequiredFeatureSet = pointer.String(string(featureGate.Spec.FeatureSet))
+	}
+
+	if clusterVersion, err := client.ClusterVersions().Get(ctx, "version", metav1.GetOptions{}); err != nil {
+		return config, err
+	} else {
+		config.Overrides = clusterVersion.Spec.Overrides
+		config.Capabilities = &clusterVersion.Status.Capabilities
+
+		// FIXME: eventually pull in GetImplicitlyEnabledCapabilities from https://github.com/openshift/cluster-version-operator/blob/86e24d66119a73f50282b66a8d6f2e3518aa0e15/pkg/payload/payload.go#L237-L240 for cases where a minor update would implicitly enable some additional capabilities.  For now, 4.13 to 4.14 will always enable MachineAPI.
```
LalatenduMohanty, line 1029: Are we not doing it because we need to move this code to API code first then change it here?

pkg/cli/admin/release/extract_tools.go
LalatenduMohanty: Sounds good to me. Can we create a Jira to track the library-go change?

wking: I've created [OTA-1010](https://issues.redhat.com/browse/OTA-1010).

LalatenduMohanty:
The PR looks good to me. I have added couple of comments, take a look. Hopefully I will add lgtm after you respond to those comments

pkg/cli/admin/release/extract.go
wking: With `--included` the extraction will have manifests that the CVO is expected to include.  Without `--included`, the extraction will have those and also have manifests that the CVO is expected to exclude.  So we need to know "is the CVO expected to include this manifest?" with `--included`, but we don't care about trying to predict the CVO without `--included`.  And because the install config is one way for us to get the information we need to try to predict the CVO, we don't need an install-config without `--included`.  That was a lot of words, but the backing Go may be more clear:

```go
		if o.Included {
			if o.InstallConfig == "" {
				inclusionConfig, err = findClusterIncludeConfig(ctx, o.RESTConfig)
			} else {
				inclusionConfig, err = findClusterIncludeConfigFromInstallConfig(ctx, o.InstallConfig)
				context = o.InstallConfig
			}
```

Folks who set `--install-config` without setting `--included` would just have the install-config ignored, so we are failing fast to point out the unnecessary argument and allow the user to re-evaluate what they're trying to accompllish.

LalatenduMohanty (approved):
/lgtm

wking:
@jiajliu, I'm adding a hold to give you time for more pre-merge testing.  Please remove the hold when you are satisfied with (OTA-968)[https://issues.redhat.com/browse/OTA-968].  I'm also fine with you removing the hold if you want to punt some of the testing to post-merge.  Whatever works for you :)

/hold

shellyyang1989:
Looks like all the QE issues against the PR have been addressed, so

/unhold
/label qe-approved

wking:
We have AWS lease capacity again.

/retest-required

cgwalters:
So Github auto-links in the web page when you reference one PR from another.

But IMO a large procedural mistake here is that comments about bugs in a PR *still make sense to add* to a closed PR because it will cause the original author of the PR and reviewers/commenters to be notified.

In this case the author happened to be notified out of band on some other internal proprietary chat.  [Avoid the temptation to default to internal proprietary chat](https://gist.github.com/cgwalters/536737496d01a8bf3c641b844f0cd75c)! 

Again, adding a comment here ensures that the original people are involved are CC'd by default.

PR #1524, opened 2023-08-03
WINC-692: Add Windows node functionality to debug command

sebsoto:
Allows oc debug to be ran against a Windows node. Adds Windows specific fields to the pod spec, if the node is detected to be Windows.

When debugging a Windows node the image option is required as we cannot ship Windows images as part of the OpenShift payload. This means we cannot set a default image, and the user must provide their own.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -65,7 +65,6 @@ const (
 	containerResourcesAnnotationPrefix = "resources.workload.openshift.io/"
 	// podWorkloadTargetAnnotationPrefix contains the prefix for the pod workload target annotation
 	podWorkloadTargetAnnotationPrefix = "target.workload.openshift.io/"
-	kubeOSNodeSelector                = "kubernetes.io/os"
 	commandLinuxShell                 = "/bin/sh"
 	commandWindowsShell               = "cmd.exe"
 )
@@ -108,6 +107,10 @@ var (
 
 		# Debug a node as an administrator
 		oc debug node/master-1
+
+		# Debug a Windows Node
+		# Note: the chosen image must match the Windows Server version (2019, 2022) of the Node
+		oc debug node/win-worker-1 --image=mcr.microsoft.com/powershell:lts-nanoserver-ltsc2022
 
 		# Launch a shell in a pod using the provided image stream tag
 		oc debug istag/mysql:latest -n openshift
@@ -539,7 +542,9 @@ func (o *DebugOptions) RunDebug() error {
 				fmt.Fprintf(o.ErrOut, "Starting pod/%s ...\n", pod.Name)
 			}
 			if o.IsNode {
-				fmt.Fprintf(o.ErrOut, "To use host binaries, run `chroot /host`\n")
+				if !(template.Spec.OS != nil && template.Spec.OS.Name == corev1.Windows) {
+					fmt.Fprintf(o.ErrOut, "To use host binaries, run `chroot /host`\n")
+				}
 			}
 		}
 
@@ -978,7 +983,7 @@ func (o *DebugOptions) getContainerCommand() []string {
 	if len(o.Command) == 1 && o.Command[0] == commandLinuxShell {
 		pod := o.Attach.Pod
 
-		if val, ok := pod.Spec.NodeSelector[kubeOSNodeSelector]; ok && val == "windows" {
+		if pod.Spec.OS != nil && pod.Spec.OS.Name == corev1.Windows {
 			return []string{commandWindowsShell}
 		}
 	}
@@ -997,12 +1002,11 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 			// TODO: allow --as-root=false to skip all the namespaces except network
 			return nil, fmt.Errorf("can't debug nodes without running as the root user")
 		}
-		if t.Labels["kubernetes.io/os"] == "windows" {
-			// Windows nodes don't yet support privileged containers, which is required for debug
-			return nil, fmt.Errorf("can't debug Windows nodes")
-		}
 		image := o.Image
 		if len(o.Image) == 0 {
+			if t.Labels[corev1.LabelOSStable] == string(corev1.Windows) {
+				return nil, fmt.Errorf("--image must be set when debugging Windows nodes")
+			}
 			imageStream := o.ImageStream
 			if len(o.ImageStream) == 0 {
 				imageStream = "openshift/tools:latest"
@@ -1020,7 +1024,7 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 		zero := int64(0)
 		isTrue := true
 		hostPathType := corev1.HostPathDirectory
-		return &corev1.PodTemplateSpec{
+		template := &corev1.PodTemplateSpec{
 			Spec: corev1.PodSpec{
 				NodeName:    t.Name,
 				HostNetwork: true,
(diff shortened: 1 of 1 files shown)


pkg/cli/debug/debug.go
```diff
@@ -978,7 +977,7 @@ func (o *DebugOptions) getContainerCommand() []string {
 	if len(o.Command) == 1 && o.Command[0] == commandLinuxShell {
 		pod := o.Attach.Pod
 
-		if val, ok := pod.Spec.NodeSelector[kubeOSNodeSelector]; ok && val == "windows" {
+		if pod.Spec.OS != nil && pod.Spec.OS.Name == corev1.Windows {
```
ardaguclu, line 980: What is the reason of the change in this line?

pkg/cli/debug/debug.go
```diff
@@ -997,12 +996,11 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 			// TODO: allow --as-root=false to skip all the namespaces except network
 			return nil, fmt.Errorf("can't debug nodes without running as the root user")
 		}
-		if t.Labels["kubernetes.io/os"] == "windows" {
-			// Windows nodes don't yet support privileged containers, which is required for debug
-			return nil, fmt.Errorf("can't debug Windows nodes")
-		}
 		image := o.Image
 		if len(o.Image) == 0 {
+			if t.Labels[corev1.LabelOSStable] == "windows" {
```
ardaguclu, line 1001:
```suggestion
			if t.Labels[corev1.LabelOSStable] == corev1.Windows {
```

pkg/cli/debug/debug.go
```diff
@@ -1063,7 +1061,19 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 					},
 				},
 			},
-		}, nil
+		}
+		if t.Labels[corev1.LabelOSStable] == "windows" {
+			template.Spec.OS = &corev1.PodOS{Name: corev1.Windows}
+			template.Spec.HostPID = false
+			template.Spec.HostIPC = false
+			containerUser := "ContainerUser"
```
ardaguclu, line 1069: From where this user comes from and why it is static?

pkg/cli/debug/debug.go
```diff
@@ -1063,7 +1061,19 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 					},
 				},
 			},
-		}, nil
+		}
+		if t.Labels[corev1.LabelOSStable] == "windows" {
+			template.Spec.OS = &corev1.PodOS{Name: corev1.Windows}
+			template.Spec.HostPID = false
```
ardaguclu, line 1067: We are setting HostPID and HostIPC false, I wonder how user will debug node from this pod? Because there seems to be no privilege to access the node components.

pkg/cli/debug/debug.go
sebsoto: The reason is that `pod.Spec.NodeSelector[kubeOSNodeSelector]` is not set on the debug pod, making this if statement never evaluate to true.
It seems that the recommended way to check if a pod is meant for Windows is to use OS field, so I opted for that method.

pkg/cli/debug/debug.go
sebsoto: HostPID and HostIPC are not usable with Windows pods:
https://kubernetes.io/docs/concepts/windows/intro/#compatibility-linux-similarities

```
...
If you set the .spec.os.name field to windows, you must not set the following fields in the .spec of that Pod:

    spec.hostPID
    spec.hostIPC
...
```

pkg/cli/debug/debug.go
sebsoto: There are only two possible users for Windows containers, ContainerUser and ContainerAdministrator.
https://learn.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-security#when-to-use-containeradmin-and-containeruser-user-accounts

ContainerAdministrator would give the debug container full root access to the VM, so I'd think ContainerUser would be preferred.

sebsoto (commented):
Thank you for the review @ardaguclu, I have pushed a change addressing one of your comments, and responded to the rest.

pkg/cli/debug/debug.go
```diff
@@ -1063,7 +1061,19 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
 					},
 				},
 			},
-		}, nil
+		}
+		if t.Labels[corev1.LabelOSStable] == "windows" {
```
ardaguclu, line 1065: nit:
```suggestion
		if t.Labels[corev1.LabelOSStable] == corev1.Windows {
```

pkg/cli/debug/debug.go
ardaguclu: In addition to your comment, according to here https://github.com/openshift/oc/blob/4eb0aab1f3183374b543fab7ef04e29b56b560c5/vendor/k8s.io/api/core/v1/types.go#L3426-L3450, these fields should be cleaned to default values which is false.

pkg/cli/debug/debug.go
ardaguclu: Good idea to use `ContainerUser` instead admin.

ardaguclu (commented):
Thanks. Changes look good to me but I think we need a pre-merge testing to move forward. @sebsoto Did you QE do pre-merge testing for this feature?.

jrvaldes:
LGTM

pkg/cli/debug/debug.go
```diff
@@ -1063,7 +1061,42 @@ func (o *DebugOptions) approximatePodTemplateForObject(object runtime.Object) (*
...
 				},
 			},
-		}, nil
+		}
+		if t.Labels[corev1.LabelOSStable] == string(corev1.Windows) {
+			template.Spec.OS = &corev1.PodOS{Name: corev1.Windows}
+			containerUser := "ContainerUser"
+			// Override all fields that will throw an error if used on a Windows Pod
+			template.Spec.HostPID = false
+			template.Spec.HostIPC = false
```
ardaguclu, line 1070: I don't think it is feasible to explicitly setting all these fields. Because their defaults are already nil or false or empty. What will happen when there is a new field arise?. I'd recommend reverting to the previous version of this PR and QE does verification on it.

sebsoto: Makes sense, I misunderstood your previous comment.
I will ask QE to do testing against the new revision

ardaguclu: ah, sorry, I should have been more clear.

jfrancoa:
@ardaguclu I have already pre-merge testing the feature. In general, it's working great I just have couple of usability comments:

* The first one is the default output intended for the Linux debug pods:
```
```

I would recommend changing in into:
```
To use host binaries, run `powershell -wd C:\host` or `pwsh.exe -wd C:\host`
```

* The usage of the --image parameter is not very intuitive. If it's the first time a user tries to debug a Windows node there might be doubts on which image to use. I would recommend modifying the `oc debug --help `  and leave an example on how to use the oc debug command for a Windows node, for example, using the nanoserver image which is quicker to spawn:
```
oc debug node/winworker-q775m --image=mcr.microsoft.com/powershell:lts-nanoserver-ltsc2022
```
Besides this, the command worked well, I didn't find any security concerns and all the host files were accessible for debugging.

sebsoto:
@ardaguclu @jfrancoa 
I've pushed the suggested changes.
I chose not to include: `To use host binaries, run powershell -wd C:\host` or `pwsh.exe -wd C:\host`, and opted instead to only print `To use host binaries, run chroot /host` for Linux debug pods.
The reasoning for this is that depending on the image the user provides that may or may not be true. So I'd rather just be more cautious and say nothing.

jfrancoa (approved):
lgtm

sebsoto:
@ardaguclu could you take another look at this?

pkg/cli/debug/debug.go
```diff
@@ -109,6 +108,10 @@ var (
 		# Debug a node as an administrator
 		oc debug node/master-1
 
+		# Debug a Windows Node
+		# Note: the chosen image must match the Windows Server version (2019, 2022) of the Node
```
ardaguclu, line 112: Is it possible to add a validation for this?

ardaguclu (commented):
Only minor comment, other than that looks good to me. Thanks.

pkg/cli/debug/debug.go
sebsoto: Not that I know of

PR #1525, opened 2023-08-11
Update openshift/api

ardaguclu:
This PR updates openshift/api to latest one to get the httpHeader field in route api.


ardaguclu:
/hold
until we get a proof that this solves `oc edit` failure route api getting.

ardaguclu:
It seems that we also need to update other dependencies as well.

ardaguclu:
According to my tests, this patch does not fix this bug https://issues.redhat.com/browse/OCPBUGS-17636 because latest version of oc also works without any problem. But syncing `oc` with the latest `openshift/api` is something we should regularly do, in order to get the latest changes on APIs.

ardaguclu:
/hold cancel
/cc @ingvagabund 
Could you please have a look at this?. Thanks.

ardaguclu:
unrelated failure;
/retest-required

PR #1526, opened 2023-08-14
Add a safety check for debugging a pod with persistent volumes

gnufied:
xref https://issues.redhat.com/browse/OCPBUGS-16357

I wonder if we can add a flag to force debugging of such pods. But I am not sure what name we can use.

**Why are we doing this?**

Running multiple pods that share a volume and concurrently write to it, is a fundamentally unsafe operation and I am afraid this will cause data loss/corruption, unless application have been designed to tolerate it.  An user using `oc debug` unwittingly may not realize that and can lose their data. 

For example - lets say I am running *mysql* in a pod and apart from the container that runs database, there is a sidecar that initializes some databases/tables. Now when I `oc debug` that pod - I could have have my existing data botched in a way that could become unrecoverable because debug pod's sidecar tried to initialize already initialized databases/tables.

There is another problem of - pods that use `ReadWriteOncePod` volumes can't be debugged this way anyways, because scheduler will prevent scheduling of such pods and kubelet will prevent mount of such volumes more than once.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -499,6 +499,12 @@ func (o *DebugOptions) RunDebug() error {
 		return nil
 	}
 
+	msg, okayToDebug := safeToDebugPod(pod)
+	if !okayToDebug {
+		klog.Errorf(msg)
+		os.Exit(1)
+	}
+
 	klog.V(5).Infof("Creating pod: %#v", pod)
 	pod, err = o.createPod(pod)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/cli/debug/debug.go
```diff
@@ -499,6 +499,12 @@ func (o *DebugOptions) RunDebug() error {
 		return nil
 	}
 
+	msg, okayToDebug := safeToDebugPod(pod)
+	if !okayToDebug {
+		klog.Errorf(msg)
```
ardaguclu, line 504: This is a breaking change because we are no more allowing **any** pod is being debugged via debug command.

ardaguclu (commented):
I don't think it is a good idea to dropping debug command support of pods having persistent volumes(not only because of it is breaking change, but also I don't think there is a sufficient reason to do that) because same problem also happens in `kubectl debug --copy-to`.
On the other hand, I understand the risk poses in here and am favor of adding a some kind of a warning. But first I'd like to understand what is the problem. We have several flags and can't we suggest user to use them to overcome this problem;

* Is this a problem stems from annotations and labels are being cleared - we can warn user to pass `--keep-annotation` or `--keep-labels` flags.
* Is this a problem that first container is being picked as default debug container - we can warn user to select correct container via `--container` flag
* Is this a problem that all containers in the pod are copied - we can warn user to pass `--one-container` flag.

gnufied:
Not to mention there are scheduling concerns. I think that a volume that has been declared `ReadWriteOncePod` will prevent scheduling of second debug pod, because truly most read-write-once volumes are not designed to be used this way.    I am afraid this was an oversight that, we allowed this to happen in the first place.

PR #1530, opened 2023-08-22
OCPBUGS-17925: pkg/cli/admin/prune/images: omit not found error for deployment configs

flavianmissi:
Allows the pruner to run on clusters without the DeploymentConfig capability.

### DIFF
--- a/pkg/cli/admin/prune/images/images.go
+++ b/pkg/cli/admin/prune/images/images.go
@@ -270,9 +270,11 @@ func (o PruneImagesOptions) Validate() error {
 	return nil
 }
 
-var errNoRegistryURLPathAllowed = errors.New("no path after <host>[:<port>] is allowed")
-var errNoRegistryURLQueryAllowed = errors.New("no query arguments are allowed after <host>[:<port>]")
-var errRegistryURLHostEmpty = errors.New("no host name specified")
+var (
+	errNoRegistryURLPathAllowed  = errors.New("no path after <host>[:<port>] is allowed")
+	errNoRegistryURLQueryAllowed = errors.New("no query arguments are allowed after <host>[:<port>]")
+	errRegistryURLHostEmpty      = errors.New("no host name specified")
+)
 
 // validateRegistryURL returns error if the given input is not a valid registry URL. The url may be prefixed
 // with http:// or https:// schema. It may not contain any path or query after the host:[port].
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/prune/images/images.go
```diff
@@ -355,7 +357,7 @@ func (o PruneImagesOptions) Run() error {
 	}
 
 	allDCs, err := o.AppsClient.DeploymentConfigs(o.Namespace).List(context.TODO(), metav1.ListOptions{})
-	if err != nil {
+	if err != nil && !kerrors.IsNotFound(err) {
```
ardaguclu, line 360: As far as I understand, `allDCs` will not be nil just like https://github.com/openshift/oc/blob/61cc186c35bd568e16dd62712956e27c40cd65a2/pkg/cli/admin/prune/images/images.go#L335 to pass to https://github.com/openshift/oc/blob/68e3d16b4426d95374701fbb0b23f0448b04f29c/pkg/cli/admin/prune/imageprune/prune.go#L658.

ardaguclu (commented):
LGTM

pkg/cli/admin/prune/images/images.go
flavianmissi: Thanks for the quick review @ardaguclu.
Do you mean that it's safe to pass `allDCs` around, thus we do not need to further modify the code?

ardaguclu: According to the `Build` resource, it seems safe but I think, we need to ensure that it won't panic

flavianmissi: Indeed. I haven't been able to try this out o a running cluster yet (getting errors on bootstrap), will ping IR QE to help us.

flavianmissi:
holding until QE has the chance to test it.
/hold

flavianmissi:
I've gotten confirmation from the bug reporter that this change fixes the issue, so I think it's okay that QE tests it after. I'll remove the hold.
/hold cancel

flavianmissi:
@ardaguclu could you give the lgtm label?

xiuwang:
/label qe-approved

Validate this oc working https://issues.redhat.com/browse/OCPBUGS-17925?focusedId=22864377&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22864377

ardaguclu:
Thanks for the fix;
/lgtm

PR #1533, opened 2023-09-05
WRKLDS-860: Add volume usage percentage checker script in must-gather's gather container

ardaguclu:
Since must-gather pod has to run on master node(because it requires to gather
information from the critical components, in case worker nodes are not accessible),
must-gather pods should not affect the functionality of the critical components
such as API server, controller manager, etc.

Currently, must-gather pod uses an emptydir to store all gathered information and
`copy` container copies this to local from the same mounted emptydir. However,
in some rare cases, must-gather pod may fill the entire emptydir allocated by
master node and cause issues for the other components residing in master node
also needs to use storage.

This PR introduces a new flag to limit the volume size usage percentage with
respect to the total volume size allocated for must-gather pod. 

This PR adds a new volume usage percentage checker script in must-gather's `gather` container.
This script will check the volume usage percentage per predefined seconds(per 5 second)
and if the volume usage percentage exceeds the number defined in maximum usage
flag(default is 30%), script will kill the gather container(to stop gathering) 
but `copy` container will continue copying.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -76,6 +76,22 @@ var (
 		# Gather information using a specific image, command, and pod directory
 		  oc adm must-gather --image=my/image:tag --source-dir=/pod/directory -- myspecial-command.sh
 	`)
+
+	volumeUsageCheckerScript = `
+echo "volume percentage checker started....."
+while true; do 
+disk_usage=$(du -s "%s" | awk '{print $1}')
+disk_space=$(df -P "%s" | awk 'NR==2 {print $2}')
+usage_percentage=$(( (disk_usage * 100) / disk_space ))
+echo "volume usage percentage $usage_percentage" 
+if [ "$usage_percentage" -gt "%d" ]; then 
+	echo "Disk usage exceeds the volume percentage of %d for mounted directory. Exiting..." 
+	# kill gathering process in gather container to prevent disk to use more.
+	pkill --signal SIGKILL -f %s
+	exit 1
+fi
+sleep 5
+done`
 )
 
 func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
@@ -101,6 +117,7 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 	cmd.Flags().StringVar(&o.SourceDir, "source-dir", o.SourceDir, "Set the specific directory on the pod copy the gathered data from.")
 	cmd.Flags().StringVar(&o.timeoutStr, "timeout", "10m", "The length of time to gather data, like 5s, 2m, or 3h, higher than zero. Defaults to 10 minutes.")
 	cmd.Flags().StringVar(&o.RunNamespace, "run-namespace", o.RunNamespace, "An existing privileged namespace where must-gather pods should run. If not specified a temporary namespace will be generated.")
+	cmd.Flags().Uint8Var(&o.VolumePercentage, "volume-percentage", o.VolumePercentage, "Specify maximum percentage of must-gather pod's allocated volume that can be used. If this limit is exceeded, must-gather will stop gathering, but still copy gathered data. Defaults to 30%.")
 	cmd.Flags().BoolVar(&o.Keep, "keep", o.Keep, "Do not delete temporary resources when command completes.")
 	cmd.Flags().MarkHidden("keep")
 
@@ -109,11 +126,12 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 
 func NewMustGatherOptions(streams genericiooptions.IOStreams) *MustGatherOptions {
 	return &MustGatherOptions{
-		SourceDir: "/must-gather/",
-		IOStreams: streams,
-		LogOut:    newPrefixWriter(streams.Out, "[must-gather      ] OUT"),
-		RawOut:    streams.Out,
-		Timeout:   10 * time.Minute,
+		SourceDir:        "/must-gather/",
+		IOStreams:        streams,
+		LogOut:           newPrefixWriter(streams.Out, "[must-gather      ] OUT"),
+		RawOut:           streams.Out,
+		Timeout:          10 * time.Minute,
+		VolumePercentage: 30,
 	}
 }
 
@@ -234,18 +252,19 @@ type MustGatherOptions struct {
 	ImageClient      imagev1client.ImageV1Interface
 	RESTClientGetter genericclioptions.RESTClientGetter
 
-	NodeName     string
-	NodeSelector string
-	HostNetwork  bool
-	DestDir      string
-	SourceDir    string
-	Images       []string
-	ImageStreams []string
-	Command      []string
-	Timeout      time.Duration
-	timeoutStr   string
-	RunNamespace string
-	Keep         bool
+	NodeName         string
+	NodeSelector     string
+	HostNetwork      bool
+	DestDir          string
+	SourceDir        string
+	Images           []string
+	ImageStreams     []string
+	Command          []string
+	Timeout          time.Duration
+	timeoutStr       string
+	RunNamespace     string
+	VolumePercentage uint8
+	Keep             bool
 
 	RsyncRshCmd string
 
@@ -270,6 +289,14 @@ func (o *MustGatherOptions) Validate() error {
 	if strings.Contains(o.DestDir, ":") {
 		return fmt.Errorf("--dest-dir may not contain special characters such as colon(:)")
 	}
+
+	if o.VolumePercentage <= 0 || o.VolumePercentage > 100 {
+		return fmt.Errorf("invalid volume usage percentage, please specify a value between 0 and 100")
+	}
+	if o.VolumePercentage >= 80 {
+		klog.Warningf("volume percentage greater than or equal to 80 might cause filling up the disk space and have an impact on other components running on master")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -108,11 +110,12 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericclioptions.IOStream
...
-		IOStreams: streams,
-		LogOut:    newPrefixWriter(streams.Out, "[must-gather      ] OUT"),
-		RawOut:    streams.Out,
-		Timeout:   10 * time.Minute,
+		SourceDir:  "/must-gather/",
+		IOStreams:  streams,
+		LogOut:     newPrefixWriter(streams.Out, "[must-gather      ] OUT"),
+		RawOut:     streams.Out,
+		Timeout:    10 * time.Minute,
+		VolumeSize: "3Gi",
```
ardaguclu, line 118: @tssurya @sferich888 you did a good amount of analysis about the must-gather pod's volume size usage on this https://github.com/openshift/must-gather/pull/370. Would you mind, if I ask your opinions about this default `3Gi`?.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -740,10 +749,10 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool) *corev1.P
 		},
 		Spec: corev1.PodSpec{
 			NodeName: node,
-			// This pod is ok to be OOMKilled but not preempted. Following the conventions mentioned at:
+			// This pod is ok to be OOMKilled and preempted. Following the conventions mentioned at:
 			// https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#priority-classes
-			// so setting priority class to system-cluster-critical
-			PriorityClassName: "system-cluster-critical",
+			// so setting priority class to openshift-user-critical
+			PriorityClassName: "openshift-user-critical",
```
ardaguclu, line 755: @ingvagabund would have any reservations about changing this to `openshift-user-critical`?

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -269,6 +273,11 @@ func (o *MustGatherOptions) Validate() error {
 	if strings.Contains(o.DestDir, ":") {
 		return fmt.Errorf("--dest-dir may not contain special characters such as colon(:)")
 	}
+
+	_, err := resource.ParseQuantity(o.VolumeSize)
+	if err != nil {
+		return fmt.Errorf("invalid --volume-limit value: %v", err)
```
ardaguclu, line 279:
```suggestion
		return fmt.Errorf("invalid --volume-size value: %v", err)
```

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: My only reservation would be inability to deploy a must-gather pod in cases the monitoring workload eats up all the available resources. For which one will not be able to run the must-gather. Nevertheless, that's a case that can be analyzed by collecting running `oc inspect` and `oc logs/get/...` commands.

Given all control plane components are of the highest priority and each master node has the minimum resources required, the scheduler will always be able to schedule new control plane components by preemption. The installer pods have [system-node-critical](https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml#L47) pc.

Though, this will deserve few lines in the release notes so users are aware of this change.

ingvagabund: @mfojtik @soltysh any objections here?

ardaguclu:
/hold cancel
we already require acknowledge-critical-fixes-only label

pkg/cli/admin/mustgather/mustgather.go
soltysh: We're talking about it on slack, I'll update once we settle. For now I'm strongly against such a change.

soltysh (changes requested):
/hold
until we reach consensus for the priority class

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: I have dug the kubernetes code base and this https://github.com/kubernetes/kubernetes/blob/02960a8253ecf89d0487bb8aacdf499c80331064/pkg/kubelet/eviction/eviction_manager.go#L573 prevents the `system-cluster-critical` priority classes are being evicted.
As a workaround to not change the priority class of must-gather;
* in a separete goroutine we can listen this log https://github.com/kubernetes/kubernetes/blob/02960a8253ecf89d0487bb8aacdf499c80331064/pkg/kubelet/eviction/eviction_manager.go#L574 via `oc adm node-logs` (assuming that this error message won't be changed in the future)
* When we see the error message, we manually kill the gather container via `oc exec -- kill 1` (there is no differentiator about the eviction reason and we assume that this is due to volume is exceeded)

This way would work as if user passes `--keep` because we are trying to keep the pod to let user downloads the logs.

@soltysh what do you think about this?

ardaguclu: But there is no guarantee that [this log](https://github.com/kubernetes/kubernetes/blob/02960a8253ecf89d0487bb8aacdf499c80331064/pkg/kubelet/eviction/eviction_manager.go#L574) will not be changed(or even removed).

I start thinking that what about we lower the priority class name to openshift-user-critical and for catastrophic cases, introduce a hidden flag(to use `system-cluster-critical` priority class name) by warning user about the possible risks.

ingvagabund: No matter which solution gets implemented, would it make sense to prioritize which resources get collected first? If the size gets limited, we should collect the most important resources first.

pkg/cli/admin/mustgather/mustgather.go
sferich888: If we collect more than 3 GB I think we have a problem; In general we collect 800 MB to 1.5 GB (if I recall). So a 3 GB limit gives us some buffer (however I don't think we really should be collecting that much data); we really need to look at how to make this smaller (but that is for a different ticket).

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: > No matter which solution gets implemented, would it make sense to prioritize which resources get collected first? If the size gets limited, we should collect the most important resources first. So whatever gets left is still usable for analysis. To at least get some pointers where to look. E.g. collect pod logs first, then their manifests, then other resources?

Sounds like a good enhancement for the future releases. We need to reevaluate the greedy resource traversal and may introduce a priority base search perhaps.

ardaguclu: @soltysh as we discussed on Slack. I changed this PR in the way that as you suggested. Please let me know what you think, when you have a chance. Thanks.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -101,6 +117,7 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 	cmd.Flags().StringVar(&o.SourceDir, "source-dir", o.SourceDir, "Set the specific directory on the pod copy the gathered data from.")
 	cmd.Flags().StringVar(&o.timeoutStr, "timeout", "10m", "The length of time to gather data, like 5s, 2m, or 3h, higher than zero. Defaults to 10 minutes.")
 	cmd.Flags().StringVar(&o.RunNamespace, "run-namespace", o.RunNamespace, "An existing privileged namespace where must-gather pods should run. If not specified a temporary namespace will be generated.")
+	cmd.Flags().Uint8Var(&o.VolumePercentage, "volume-percentage", o.VolumePercentage, "Specify what percentage of must-gather pod's allocated volume can be maximum used. 30% will be used as default and if this limit exceeds, must-gather will stop gathering but copying will continue until it is done.")
```
soltysh, line 120:
```suggestion
	cmd.Flags().Uint8Var(&o.VolumePercentage, "volume-percentage", o.VolumePercentage, "Specify maximum percentage of must-gather pod's allocated volume that can be used. If this limit is exceeded, must-gather will stop gathering, but still copy gathered data. Defaults to 30%.")
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -270,6 +289,17 @@ func (o *MustGatherOptions) Validate() error {
 	if strings.Contains(o.DestDir, ":") {
 		return fmt.Errorf("--dest-dir may not contain special characters such as colon(:)")
 	}
+
+	if o.VolumePercentage > 100 {
+		return fmt.Errorf("volume usage percentage value can not be greater than 100")
+	}
+	if o.VolumePercentage == 0 {
+		return fmt.Errorf("invalid volume usage percentage, please specify a value between 0 and 100")
+	}
```
soltysh, line 298: I'd squash these two checks into:
```golang
if o.VolumePercentage <= 0 || o.VolumePercentage > 100 {
    return fmt.Errorf("invalid volume usage percentage, please specify a value between 0 and 100")
}
```

soltysh (commented):
Minor nits and this should be good.

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: Done

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: Updated.

soltysh (approved):
/lgtm

ardaguclu:
/hold cancel
I think, pre-merge tests have been completed and no issue found.

ardaguclu:
unrelated;
/retest

PR #1534, opened 2023-09-07
OCPBUGS-15772: Decouple new-app and set commands scheme from global scheme

ardaguclu:
Currently, route schema is registered for all commands in oc to
use it as native resource. Route is managed by openshift-apiserver and
up to this now, it works without any problem as well as vendoring
is happening in oc.

However, microshift does not have openshift-apiserver and that's why,
some commands behave differently whether object is native or CRD such
as `oc edit route`, `oc patch route`, etc. and these commands are failing on microshift.

To make `oc edit route` working on microshift, we have to remove registration
of routes.openshift.io and the rest of commands should work without any
problem embracing routes.openshift.io as CRD except that the `oc new-app`
and `oc set route-backends` commands. `oc new-app` must register Routes 
because there are templates creating route resources. 
In order to keep route templates working in
oc new-app command, this PR adds route.openshift.io registration just
for oc new-app and oc set route-backends and removes from the others.

### DIFF
--- a/cmd/oc/oc.go
+++ b/cmd/oc/oc.go
@@ -14,8 +14,6 @@ import (
 	"k8s.io/kubectl/pkg/scheme"
 
 	"github.com/openshift/oc/pkg/cli"
-	"github.com/openshift/oc/pkg/cli/set"
-	"github.com/openshift/oc/pkg/helpers/newapp"
 	schemehelper "github.com/openshift/oc/pkg/helpers/scheme"
 	"github.com/openshift/oc/pkg/version"
 
@@ -43,9 +41,8 @@ func main() {
 
 	// the kubectl scheme expects to have all the recognizable external types it needs to consume.  Install those here.
 	// We can't use the "normal" scheme because apply will use that to build stategic merge patches on CustomResources
+	// new-app and set commands don't use this global scheme. Instead they use their own scheme.
 	schemehelper.InstallSchemes(scheme.Scheme)
-	newapp.InstallSchemes()
-	set.InstallSchemes()
 
 	basename := filepath.Base(os.Args[0])
 	command := cli.CommandFor(basename)
--- a/cmd/oc/oc.go
+++ b/cmd/oc/oc.go
@@ -5,32 +5,19 @@ import (
 	"path/filepath"
 	"runtime"
 
+	"github.com/openshift/oc/pkg/helpers/newapp"
+
+	"k8s.io/kubectl/pkg/scheme"
+
 	"github.com/spf13/pflag"
 
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"github.com/openshift/library-go/pkg/serviceability"
 	kcli "k8s.io/component-base/cli"
 	"k8s.io/component-base/logs"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
-	"k8s.io/kubectl/pkg/scheme"
-
-	"github.com/openshift/api/apps"
-	"github.com/openshift/api/authorization"
-	"github.com/openshift/api/build"
-	"github.com/openshift/api/image"
-	"github.com/openshift/api/oauth"
-	"github.com/openshift/api/project"
-	quotav1 "github.com/openshift/api/quota/v1"
-	"github.com/openshift/api/route"
-	securityv1 "github.com/openshift/api/security/v1"
-	"github.com/openshift/api/template"
-	"github.com/openshift/api/user"
-	"github.com/openshift/library-go/pkg/serviceability"
 
 	"github.com/openshift/oc/pkg/cli"
-	"github.com/openshift/oc/pkg/helpers/legacy"
+	schemehelper "github.com/openshift/oc/pkg/helpers/scheme"
 	"github.com/openshift/oc/pkg/version"
 
 	// Import to initialize client auth plugins.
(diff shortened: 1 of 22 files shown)


ardaguclu:
/hold
until code reviews are done

pkg/cli/set/routebackends.go
```diff
@@ -93,7 +92,7 @@ type BackendsOptions struct {
 
 func NewBackendsOptions(streams genericclioptions.IOStreams) *BackendsOptions {
```
atiratree, line 93: I have not, but can you please test `oc set route-backends` properly? It seems we are using `scheme.DefaultJSONEncoder()`  in `CalculatePatchesExternal` which is used here and it does not have route in the schema as it uses the default one..

pkg/cli/set/scheme.go
```diff
@@ -0,0 +1,27 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	"github.com/openshift/api"
+	"github.com/openshift/api/route"
+
+	schemehelper "github.com/openshift/oc/pkg/helpers/scheme"
+)
+
+var setCustomScheme = runtime.NewScheme()
```
atiratree, line 13: this name is still unfortunately ambiguous,  can we for example name it `setCmdScheme` ?

cmd/oc/oc.go
```diff
@@ -57,18 +43,9 @@ func main() {
...
-	utilruntime.Must(oauth.Install(scheme.Scheme))
-	utilruntime.Must(project.Install(scheme.Scheme))
-	utilruntime.Must(installNonCRDQuota(scheme.Scheme))
-	utilruntime.Must(route.Install(scheme.Scheme))
-	utilruntime.Must(installNonCRDSecurity(scheme.Scheme))
-	utilruntime.Must(template.Install(scheme.Scheme))
-	utilruntime.Must(user.Install(scheme.Scheme))
-	legacy.InstallExternalLegacyAll(scheme.Scheme)
+	schemehelper.InstallSchemes(scheme.Scheme)
+	newapp.InstallSchemes()
```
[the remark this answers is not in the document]
ardaguclu: I added in here just for the readability because we need to warn code reader about new-app and set commands don't use global scheme. But I can also move these into their own init functions and can add a comment about this. WDYT?

pkg/cli/newapp/newapp.go
```diff
@@ -443,7 +442,7 @@ func (o *AppOptions) RunNewApp() error {
 			continue
 		}
 
-		obj, err := scheme.Scheme.New(unstructuredObj.GroupVersionKind())
+		obj, err := newapp.NewAppScheme.New(unstructuredObj.GroupVersionKind())
```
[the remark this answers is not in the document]
ardaguclu: Code in master branch uses 2 different schemes in new-app due to some reason(I need to check why). I'd prefer using separate schemes in this PR and try to combine them in a follow up PR?

pkg/cli/set/scheme.go
ardaguclu: Agree and `setCmdScheme` sounds good to me

pkg/cli/set/routebackends.go
ardaguclu: I have verified that `oc set route-backends` works without any problem with this patch. I think, it works with `scheme.DefaultJSONEncoder()` because default json encoder has a fallback mechanism for unregistered resources https://github.com/openshift/oc/blob/95512488a135d12d17cc271ff5fe9d57baa10baf/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/helpers.go#L489

cmd/oc/oc.go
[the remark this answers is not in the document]
atiratree: ah, you mean in a case somebody wants to modify the `schemehelper.InstallSchemes` functions? Ack, I think code-wise it would be better to move, and we can document this in all the places necessary.

pkg/cli/newapp/newapp.go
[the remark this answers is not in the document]
atiratree: ok, will leave it up to you

pkg/cli/set/routebackends.go
atiratree: hm, even if it works, if I look at the implementation in  
https://github.com/openshift/oc/blob/795bf1a6260847ecfc612da2ab11ea2d6e07da16/vendor/k8s.io/apimachinery/pkg/runtime/serializer/versioning/versioning.go#L217

(hopefully the correct file) I can see bunch of logic for the recognized types (eg. version conversion), I do not fully understand the implications that might occur in the future ,but think it would be safer to either pass the scheme or the encoder to `CalculatePatchesExternal` so each caller can use their scheme

ardaguclu: I agree that safest would be to pass the scheme into CalculatePatchesExternal, I'll update with that.

ardaguclu:
@atiratree could you PTAL, when you have a chance?. Thanks.

pkg/cli/set/scheme.go
```diff
@@ -0,0 +1,36 @@
...
+	// as native object which is normally managed
+	// by openshift-apiserver(microshift has no openshift-apiserver).
+	// But set route-backends command requires to
+	// register route to let some application templates including
+	// route continue working. That's why, we are manually
+	// registering route in here.
+	utilruntime.Must(route.Install(setCmdScheme))
+}
+
+func DefaultJSONEncoder() runtime.Encoder {
```
atiratree, line 34: I think the name is too generic and could be easily confused elsewhere

what about using the same prefix? And does it have to be public as well?

 ```suggestion
func setCmdSONEncoder() runtime.Encoder {
```

File: pkg/helpers/newapp/scheme.go [hunk 6]
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/openshift/api/route"
+
+	schemehelper "github.com/openshift/oc/pkg/helpers/scheme"
+)
+
+var NewAppScheme = runtime.NewScheme()
+
+var NewAppCodecs = serializer.NewCodecFactory(NewAppScheme)
+
+func InstallSchemes() {
```
atiratree, line 18: instead of having an init function here?

the advantages for having it here:
- contains the logic in a single place
- simplify the dependency graph between the packages

pkg/cli/set/scheme.go
ardaguclu: Done

pkg/helpers/newapp/scheme.go
ardaguclu: You are right, I updated.

ardaguclu:
errors seem not related to this PR;
/retest

pkg/helpers/newapp/scheme.go
```diff
@@ -0,0 +1,34 @@
...
+	// by openshift-apiserver(microshift has no openshift-apiserver).
+	// But new-app command requires to
+	// register route to let some application templates including
+	// route continue working. That's why, we are manually
+	// registering route in here.
+	utilruntime.Must(route.Install(NewAppScheme))
+}
+
+func init() {
+	InstallSchemes()
```
atiratree, line 33: nit: is having the inner function necessary?

atiratree (commented):
one nit, otherwise lgtm

pkg/helpers/newapp/scheme.go
ardaguclu: indeed, this function becomes redundant. Removed.

atiratree:
Build image cli-artifacts failed... seems due to `unauthorized: unable to validate token: ServiceUnavailable` - hopefully unrelated

/lgtm
/retest-required

PR #1537, opened 2023-09-13
The latest tags are openshift-clients-*.

adelton:
Avoid 4.2.0 being considered the current version.

We force the output to the vX.Y.Z format because that's what the rest of the code assumes.

### DIFF
--- a/vendor/github.com/openshift/build-machinery-go/make/lib/golang.mk
+++ b/vendor/github.com/openshift/build-machinery-go/make/lib/golang.mk
@@ -50,7 +50,7 @@ GO_TEST_FLAGS ?=-race
 
 GO_LD_EXTRAFLAGS ?=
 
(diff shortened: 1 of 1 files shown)


adelton:
The ci/prow/verify-deps at https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1537/pull-ci-openshift-oc-master-verify-deps/1701833325905186816 reports fail with
```
ERROR: Content of 'vendor/' directory doesn't match 'go.mod' configuration and the overrides in 'deps.diff'! 
```
However, `find . -name deps.diff` in the checkout does not find anything, so I'd need some help finding what the failure talks about.

adelton:
Hello @ardaguclu, @mfojtik, so we really not want this fixed?

PR #1538, opened 2023-09-13
poc for a dynamic token refresher

grokspawn:
This is a draft of a solution for something that is ugly as hell but works to refresh tokens during oc-mirror pulls.  There's definitely ways to better fit this into context, but I was trying to hop the fewest number of repos while researching a problem experienced by oc-mirror. 

This approach is not intended to merge.  It's just a "least disruptive" hack to demonstrate that a fix in this area would resolve the issue.

### DIFF
--- a/pkg/cli/image/manifest/dockercredentials/credential_store_factory.go
+++ b/pkg/cli/image/manifest/dockercredentials/credential_store_factory.go
@@ -1,10 +1,12 @@
 package dockercredentials
 
 import (
+	"net/url"
+	"sync"
+
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/distribution/distribution/v3/registry/client/auth"
 	"github.com/docker/docker/api/types"
-	dockerregistry "github.com/docker/docker/registry"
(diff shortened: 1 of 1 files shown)


grokspawn:
/test all 
... wanting to see how this interacts with CI ... at least once.

pkg/cli/image/manifest/dockercredentials/credential_store_factory.go
```diff
@@ -38,9 +40,48 @@ func (c *credentialStoreFactory) CredentialStoreFor(image string) auth.Credentia
 		return nocreds
 	}
 
-	return dockerregistry.NewStaticCredentialStore(&types.AuthConfig{
+	return NewDynamicCredentialStore(&types.AuthConfig{
```
ardaguclu, line 43: Although it sounds reasonable, this function is not only used by oc-mirror and I'd prefer leaving it as `dockerregistry.NewStaticCredentialStore` because there is no complaints about oc commands not refresh tokens.

PR #1540, opened 2023-09-15
OCPBUGS-7465: Introduce refreshable dynamic credential store for image manifests

ardaguclu:
This PR is basically identical to the proof of concept https://github.com/openshift/oc/pull/1538 for dynamic credentials store(just includes more reasoning in description and the removal of `github.com/docker/docker/registry` because it seems that only usage of this library in oc is this code bits.). 

We have to do this change for oc-mirror(https://issues.redhat.com/browse/OCPBUGS-7465) because oc-mirror requires to refresh token, otherwise it fails. But we exactly don't know which functions are used in oc by oc-mirror and it seems that `DockerStaticCredentialsStore` https://github.com/docker/engine/blob/8955d8da8951695a98eb7e15bead19d402c6eb27/registry/auth.go#L43-L66 is so primitive that oc can entirely switch to `DynamicCredentialsStore` in this PR. 

In any case, if noone calls `SetRefreshToken` function in credentials store, token will not be refreshed and if any one calls `SetRefreshToken` which implies that token is being tried to be refreshed and we can safely refresh it.


ardaguclu:
If it is proved that the `DynamicCredentialsStore` solves the current problem, oc-mirror can pass `NewCredentialStoreFactory(path, true)`

sherine-k:
thanks @ardaguclu @tsmetana 
we'll keep watching and use the new version of the oc module, with `NewCredentialStoreFactory(path, true)` to fix OCPBUGS-7465

ardaguclu:
@sherine-k just for clarification;
Are you going to import this patch without waiting it to be merged and check that problem is resolved?. If this patch does not solve the problem, no point of merging it.

sherine-k:
Hi @ardaguclu 
I'll reference the PR for the test as soon as I have it. We'll ask our QE to test.
Thanks so much

grokspawn:
There are a couple of things about my original PR that I didn't like (but were -- I guess -- fine for a proof of concept). 
1. the `DynamicCredentialStore` was not in dockerregistry like the `StaticCredentialStore`
2. the mutex fails to coordinate the roundtrippers' token refresh: all roundtrippers individually refresh the token (it only happens about every 15 minutes, and there are ~4 of them in the tie, so it's probably OK for now)

I appreciate you running with this, @ardaguclu .

soltysh (approved):
/lgtm

ardaguclu:
I'd like to see the signal from ipv6 job, but it seems that it is permafailing and we have to move forward;
/hold cancel

Issue #1531, opened 2023-08-22
oc debug should default to kubectl.kubernetes.io/default-container value


holzman:
This is pretty minor, but IMHO `oc debug` should work the same as `oc exec` and default to the container name in `kubectl.kubernetes.io/default-container` annotation if set.

PR #1541, opened 2023-09-15
oc debug: default to kubectl.kubernetes.io/default-container

holzman:
If the pod annotation "kubectl.kubernetes.io/default-container" is set, default to it when using oc debug.

### DIFF
--- a/pkg/cli/debug/debug.go
+++ b/pkg/cli/debug/debug.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
+        "k8s.io/kubectl/pkg/cmd/util/podcmd"
 
 	kappsv1 "k8s.io/api/apps/v1"
 	kappsv1beta1 "k8s.io/api/apps/v1beta1"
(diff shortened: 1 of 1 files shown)


holzman:
Fixes #1531

pkg/cli/debug/debug.go
```diff
@@ -455,17 +456,24 @@ func (o *DebugOptions) RunDebug() error {
 	pod.Name, pod.Namespace = fmt.Sprintf("%s-debug-%s", generateapp.MakeSimpleName(infos[0].Name), utilrand.String(5)), ns
 	o.Attach.Pod = pod
 
+
 	if len(o.ContainerName) == 0 && len(pod.Spec.Containers) > 0 {
+		c, err := podcmd.FindOrDefaultContainerByName(pod, o.ContainerName, o.Quiet, o.ErrOut)
```
ardaguclu, line 461: @holzman thanks for spending effort on this PR. 

As stated in https://github.com/openshift/oc/blob/9f903f99a6650da83a0409f81aff3ced0acaad38/pkg/cli/debug/debug.go#L245, `oc debug` particularly defaults to first container and if user wants to use another pod, user can specify it via `container` flag. This change may break some scripts possibly relying on using the first container in pod rather than the one with annotation.

holzman: Yes, but that behavior predates the default container annotations. I'd argue that ignoring the annotation is now a bug rather than a feature.

PR #1547, opened 2023-09-18
WRKLDS-806: Bump kube dependencies to 1.28.2

ardaguclu:
This PR bumps k8s dependencies to 1.28.2 for OCP 4.15.


ardaguclu:
Version failure in ipv6 is due to the origin refers to 1.27.1. 
/hold
until origin is bumped to 1.28.2

ardaguclu:
Since the o/o bump has been merged;
/retest

ardaguclu:
ipv6 failure is fixed by https://github.com/openshift/origin/pull/28273/

ardaguclu:
@atiratree could you PTAL, when you have a chance?. Thanks.

pkg/cli/version/version.go
```diff
@@ -53,18 +53,18 @@ type VersionOptions struct {
 	oClient         configv1client.ClusterVersionsGetter
```
atiratree, line 53: we should update the examples as well

pkg/cli/cli.go
```diff
@@ -143,7 +141,7 @@ func NewDefaultOcCommand(o kubecmd.KubectlOptions) *cobra.Command {
 		if kcmdutil.CmdPluginAsSubcommand.IsEnabled() {
 			// Command exists(e.g. kubectl create), but it is not certain that
 			// subcommand also exists (e.g. kubectl create networkpolicy)
-			if _, ok := allowedCmdsSubcommandPlugin[foundCmd.Name()]; ok {
+			if kubecmd.IsSubcommandPluginAllowed(foundCmd.Name()) {
```
atiratree, line 144: +1 for bringing this in. 

Btw, would you mind looking at the plugin completion?

https://github.com/kubernetes/kubernetes/blob/1020678366f741834f06b9f0087e57e48f93ffba/staging/src/k8s.io/kubectl/pkg/cmd/cmd.go#L329

https://github.com/kubernetes/kubernetes/blob/1020678366f741834f06b9f0087e57e48f93ffba/staging/src/k8s.io/kubectl/pkg/cmd/cmd.go#L472

It seems we are missing it.

pkg/cli/version/version.go
ardaguclu: Done

pkg/cli/cli.go
ardaguclu: Updated code with the shell completion for plugins. But this https://github.com/kubernetes/kubernetes/blob/1020678366f741834f06b9f0087e57e48f93ffba/staging/src/k8s.io/kubectl/pkg/cmd/cmd.go#L472 seems to be added in 1.29 and we can get this in in next bump

atiratree: Ah, I see. Thanks for adding the first part.

PR #1548, opened 2023-09-20
OCPBUGS-19517: remove unsupported commands from microshift docs

dhellmann:
We missed some of the commands that are not supported on MicroShift in
the last review. This change removes commands related to deployment
configs and image streams. It also removes the debug command, which
needs a release payload to determine the toolbox image.


tools/gendocs/gendocs/microshift.go
```diff
@@ -66,11 +67,24 @@ var microshiftCommands = sets.NewString(
 	"oc registry info",
 	"oc registry login",
 	"oc replace",
+	"oc rollback",
+	"oc rollback",
```
ardaguclu, line 71: nit: `oc rollback` appears 2 times.

ardaguclu (commented):
/lgtm

tools/gendocs/gendocs/microshift.go
dhellmann: Since CI failed I went ahead and fixed this one.

PR #1550, opened 2023-09-26
ACM-7713: Skip related objects which are invalid

JustinKuli:
If a custom resource has a `status.relatedObject` list with items that are not `ObjectReference`s, then the `inspect` command would have an error while trying to gather that resource. Now, those invalid items are skipped.

Refs:
 - https://issues.redhat.com/browse/ACM-7713

For example, in open-cluster-management.io, we have a ConfigurationPolicy custom resource which is currently un-gatherable via `oc adm inspect`, because its related objects [are not ObjectReferences](https://github.com/stolostron/config-policy-controller/blob/f94c875ae12c0b158be788f0d6cdae737e59b49b/api/v1/configurationpolicy_types.go#L276).

### DIFF
--- a/pkg/cli/admin/inspect/resource.go
+++ b/pkg/cli/admin/inspect/resource.go
@@ -2,14 +2,14 @@ package inspect
 
 import (
 	"fmt"
-	routev1 "github.com/openshift/api/route/v1"
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"os"
 	"path"
 
 	configv1 "github.com/openshift/api/config/v1"
+	routev1 "github.com/openshift/api/route/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/errors"
@@ -186,11 +186,12 @@ func toStructuredObject[T any, TList any](obj runtime.Object) (runtime.Object, e
 }
 
 func gatherGenericObject(context *resourceContext, info *resource.Info, o *InspectOptions) error {
+	errs := []error{}
 	unstr, ok := info.Object.(*unstructured.Unstructured)
 	if ok {
 		// obtain associated objects for the current resource
 		if err := gatherRelatedObjects(context, unstr, o); err != nil {
-			return err
+			errs = append(errs, err)
 		}
 	}
 
@@ -199,10 +200,15 @@ func gatherGenericObject(context *resourceContext, info *resource.Info, o *Inspe
 	filename := filenameForInfo(info)
 	// ensure destination path exists
 	if err := os.MkdirAll(dirPath, os.ModePerm); err != nil {
-		return err
+		errs = append(errs, err)
+		return errors.NewAggregate(errs)
 	}
 
-	return o.fileWriter.WriteFromResource(path.Join(dirPath, filename), info.Object)
+	if err := o.fileWriter.WriteFromResource(path.Join(dirPath, filename), info.Object); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errors.NewAggregate(errs)
 }
 
 func gatherRelatedObjects(context *resourceContext, unstr *unstructured.Unstructured, o *InspectOptions) error {


JustinKuli:
Opened https://issues.redhat.com/browse/OCPBUGS-20142 for this

pkg/cli/admin/inspect/resource.go
```diff
@@ -275,13 +275,18 @@ func obtainRelatedObjects(obj *unstructured.Unstructured) ([]*configv1.ObjectRef
...
 	relatedObjs := []*configv1.ObjectReference{}
-	for _, relatedObj := range val {
+	for i, relatedObj := range val {
 		ref := &configv1.ObjectReference{}
 		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(relatedObj.(map[string]interface{}), ref); err != nil {
 			return nil, err
 		}
-		relatedObjs = append(relatedObjs, ref)
-		klog.V(1).Infof("    Found related object %q...\n", objectReferenceToString(ref))
+		refStr := objectReferenceToString(ref)
```
ardaguclu, line 283: Thanks for opening this PR. Currently if ref is not objectreference, it will be skipped and must-gather will continue for other resources. In the bug description, it says that `ConfigurationPolicy` should be collected but I'm not sure that this change will make `ConfigurationPolicy` is gathered.

JustinKuli: It looks like the `gatherMoreObjects` function would continue after trying with the empty `ObjectReference` in this situation, saving the error to report later. But, [in the gatherGenericObject function](https://github.com/openshift/oc/blob/d2f8bfe9c582228f7cffd03834f6872dffdde623/pkg/cli/admin/inspect/resource.go#L193), it returns that error before writing the data for the root object (the `ConfigurationPolicy`, in my case).

The change I proposed removes the error. It might be better to not return immediately in `gatherGenericObject`, but I am not sure what other kinds of errors might appear there, which *should* cause it return immediately.

I understand testing with other people's CRDs is hard, so I have run `oc adm inspect configurationpolicies.policy.open-cluster-management.io` using a build with this change, and it did successfully gather the `ConfigurationPolicy`, if that makes you feel more confident.

ardaguclu: Thanks, I'm leaning towards what you said;

> The change I proposed removes the error. It might be better to not return immediately in gatherGenericObject, but I am not sure what other kinds of errors might appear there, which should cause it return immediately.

We only want to write the yaml of this object, when it is relatedobjects are totally not related. Maybe as you said, we can write the yaml and after that return error. WDYT?

JustinKuli: Yeah. I thought writing the YAML and then handling multiple possible errors would be annoying. But it doesn't seem to bad actually. Maybe something like this?
```diff
diff --git a/pkg/cli/admin/inspect/resource.go b/pkg/cli/admin/inspect/resource.go
index 3e0f061d8..a9f89097a 100644
--- a/pkg/cli/admin/inspect/resource.go
+++ b/pkg/cli/admin/inspect/resource.go
@@ -1,6 +1,7 @@
 package inspect

 import (
+	stderrors "errors"
 	"fmt"
 	"os"
 	"path"
@@ -186,12 +187,12 @@ func toStructuredObject[T any, TList any](obj runtime.Object) (runtime.Object, e
 }

 func gatherGenericObject(context *resourceContext, info *resource.Info, o *InspectOptions) error {
+	var gatherErr error
+
 	unstr, ok := info.Object.(*unstructured.Unstructured)
 	if ok {
 		// obtain associated objects for the current resource
-		if err := gatherRelatedObjects(context, unstr, o); err != nil {
-			return err
-		}
+		gatherErr = gatherRelatedObjects(context, unstr, o)
 	}

 	// save the current object to disk
@@ -199,10 +200,12 @@ func gatherGenericObject(context *resourceContext, info *resource.Info, o *Inspe
 	filename := filenameForInfo(info)
 	// ensure destination path exists
 	if err := os.MkdirAll(dirPath, os.ModePerm); err != nil {
-		return err
+		return stderrors.Join(err, gatherErr)
 	}

-	return o.fileWriter.WriteFromResource(path.Join(dirPath, filename), info.Object)
+	writeErr := o.fileWriter.WriteFromResource(path.Join(dirPath, filename), info.Object)
+
+	return stderrors.Join(writeErr, gatherErr)
 }

 func gatherRelatedObjects(context *resourceContext, unstr *unstructured.Unstructured, o *InspectOptions) error {
```

ardaguclu: Sounds reasonable. Could you please use error aggregator like in https://github.com/openshift/oc/blob/fe0d9d929496b6dfe8c4fb931ca7954f122b179c/pkg/cli/admin/inspect/resource.go#L65 instead of `stderrors`?

JustinKuli:
I set up my cluster with some CRs to demonstrate this issue, and compare with my proposed changes.

With a usual build of `oc`:
```
❯ oc adm inspect configurationpolicies.policy.open-cluster-management.io --all-namespaces --v=1
I1027 11:21:07.358461   20162 resource.go:167] Gathering related object reference information for "ConfigurationPolicy/policy-namespace-jkuli"...
I1027 11:21:07.358529   20162 resource.go:182]     Found related object ""...
I1027 11:21:07.358539   20162 resource.go:182]     Found related object ""...
I1027 11:21:19.095819   20162 resource.go:167] Gathering related object reference information for "ConfigurationPolicy/policy-pod-jkuli"...
I1027 11:21:19.095850   20162 resource.go:182]     Found related object ""...
Wrote inspect data to inspect.local.41149464613221541.
error: errors ocurred while gathering data:
    skipping gathering  due to error: the server doesn't have a resource type ""

❯ echo $?
1

❯ tree inspect.local.41149464613221541 -h
[ 128]  inspect.local.41149464613221541
├── [3.3K]  event-filter.html
└── [ 105]  timestamp

1 directory, 2 files
```

The command exits with an error, but the logs are not particularly helpful for understanding what happened, and the ConfigurationPolicy yaml is not written to a file.

With a build with my proposed changes:
```
❯ ./oc adm inspect configurationpolicies.policy.open-cluster-management.io --all-namespaces --v=1
I1027 11:26:53.570401   20593 resource.go:275] Gathering related object reference information for "ConfigurationPolicy/policy-namespace-jkuli"...
I1027 11:26:53.572214   20593 resource.go:275] Gathering related object reference information for "ConfigurationPolicy/policy-pod-jkuli"...
Wrote inspect data to inspect.local.1274326490368133161.
error: errors occurred while gathering data:
    [skipping .status.relatedObjects[0] in "ConfigurationPolicy/policy-namespace-jkuli" because the ObjectReference is invalid, skipping .status.relatedObjects[1] in "ConfigurationPolicy/policy-namespace-jkuli" because the ObjectReference is invalid, skipping .status.relatedObjects[0] in "ConfigurationPolicy/policy-pod-jkuli" because the ObjectReference is invalid]

❯ echo $?
1

❯ tree inspect.local.1274326490368133161 -h
[ 160]  inspect.local.1274326490368133161
├── [3.3K]  event-filter.html
├── [  96]  namespaces
│   └── [  96]  local-cluster
│       └── [  96]  policy.open-cluster-management.io
│           └── [ 128]  configurationpolicies
│               ├── [4.0K]  policy-namespace-jkuli.yaml
│               └── [3.7K]  policy-pod-jkuli.yaml
└── [ 104]  timestamp

5 directories, 4 files
```

The command exits with an error (unchanged), the logs are slightly different, and the ConfigurationPolicies are written to files.

pkg/cli/admin/inspect/resource.go
```diff
@@ -274,15 +280,23 @@ func obtainRelatedObjects(obj *unstructured.Unstructured) ([]*configv1.ObjectRef
...
-		relatedObjs = append(relatedObjs, ref)
-		klog.V(1).Infof("    Found related object %q...\n", objectReferenceToString(ref))
+
+		refStr := objectReferenceToString(ref)
+		if refStr != "" {
+			klog.V(1).Infof("    Found related object %q...\n", refStr)
+			relatedObjs = append(relatedObjs, ref)
+		} else {
+			errs = append(errs, fmt.Errorf("skipping .status.relatedObjects[%v] in %q because the ObjectReference is invalid",
+				i, unstructuredToString(obj)))
```
JustinKuli, line 297: If an error is not generated here, and the invalid related object is just skipped instead, then the command exits cleanly. Personally, I think that would be a bad change in behavior.

```
❯ ./oc adm inspect configurationpolicies.policy.open-cluster-management.io --all-namespaces --v=1
I1027 11:31:04.951457   20922 resource.go:275] Gathering related object reference information for "ConfigurationPolicy/policy-namespace-jkuli"...
I1027 11:31:04.953215   20922 resource.go:275] Gathering related object reference information for "ConfigurationPolicy/policy-pod-jkuli"...
Wrote inspect data to inspect.local.8438621565858530944.

❯ echo $?
0

❯ tree inspect.local.8438621565858530944 -h
[ 160]  inspect.local.8438621565858530944
├── [3.3K]  event-filter.html
├── [  96]  namespaces
│   └── [  96]  local-cluster
│       └── [  96]  policy.open-cluster-management.io
│           └── [ 128]  configurationpolicies
│               ├── [4.0K]  policy-namespace-jkuli.yaml
│               └── [3.7K]  policy-pod-jkuli.yaml
└── [ 104]  timestamp

5 directories, 4 files
```

pkg/cli/admin/inspect/resource.go
```diff
@@ -274,15 +280,23 @@ func obtainRelatedObjects(obj *unstructured.Unstructured) ([]*configv1.ObjectRef
 		return nil, nil
 	}
 
+	errs := []error{}
```
ardaguclu, line 283: I don't think we would need any changes in this function.

ardaguclu: Our only goal is to write the yaml of object in cases of error is returned. The rest piece of code should stay as is. Am I missing something?

ardaguclu: Could you please point me the line of where `gatherRelatedObjects` function return error?

JustinKuli: Here: https://github.com/openshift/oc/blob/3544c64b893926687ae799475168d6389022c121/pkg/cli/admin/inspect/resource.go#L226

`gatherMoreObjects` has an error because the ObjectReference is empty, because the CR's status.relatedObjects are not real ObjectReferences

ardaguclu: I think, we can leave this as is. Same error (although it is not clear) returns with/without this patch.

ardaguclu:
Thanks @JustinKuli 
/lgtm

PR #1554, opened 2023-10-03
pkg/cli/admin/update: Add tech-preview 'oc adm upgrade status'

wking:
Initial scaffolding, based on the existing code for status rendering in `oc adm upgrade`.  I'm gating the new subcommand behind an `OC_ENABLE_CMD_UPGRADE_STATUS` feature gate to avoid surprising users with a volatile command while the implementation settles down.

```console
$ ./oc adm upgrade  status
error: unknown command "status"
See 'oc adm upgrade -h' for help and examples
$ OC_ENABLE_CMD_UPGRADE_STATUS=true ./oc adm upgrade --help
...
Available Commands:
  channel       Set or clear the update channel
  status        Display the status of current cluster version updates.
...
$ OC_ENABLE_CMD_UPGRADE_STATUS=true ./oc adm upgrade status

  Reason:
  Message: Cluster version is 4.13.10
```

### DIFF
--- a/pkg/cli/admin/upgrade/status/status.go
+++ b/pkg/cli/admin/upgrade/status/status.go
@@ -1,0 +1,107 @@
+// Package status displays the status of current cluster version updates.
+package status
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


wking:
Rerolled to be `oc adm upgrade status` instead of `oc adm update status` with 90a8eaba9 -> 701c2eee9, because it seems like that's part of what folks want.  It's easy to move back and forth if we change our mind later, at least as long as it's a tech-preview subcommand.

LalatenduMohanty (approved):
/lgtm

PR #1555, opened 2023-10-03
OCPBUGS-20063: regeneratemco: explicitly check for PlatformStatus field

yuqi-zhang:
It turns out that cfg.Status.PlatformStatus.$PLATFORM can be nil, so add this check to not panic.

If the IP field doesn't exist we will go add the api-int url so this should otherwise not be a problem.


yuqi-zhang:
Test failure should be unrelated.

@ardaguclu would you be able to lgtm? Thanks!

PR #1557, opened 2023-10-04
OCPBUGS-9340: pkg/cli/admin/upgrade: Warn on unrecognized subcommands

wking:
Avoid confusion like we had with `oc adm upgrade channel`, where folks running older `oc` (where `channel` didn't exist) had the command no-op and exit 0.  With this change, that sort of thing will give:

```console
$ oc adm upgrade does-not-exist
error: unknown command "does-not-exist"
See 'oc adm upgrade -h' for help and examples
```

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -135,6 +135,8 @@ type Options struct {
 }
 
 func (o *Options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
For future reference https://pkg.go.dev/k8s.io/kubectl/pkg/cmd/util#RequireNoArguments

evakhoni:
pre-merge verification passed
/label qe-approved

ardaguclu:
This https://github.com/openshift/oc/pull/1559 should fix unit test failure.

PR #1559, opened 2023-10-06
OCPBUGS-20181: Use quay redis image instead docker mysql

ardaguclu:
docker mysql image frequently fails and causes unit test job's failure. This PR uses quay redis image instead docker image for the sake of unit test job's stability.


ardaguclu:
@atiratree could you please tag this?. I think, there was a hiccup due to an issue in docker and in any case, it would be good to use quay images.

PR #1572, opened 2023-10-12
OCPBUGS-20474: Set ImportPolicy to PreserveOriginal to honor --keep-manifest-list when mirroring a payload to an image stream

aleskandro:
When mirroring a multiarch payload via `oc adm release mirror --from=... --to-image-stream=release --keep-manifest-list=true`, the ImageStreamTags are created with the default ImportPolicy.ImportMode (Legacy). Therefore, the release is not mirrored as multiarch/manifest-list, but will be mirrored as a single-arch manifest depending on the architecture of the cluster's node running the internal registry.

### DIFF
--- a/pkg/cli/admin/release/mirror.go
+++ b/pkg/cli/admin/release/mirror.go
@@ -681,6 +681,10 @@ func (o *MirrorOptions) Run(ctx context.Context) error {
 		hasErrors := make(map[string]error)
 		maxPerIteration := 12
 
+		importMode := imagev1.ImportModeLegacy
+		if o.KeepManifestList {
+			importMode = imagev1.ImportModePreserveOriginal
+		}
 		for retries := 4; (len(remaining) > 0 || len(hasErrors) > 0) && retries > 0; {
 			if len(remaining) == 0 {
 				for _, mapping := range mappings {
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/release/mirror.go
```diff
@@ -681,6 +681,10 @@ func (o *MirrorOptions) Run(ctx context.Context) error {
 		hasErrors := make(map[string]error)
 		maxPerIteration := 12
 
+		importMode := imagev1.ImportModeLegacy
+		if o.KeepManifestList {
+			importMode = imagev1.ImportModePreserveOriginal
```
ardaguclu, line 686: Sounds reasonable and `istag` seems more aligned with what is desired. What will be the impact on the mirrored payload?

aleskandro: Hi, as in the reported logs in the jira ticket:

### without this change

ManifestList payload: it's 'unpacked', and we import the manifests for the same architecture as control plane of the cluster.
Manifest payload: no impact

### With this change

ManifestList payload: imported as ManifestList
Manifest payload: no impact

aleskandro: However, PreserveOriginal can eventually become the default... I could change the object only when keep-manifest-list is set to make this change future-proof... wdyt?

ardaguclu: In my opinion, this way is more readable, instead of a hidden default. In any case, there will be a PR, when the default is changed and we can make change here too.

zhouying7780:
Done pre-merge test , the fix works well

PR #1580, opened 2023-10-16
adm release new with --mirror --from-image-stream should not require a public route for the internal registry

aleskandro:
Commit 57ce6b4 introduced the `--reference-mode=source` option in the `oc adm release new` command to preserve image location while creating a new release image. The author removed a TODO comment regarding support for internal and referential images, although the commit was focusing primarily on referential mode.

Today, if a user sets `--reference-mode` to a value other than `source`, we require the release ImageStream to have an external-facing repository based on the `.Status.PublicDockerImageRepository` field not being empty. The command fails with:

> only image streams or releases with public image repositories can be the source for releases when using the default --reference-mode

However, users might run this command in a pod, capable of reaching the internal registry url without the need to expose it publicly. This change introduces logic to check for the presence of the `--mirror` option. If `--mirror` is used, and the `oc adm release new` command is executed within the cluster itself, there is no longer a need to expose an external route for the internal registry. This adjustment simplifies the process when external access to the internal registry is unnecessary.

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -559,7 +559,7 @@ func (o *NewOptions) Run(ctx context.Context) error {
 			inputIS.Annotations = make(map[string]string)
 		}
 		inputIS.Annotations[annotationBuildVersions] = extraComponentVersions.String()
-		if err := resolveImageStreamTagsToReferenceMode(inputIS, is, o.ReferenceMode, exclude); err != nil {
+		if err := resolveImageStreamTagsToReferenceMode(inputIS, is, o.ReferenceMode, exclude, len(o.Mirror) > 0); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


aleskandro:
cc @LorbusChris

pkg/cli/admin/release/new.go
```diff
@@ -807,11 +807,15 @@ func resolveImageStreamTagsToReferenceMode(inputIS, is *imageapi.ImageStream, re
 
 		externalFn := func(source, image string) string {
 			// filter source URLs
-			if len(source) > 0 && len(internal) > 0 && strings.HasPrefix(source, internal) {
+			if !toMirror && len(source) > 0 && len(internal) > 0 && strings.HasPrefix(source, internal) {
 				klog.V(2).Infof("Can't use source %s because it points to the internal registry", source)
 				source = ""
 			}
-			// default to the external registry name
+			// if --mirror is set and the referenceMode is not 'source', use the internal registry as image source.
```
ardaguclu, line 814: `mirror` field can only be set for the referenceMode public (see: https://github.com/openshift/oc/blob/1f30e7346ee765cb27d4ac451720e8fd30a01be8/pkg/cli/admin/release/new.go#L284) and public referenceMode is self-descriptive that can force using external. Besides, I'm not sure I fully measure the consequences of this change.  

Yes, it makes sense from the pod as explained but it still (incorrectly) may fall back to internal registry even in the connected environment instead surfacing to external registry.

PR #1582, opened 2023-10-17
create/identity: fail gracefully when no user API exists

stevekuznetsov:


pkg/cli/create/identity.go
```diff
@@ -78,6 +80,11 @@ func (o *CreateIdentityOptions) Complete(cmd *cobra.Command, f genericclioptions
 		return err
 	}
 
+	o.DiscoveryClient, err = f.ToDiscoveryClient()
```
stevekuznetsov, line 83: This is using the discovery cache.

pkg/helpers/cmd/errors.go
```diff
@@ -30,3 +32,67 @@ func CheckPodSecurityErr(err error) {
...
+		kcmdutil.CheckErr(originalError)
+		return
+	}
+
+	// in the case of a NotFound error, either the GR is not found or the object iself
+	// is not found (when the action was a mutation of existing content); make sure it's
+	// a missing GR by cross-checking discovery
+	_, resourceLists, err := discoveryClient.ServerGroupsAndResources()
+	if err != nil {
+		kcmdutil.CheckErr(utilerrors.NewAggregate([]error{originalError, err}))
```
deads2k, line 76: no, don't add a non-sensical error for the user.  Just the original

pkg/helpers/cmd/errors.go
```diff
@@ -30,3 +32,67 @@ func CheckPodSecurityErr(err error) {
...
+	if details == nil {
+		// we're not going to be able to detect whether the GR is registered
+		kcmdutil.CheckErr(originalError)
+		return
+	}
+
+	// in the case of a NotFound error, either the GR is not found or the object iself
+	// is not found (when the action was a mutation of existing content); make sure it's
+	// a missing GR by cross-checking discovery
+	_, resourceLists, err := discoveryClient.ServerGroupsAndResources()
```
deads2k, line 74: I think we can have errors and still have some data, so we probably want to search the data and ignore the error.

pkg/cli/create/identity.go
```diff
@@ -58,7 +60,7 @@ func NewCmdCreateIdentity(f genericclioptions.RESTClientGetter, streams generici
 		Example: identityExample,
 		Run: func(cmd *cobra.Command, args []string) {
 			cmdutil.CheckErr(o.Complete(cmd, f, args))
-			cmdutil.CheckErr(o.Run())
+			ocmdhelpers.CheckGroupResourceNotRegisteredErr(o.Run(), o.DiscoveryClient, "Identity creation", ocmdhelpers.CustomOIDCReason)
```
deads2k, line 63: make the helper specifically for `"Identity creation", ocmdhelpers.CustomOIDCReason`.  We can create another custom one in a few years when we need it.

pkg/helpers/cmd/errors.go
stevekuznetsov: Ah, because it's aggregated? Gotcha. Let me do that.

pkg/helpers/cmd/errors.go
```diff
@@ -1,11 +1,13 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
stevekuznetsov: GoLand lied to me! I'll update.

pkg/helpers/cmd/errors.go
```diff
@@ -30,3 +32,67 @@ func CheckPodSecurityErr(err error) {
...
+		return
+	}
+	details := status.Status().Details
+	if details == nil {
+		// we're not going to be able to detect whether the GR is registered
+		kcmdutil.CheckErr(originalError)
+		return
+	}
+
+	// in the case of a NotFound error, either the GR is not found or the object iself
```
deads2k, line 71: only need to check for the groups served by the oauth-apiserver.

pkg/cli/create/identity.go
stevekuznetsov: Had this for `oc adm` user-related commands, `oc create useridentitymapping`, etc - which I would add once we liked the shape.

pkg/helpers/cmd/errors.go
```diff
@@ -30,3 +35,80 @@ func CheckPodSecurityErr(err error) {
...
+
+	// only NotFound errors can stem from missing GRs, so short-circuit if we're
+	// dealing with any other error
+	if !apierrors.IsNotFound(originalError) {
+		kcmdutil.CheckErr(originalError)
+		return
+	}
+
+	status, ok := originalError.(apierrors.APIStatus)
+	if !ok && !errors.As(originalError, &status) {
```
ardaguclu, line 68: Shouldn't this be
```suggestion
	if !ok || !errors.As(originalError, &status) {
```
instead?

ardaguclu: I wonder this case is already covered in `IsNotFound` above?

pkg/helpers/cmd/errors.go
```diff
@@ -30,3 +35,80 @@ func CheckPodSecurityErr(err error) {
...
+			// n.b. metav1.Status does not specify a version, so the best we can do is detect that
+			// *some* related GroupResource (at some version) exists on the server
+			if resource.Name == details.Kind && resource.Group == details.Group {
+				// the GVK does exist on the server, so the NotFound must have been for a missing resource
+				kcmdutil.CheckErr(originalError)
+				return
+			}
+		}
+	}
+	if err != nil {
```
ardaguclu, line 104: What we lose if we ignore this error check?. Because it may change the error message from user not found to more non graceful.

pkg/cli/create/identity.go
```diff
@@ -58,7 +60,7 @@ func NewCmdCreateIdentity(f genericclioptions.RESTClientGetter, streams generici
 		Example: identityExample,
 		Run: func(cmd *cobra.Command, args []string) {
 			cmdutil.CheckErr(o.Complete(cmd, f, args))
-			cmdutil.CheckErr(o.Run())
+			ocmdhelpers.CustomOIDCDerivedErrorChecker(o.DiscoveryClient, "Identity creation")(o.Run())
```
ardaguclu, line 63: I think, `Identity creation` is passed to point out which command fails. But it is very likely that user knows which command is executed and maybe we can remove this parameter(from also error message). WDYT?

pkg/helpers/cmd/errors.go
```diff
@@ -30,3 +35,80 @@ func CheckPodSecurityErr(err error) {
...
+	}
+}
+
+const (
+	CustomOIDCReason = "This may occur if an external OIDC provider is configured."
+)
+
+// CheckGroupResourceNotRegisteredErr checks to see if originalError stems from the fact that
+// the GR in question is not registered on the server.
+func CheckGroupResourceNotRegisteredErr(originalError error, discoveryClient discovery.DiscoveryInterface, what, why string, groups ...string) {
```
ardaguclu, line 51: It is worth to check that discoveryClient is initialized by the command

pkg/helpers/cmd/errors.go
stevekuznetsov: No, the logic as written is correct. If it's not cast-able *and* `errors.As` does not work, then we short-circuit and return.

pkg/helpers/cmd/errors.go
stevekuznetsov: If this error occurred, we can't be sure that not finding the GR in discovery really means the resource is missing, so we fall back to `kcmdutil.CheckErr(originalError)` - which is the original error, not the one from discovery, so nothing is changed, I think?

ardaguclu: You are right, I missed `kcmdutil.CheckErr(originalError)`

ardaguclu:
Thanks @stevekuznetsov looks good to me. I'm deferring to @deads2k in case he may want to look at.

pkg/helpers/cmd/errors.go
```diff
@@ -30,3 +35,75 @@ func CheckPodSecurityErr(err error) {
...
+		return
+	}
+
+	status, ok := originalError.(apierrors.APIStatus)
+	if !ok && !errors.As(originalError, &status) {
+		// this should never happen, since `apierrors.IsNotFound()` succeeded above
+		kcmdutil.CheckErr(originalError)
+		return
+	}
+	details := status.Status().Details
```
deads2k, line 68: if ok==false, why is this line accurate?

stevekuznetsov: Either `ok` is true, or `errors.As` returns true, in both cases `status` is a non-nil `apierrors.APIStatus`. If both fail, we exit on L66.

deads2k:
looks like it does what's desired to me.  Will give @ardaguclu a day or so.

PR #1585, opened 2023-10-19
oc/*: gracefully fail when user group is not configured

stevekuznetsov:
All commands using the `userv1` clients should be wrapped now.

```
$ for file in $( git grep -l 'clientset/versioned/typed/user/v1' | grep -v vendor ); do if ! grep -q 'CheckOAuthDisabledErr' $file; then echo $file; fi; done
pkg/cli/admin/groups/sync/interfaces.go    # just interfaces, no impl
pkg/cli/admin/prune/auth/group_test.go     # test
pkg/cli/admin/prune/auth/user.go           # helper func, caller handles it
pkg/cli/admin/prune/auth/user_test.go      # test
pkg/helpers/describe/describer.go          # TODO: need to figure this out
pkg/helpers/groupsync/grouplister.go       # helper func, caller handles it
pkg/helpers/groupsync/grouplister_test.go  # test
pkg/helpers/groupsync/grouppruner.go       # helper func, caller handles it
pkg/helpers/groupsync/grouppruner_test.go  # test
pkg/helpers/groupsync/groupsyncer.go       # helper func, caller handles it
pkg/helpers/groupsync/groupsyncer_test.go  # test
pkg/helpers/project/whoami.go              # helper func, caller handles it
```

Follow-ups:

 - [ ] determine how to handle the describer functions
 - [ ] also handle the oauth group

### DIFF
--- a/pkg/cli/admin/groups/sync/prune.go
+++ b/pkg/cli/admin/groups/sync/prune.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/go-ldap/ldap/v3"
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/discovery"
 
 	kerrs "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -16,6 +17,7 @@ import (
 	legacyconfigv1 "github.com/openshift/api/legacyconfig/v1"
 	userv1typedclient "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
 	"github.com/openshift/library-go/pkg/security/ldapclient"
+	ocmdhelpers "github.com/openshift/oc/pkg/helpers/cmd"
 	syncgroups "github.com/openshift/oc/pkg/helpers/groupsync"
 	ldapsync "github.com/openshift/oc/pkg/helpers/groupsync/ldap"
 )
@@ -65,7 +67,8 @@ type PruneOptions struct {
 	Confirm bool
 
 	// GroupClient is the interface used to interact with OpenShift Group objects
-	GroupClient userv1typedclient.GroupsGetter
+	GroupClient     userv1typedclient.GroupsGetter
+	DiscoveryClient discovery.DiscoveryInterface
 
 	genericiooptions.IOStreams
 }
@@ -87,7 +90,7 @@ func NewCmdPruneGroups(name, fullName string, f kcmdutil.Factory, streams generi
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Complete(f, cmd, args))
 			kcmdutil.CheckErr(o.Validate())
-			kcmdutil.CheckErr(o.Run())
+			ocmdhelpers.CheckOAuthDisabledErr(o.Run(), o.DiscoveryClient)
 		},
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/cli/admin/groups/sync/prune.go
```diff
@@ -5,7 +5,9 @@ import (
 	"fmt"
 
 	"github.com/go-ldap/ldap/v3"
+	ocmdhelpers "github.com/openshift/oc/pkg/helpers/cmd"
```
ardaguclu, line 8: nit: should be in oc/pkg group with the end of imports

pkg/cli/admin/verifyimagesignature/verify-signature.go
```diff
@@ -14,7 +14,9 @@ import (
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/signature"
 	sigtypes "github.com/containers/image/v5/types"
+	ocmdhelpers "github.com/openshift/oc/pkg/helpers/cmd"
```
ardaguclu, line 17: nit: same

pkg/cli/whoami/whoami.go
```diff
@@ -68,7 +71,7 @@ func NewCmdWhoAmI(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Complete(f))
 			kcmdutil.CheckErr(o.Validate())
-			kcmdutil.CheckErr(o.Run())
+			ocmdhelpers.CheckOAuthDisabledErr(o.Run(), o.DiscoveryClient)
```
ardaguclu, line 74: no need to add this check in here because we will support both cases(catch notfound error internally and use different endpoint)

pkg/cli/admin/verifyimagesignature/verify-signature.go
```diff
@@ -106,7 +109,7 @@ func NewCmdVerifyImageSignature(f kcmdutil.Factory, streams genericiooptions.IOS
 		Example: verifyImageSignatureExample,
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Validate())
-			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			ocmdhelpers.CheckOAuthDisabledErr(o.Complete(f, cmd, args), o.DiscoveryClient)
```
ardaguclu, line 112: It is interesting to see this command but after checking, indeed it relies on user object.

pkg/cli/whoami/whoami.go
ardaguclu: whoami should work with external OIDC

pkg/cli/admin/groups/sync/prune.go
stevekuznetsov: Do you have an import formatter enabled on the repo? I think you (and anyone making PRs) would appreciate your preferences being automated so these kinds of comments don't need to happen. OpenShift has an import formatter that I've seen other repos use.

stevekuznetsov: https://github.com/openshift-eng/openshift-goimports

ardaguclu:
Thanks @stevekuznetsov 

/lgtm

PR #1588, opened 2023-10-23
WRKLDS-874: oc whoami: Ask from selfsubjectreview and fallback to internal oauth

ardaguclu:
Currently, `oc whoami` assumes that internal oauth is enabled and that's
why, it fails when an error is gotten from server. However, cluster might
be configured to rely on external OIDC issuer and internal oauth server
might be disabled. Even in that case, `oc whoami` should return the
correct data.

Same rules also apply for project whoami. It can use
selfsubjectreview endpoint as well.

This PR prioritizes asking to selfsubjectreview endpoint to
get this information and if selfsubjectreview fails, it falls back to
oauth server's user object.

### DIFF
--- a/pkg/helpers/project/whoami.go
+++ b/pkg/helpers/project/whoami.go
@@ -6,8 +6,12 @@ import (
 	"encoding/base64"
 	"fmt"
 
+	"k8s.io/klog/v2"
+
+	v1 "k8s.io/api/authentication/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	authenticationv1client "k8s.io/client-go/kubernetes/typed/authentication/v1"
 	restclient "k8s.io/client-go/rest"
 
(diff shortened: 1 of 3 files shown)


ardaguclu:
/hold
until it is verified.

ardaguclu:
/hold
I'll change the API call order between old and new

pkg/helpers/project/whoami.go
```diff
@@ -6,15 +6,35 @@ import (
...
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	authenticationv1client "k8s.io/client-go/kubernetes/typed/authentication/v1"
 	restclient "k8s.io/client-go/rest"
 
 	userv1 "github.com/openshift/api/user/v1"
 	userv1typedclient "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
 )
 
 func WhoAmI(clientConfig *restclient.Config) (*userv1.User, error) {
```
deads2k, line 21: huh, didn't know we had this.  Good catch.

PR #1584, opened 2023-10-19
oc login external OIDC provider integration

ardaguclu:
SSIA

PR #1590, opened 2023-10-25
WRKLDS-875: oc login: Add oidc and azure external issuers support

ardaguclu:
This PR supersedes https://github.com/openshift/oc/pull/1584 because it doesn't vendor any plugin and works for azure and oidc authentication via credentials exec plugins.

This PR also adds token clean up in cache dir for oc logout.

### DIFF
--- a/pkg/cli/login/login.go
+++ b/pkg/cli/login/login.go
@@ -17,7 +17,9 @@ import (
 	"k8s.io/kubectl/pkg/util/term"
 
 	"github.com/openshift/library-go/pkg/oauth/tokenrequest"
+
 	"github.com/openshift/oc/pkg/helpers/flagtypes"
+	"github.com/openshift/oc/pkg/helpers/oidc"
 )
 
 var (
@@ -47,6 +49,12 @@ var (
 
 		# Log in to the given server through a browser
 		oc login localhost:8443 --web --callback-port 8280
+
+		# Log in to the given server uses an external OIDC issuer for authentication
+		oc login localhost:8443 --external-auth-type=oidc --external-client-id=client-id --external-extra-args="--grant-type=authcode"
+
+		# Log in to the given server uses Azure as an external issuer for authentication
+		oc login localhost:8443 --external-auth-type=azure --external-client-id=client-id --external-extra-args="--tenant-id=user-tenant-id"
 	`)
 )
 
@@ -89,6 +97,11 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.
 
 	cmds.Flags().BoolVarP(&o.WebLogin, "web", "w", o.WebLogin, "Login with web browser. Starts a local HTTP callback server to perform the OAuth2 Authorization Code Grant flow. Use with caution on multi-user systems, as the server's port will be open to all users.")
 	cmds.Flags().Int32VarP(&o.CallbackPort, "callback-port", "c", o.CallbackPort, "Port for the callback server when using --web. Defaults to a random open port")
+
+	cmds.Flags().StringVar(&o.OIDCAuthType, "external-auth-type", o.OIDCAuthType, "Experimental: Specify the authentication type for external issuers to choose which plugin will be used for authentication. Valid values are oidc and azure.")
+	cmds.Flags().StringArrayVar(&o.OIDCExtraArgs, "external-extra-args", o.OIDCExtraArgs, "Experimental: Set extra arguments or overwrite the default ones for external OIDC plugins.")
+	cmds.Flags().StringVar(&o.OIDCClientID, "external-client-id", o.OIDCClientID, "Experimental: OIDC client id to be used for authentication. Required if external OIDC is used.")
+
 	return cmds
 }
 
@@ -143,6 +156,10 @@ func (o *LoginOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []s
 	o.InsecureTLS = kcmdutil.GetFlagBool(cmd, "insecure-skip-tls-verify")
 	o.Token = kcmdutil.GetFlagString(cmd, "token")
 
+	o.OIDCAuthType = kcmdutil.GetFlagString(cmd, "external-auth-type")
+	o.OIDCClientID = kcmdutil.GetFlagString(cmd, "external-client-id")
+	o.OIDCExtraArgs = kcmdutil.GetFlagStringArray(cmd, "external-extra-args")
+
 	o.DefaultNamespace, _, _ = f.ToRawKubeConfigLoader().Namespace()
 
 	o.PathOptions = kclientcmd.NewDefaultPathOptions()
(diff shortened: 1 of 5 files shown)


pkg/cli/login/login.go
```diff
@@ -89,6 +89,12 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.
 
 	cmds.Flags().BoolVarP(&o.WebLogin, "web", "w", o.WebLogin, "Login with web browser. Starts a local HTTP callback server to perform the OAuth2 Authorization Code Grant flow. Use with caution on multi-user systems, as the server's port will be open to all users.")
 	cmds.Flags().Int32VarP(&o.CallbackPort, "callback-port", "c", o.CallbackPort, "Port for the callback server when using --web. Defaults to a random open port")
+
+	cmds.Flags().StringVar(&o.OIDCAuthType, "auth-type", o.OIDCAuthType, "Auth type for oidc login to choose which pluging will be used for authentication. Valid values are oidc and azure.")
```
ardaguclu, line 93:
```suggestion
	cmds.Flags().StringVar(&o.OIDCAuthType, "auth-type", o.OIDCAuthType, "Auth type for oidc login to choose which plugin will be used for authentication. Valid values are oidc and azure.")
```

pkg/cli/login/loginoptions.go
```diff
@@ -216,6 +227,78 @@ func (o *LoginOptions) gatherAuthInfo() error {
...
+				InteractiveMode: kclientcmdapi.IfAvailableExecInteractiveMode,
+			}
+		}
+		me, err := project.WhoAmI(clientConfig)
+		if err != nil {
+			if kerrors.IsUnauthorized(err) {
+				return fmt.Errorf("The token provided is invalid or expired.\n\n")
+			} else if strings.Contains(err.Error(), "It looks like you are trying to use a client-go credential plugin that is not installed.") {
+				if o.OIDCAuthType == "oidc" {
+					return fmt.Errorf("please be sure that the oidc-login plugin is installed. You can install it via oc krew install oidc-login command")
```
ardaguclu, line 288: We can append this message to err which also contains valuable information

pkg/cli/login/login.go
```diff
@@ -177,6 +188,24 @@ func (o LoginOptions) Validate(cmd *cobra.Command, serverFlag string, args []str
 		return errors.New("--web cannot be used along with --username, --password or --token")
 	}
 
+	if len(o.OIDCClientID) > 0 {
+		if len(o.OIDCAuthType) == 0 {
+			return errors.New("auth type should be set and accepted values are oidc and azure")
```
ardaguclu, line 193: We are reserving default value for openshift in the future.

ardaguclu:
Unrelated errors;
/retest

pkg/cli/login/login.go
```diff
@@ -89,6 +89,12 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.
 
 	cmds.Flags().BoolVarP(&o.WebLogin, "web", "w", o.WebLogin, "Login with web browser. Starts a local HTTP callback server to perform the OAuth2 Authorization Code Grant flow. Use with caution on multi-user systems, as the server's port will be open to all users.")
 	cmds.Flags().Int32VarP(&o.CallbackPort, "callback-port", "c", o.CallbackPort, "Port for the callback server when using --web. Defaults to a random open port")
+
+	cmds.Flags().StringVar(&o.OIDCAuthType, "external-auth-type", o.OIDCAuthType, "Specify the authentication type for external issuers to choose which plugin will be used for authentication. Valid values are oidc and azure.")
```
ardaguclu, line 93: It would be a good idea to mark these flags as experimental for 4.15 and even in 4.16

pkg/cli/login/loginoptions.go
```diff
@@ -512,3 +618,39 @@ func (o *LoginOptions) serverProvided() bool {
 func (o *LoginOptions) tokenProvided() bool {
 	return len(o.Token) > 0
 }
+
+func (o *LoginOptions) getIssuerUrlForOIDC() (*oauthdiscovery.OauthAuthorizationServerMetadata, error) {
```
stlaz, line 622: https://github.com/coreos/go-oidc/blob/v3/oidc/oidc.go#L202

pkg/cli/logout/logout.go
```diff
@@ -110,13 +113,18 @@ func (o LogoutOptions) Validate(args []string) error {
...
+		if o.Config.ExecProvider == nil {
+			return errors.New("You must have a token in order to logout.")
+		}
 	}
 
 	return nil
 }
 
 func (o LogoutOptions) RunLogout() error {
+	if o.Config.ExecProvider != nil {
```
stlaz, line 125: is it possible to to have a kubeconfig where there are multiple clusters configured with some of them using the execprovider for login and some of them don't?

stlaz (changes requested):

pkg/cli/login/loginoptions.go
ardaguclu: ah, it is nice to use that instead of copying. Thanks.

pkg/cli/logout/logout.go
ardaguclu: Each context in kubeconfig has it's own login mechanism. Some case use exec plugin and some can use token based.

pkg/cli/login/login.go
```diff
@@ -89,6 +97,11 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.
 
 	cmds.Flags().BoolVarP(&o.WebLogin, "web", "w", o.WebLogin, "Login with web browser. Starts a local HTTP callback server to perform the OAuth2 Authorization Code Grant flow. Use with caution on multi-user systems, as the server's port will be open to all users.")
 	cmds.Flags().Int32VarP(&o.CallbackPort, "callback-port", "c", o.CallbackPort, "Port for the callback server when using --web. Defaults to a random open port")
+
+	cmds.Flags().StringVar(&o.OIDCAuthType, "external-auth-type", o.OIDCAuthType, "Experimental: Specify the authentication type for external issuers to choose which plugin will be used for authentication. Valid values are oidc and azure.")
```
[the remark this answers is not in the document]
ardaguclu: @stlaz thanks for review. Could you please elaborate what is your suggestion?. Users can pass whatever they want that specific to their plugins via `OIDCExtraArgs`.

pkg/cli/login/login.go
```diff
@@ -177,6 +194,31 @@ func (o LoginOptions) Validate(cmd *cobra.Command, serverFlag string, args []str
...
+		return errors.New("--external-auth-type, --external-extra-args can only be used for external issuers and --external-client-id is required")
+	}
+
+	if len(o.OIDCClientID) > 0 {
+		if len(o.OIDCAuthType) == 0 {
+			return errors.New("auth type should be set and accepted values are oidc and azure")
+		}
+
+		if o.OIDCAuthType == AzureAuthType {
+			found, _ := oidc.ExternalOIDCExtraArgIsSet(o.OIDCExtraArgs, "--tenant-id")
```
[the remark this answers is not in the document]
ardaguclu: I agree with you. In any case, it is mandatory for azure kubelogin plugin and I think, we don't need to check this in here because plugin will return error if `tenant-id` is not set.

pkg/cli/login/loginoptions.go
ardaguclu: After thinking on this, I'd not prefer importing a new go module just for one helper function(more CVEs filed by scanners) unless I have to.

Instead I think, we can wrap this https://github.com/openshift/library-go/blob/cfcf2b4fbc878cad84f0552dfdabe1642b985f23/pkg/oauth/tokenrequest/request_token.go#L161-L200 to a helper function and oc can consume it seems like a better idea. WDYT?.

ardaguclu:
unrelated failure;
/retest

ardaguclu:
As we've decided how are going to move forward, we don't need this PR anymore;

/close

PR #1595, opened 2023-11-02
OTA-1035: Allow testing `adm upgrade status` on mock data

petr-muller:
While `oc adm upgrade status` is under development, logic is client-side, and
behind feature gate, it is useful to avoid relying on an actual upgrading
cluster in an interesting state at hand.

This will also allow us to capture interesting "stuck" upgrade cases and
back-test further development on previously seen interesting states.

I have added a simple improvement commit to this PR (emiting `<none>` when a reason or message is empty) to illustrate:

```console
$ OC_ENABLE_CMD_UPGRADE_STATUS=true ./oc --context build01 adm upgrade status
error: You must be logged in to the server (Unauthorized)
// my $KUBECONFIG does not have an authorized b01 context so I cannot test 

$ OC_ENABLE_CMD_UPGRADE_STATUS=true ./oc --context build01 adm upgrade status \
  --mock-clusteroperators=clusteroperators-clusteroperatorlist.yaml \
  --mock-clusterversion=clusterversion.yaml

  Reason: <none>
  Message: Cluster version is 4.14.1
```

### DIFF
--- a/pkg/cli/admin/upgrade/status/status.go
+++ b/pkg/cli/admin/upgrade/status/status.go
@@ -4,12 +4,16 @@ package status
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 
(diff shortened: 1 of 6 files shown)


pkg/cli/admin/upgrade/status/status.go
```diff
@@ -40,12 +44,23 @@ func New(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command
 		},
 	}
 
+	flags := cmd.Flags()
+	// TODO: We can remove these flags once the idea about `oc adm upgrade status` stabilizes and the command
+	//       is promoted out of the OC_ENABLE_CMD_UPGRADE_STATUS feature gate
+	flags.StringVar(&o.mockCvPath, "mock-clusterversion", "", "Path to a YAML ClusterVersion object to use for testing (will be removed later).")
+	flags.StringVar(&o.mockOperatorsPath, "mock-clusteroperators", "", "Path to a YAML ClusterOperatorList to use for testing (will be removed later).")
```
wking, line 51: Did we want to add some fixtures to this to unit test known configs vs. expected output?  Because running `go test ./pkg/cli/admin/upgrade/status` and having it consume mock data other folks have written and confirm that I haven't broken the output other folks were expecting in those situations would be nice.  Mainly to start building a corpus of situations we want to cover, and as a way to track how our output in those situations evolves over time.

Also, no worries if you think those fixtures sound like overkill, or if they sound useful in the future, but not worth picking up yet.  I'm also fine with a middle ground where we start a directory like `pkg/cli/admin/upgrade/status/fixtures/happy-mid-update`, etc. with the mock  content in it, even if we don't version-control expected output yet.

wking (approved):
Looks good to me.  Adding a hold in case you want to tweak in response to [this][1], but feel free to lift the hold without changes if none of that sounds right-now appealing ;)

/lgtm
/hold

[1]: https://github.com/openshift/oc/pull/1595#discussion_r1385381838

pkg/cli/admin/upgrade/status/status.go
petr-muller: That sounds great, will do!

petr-muller:
/lgtm cancel

I want to address Trevor's suggestion, canceling lgtm to avoid mistakenly lifting the hold

petr-muller:
```
: Run test unit - container test expand_more
: TestNewAppRunBuilds
: TestNewAppRunBuilds/successful_input_image_source_build_with_a_repository
```
The above really does not seem like it relates to my change?

/test unit

wking (approved):
/lgtm

PR #1597, opened 2023-11-03
OTA-1035: `adm upgrade status` Add control plane update status

petr-muller:
TBD

### DIFF
--- a/pkg/cli/admin/upgrade/status/controlplane.go
+++ b/pkg/cli/admin/upgrade/status/controlplane.go
@@ -8,6 +8,13 @@ import (
 	v1 "github.com/openshift/api/config/v1"
 )
 
+type assessmentState string
+
+const (
+	assessmentStateProgressing assessmentState = "Progressing"
+	assessmentStateCompleted   assessmentState = "Completed"
+)
+
 type operators struct {
 	Total       int
 	Available   int
@@ -16,11 +23,14 @@ type operators struct {
 }
 
 type controlPlaneStatusDisplayData struct {
-	Operators operators
+	Assessment assessmentState
+	Completion float64
+	Operators  operators
 }
 
 func assessControlPlaneStatus(cv *v1.ClusterVersion, operators []v1.ClusterOperator) controlPlaneStatusDisplayData {
 	var displayData controlPlaneStatusDisplayData
+	var completed int
 
 	targetVersion := cv.Status.Desired.Version
 
@@ -41,6 +51,8 @@ func assessControlPlaneStatus(cv *v1.ClusterVersion, operators []v1.ClusterOpera
 		for _, version := range operator.Status.Versions {
 			if version.Name == "operator" && version.Version == targetVersion {
 				updated = true
+				completed++
+				break
 			}
 		}
 		displayData.Operators.Total++
@@ -56,6 +68,13 @@ func assessControlPlaneStatus(cv *v1.ClusterVersion, operators []v1.ClusterOpera
 		}
 	}
 
+	if completed == displayData.Operators.Total {
+		displayData.Assessment = assessmentStateCompleted
+	} else {
+		displayData.Assessment = assessmentStateProgressing
+	}
+
+	displayData.Completion = float64(completed) / float64(displayData.Operators.Total) * 100.0
 	return displayData
 }
 
@@ -66,5 +85,7 @@ func (d *controlPlaneStatusDisplayData) Write(f io.Writer) error {
 }
 
 const controlPlaneStatusTemplateRaw = `= Control Plane =
+Assessment:      {{ .Assessment }}
+Completion:      {{ printf "%.0f" .Completion }}%
 Operator Status: {{ .Operators.Total }} Total, {{ .Operators.Available }} Available, {{ .Operators.Progressing }} Progressing, {{ .Operators.Degraded }} Degraded
 `
--- a/pkg/cli/admin/upgrade/status/controlplane.go
+++ b/pkg/cli/admin/upgrade/status/controlplane.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"strings"
 	"text/template"
+	"time"
 
 	v1 "github.com/openshift/api/config/v1"
 )
@@ -25,10 +26,11 @@ type operators struct {
 type controlPlaneStatusDisplayData struct {
 	Assessment assessmentState
 	Completion float64
+	Duration   time.Duration
 	Operators  operators
 }
 
-func assessControlPlaneStatus(cv *v1.ClusterVersion, operators []v1.ClusterOperator) controlPlaneStatusDisplayData {
+func assessControlPlaneStatus(cv *v1.ClusterVersion, operators []v1.ClusterOperator, at time.Time) controlPlaneStatusDisplayData {
 	var displayData controlPlaneStatusDisplayData
 	var completed int
 
@@ -74,6 +76,15 @@ func assessControlPlaneStatus(cv *v1.ClusterVersion, operators []v1.ClusterOpera
 		displayData.Assessment = assessmentStateProgressing
 	}
 
+	if len(cv.Status.History) > 0 {
+		currentHistoryItem := cv.Status.History[0]
+		if currentHistoryItem.State == v1.CompletedUpdate {
+			displayData.Duration = currentHistoryItem.CompletionTime.Time.Sub(currentHistoryItem.StartedTime.Time)
+		} else {
+			displayData.Duration = at.Sub(currentHistoryItem.StartedTime.Time)
+		}
+	}
+
 	displayData.Completion = float64(completed) / float64(displayData.Operators.Total) * 100.0
 	return displayData
 }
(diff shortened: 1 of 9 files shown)


pkg/cli/admin/upgrade/status/controlplane.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, operator := range operators {
+		displayData.Operators.Total++
+		for _, condition := range operator.Status.Conditions {
+			switch {
+			case condition.Type == v1.OperatorAvailable:
+				if condition.Status == v1.ConditionTrue {
+					displayData.Operators.Available++
+				}
+			case condition.Type == v1.OperatorProgressing && condition.Status == v1.ConditionTrue:
+				displayData.Operators.Progressing++
```
wking, line 43: I doubt `Progressing` is much use, because some ClusterOperator go `Progressing=True` on unrelated-to-update events like the cluster adding a compute node.  But :shrug:, also only a few digits in the output, so I will survive if other folks think it adds value to display this count.

pkg/cli/admin/upgrade/status/controlplane.go
```diff
@@ -0,0 +1,78 @@
...
+			case condition.Type == v1.OperatorDegraded && condition.Status == v1.ConditionTrue:
+				displayData.Operators.Degraded++
+			}
+		}
+
+		for _, version := range operator.Status.Versions {
+			if version.Name == "operator" {
+				if version.Version == cv.Spec.DesiredUpdate.Version {
```
wking, line 51: I think we should use `cv.Status.Desired.Version` here.  The `spec` property may not set a version (e.g. via a `--to-image` update request).  And the `spec` property may not be accepted as a new update target, while the `status` property will (or at least should) only include releases once they've been accepted as a target.

pkg/cli/admin/upgrade/status/controlplane.go
```diff
@@ -0,0 +1,78 @@
...
+
+var controlPlaneStatusTemplate = template.Must(template.New("controlPlaneStatus").Parse(controlPlaneStatusTemplateRaw))
+
+func renderControlPlaneStatus(f io.Writer, displayData controlPlaneStatusDisplayData) error {
+	return controlPlaneStatusTemplate.Execute(f, displayData)
+}
+
+const controlPlaneStatusTemplateRaw = `= Control Plane =
+Assessment:      {{ .Assessment }}
+Completion:      {{ printf "%.0f" .Completion }}%
```
wking, line 76: To avoid confusion vs. the `Progressing` % completion (which is based on manifest counts), it's probably worth adding qualifying text here that explains that it's a ClusterOperator percentage.  Something like:

```
Completion:      {{ printf "%.0f" .Completion }}% ({{ printf %d .ClusterOperatorCompletedCount }} of {{ printf %d .ClusterOperatorManagedCount }} cluster operators are {{ printf "%s" .TargetVersion }})

File: pkg/cli/admin/upgrade/status/controlplane.go (hunk 1 above)
Comment (petr-muller) [reply pkg/cli/admin/upgrade/status/controlplane.go:43]: I like `Progressing` as an indicator of "something is moving" which I think users generally like - "Progressing: 0" is something that I would hope to avoid in a healthy upgrade.

I agree about imprecision - I was thinking about a followup card about how exactly we would need to compute this.

I personally dislike the `Available` bit - this number is not interesting in most cases, until something goes unavailable, and *then* I think the amount of unavailable ones (1? 10?) is more interesting than Available getting lower.

File: pkg/cli/admin/upgrade/status/controlplane.go [hunk 4]
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+	}
+	if completed == displayData.Operators.Total {
+		displayData.Assessment = assessmentStateCompleted
+	} else {
+		displayData.Assessment = assessmentStateProgressing
+	}
+
+	displayData.Completion = float64(completed) / float64(displayData.Operators.Total) * 100.0
```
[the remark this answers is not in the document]
petr-muller: Good point about OLM-installed operators also having `ClusterOperator` resources, I did not know that. Adding a filter for CVO annotation is a good suggestiong :+1:

pkg/cli/admin/upgrade/status/controlplane.go
petr-muller: Yes! I have this discrepancy in my notes, we'll need to deal with that somehow.

pkg/cli/admin/upgrade/status/controlplane.go
wking: > "Progressing: 0" is something that I would hope to avoid in a healthy upgrade.

Does this happen?  I guess there could be times when the CVO is working with non-ClusterOperator resources where we are updating but have no ClusterOperator attempting to move.  But in that case, I'd want the `status` command to be talking to me about what the CVO was stuck on, not waiting for me to guess from `Progressing: 0`  that it was stuck before I started to do my own digging.

> I personally dislike the Available bit - this number is not interesting in most cases, until something goes unavailable, and then I think the amount of unavailable ones (1? 10?) is more interesting than Available getting lower.

Good point.

petr-muller: 1. The current iteration only counts a cluster operator as progressing if it has `Progressing=True` and is still at old version. I think that's good enough? I still think it is useful as an indicator that something is moving, so I think we should keep it, at least for now.
2. I kept even `Available: X` for now, but I really think we should present this better, it is mostly just clutter.
3. We actually spend a LOT of time in `Progressing: 0` state, check out the traces in https://docs.google.com/document/d/1rWKPA0DrBU6aA4fsolEfOJIrenzeO5kVKtCoDsMum50/edit?usp=sharing. It seems that most COs will bump themselves right away and don't spend enough time in that state?

pkg/cli/admin/upgrade/status/controlplane.go
petr-muller: Done in current HEAD 12bd58bcaf3a731aff4e54fc13a53278989bd7b0

pkg/cli/admin/upgrade/status/controlplane.go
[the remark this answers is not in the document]
petr-muller: Done in current HEAD 12bd58bcaf3a731aff4e54fc13a53278989bd7b0. It seems that some clusters do not have their CO resource annotated at all though, that may be a bug in CVO (check out [b02 trace](https://docs.google.com/document/d/1rWKPA0DrBU6aA4fsolEfOJIrenzeO5kVKtCoDsMum50/edit#heading=h.igm7fpl6uf1z))

pkg/cli/admin/upgrade/status/controlplane.go
petr-muller: As long as we show the completion on the top line (which is manifest-based and I am increasingly convinced it is really confusing for users, because payload manifests are not a user-facing concept, we dont even say 303 of 867 of what, and the percentage resets several times during the upgrade), having two numbers will be always be confusing...

I'm not totally against being transparent and say that we count COs here, but I'm personally leaning towards making this number opaque, so that we can change its implementation if we find more precise notions of completeness (maybe really count manifests?).

pkg/cli/admin/upgrade/status/controlplane.go
```diff
@@ -0,0 +1,105 @@
...
+			}
+		}
+		displayData.Operators.Total++
+		for _, condition := range operator.Status.Conditions {
+			switch {
+			case condition.Type == v1.OperatorAvailable:
+				if condition.Status == v1.ConditionTrue {
+					displayData.Operators.Available++
+				}
+			case condition.Type == v1.OperatorProgressing && condition.Status == v1.ConditionTrue && !updated:
```
wking, line 67: nit: `Available` splits  this into `case` and an `if`, but  here and for `Degraded` you check both `Type` and `Status` in the `case`.  Not a big deal, but you may want to consolidate around one or the other approach, just to make it slightly easier code-reading.

pkg/cli/admin/upgrade/status/controlplane.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+
+	displayData.Completion = float64(completed) / float64(displayData.Operators.Total) * 100.0
+	return displayData
+}
+
+var controlPlaneStatusTemplate = template.Must(template.New("controlPlaneStatus").Parse(controlPlaneStatusTemplateRaw))
+
+func renderControlPlaneStatus(f io.Writer, displayData controlPlaneStatusDisplayData) error {
```
wking, line 96: nit: feels like this could be a method on `controlPlaneStatusDisplayData`, like `func (s *controlPlaneStatusDisplayData) Write(f io.Writer) error`.  But :shrug:, I can't point to a reason for that beyond "feels kindof like `String()`, and that's usually a method", so I have no problems with this staying a function either.

wking (approved):
/lgtm
/hold in case you  want to make any futher pivots.  Lift the hold whenever you like

petr-muller:
/hold cancel

Thanks for reviews, addressed!

pkg/cli/admin/upgrade/status/controlplane.go
petr-muller: Good catch, thanks!

pkg/cli/admin/upgrade/status/controlplane.go
petr-muller: Good catch, thanks!

petr-muller:
`ci/prow/security`: I have reviewed https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc/1597/pull-ci-openshift-oc-master-security/1729882511145701376 and there are no vulnerabilities reported in files from this PR.

PR #1600, opened 2023-11-20
[must-gather] Add support of extra arguments for collection script

black-dragon74:
This patch adds the ability to pass additional arguments to the must-gather script. This is an upcoming feature meant to modularize and optimize the mg collection process.

Note: The current implementation does not forward the args to the deployed pod when explicitly using a custom entry-point.

This is done to prevent the args overlap when using custom collection scripts.

Regards

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -58,6 +58,9 @@ var (
 	mustGatherExample = templates.Examples(`
 		# Gather information using the default plug-in image and command, writing into ./must-gather.local.<rand>
 		  oc adm must-gather
+		
+		# Pass additional arguments to the must-gather script (if supported by the image)
+		oc adm must-gather --args=""
 
 		# Gather information with a specific local folder to copy to
 		  oc adm must-gather --dest-dir=/local/directory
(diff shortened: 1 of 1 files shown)


black-dragon74:
cc: @yati1998

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -95,6 +98,7 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 	cmd.Flags().BoolVar(&o.HostNetwork, "host-network", o.HostNetwork, "Run must-gather pods as hostNetwork: true - relevant if a specific command and image needs to capture host-level data")
 	cmd.Flags().StringSliceVar(&o.Images, "image", o.Images, "Specify a must-gather plugin image to run. If not specified, OpenShift's default must-gather image will be used.")
 	cmd.Flags().StringSliceVar(&o.ImageStreams, "image-stream", o.ImageStreams, "Specify an image stream (namespace/name:tag) containing a must-gather plugin image to run.")
+	cmd.Flags().StringVar(&o.GatherArgs, "args", "", "Extra arguments to pass to the must-gather script.")
```
ardaguclu, line 101: As far as I know, there wasn't any request to pass arguments to the must-gather images(including the default one) and it seems like a requirement very specific to a custom must-gather image. 

To preserve simplicity, I'd not prefer adding a new flag in must-gather command, instead I'd recommend using environment variables to manage.

sferich888: The arguments are also only passed to the 'gather' script. 
Why not just use `oc adm must-gather --image IMAGE -- [gather|OTHER_SCRIPT] ARGUMENT` to run the command?

black-dragon74: Ack, As this functionality is only supported my odf-must-gather, we will advise customers to use the custom command syntax as of now. The only redundancy I see going this way is providing the abs path to the gather script before the args. `-- /usr/bin/gather --args`.

Regards

black-dragon74:
Closing the PR as it is not needed.

PR #1601, opened 2023-11-23
OCPBUGS-23071: Mention inspection completion in the error message

ardaguclu:
When error occurs during inspection, current message `errors occurred while gathering data` may mislead
users about that the command short cuts inspection process and exits. However, in reality, command prints all the logs
in aggregated format after completing the inspection.

This PR basically just adds `inspection completed` message ahead of the error logs to emphasize to user that
inspection is completed in best effort and the rest is the errors we get.


ardaguclu:
@ingvagabund could you please have a look at, when you have a chance?. Thanks.

/tide refresh

PR #1608, opened 2023-11-29
CORS-2785: extract openshift-install from installer-artifacts

r4f4:
We are decoupling installer-artifacts from installer, so they can be built in parallel both in CI and during release build. As a consequence, the openshift-install binary should be extracted from the installer-artifacts image exclusively.


r4f4:
/retitle CORS-2785: extract openshift-install from installer-artifacts

r4f4:
This might have been a bit premature. I'll revisit this idea in the future.
/close

PR #1612, opened 2023-12-05
OCPBUGS-24375: Overwrite template's namespace with the explicit one

ardaguclu:
https://github.com/openshift/oc/pull/1318 fixes only when the template is passed
directly. But the problem still exists, when the template is passed in a file.

This PR overwrites the namespace in template with the one passed explicitly
in command, when template's namespace and explicit namespace is mismatched.


soltysh (approved):
/lgtm

PR #1614, opened 2023-12-05
OCPBUGS-24456: Be more verbose about must-gather failure

andreaskaris:
Print an explicit failure message, in red and in bold, at the end of a failed must-gather run.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-24456


andreaskaris:
image

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -92,6 +92,11 @@ sleep 5
 done`
 )
 
+const (
+	redColor   = "\u001b[31;1m"
```
ardaguclu, line 96: This is unix specific and I'd not prefer to add something os specific.

andreaskaris: Hi, thanks for having a look at this. I'll see if I can find something platform agnostic without pulling in too many dependencies. However, just to make you aware of this:

image

That color is coming from vendor/k8s.io/client-go/rest/warnings.go which is doing the exact same thing:
```
  102 const (                                                                                                                 
  103     yellowColor = "\u001b[33;1m"                                                                                        
  104     resetColor  = "\u001b[0m"                                                                                           
  105 )                                                                                                                       
  106                                                                                                                         
  107 // HandleWarningHeader prints warnings with code=299 to the configured writer.                                          
  108 func (w *warningWriter) HandleWarningHeader(code int, agent string, message string) {                                   
  109     if code != 299 || len(message) == 0 {                                                                               
  110         return                                                                                                          
  111     }                                                                                                                   
  112                                                                                                                         
  113     w.writtenLock.Lock()                                                                                                
  114     defer w.writtenLock.Unlock()                                                                                        
  115                                                                                                                         
  116     if w.opts.Deduplicate {                                                                                             
  117         if _, alreadyWritten := w.written[message]; alreadyWritten {                                                    
  118             return                                                                                                      
  119         }                                                                                                               
  120         w.written[message] = struct{}{}                                                                                 
  121     }                                                                                                                   
  122     w.writtenCount++                                                                                                    
  123                                                                                                                         
  124     if w.opts.Color {                                                                                                   
  125         fmt.Fprintf(w.out, "%sWarning:%s %s\n", yellowColor, resetColor, message)                                       
  126     } else {                                                                                                            
  127         fmt.Fprintf(w.out, "Warning: %s\n", message)                                                                    
  128     }                                                                                                                   
  129 }  
```

andreaskaris:
With the latest changes:

Linux:
image

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -92,6 +93,11 @@ sleep 5
...
```
andreaskaris, line 97: This is specific to terminals that support ANSI color codes, but o.fatal checks `kterm.AllowsColorOutput(o.IOStreams.ErrOut)`; meaning that color will be printed only for terminals that support color output.

The same approach is already used when a "Warning:" message is printed in yellow - in the libraries that `oc must-gather` is using in k8s.io/client-go/rest/warnings.go:
https://github.com/kubernetes/client-go/blob/0a514be22c2b4718e71c09dbd6fc3841f5879432/rest/warnings.go#L124

The Color value is set by NewOcCommand:
https://github.com/openshift/oc/blob/36f5a453c756a2a1da548fb996bdfac98cd1df3d/pkg/cli/cli.go#L174

Issue #1615, opened 2023-12-05
`oc logout` should not prompt for username and password when logged out


datho7561:
If I run `oc logout` when the current kube context is set to an OpenShift cluster but I'm logged out, `oc` prompts for a username and password.

The use case is [the integration test suite for vscode-openshift-tools](https://github.com/redhat-developer/vscode-openshift-tools/blob/main/test/integration/ocWrapper.test.ts#L27). We are using `oc` to log in and out of the OpenShift cluster. In the integration tests, we call `oc logout` at the beginning of the test suite regardless of if the user is logged in or not, in order to be robust and handle the case that the developer was logged in when launching the test suite.

Version information:

Client Version: 4.14.3
Kustomize Version: v5.0.1
Kubernetes Version: v1.26.6+73ac56

ardaguclu:
Patch has been merged and issue should be resolved. I'm closing;

/close

PR #1619, opened 2023-12-07
OCPBUGS-24225: Remove deprecated password defaulting in default config flag

ardaguclu:
It seemed reasonable to add `WithDeprecatedPasswordFlag` to sync with kubectl in https://github.com/openshift/oc/pull/1424#discussion_r1269268904. However, 
unexpectedly this caused a slight behavioral change in oc command as filed in
https://github.com/openshift/oc/issues/1615 and the referenced issue to this PR.

The reason of the failure is when password is not nil(instead it is empty string ""), 
config flags prompts user to enter username and password to move forward https://github.com/kubernetes/kubernetes/blob/a504aed54d028dbc8ea2508142c94d309f5f1ec6/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L235. However, prior oc behavior is
password should be nil and it returns non interactive config https://github.com/kubernetes/kubernetes/blob/a504aed54d028dbc8ea2508142c94d309f5f1ec6/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L233 which will end up returning to user
about not found error of the resource.

This PR removes `WithDeprecatedPasswordFlag` from the default to preserve
old behavior. This change is safe because we don't use it anyway.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -98,7 +98,7 @@ var (
 )
 
 func defaultConfigFlags() *genericclioptions.ConfigFlags {
-	return genericclioptions.NewConfigFlags(true).WithDeprecatedPasswordFlag().WithDiscoveryBurst(350).WithDiscoveryQPS(50.0)
+	return genericclioptions.NewConfigFlags(true).WithDiscoveryBurst(350).WithDiscoveryQPS(50.0)
 }
 
 func NewDefaultOcCommand(o kubecmd.KubectlOptions) *cobra.Command {


ardaguclu:
/jira refresh
/assign @atiratree 
could you PTAL this change, when you have any availability?. Thanks.

atiratree:
/lgtm

@ardaguclu makes sense, we do not need to be compatible with kubectl in this regard - especially when it makes problems. Thanks for the fix.

We should probably backport this one, right?

ardaguclu:
Yes, good point;

/cherry-pick release-4.15

PR #1622, opened 2023-12-10
OCPBUGS-24834: Bump api to include the new CloudCredential cap

fxierh:
CCO was made optional in https://issues.redhat.com/browse/OCPEDGE-69. CloudCredential was introduced as a new capability to openshift/api. 

We need to bump openshift/api to include the new cap so 
`oc adm release extract --credentials-requests --included --install-config=install-config.yaml ...` 
recognizes it and is able to extract all relevant credentials requests.


ardaguclu:
This bump should be done in 4.16 anyway. Thanks.

/lgtm

PR #1633, opened 2023-12-13
CNV-37284: Add --all-images flag to must-gather

dharmit:
This PR is based on the discussions in https://github.com/openshift/enhancements/pull/1487.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -15,19 +15,22 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
-	"k8s.io/klog/v2"
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"k8s.io/cli-runtime/pkg/printers"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/logs"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/polymorphichelpers"
@@ -41,7 +44,6 @@ import (
 	"github.com/openshift/library-go/pkg/image/imageutil"
 	imagereference "github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/library-go/pkg/operator/resource/retry"
-
 	"github.com/openshift/oc/pkg/cli/admin/inspect"
 	"github.com/openshift/oc/pkg/cli/rsync"
 	ocmdhelpers "github.com/openshift/oc/pkg/helpers/cmd"
@@ -96,6 +98,13 @@ sleep 5
 done`
 )
 
+const (
+	// number of concurrent must-gather Pods to run if --all-images or multiple --image are provided
+	concurrentMG = 4
+	// annotation to look for in ClusterServiceVersions and ClusterOperators when using --all-images
+	mgAnnotation = "operators.openshift.io/must-gather-image"
+)
+
 func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
 	o := NewMustGatherOptions(streams)
 	cmd := &cobra.Command{
@@ -115,6 +124,7 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 	cmd.Flags().BoolVar(&o.HostNetwork, "host-network", o.HostNetwork, "Run must-gather pods as hostNetwork: true - relevant if a specific command and image needs to capture host-level data")
 	cmd.Flags().StringSliceVar(&o.Images, "image", o.Images, "Specify a must-gather plugin image to run. If not specified, OpenShift's default must-gather image will be used.")
 	cmd.Flags().StringSliceVar(&o.ImageStreams, "image-stream", o.ImageStreams, "Specify an image stream (namespace/name:tag) containing a must-gather plugin image to run.")
+	cmd.Flags().BoolVar(&o.AllImages, "all-images", o.AllImages, `Collect must-gather using the default image for all Operators on the cluster annotated with `+mgAnnotation)
 	cmd.Flags().StringVar(&o.DestDir, "dest-dir", o.DestDir, "Set a specific directory on the local machine to write gathered data to.")
 	cmd.Flags().StringVar(&o.SourceDir, "source-dir", o.SourceDir, "Set the specific directory on the pod copy the gathered data from.")
 	cmd.Flags().StringVar(&o.timeoutStr, "timeout", "10m", "The length of time to gather data, like 5s, 2m, or 3h, higher than zero. Defaults to 10 minutes.")
@@ -152,6 +162,9 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 	if o.ConfigClient, err = configclient.NewForConfig(o.Config); err != nil {
 		return err
 	}
+	if o.DynamicClient, err = dynamic.NewForConfig(o.Config); err != nil {
+		return err
+	}
 	if o.ImageClient, err = imagev1client.NewForConfig(o.Config); err != nil {
 		return err
 	}
@@ -178,6 +191,25 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 	if len(o.DestDir) == 0 {
 		o.DestDir = fmt.Sprintf("must-gather.local.%06d", rand.Int63())
 	}
+	// TODO: this should be in Validate() method, but added here because of the call to o.completeImages() below
+	if o.AllImages {
+		errStr := fmt.Sprintf("and --all-images are mutually exclusive: please specify one or the other")
+		if len(o.Images) != 0 {
+			return fmt.Errorf("--image %s", errStr)
+		}
+		if o.HostNetwork {
+			return fmt.Errorf("--host-network %s", errStr)
+		}
+		if len(o.ImageStreams) != 0 {
+			return fmt.Errorf("--image-streams %s", errStr)
+		}
+		if o.NodeName != "" {
+			return fmt.Errorf("--node-name %s", errStr)
+		}
+		if o.RunNamespace != "" {
+			return fmt.Errorf("--run-namespace %s", errStr)
+		}
+	}
 	if err := o.completeImages(); err != nil {
 		return err
 	}
@@ -201,7 +233,7 @@ func (o *MustGatherOptions) completeImages() error {
 			return fmt.Errorf("unable to resolve image stream '%v': %v", imageStream, err)
 		}
 	}
-	if len(o.Images) == 0 {
+	if len(o.Images) == 0 || o.AllImages {
 		var image string
 		var err error
 		if image, err = o.resolveImageStreamTag("openshift", "must-gather", "latest"); err != nil {
@@ -210,9 +242,57 @@ func (o *MustGatherOptions) completeImages() error {
 		}
 		o.Images = append(o.Images, image)
 	}
+	if o.AllImages {
+		// find all csvs and clusteroperators with the annotation "operators.openshift.io/must-gather-image"
+		pluginImages := make(map[string]struct{})
+		var err error
+
+		pluginImages, err = o.annotatedCSVs()
+		if err != nil {
+			return err
+		}
+
+		cos, err := o.ConfigClient.ConfigV1().ClusterOperators().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		for _, item := range cos.Items {
+			ann := item.GetAnnotations()
+			if v, ok := ann[mgAnnotation]; ok {
+				pluginImages[v] = struct{}{}
+			}
+		}
+		// delete the default image to avoid duplication in case an Operator had it in its annotation
+		delete(pluginImages, o.Images[0])
+		for i := range pluginImages {
+			o.Images = append(o.Images, i)
+		}
+	}
 	o.log("Using must-gather plug-in image: %s", strings.Join(o.Images, ", "))
 
 	return nil
+}
+
+func (o *MustGatherOptions) annotatedCSVs() (map[string]struct{}, error) {
+	var csvGVR = schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
+		Version:  "v1alpha1",
+		Resource: "clusterserviceversions",
+	}
+	pluginImages := make(map[string]struct{})
+
+	csvs, err := o.DynamicClient.Resource(csvGVR).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, item := range csvs.Items {
+		ann := item.GetAnnotations()
+		if v, ok := ann[mgAnnotation]; ok {
+			pluginImages[v] = struct{}{}
+		}
+	}
+	return pluginImages, nil
 }
 
 func (o *MustGatherOptions) resolveImageStreamTagString(s string) (string, error) {
@@ -255,6 +335,7 @@ type MustGatherOptions struct {
 	Config           *rest.Config
 	Client           kubernetes.Interface
 	ConfigClient     configclient.Interface
+	DynamicClient    dynamic.Interface
 	ImageClient      imagev1client.ImageV1Interface
 	RESTClientGetter genericclioptions.RESTClientGetter
 
@@ -264,6 +345,7 @@ type MustGatherOptions struct {
 	DestDir          string
 	SourceDir        string
 	Images           []string
+	AllImages        bool
 	ImageStreams     []string
 	Command          []string
 	Timeout          time.Duration
@@ -321,7 +403,7 @@ func (o *MustGatherOptions) Validate() error {
 	return nil
 }
 
-// Run creates and runs a must-gather pod.d
+// Run creates and runs a must-gather pod
 func (o *MustGatherOptions) Run() error {
 	var errs []error
 
@@ -414,14 +496,7 @@ func (o *MustGatherOptions) Run() error {
 				return err
 			}
 			for _, node := range nodes.Items {
-				pod, err := o.Client.CoreV1().Pods(ns.Name).Create(context.TODO(), o.newPod(node.Name, image, hasMaster), metav1.CreateOptions{})
-				if err != nil {
-					// ensure the errors bubble up to BackupGathering method for display
-					errs = []error{err}
-					return err
-				}
-				o.log("pod: %s on node: %s for plug-in image %s created", pod.Name, node.Name, image)
-				pods = append(pods, pod)
+				pods = append(pods, o.newPod(node.Name, image, hasMaster))
 			}
 		} else {
 			if o.NodeName != "" {
@@ -431,14 +506,7 @@ func (o *MustGatherOptions) Run() error {
 					return err
 				}
 			}
-			pod, err := o.Client.CoreV1().Pods(ns.Name).Create(context.TODO(), o.newPod(o.NodeName, image, hasMaster), metav1.CreateOptions{})
-			if err != nil {
-				// ensure the errors bubble up to BackupGathering method for display
-				errs = []error{err}
-				return err
-			}
-			o.log("pod for plug-in image %s created", image)
-			pods = append(pods, pod)
+			pods = append(pods, o.newPod(o.NodeName, image, hasMaster))
 		}
 	}
 
@@ -450,65 +518,31 @@ func (o *MustGatherOptions) Run() error {
 	}
 	defer o.logTimestamp()
 
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+
 	var wg sync.WaitGroup
-	wg.Add(len(pods))
 	errCh := make(chan error, len(pods))
+
 	for _, pod := range pods {
-		go func(pod *corev1.Pod) {
+		queue.Add(pod)
+	}
+	queue.ShutDownWithDrain()
+
+	wg.Add(concurrentMG)
+	for i := 0; i < concurrentMG; i++ {
+		go func() {
 			defer wg.Done()
-
-			if len(o.RunNamespace) > 0 && !o.Keep {
-				defer func() {
-					// must-gather runs in its own separate namespace as default , so after it is completed
-					// it deletes this namespace and all the pods are removed by garbage collector.
-					// However, if user specifies namespace via `run-namespace`, these pods need to
-					// be deleted manually.
-					err = o.Client.CoreV1().Pods(o.RunNamespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
-					if err != nil {
-						klog.V(4).Infof("pod deletion error %v", err)
-					}
-				}()
-			}
-
-			log := o.newPodOutLogger(o.Out, pod.Name)
-
-			// wait for gather container to be running (gather is running)
-			if err := o.waitForGatherContainerRunning(pod); err != nil {
-				log("gather did not start: %s", err)
-				errCh <- fmt.Errorf("gather did not start for pod %s: %s", pod.Name, err)
-				return
-			}
-			// stream gather container logs
-			if err := o.getGatherContainerLogs(pod); err != nil {
-				log("gather logs unavailable: %v", err)
-			}
-
-			// wait for pod to be running (gather has completed)
-			log("waiting for gather to complete")
-			if err := o.waitForGatherToComplete(pod); err != nil {
-				log("gather never finished: %v", err)
-				if exiterr, ok := err.(*exec.CodeExitError); ok {
-					errCh <- exiterr
-				} else {
-					errCh <- fmt.Errorf("gather never finished for pod %s: %s", pod.Name, err)
+			for {
+				pod, quit := queue.Get()
+				if quit {
+					return
 				}
-				return
+				defer queue.Done(pod)
+				if err := o.processNextWorkItem(ns.Name, pod.(*corev1.Pod)); err != nil {
+					errCh <- err
+				}
 			}
-
-			// copy the gathered files into the local destination dir
-			log("downloading gather output")
-			pod, err = o.Client.CoreV1().Pods(pod.Namespace).Get(context.TODO(), pod.Name, metav1.GetOptions{})
-			if err != nil {
-				log("gather output not downloaded: %v\n", err)
-				errCh <- fmt.Errorf("unable to download output from pod %s: %s", pod.Name, err)
-				return
-			}
-			if err := o.copyFilesFromPod(pod); err != nil {
-				log("gather output not downloaded: %v\n", err)
-				errCh <- fmt.Errorf("unable to download output from pod %s: %s", pod.Name, err)
-				return
-			}
-		}(pod)
+		}()
 	}
 	wg.Wait()
 	close(errCh)
@@ -532,6 +566,68 @@ func (o *MustGatherOptions) Run() error {
 	}
 
 	return errors.NewAggregate(errs)
+}
+
+// processNextWorkItem creates & processes the must-gather pod and returns error if any
+func (o *MustGatherOptions) processNextWorkItem(ns string, pod *corev1.Pod) error {
+	var err error
+	pod, err = o.Client.CoreV1().Pods(ns).Create(context.TODO(), pod, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	if o.NodeSelector != "" {
+		o.log("pod: %s on node: %s for plug-in image %s created", pod.Name, pod.Spec.NodeName, pod.Spec.Containers[0].Image)
+	} else {
+		o.log("pod for plug-in image %s created", pod.Spec.Containers[0].Image)
+	}
+	if len(o.RunNamespace) > 0 && !o.Keep {
+		defer func() {
+			// must-gather runs in its own separate namespace as default , so after it is completed
+			// it deletes this namespace and all the pods are removed by garbage collector.
+			// However, if user specifies namespace via `run-namespace`, these pods need to
+			// be deleted manually.
+			err = o.Client.CoreV1().Pods(o.RunNamespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+			if err != nil {
+				klog.V(4).Infof("pod deletion error %v", err)
+			}
+		}()
+	}
+
+	log := o.newPodOutLogger(o.Out, pod.Name)
+
+	// wait for gather container to be running (gather is running)
+	if err := o.waitForGatherContainerRunning(pod); err != nil {
+		log("gather did not start: %s", err)
+		return fmt.Errorf("gather did not start for pod %s: %s", pod.Name, err)
+
+	}
+	// stream gather container logs
+	if err := o.getGatherContainerLogs(pod); err != nil {
+		log("gather logs unavailable: %v", err)
+	}
+
+	// wait for pod to be running (gather has completed)
+	log("waiting for gather to complete")
+	if err := o.waitForGatherToComplete(pod); err != nil {
+		log("gather never finished: %v", err)
+		if exiterr, ok := err.(*exec.CodeExitError); ok {
+			return exiterr
+		}
+		return fmt.Errorf("gather never finished for pod %s: %s", pod.Name, err)
+	}
+
+	// copy the gathered files into the local destination dir
+	log("downloading gather output")
+	pod, err = o.Client.CoreV1().Pods(pod.Namespace).Get(context.TODO(), pod.Name, metav1.GetOptions{})
+	if err != nil {
+		log("gather output not downloaded: %v\n", err)
+		return fmt.Errorf("unable to download output from pod %s: %s", pod.Name, err)
+	}
+	if err := o.copyFilesFromPod(pod); err != nil {
+		log("gather output not downloaded: %v\n", err)
+		return fmt.Errorf("unable to download output from pod %s: %s", pod.Name, err)
+	}
+	return nil
 }
 
 func (o *MustGatherOptions) newPodOutLogger(out io.Writer, podName string) func(string, ...interface{}) {
--- a/pkg/cli/admin/mustgather/mustgather_test.go
+++ b/pkg/cli/admin/mustgather/mustgather_test.go
@@ -8,28 +8,36 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"k8s.io/cli-runtime/pkg/printers"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/utils/diff"
 
+	configv1 "github.com/openshift/api/config/v1"
 	imagev1 "github.com/openshift/api/image/v1"
+	configv1fake "github.com/openshift/client-go/config/clientset/versioned/fake"
 	imageclient "github.com/openshift/client-go/image/clientset/versioned/fake"
 )
 
 func TestImagesAndImageStreams(t *testing.T) {
 
 	testCases := []struct {
-		name           string
-		images         []string
-		imageStreams   []string
-		expectedImages []string
-		objects        []runtime.Object
+		name                string
+		images              []string
+		imageStreams        []string
+		expectedImages      []string
+		imageObjects        []runtime.Object
+		unstructuredObjects []runtime.Object
+		coObjects           []runtime.Object
+		allImages           bool
 	}{
 		{
 			name: "Default",
-			objects: []runtime.Object{
+			imageObjects: []runtime.Object{
 				newImageStream("openshift", "must-gather", withTag("latest", "registry.test/must-gather:1.0.0")),
 			},
 			expectedImages: []string{"registry.test/must-gather:1.0.0"},
@@ -47,7 +55,7 @@ func TestImagesAndImageStreams(t *testing.T) {
 			name:           "MultipleImageStreams",
 			imageStreams:   []string{"test/one:a", "test/two:a", "test/three:a", "test/three:b"},
 			expectedImages: []string{"one@a", "two@a", "three@a", "three@b"},
-			objects: []runtime.Object{
+			imageObjects: []runtime.Object{
 				newImageStream("test", "one", withTag("a", "one@a")),
 				newImageStream("test", "two", withTag("a", "two@a")),
 				newImageStream("test", "three",
@@ -61,10 +69,28 @@ func TestImagesAndImageStreams(t *testing.T) {
 			images:         []string{"one", "two"},
 			imageStreams:   []string{"test/three:a", "test/four:a"},
 			expectedImages: []string{"one", "two", "three@a", "four@a"},
-			objects: []runtime.Object{
+			imageObjects: []runtime.Object{
 				newImageStream("test", "three", withTag("a", "three@a")),
 				newImageStream("test", "four", withTag("a", "four@a")),
 			},
+		},
+		{
+			name:      "AllImagesOnlyCSVs",
+			allImages: true,
+			unstructuredObjects: []runtime.Object{
+				newClusterServiceVersion("csva", "test/csv:a", true),
+				newClusterServiceVersion("csvb", "test/csv:b", false),
+			},
+			expectedImages: []string{"registry.redhat.io/openshift4/ose-must-gather:latest", "test/csv:a"},
+		},
+		{
+			name:      "AllImagesOnlyCOs",
+			allImages: true,
+			coObjects: []runtime.Object{
+				newClusterOperator("coa", "test/co:a", true),
+				newClusterOperator("cob", "test/co:b", false),
+			},
+			expectedImages: []string{"registry.redhat.io/openshift4/ose-must-gather:latest", "test/co:a"},
 		},
 	}
 
@@ -72,11 +98,22 @@ func TestImagesAndImageStreams(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			options := MustGatherOptions{
 				IOStreams:    genericiooptions.NewTestIOStreamsDiscard(),
+				ConfigClient: configv1fake.NewSimpleClientset(tc.coObjects...),
 				Client:       fake.NewSimpleClientset(),
-				ImageClient:  imageclient.NewSimpleClientset(tc.objects...).ImageV1(),
+				DynamicClient: dynamicfake.NewSimpleDynamicClient(func() *runtime.Scheme {
+					s := runtime.NewScheme()
+					s.AddKnownTypeWithName(schema.GroupVersionKind{
+						Group:   "operators.coreos.com",
+						Version: "v1alpha1",
+						Kind:    "ClusterServiceVersionList",
+					}, &unstructured.UnstructuredList{})
+					return s
+				}(), tc.unstructuredObjects...),
+				ImageClient:  imageclient.NewSimpleClientset(tc.imageObjects...).ImageV1(),
 				Images:       tc.images,
 				ImageStreams: tc.imageStreams,
 				LogOut:       genericiooptions.NewTestIOStreamsDiscard().Out,
+				AllImages:    tc.allImages,
 			}
 			err := options.completeImages()
 			if err != nil {
@@ -108,6 +145,29 @@ func withTag(tag, reference string) func(*imagev1.ImageStream) *imagev1.ImageStr
 		})
 		return imageStream
 	}
+}
+
+func newClusterServiceVersion(name, image string, annotate bool) *unstructured.Unstructured {
+	u := &unstructured.Unstructured{}
+	u.SetName(name)
+	u.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "operators.coreos.com",
+		Version: "v1alpha1",
+		Kind:    "ClusterServiceVersion",
+	})
+	if annotate {
+		u.SetAnnotations(map[string]string{"operators.openshift.io/must-gather-image": image})
+	}
+	return u
+}
+
+func newClusterOperator(name, image string, annotate bool) *configv1.ClusterOperator {
+	c := &configv1.ClusterOperator{}
+	c.SetName(name)
+	if annotate {
+		c.SetAnnotations(map[string]string{"operators.openshift.io/must-gather-image": image})
+	}
+	return c
 }
 
 func TestGetNamespace(t *testing.T) {


dharmit:
Trying to add unit tests to this PR, but I am hitting a wall.

I've created a separate commit for the unit tests. Added the `github.com/operator-framework/api` dependency to go.mod. Running the tests fails in `mustgather_test.go` at the step:
```go
ImageClient:  imageclient.NewSimpleClientset(tc.objects...).ImageV1(),
```
It seems to fail because the `tc.objects` in this case has a `ClusterServiceVersion` and not an image or imagestream. And the error message below makes sense to me:
```go
panic: no kind is registered for the type v1alpha1.ClusterServiceVersion in scheme "github.com/openshift/client-go/image/clientset/versioned/fake/register.go:14" [recovered]
	panic: no kind is registered for the type v1alpha1.ClusterServiceVersion in scheme "github.com/openshift/client-go/image/clientset/versioned/fake/register.go:14"
```
But I don't understand how to address this. Should we:
1. Add `ClusterServiceVersion` to scheme? If yes, where's this scheme defined? I spent a crazy amount of time, but couldn't figure out. :(
2. Add an `OperatorClient` like the `ImageClient`? I tried this, but that's panicking as well.

@ardaguclu @soltysh can you help?

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +211,73 @@ func (o *MustGatherOptions) completeImages() error {
...
+		var csvs []byte
+		var cos []byte
+		var err error
+		var csvPath = "/apis/operators.coreos.com/v1alpha1/clusterserviceversions"
+		var coPath = "/apis/config.openshift.io/v1/clusteroperators"
+		var pluginImages = map[string]struct{}{}
+
+		csvs, err = o.Client.Discovery().RESTClient().Get().
+			AbsPath(csvPath).
+			DoRaw(context.TODO())
```
ardaguclu, line 225: These raw requests look to me quite _hacky_. Can't we filter these resources based on label selectors or field selectors?

dharmit: > These raw requests look to me quite hacky.

Do you suggest using ClientSet instead?

> Can't we filter these resources based on label selectors or field selectors?

We're looking for a specific annotation, and based on my web search-fu, there isn't a way to filter the resources based on annotations. Let me know if you have any ideas.

pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -66,6 +70,15 @@ func TestImagesAndImageStreams(t *testing.T) {
 				newImageStream("test", "four", withTag("a", "four@a")),
 			},
 		},
+		{
```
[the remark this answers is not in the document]
dharmit: > of course, this should be registered in like fakeregister for unit tests.

Yep, and that fakeregister is the thing I'm unable to find.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +211,73 @@ func (o *MustGatherOptions) completeImages() error {
...
+
+		for i := range pluginImages {
+			o.Images = append(o.Images, i)
+		}
+	}
 	o.log("Using must-gather plug-in image: %s", strings.Join(o.Images, ", "))
 	return nil
 }
 
+func annotatedOperators(data []byte) (map[string]struct{}, error) {
```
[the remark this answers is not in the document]
dharmit: We're trying to find based on existing annotations, not trying to annotate. Your comment seems to be about the latter. Correct me if I am wrong.

pkg/cli/admin/mustgather/mustgather_test.go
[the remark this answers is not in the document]
ardaguclu: This is the example of registering a type in unit tests as fake https://github.com/kubernetes/kubernetes/blob/a504aed54d028dbc8ea2508142c94d309f5f1ec6/staging/src/k8s.io/kubectl/pkg/cmd/testing/fake.go#L339. Let me know if this doesn't work.

dharmit:
@ardaguclu I have modified the PR quite a bit from the initial stages. However, I have still kept the commits so that one could have a look.

It now uses a Dynamic client to fetch CSVs. That way the dependency on Operator code could be avoided in the oc repo. I've tried to make little use of generics, since the code is similar. Finally, the unit test initializes more fake clients than earlier.

PTAL and share your feedback.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -412,11 +502,16 @@ func (o *MustGatherOptions) Run() error {
 	defer o.logTimestamp()
 
 	var wg sync.WaitGroup
-	wg.Add(len(pods))
+	w := make(chan struct{}, concurrentMG)
```
ardaguclu, line 505: Why do we need this?. Are we running 4 must-gather pods or are we running 4 different processes in the same must-gather pod?. I think, we can achieve by only using `wg`

pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -5,12 +5,17 @@ import (
 	"reflect"
 	"testing"
 
+	configv1 "github.com/openshift/api/config/v1"
```
ardaguclu, line 8: openshift related dependencies should go after k8s as a separate group

go.mod
```diff
@@ -81,7 +81,7 @@ require (
 	github.com/docker/docker-credential-helpers v0.8.0 // indirect
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/docker/go-metrics v0.0.1 // indirect
-	github.com/emicklei/go-restful/v3 v3.10.1 // indirect
+	github.com/emicklei/go-restful/v3 v3.10.2 // indirect
```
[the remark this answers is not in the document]
dharmit: Nothing in particular. It's a result of adding operator-framework dependencies, running a bunch of `go mod vendor` and `go mod tidy` commands, and then removing the operator-framework discussion after the conversation with you.

Do you prefer I remove it?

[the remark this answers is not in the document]
ardaguclu: > Do you prefer I remove it?

Yes, please. Thanks.

pkg/cli/admin/mustgather/mustgather.go
dharmit: This is because we want to limit the must-gather to startup only 4 pods at a time. https://github.com/openshift/enhancements/pull/1487#discussion_r1402228181

go.mod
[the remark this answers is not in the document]
dharmit: Done

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -15,6 +15,8 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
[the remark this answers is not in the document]
dharmit: Done

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -111,6 +117,7 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 	cmd.Flags().BoolVar(&o.HostNetwork, "host-network", o.HostNetwork, "Run must-gather pods as hostNetwork: true - relevant if a specific command and image needs to capture host-level data")
 	cmd.Flags().StringSliceVar(&o.Images, "image", o.Images, "Specify a must-gather plugin image to run. If not specified, OpenShift's default must-gather image will be used.")
 	cmd.Flags().StringSliceVar(&o.ImageStreams, "image-stream", o.ImageStreams, "Specify an image stream (namespace/name:tag) containing a must-gather plugin image to run.")
+	cmd.Flags().BoolVar(&o.AllImages, "all-images", o.AllImages, "Collect must-gather using the default image for all Operators on the cluster annotated with \"operators.openshift.io/must-gather-image\"")
```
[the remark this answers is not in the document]
dharmit: Thanks, done.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +217,84 @@ func (o *MustGatherOptions) completeImages() error {
 		}
 		o.Images = append(o.Images, image)
 	}
+	if o.AllImages {
+		var pluginImages = map[string]struct{}{}
+
+		// find all csvs and clusteroperators with the annotation "operators.openshift.io/must-gather-image"
+		images, err := annotatedCSVs(o)
```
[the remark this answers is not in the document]
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +217,84 @@ func (o *MustGatherOptions) completeImages() error {
...
+		if err != nil {
+			return err
+		}
+		for k := range images {
+			pluginImages[k] = images[k]
+		}
+
+		images, err = annotatedCOs(o)
```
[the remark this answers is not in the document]
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +217,84 @@ func (o *MustGatherOptions) completeImages() error {
...
+		if err != nil {
+			return err
+		}
+		for k := range images {
+			pluginImages[k] = images[k]
```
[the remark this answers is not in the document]
dharmit: `pluginImages` is a map. So there won't be duplicates. Keys are the image names and values are `struct{}{}`

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +217,84 @@ func (o *MustGatherOptions) completeImages() error {
...
+		Resource: "clusterserviceversions",
+	}
+
+	csvs, err := o.DynamicClient.Resource(csvGVR).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, item := range csvs.Items {
+		img := annotated(&item)
```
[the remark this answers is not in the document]
dharmit: I have removed the generics. But I am not sure if we can use the exact same function. Having a separate `annotatedCSVs` and `annotatedCOs` made more sense to me. Or maybe I'm missing something that you're trying to convey.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +217,84 @@ func (o *MustGatherOptions) completeImages() error {
...
+
+func annotatedCOs(o *MustGatherOptions) (map[string]struct{}, error) {
+	var pluginImages = map[string]struct{}{}
+	cos, err := o.ConfigClient.ConfigV1().ClusterOperators().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, item := range cos.Items {
+		img := annotated(&item)
```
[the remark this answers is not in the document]
dharmit: same

pkg/cli/admin/mustgather/mustgather.go
dharmit: @ardaguclu can you please mark this bit resolved?

pkg/cli/admin/mustgather/mustgather_test.go
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -171,6 +178,31 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+			return fmt.Errorf("--image-streams %s", errStr)
+		}
+		if o.NodeName != "" {
+			return fmt.Errorf("--node-name %s", errStr)
+		}
+		if o.RunNamespace != "" {
+			return fmt.Errorf("--run-namespace %s", errStr)
+		}
+		if o.SourceDir != "/must-gather/" {
+			return fmt.Errorf("--sourec-dir %s", errStr)
```
ardaguclu, line 200:
```suggestion
			return fmt.Errorf("--source-dir %s", errStr)
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -171,6 +178,31 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+		if len(o.ImageStreams) != 0 {
+			return fmt.Errorf("--image-streams %s", errStr)
+		}
+		if o.NodeName != "" {
+			return fmt.Errorf("--node-name %s", errStr)
+		}
+		if o.RunNamespace != "" {
+			return fmt.Errorf("--run-namespace %s", errStr)
+		}
+		if o.SourceDir != "/must-gather/" {
```
ardaguclu, line 199: Why we don't allow changing the source directory?

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -171,6 +178,31 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 	if len(o.DestDir) == 0 {
 		o.DestDir = fmt.Sprintf("must-gather.local.%06d", rand.Int63())
 	}
+	// TODO: this should be in Validate() method, but added here because of the call to o.completeImages() below
+	if o.AllImages {
+		errStr := fmt.Sprintf("and --all-images are mutually exclusive: please specify one or the other")
+		if len(o.Images) != 0 {
```
ardaguclu, line 184: nit: maybe switch looks nicer than `if`s

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -171,6 +178,31 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+		if o.NodeName != "" {
+			return fmt.Errorf("--node-name %s", errStr)
+		}
+		if o.RunNamespace != "" {
+			return fmt.Errorf("--run-namespace %s", errStr)
+		}
+		if o.SourceDir != "/must-gather/" {
+			return fmt.Errorf("--sourec-dir %s", errStr)
+		}
+		if o.VolumePercentage != 30 {
```
ardaguclu, line 202: Why don't we allow changing the volume percentage. On the contrary, I think we should allow it.

pkg/cli/admin/mustgather/mustgather.go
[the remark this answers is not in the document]
ardaguclu: I think this code block can be simplified to something like this;

```go
// find all csvs and clusteroperators with the annotation "operators.openshift.io/must-gather-image"
	pluginImages, err := o.annotatedCSVs()
	if err != nil {
		return err
	}

	images, err = o.annotatedCOs()
	if err != nil {
		return err
	}
	for k := range images {
		pluginImages[k] = struct{}{}
	}

	for image := range pluginImages {
		o.Images = append(o.Images, image)
	}
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +235,78 @@ func (o *MustGatherOptions) completeImages() error {
...
+		for i := range pluginImages {
+			o.Images = append(o.Images, i)
+		}
+	}
 	o.log("Using must-gather plug-in image: %s", strings.Join(o.Images, ", "))
 	return nil
 }
 
+func (o *MustGatherOptions) annotatedCSVs() (map[string]struct{}, error) {
+	var annotation = "operators.openshift.io/must-gather-image"
```
ardaguclu, line 267: can be moved into `const` group at the beginning of the file.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +235,78 @@ func (o *MustGatherOptions) completeImages() error {
...
+	var csvGVR = schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
+		Version:  "v1alpha1",
+		Resource: "clusterserviceversions",
+	}
+	var pluginImages = map[string]struct{}{}
```
ardaguclu, line 273:
```suggestion
	var pluginImages = make(map[string]struct{}{})
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +235,78 @@ func (o *MustGatherOptions) completeImages() error {
...
+
+	csvs, err := o.DynamicClient.Resource(csvGVR).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, item := range csvs.Items {
```
ardaguclu, line 280:
```suggestion
	for _, item := range csvs.Items {
	ann := item.GetAnnotations()
	if v, ok := ann[annotation]; !ok {
	continue
	}
	pluginImages[v] = struct{}{}
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +235,78 @@ func (o *MustGatherOptions) completeImages() error {
...
+		if v, ok := ann[annotation]; ok {
+			pluginImages[v] = struct{}{}
+		} else {
+			continue
+		}
+	}
+	return pluginImages, nil
+}
+
+func (o *MustGatherOptions) annotatedCOs() (map[string]struct{}, error) {
```
ardaguclu, line 291: I mean there are a few lines of code and it doesn't do anything fancy, we can take these lines into the caller and remove `annotatedCOs` function

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: wait group is incremented by the number of pods(which is 4) and we don't need additional channel(`w`)

ardaguclu: We just need a `wg.Wait`

pkg/cli/admin/mustgather/mustgather.go
dharmit: Discussion happened here - https://github.com/openshift/enhancements/pull/1487#discussion_r1402232967. 

If you recommend letting users specify this particular option, I can do so.

pkg/cli/admin/mustgather/mustgather.go
dharmit: OK, I'll remove this check.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Yeah, I had it as a const earlier. But then modified some commits and it got lost in rebases. :)

pkg/cli/admin/mustgather/mustgather.go
dharmit: > wait group is incremented by the number of pods(which is 4) and we don't need additional channel(`w`)

Don't we need the channel to ensure that there are at most 4 pods running at any given time? 🤔 

How do you suggest we ensure that otherwise?

pkg/cli/admin/mustgather/mustgather.go
dharmit: But a `switch` can be useful when checking for different possibility of same variable. Here we are checking for different flags in each `if` statement. I don't think we can replace that with `switch` in this particular case.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Just curious as to what's the reason/benefit in doing so?

ardaguclu: map types should be initialized otherwise they panicks. Actually this one is even better;

```go
pluginImages := make(map[string]struct{}{})
```

pkg/cli/admin/mustgather/mustgather.go
dharmit: How about below instead?
```go
for _, item := range csvs.Items {
    ann := item.GetAnnotations()
    if v, ok := ann[mgAnnotation]; ok {
        pluginImages[v] = struct{}{}
    }
}
```

ardaguclu: That is even better. Thanks.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Yeah, I agree. It just feels cleaner this way. But then you are the maintainer here, so I'll go with your suggestion. :)

pkg/cli/admin/mustgather/mustgather.go
dharmit: @ardaguclu tagging you just in case this slipped

dharmit: @ardaguclu PTAL https://go.dev/play/p/r8IjXNTNtfR. Please run it with and without the channel and let me know your thoughts.

ardaguclu: hmm, now I got the point. Thanks. I wonder is there any simpler solution to always have 4 pods in any time frame because this mechanism looks to me hard to read.

dharmit: [semaphore package](https://pkg.go.dev/golang.org/x/sync/semaphore), maybe. But I, personally, find this simpler than using the semaphore package.

pkg/cli/admin/mustgather/mustgather.go
dharmit: > Why we don't allow changing the source directory?

Based on its help message, I suppose it's not possible to specify a source dir for multiple images? So using it in conjunction with `--all-images` will look for it on all the Pods?

I'm not 100% sure. If you recommend, I'd remove it.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Nice catch! Done.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Removed.

pkg/cli/admin/mustgather/mustgather.go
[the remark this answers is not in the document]
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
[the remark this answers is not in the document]
dharmit: I think we're in agreement on `annotatedCSVs`, `annotatedCOs` and the use of generics. Hence, marking resolved.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Done

pkg/cli/admin/mustgather/mustgather.go
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Thanks. Done.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Removed. :)

pkg/cli/admin/mustgather/mustgather.go
dharmit: Do you think adding a code comment would help? I recently did so for the two new constants being introduced in this PR.

ardaguclu: If it is ok for you I think we can try with semaphore package but I'm not sure as well.

ardaguclu (commented):
Code looks good to me and thanks for the effort. I'll defer to @soltysh @ingvagabund for approval

pkg/cli/admin/mustgather/mustgather.go
dharmit: TBH, I'm OK to try it since it'll be a new thing to learn for me. 😉 

However, I've observed more use of this pattern vs. semaphore package. Semaphore package, I think, shines when we need weighted bounded concurrency. Ours is a use case of bounded concurrency.

If you don't mind, I'd like @soltysh & @ingvagabund to chime in before we decide.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +237,57 @@ func (o *MustGatherOptions) completeImages() error {
 		}
 		o.Images = append(o.Images, image)
 	}
+	if o.AllImages {
+		// find all csvs and clusteroperators with the annotation "operators.openshift.io/must-gather-image"
+		pluginImages := make(map[string]struct{})
+		var err error
+
+		pluginImages, err = o.annotatedCSVs()
```
ingvagabund, line 245: `pluginImages` needs to be extended with `o.Images[0]` just in case the detected images have it duplicated.

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: Alternatively, you can create a queue of pods and have concurrentMG number of workers to process the items in the queue. Which is a well-establish solution and produces a nice looking code. E.g. https://github.com/openshift/library-go/blob/master/pkg/controller/factory/base_controller.go#L109-L119 and https://github.com/openshift/library-go/blob/master/pkg/controller/factory/base_controller.go#L246.

dharmit: @ingvagabund not gonna lie, I understand the worker pool pattern, but I'm a bit hung here.
Do you want me to use the `baseController` implementation that you've linked to, or can I try doing something simpler?
The `baseController` implementation isn't making complete sense to me. Especially, how is it limiting the workers to a certain number. Hence, I'm hung.

ingvagabund: It was an example. Any worker pool pattern works.

pkg/cli/admin/mustgather/mustgather.go
dharmit: IIUC, `--source-dir` flag shouldn't be specified with `--all-images` flag, right?

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: > I was unable to find how/if/where this limit was being set.

passed through `workers` variable

> Looking at baseController it looks like a way to implement/write k8s controllers. I don't think it needs a way to limit workers built into it.

Only the queue and worker loop from the base controller is relevant.

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: https://github.com/openshift/enhancements/pull/1487#discussion_r1402232967 does not forbid that. The statement in https://github.com/openshift/enhancements/pull/1487/files#diff-890eb7a924799156fec30313ddd05cb387f5a9255eedbe00f899639bcff70700R128-R129 is to restrictive wrt. --source-dir. Specifying `--source-dir` with --all-images is fine. Worth updating the enhancement text to allow the flag as well.

dharmit:
@ingvagabund @ardaguclu I think there's a problem with the current PR. According to the enhancement, we want to spin up a limited number of Pods when the user passes multiple `--images` or `--all-images` flag.

Currently, the pod creation is happening in the `for _, image := range o.Images` loop. I think we should defer the pod creation to the goroutines that we'll spin up via worker pool, and only append the result of `o.newPod` to the `pods` slice. WDYT?

ingvagabund:
>  I think we should defer the pod creation to the goroutines 

Yes

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -583,6 +529,65 @@ func (o *MustGatherOptions) Run() error {
 	return errors.NewAggregate(errs)
 }
 
+// podProcessor processes Pods created for collecting each must-gather
+func (o *MustGatherOptions) podProcessor(ns string, podChan chan *corev1.Pod, errCh chan error, wg *sync.WaitGroup) {
```
ingvagabund, line 519: Let's stick with `runWorker` and `processNextWorkItem` naming.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -583,6 +529,65 @@ func (o *MustGatherOptions) Run() error {
...
+	var err error
+	for pod := range podChan {
+		if len(o.RunNamespace) > 0 && !o.Keep {
```
ingvagabund, line 532: Can you put the block under `processNextWorkItem` method? To create something like:
```go
func (o *MustGatherOptions) processNextWorkItem(...) {}

func (o *MustGatherOptions) runWorker(ns string, podChan chan *corev1.Pod, errCh chan error, wg *sync.WaitGroup) {
	for pod := range podChan {
		o.processNextWorkItem(...)
	}
}
```

ingvagabund (changes requested):

pkg/cli/admin/mustgather/mustgather.go
dharmit: Ack.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -533,6 +519,16 @@ func (o *MustGatherOptions) Run() error {
 func (o *MustGatherOptions) podProcessor(ns string, podChan chan *corev1.Pod, errCh chan error, wg *sync.WaitGroup) {
 	var err error
 	for pod := range podChan {
+		_, err = o.Client.CoreV1().Pods(ns).Create(context.TODO(), pod, metav1.CreateOptions{})
+		if err != nil {
+			errCh <- err
+			return
+		}
```
dharmit, line 526: @ingvagabund in "master" branch, we exit immediately if creation of any pod fails. In above code, we return from the goroutine where pod creation fails. I don't think it's accurate because other goroutines would still keep running.

One thing I could think of is to use [`errgroup.Group`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) which will error out if any of the goroutines returns an error.

Another could be to use a `select` statement that would take over the control and stop goroutines (using `context.Cancel`.

What do you think?

ardaguclu: > we return from the goroutine where pod creation fails. I don't think it's accurate because other goroutines would still keep running.

Other goroutines can continue working and we can return the aggregated error at the end. Otherwise, if there is a problematic image, that directly affects the other must-gather images processes.

dharmit: > Other goroutines can continue working and we can return the aggregated error at the end

OK. In that case, I would prefer we don't `return` and simply put the `err` on `errCh`. So that we still have `concurrentMG` number of goroutines running. Basically, just remove `return` from the code I have highlighted.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Done

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -111,6 +119,7 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 	cmd.Flags().BoolVar(&o.HostNetwork, "host-network", o.HostNetwork, "Run must-gather pods as hostNetwork: true - relevant if a specific command and image needs to capture host-level data")
 	cmd.Flags().StringSliceVar(&o.Images, "image", o.Images, "Specify a must-gather plugin image to run. If not specified, OpenShift's default must-gather image will be used.")
 	cmd.Flags().StringSliceVar(&o.ImageStreams, "image-stream", o.ImageStreams, "Specify an image stream (namespace/name:tag) containing a must-gather plugin image to run.")
+	cmd.Flags().BoolVar(&o.AllImages, "all-images", o.AllImages, `Collect must-gather using the default image for all Operators on the cluster annotated with "operators.openshift.io/must-gather-image"`)
```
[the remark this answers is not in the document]
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Done. PTAL if you find the approach OK. I am deleting the key from the map before we start appending:
```go
// delete the default image to avoid duplication in case an Operator had it in its annotation
delete(pluginImages, o.Images[0])
for i := range pluginImages {
    o.Images = append(o.Images, i)
}
```

pkg/cli/admin/mustgather/mustgather.go
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -495,6 +513,79 @@ func (o *MustGatherOptions) Run() error {
 	return errors.NewAggregate(errs)
 }
 
+// runWorker processes Pods created for collecting each must-gather
+func (o *MustGatherOptions) runWorker(ns string, podChan chan *corev1.Pod, errCh chan error, wg *sync.WaitGroup) {
+	for pod := range podChan {
+		o.processNextWorkItem(ns, pod, errCh)
```
ingvagabund, line 519: Better to return error from `o.processNextWorkItem` and add it to `errCh` here.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -412,65 +479,16 @@ func (o *MustGatherOptions) Run() error {
 	defer o.logTimestamp()
 
 	var wg sync.WaitGroup
-	wg.Add(len(pods))
 	errCh := make(chan error, len(pods))
+	podChan := make(chan *corev1.Pod)
+	wg.Add(concurrentMG)
```
ingvagabund, line 484: `wg` is no longer needed.

dharmit: Why? How do you suggest we wait for the goroutines to complete in that case?

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -203,10 +234,58 @@ func (o *MustGatherOptions) completeImages() error {
...
+			return err
+		}
+		for _, item := range cos.Items {
+			ann := item.GetAnnotations()
+			if v, ok := ann[mgAnnotation]; ok {
+				pluginImages[v] = struct{}{}
+			}
+		}
+		// delete the default image to avoid duplication in case an Operator had it in its annotation
+		delete(pluginImages, o.Images[0])
```
ingvagabund, line 258: What about the case where the default image is not added through annotations?

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: Deleting will not cover the case where the default image is not added through annotation.

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: You are right. Please ignore my comment.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Sorry, I don't follow you. By "default image", are we both referring to below:
```go
image = "registry.redhat.io/openshift4/ose-must-gather:latest"
```
If yes, the only place I see it being added to `o.Images` is in the block above:
```go
if len(o.Images) == 0 || o.AllImages
```
If none of the annotations are adding the same image, the `delete` operation would delete nothing. Are you suggesting adding a check like:
```go
if _, ok := pluginImages[o.Images]; ok {
    delete(pluginImages, o.Images[0])
}
```

pkg/cli/admin/mustgather/mustgather.go
dharmit: Ack, marking resolved.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Marking this long discussion resolved since worker pool pattern is implemented. Request you to start a new discussion if you have any feedback around this part of the code. 😆

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: Oh, I see. That works as well.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Marking resolved based on https://github.com/openshift/oc/pull/1633#pullrequestreview-1818164255.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -412,65 +479,16 @@ func (o *MustGatherOptions) Run() error {
 	defer o.logTimestamp()
 
 	var wg sync.WaitGroup
-	wg.Add(len(pods))
 	errCh := make(chan error, len(pods))
+	podChan := make(chan *corev1.Pod)
+	wg.Add(concurrentMG)
+	for i := 0; i < concurrentMG; i++ {
```
ardaguclu, line 485: I'm not comfortable with the way we handle this concurrency pattern and I'd like to spend more time to investigate this. Meanwhile, could you please explain;

For example, we have 5 images in our slice and according to this loop we have processed the first 4 in a separate goroutine. Where is iterating to process the 5th image?.

ardaguclu: I think, we can achieve always running the max number of pods at the same time by using the buffered channels. Please let me know what you think https://go.dev/play/p/fv6cGIqxG3W

ardaguclu: Something like this;
```go
...
concurrentChan := make(chan struct{}, concurrentMG)
for _, pod := range pods {
wg.Add(1)
go func(pod *corev1 Pod({
   defer wg.Done()
   concurrentChan <- struct{}{}
   // runWorker
   <-concurrentChan
})
wg.Wait()
close(errCh)
...
```

dharmit: > I'm not comfortable with the way we handle this concurrency pattern and I'd like to spend more time to investigate this.

Please let me know if I could do anything to help here.



> For example, we have 5 images in our slice and according to this loop we have processed the first 4 in a separate goroutine. Where is iterating to process the 5th image?.

The iteration is happening in:
```go
for _, pod := range pods {
    podChan <- pod
}
```

> I think, we can achieve always running the max number of pods at the same time by using the buffered channels. Please let me know what you think https://go.dev/play/p/fv6cGIqxG3W

Yes, we can. And it's _almost_ the same as what I had earlier except that I had a maximum `concurrentMG` number of goroutines running at any given time, while your suggestion has `len(pods)` number of goroutines spinning up immediately and only `concurrentMG` number of goroutines being processed at a given time. If we go with semaphore pattern, I'd prefer former against latter.

However, @ingvagabund suggested implementing worker pool pattern (which you see in the code right now).

I'd prefer either the semaphore pattern I had earlier or this worker pool pattern, since it spins up maximum `concurrentMG` goroutines at any given time. But I'll let you folks decide and let me know which one you are comfortable with maintaining.

dharmit: > which code portion runs the 5th image;
> 
> ```go
> for _, pod := range pods {
>     podChan <- pod # who is sending the pod for 5th image?.
> }
> ```

OK, let me try. The two code blocks below are adding the number of Pods to be created and processed:
```go
pods = append(pods, o.newPod(node.Name, image, hasMaster))
pods = append(pods, o.newPod(o.NodeName, image, hasMaster))
```

So if there are 5/10/20 pods, all of them are being put on the `podChan` for processing. Some code needs to read from this `podChan` and then do the processing.

> Because iteration in pod creation is limited by concurrentMG;
> 
> ```go
> for i := 0; i < concurrentMG; i++ {
> ```

This code is now only limiting the number of worker goroutines, not the number of pods to spin up. Within this `for` loop, we call `o.runWorker`. And within `o.runWorker`, we read from the above `podChan` till the channel is empty. So the 5th Pod that you're talking about gets processed by `runWorker` method because it reads the channel till there's nothing left in it to read.

dharmit: PTAL at this sample program of smaller size and see if it makes sense - https://go.dev/play/p/_jNPiZilviv

ingvagabund: @ardaguclu if I read your example in https://go.dev/play/p/fv6cGIqxG3W correctly it runs 7 goroutines but only 4 of them do an actual work of processing pods.

ingvagabund: Adding `fmt.Printf("Running %v goroutine\n", idx)` line into https://go.dev/play/p/fv6cGIqxG3W example through https://go.dev/play/p/7M9ATLUTkjk:
```
Running 6 goroutine
Running 2 goroutine
Running 0 goroutine
Running 1 goroutine
Running 4 goroutine
Running 3 goroutine
Running 5 goroutine
chan is released 2009-11-10 23:00:01 +0000 UTC m=+1.000000001
Function 1
chan is released 2009-11-10 23:00:01 +0000 UTC m=+1.000000001
Function 7
chan is released 2009-11-10 23:00:01 +0000 UTC m=+1.000000001
Function 3
chan is released 2009-11-10 23:00:01 +0000 UTC m=+1.000000001
Function 2
chan is released 2009-11-10 23:00:02 +0000 UTC m=+2.000000001
Function 5
chan is released 2009-11-10 23:00:02 +0000 UTC m=+2.000000001
Function 4
chan is released 2009-11-10 23:00:02 +0000 UTC m=+2.000000001
Function 6
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -412,65 +479,16 @@ func (o *MustGatherOptions) Run() error {
...
+		go o.runWorker(ns.Name, podChan, errCh, &wg)
+	}
 	for _, pod := range pods {
```
ingvagabund, line 488: Can you move
```
	for _, pod := range pods {
		podChan <- pod
	}
	close(podChan)
```
before running
```
	for i := 0; i < concurrentMG; i++ {
		go o.runWorker(ns.Name, podChan, errCh, &wg)
	}
```
? This will make the code easier to read.

dharmit: Sure

dharmit: Done.

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: https://go.dev/play/p/F0pzLGL878- for showing multiple goroutines looping over the same channel:
```
Spinning up a goroutine...
Spinning up a goroutine...
(1) 0
(0) 1
(0) 2
(1) 3
(0) 4
Done
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -412,64 +479,17 @@ func (o *MustGatherOptions) Run() error {
...
-			}
-			if err := o.copyFilesFromPod(pod); err != nil {
-				log("gather output not downloaded: %v\n", err)
-				errCh <- fmt.Errorf("unable to download output from pod %s: %s", pod.Name, err)
-				return
-			}
-		}(pod)
+	wg.Add(concurrentMG)
+	for i := 0; i < concurrentMG; i++ {
+		go o.runWorker(ns.Name, podChan, errCh, &wg)
```
ingvagabund, line 492: Nit: given the current implementation does not use a queue this could be inlined to avoid passing the arguments:
```
for i := 0; i < concurrentMG; i++ {
	go func() {
		for pod := range podChan {
			if err := o.processNextWorkItem(ns.Name, pod); err != nil {
				errCh <- err
			}
		}
		wg.Done()
	}()
}
```

dharmit: Makes sense. May I ask what's the difference between a queue pattern and worker pool pattern? My web search-fu made me believe that they are different terms for the same thing. Feel free to share a link to RTFM. :)

ingvagabund: My original intention was to use https://github.com/openshift/library-go/blob/964bcb3f545c24f15294fd8ba914529bf4fe8c4d/pkg/controller/factory/base_controller.go#L246 for popping the pods/items. So in case there's e.g. an error and it make sense to re-queue the item back, we can use https://github.com/openshift/library-go/blob/964bcb3f545c24f15294fd8ba914529bf4fe8c4d/pkg/controller/factory/base_controller.go#L271. Though, this can be added later once we have a case for it.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Marking this resolved since we're going with different approach now.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -492,26 +493,30 @@ func (o *MustGatherOptions) Run() error {
...
-	go func() {
-		for _, pod := range pods {
-			podChan <- pod
-		}
-		close(podChan)
-	}()
+	for _, pod := range pods {
+		queue.Add(pod)
+	}
+	queue.ShutDownWithDrain()
```
dharmit, line 504: @ingvagabund @ardaguclu I'm not sure if this is the right thing to do. But without doing this, the below call wait indefinitely because the queue is 1) empty 2) not marked for shutdown:
```go
pod, quit := queue.Get()
```

ingvagabund: Did you try `queue.Len()` with mutexes?

ingvagabund: `Len()` uses locks (https://github.com/openshift/oc/blob/master/vendor/k8s.io/client-go/util/workqueue/queue.go#L187) so no need for mutex.

ingvagabund:
```
// ShutDownWithDrain will cause q to ignore all new items added to it. As soon
// as the worker goroutines have "drained", i.e: finished processing and called
// Done on all existing items in the queue; they will be instructed to exit and
// ShutDownWithDrain will return. Hence: a strict requirement for using this is;
// your workers must ensure that Done is called on all items in the queue once
// the shut down has been initiated, if that is not the case: this will block
// indefinitely. It is, however, safe to call ShutDown after having called
// ShutDownWithDrain, as to force the queue shut down to terminate immediately
// without waiting for the drainage.
```
actually looks ok to use :) That's what we want.

dharmit:
Looks like Azure issue maybe? Triggering retest to see if it's a flake or real.
```
clusteroperator/storage is not available (AzureDiskCSIDriverOperatorCRAvailable: AzureDiskDriverControllerServiceControllerAvailable: Waiting for Deployment
AzureFileCSIDriverOperatorCRAvailable: AzureFileDriverControllerServiceControllerAvailable: Waiting for Deployment) because AzureDiskCSIDriverOperatorCRDegraded: All is well
AzureFileCSIDriverOperatorCRDegraded: All is well

error: gather did not start for pod must-gather-xgwrz: timed out waiting for the condition 
```

/test e2e-agnostic-ovn-cmd

Looks like infra/network issue?
```
 + timeout 70s curl --retry-all-errors --retry-delay 60 --retry 1 -kfX POST -H 'X-OFCIRTOKEN: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX' 'https://ofcir-service.ofcir-system.svc.cluster.local/v1/ofcir?name=pull-ci-openshift-oc-master-e2e-metal-ipi-ovn-ipv6/1747989513453441024&type=host' -o /tmp/secret/cir
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0    27    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
Warning: Problem (retrying all errors). Will retry in 60 seconds. 1 retries 
Warning: left.

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0    27    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
+ return 1
+ exit_with_failure 'Failed to create ci resource: ipi-ci-op-k3hrpdk0-0e466-1747989513453441024' 
```

/test e2e-metal-ipi-ovn-ipv6

pkg/cli/admin/mustgather/mustgather.go
dharmit: Marking resolved since I have implemented using the suggested approach.

dharmit:
> we need a pre-merge testing?

Do you mean taking it for a spin against a cluster with a bunch of ClusterOperators/CSVs annotated with the annotation we're introducing in this PR?

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"k8s.io/client-go/util/workqueue"
```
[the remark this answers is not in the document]
dharmit: Ack, I'll correct this and sqaush the first two commits into one.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -425,65 +493,31 @@ func (o *MustGatherOptions) Run() error {
 	}
 	defer o.logTimestamp()
 
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
```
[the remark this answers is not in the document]
dharmit: I definitely spent non-trivial time in understanding it and learning more about [`DefaultControllerRateLimiter`](https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/). Not complaining. I learnt something. ;)

pkg/cli/admin/mustgather/mustgather.go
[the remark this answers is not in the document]
dharmit: Fixed.

pkg/cli/admin/mustgather/mustgather.go
dharmit: Marking resolved as it looks like we agree with the approach here.

ardaguclu:
This looks good to me. Thanks for the time and effort you spent on this issue;
/approve

for the final tag, I'm deferring to @ingvagabund

ardaguclu:
Isn't there any Jira item to link to this issue for the tracking purposes?.

dharmit:
> Isn't there any Jira item to link to this issue for the tracking purposes?.

We're tracking this on https://issues.redhat.com/browse/CNV-37284.

ardaguclu:
Would it make sense to add `CNV-37284: ` as a prefix in PR title to get a valid-jira label?.

dharmit:
> Would it make sense to add `CNV-37284: ` as a prefix in PR title to get a valid-jira label?.

Absolutely. Done.

ingvagabund:
/hold
for pre-merge testing

PR #1636, opened 2023-12-15
OTA-1087: Add upgade health section to `adm update status`

petr-muller:
Add a section with upgrade health insights to `adm update status` output, and as a proof-of-concept, populate it with insights about degraded and unavailable cluster operators during the upgrade.

The CO insights do not report when a CO is in a problematic state for a short amount of time (5m as an initial value), then a warning insight is show for them, and once the condition persists over a higher threshold (20m for unavailable and 40m for degraded), the insight is bumped to error.

The names, structure and implementation generally follow the concepts and ideas from the "OpenShift Update Concepts" slides deck, and the update insights data structure follows [Justin's design](https://github.com/jupierce/oc-update-design-ideas/blob/main/status.md#get-hooksconfigopenshiftiov1alphaupdateinsights), with minor modifications (and also not fully implemented, some parts are not needed yet).


pkg/cli/admin/upgrade/status/controlplane.go
```diff
@@ -30,9 +31,52 @@ type controlPlaneStatusDisplayData struct {
 	Operators  operators
 }
 
-func assessControlPlaneStatus(cv *v1.ClusterVersion, operators []v1.ClusterOperator, at time.Time) controlPlaneStatusDisplayData {
+const (
+	unavailableWarningThreshold = 5 * time.Minute
+	unavailableErrorThreshold   = 20 * time.Minute
+	degradedWarningThreshold    = 5 * time.Minute
```
LalatenduMohanty, line 37: I would suggest to wait for longer initially , example 15 mins. If the status is still degraded we can consider this as warning. In future if we feel that the warning should be given after 5 mins , we can do that. Having a longer time for the first iteration will reduce false positives.

pkg/cli/admin/upgrade/status/health.go
```diff
@@ -0,0 +1,144 @@
...
+func (l impactLevel) String() string {
+	switch l {
+	case infoImpactLevel:
+		return "Info"
+	case warningImpactLevel:
+		return "Warning"
+	case errorImpactLevel:
+		return "Error"
+	case criticalInfoLevel:
+		return "Critical"
```
LalatenduMohanty, line 53: What is the difference between error and critical? We also need go doc for each type. It is almost like an API because once we define these we can not change it significantly. Users will setup automation or monitoring system based on these.

pkg/cli/admin/upgrade/status/health.go
```diff
@@ -0,0 +1,144 @@
...
+
+const (
+	noneImpactType                    impactType = "None"
+	apiAvailabilityImpactType         impactType = "API Availability"
+	clusterCapacityImpactType         impactType = "Cluster Capacity"
+	applicationAvailabilityImpactType impactType = "Application Availability"
+	applicationOutageImpactType       impactType = "Application Outage"
+	dataLossImpactType                impactType = "Data Loss"
+	updateSpeedImpactType             impactType = "Update Speed"
+	updateStalledImpactType           impactType = "Update Stalled"
```
LalatenduMohanty, line 69: These looks good but need to think about this more and we should get this reviewed by someone from API team. We should review this with Justin and Scott as well.

LalatenduMohanty (changes requested):

pkg/cli/admin/upgrade/status/controlplane.go
petr-muller: I think 5m is enough to smooth over useless blips and if somehting persists over 5m then I think it is worth showing..

Being more sensitive (while smoothing over blips) is better for demos and getting feedback (people will feedback on what they see, not as much on what they miss) and it is generally easier to notice that we are still too sensitive (and smooth over more) than noticing that we would like more output should be actually smoothing less.

pkg/cli/admin/upgrade/status/health.go
petr-muller: I basically followed common logging levels where you usually have one level above error. I imagine critical is something where the cluster or data is actually at risk of being destroyed or lost. I'm not married to this. We are not committed to the client-side "api" but I guess adding notes about what we, as implementors, intend to use these levels for is useful, will add.

pkg/cli/admin/upgrade/status/health.go
petr-muller: These come from Justin's slide deck, slide 18, "Health Nofifications". So we can consider them reviewed I thing :)

pkg/cli/admin/upgrade/status/health.go
petr-muller: Added a commit with godocs

pkg/cli/admin/upgrade/status/examples/4.14.1-degraded.output
```diff
@@ -0,0 +1,21 @@
...
...fe: etcd cluster has quorum of 2 and 2 healthy members which is not fault tolerant: [{Member:ID:12895393557789359222 name:"ip-10-0-73-118.ec2.internal" peerURLs:"https://10.0.73.118:2380" clientURLs:"https://10.0.73.118:2379"  Healthy:true Took:1.725492ms Error:<nil>} {Member:ID:13608765340770574953 name:"ip-10-0-0-60.ec2.internal" peerURLs:"https://10.0.0.60:2380" clientURLs:"https://10.0.0.60:2379"  Healthy:true Took:1.542919ms Error:<nil>} {Member:ID:18044478200504924924 name:"ip-10-0-12-74.ec2.internal" peerURLs:"https://10.0.12.74:2380" clientURLs:"https://10.0.12.74:2379"  Healthy:false Took: Error:create client failure: failed to make etcd client for endpoints [https://10.0.12.74:2379]: context deadline exceeded}] // EtcdMembersDegraded: 2 of 3 members are available, ip-10-0-12-74.ec2.internal is unhealthy // NodeControllerDegraded: The master nodes not ready: node "ip-10-0-12-74.ec2.internal" not ready since 2023-11-03 16:28:43 +0000 UTC because KubeletNotReady (container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: No CNI configuration file in /etc/kubernetes/cni/net.d/. Has your network provider started?)
```
petr-muller, line 20: This shows that simply taking CO condition's `message`  is not the most friendly UX we can have, I think we need a followup to define whether there are constraints on the insights, such as max length, or whether the presentation should try to shorten (ellipsize? summarize?) these messages somehow. This PR is already large, so I think that needs to be a followup, rather than an item to be done right now.

pkg/cli/admin/upgrade/status/controlplane.go
LalatenduMohanty: > Being more sensitive (while smoothing over blips) is better for demos and getting feedback (people will feedback on what they see, not as much on what they miss) and it is generally easier to notice that we are still too sensitive (and should smooth over more) than noticing that we would like more output (and should be actually smoothing less).

I agree that for demos and getting feedback/testing 5m is better. However when we GA  this to customer 15 min might be a better frequency IMO. Mostly because we do not want customers get too worked up seeing changes in every 5 mins. Also we will learn somethings when we test this out in real life scenarios. So I am fine with 5mins for now.

pkg/cli/admin/upgrade/status/health.go
LalatenduMohanty: The slide deck was mostly for suggesting a new design which can be used to solve this. Also the purpose of the slide deck is to get feedback from stakeholders. I do not think folks reviewed it from a point of view of whether that is technically accurate etc . @jupierce @sdodson you might be interested to review this.

petr-muller: Is it worth blocking the PR though? I used the list because taking it from Justin's slides was easy. I can capture your concern as a followup task but at this point I'd rather remove the consts (code only uses "API availability" and "None") than become blocked on their review. I think that changing this list is still easy enough, this work focuses more on creating the client side prototype we can learn from, at this point we're not creating the full-fledged API yet.

LalatenduMohanty (approved):
/lgtm

PR #1642, opened 2023-12-19
OTA-492: pkg/cli/admin/upgrade/rollback: New rollback command

wking:
The "which release to roll back to?" criteria are matched to openshift/cluster-version-operator@12fdd0456b
(openshift/cluster-version-operator#996), where, based on [OTA-553](https://issues.redhat.com/browse/OTA-553), Lala, Scott, and Subin all agreed that rollbacks to the cluster's previous release should not be blocked, as long as the previous and current version had the same `major.minor` version (i.e. "within a z-stream", using `x.y.z` naming for `major.minor.patch`).  We agreed to ignore whether the previous history entry is `Completed` or `Partial`, expecting that most users who are interested in rolling back will make that decision at the first sign of trouble, and not after attempting a few roll-forward retargets.

I'm currently blocking `Progressing!=False` from rolling back, to align with our CI coverage in openshift/release@f4d37e7f87 (openshift/release#45115).  We could lift this restriction if we wanted to allow rollbacks to be requested from the middle of an in-progress update, but if we do, we'd probably want to add matching CI to cover that use-case.

The new subcommand is hidden, as Lala requested in [OTA-492](https://issues.redhat.com/browse/OTA-492)'s `Description` on 2021-09-22.  I have not put it behind a feature-gate environment variable in this commit, but that's also an option if we wanted to make it even less reachable.

### DIFF
--- a/pkg/cli/admin/upgrade/rollback/rollback.go
+++ b/pkg/cli/admin/upgrade/rollback/rollback.go
@@ -1,0 +1,165 @@
+// Package rollback initiates a rollback to a previous release.
+package rollback
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/blang/semver"
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


petr-muller (approved):
LGTM

/hold
Holding to allow more reviews or merging after the break, if needed. Feel free to unhold to merge.

wking:
This is not a rush, so merging post-break would be fine.  But on the other hand, if this looks good to you, I expect any further feedback would turn up changes that could use "I'm making this small pivot because..." commit messages to explain the nuance.  And new-pull and reroll-this-pull both have the same CI cost.  So I'll merge, and we can tweak later during the 4.16 dev cycle if necessary.

/hold cancel

PR #1643, opened 2023-12-20
OCPBUGS-25703: validate tag name on creation

flavianmissi:
skip validation for deletion as existing tags might have been created with an invalid name and users must be able to delete them.


ardaguclu:
Sounds reasonable. Thanks.

/lgtm
/retest-required

PR #1653, opened 2024-01-16
WRKLDS-1010: pkg/cli/admin/inspect: use since/since-time for rotated logs

flavianmissi:
this solution is not very robust, but rotated logs can considerably increase the size of the gathered logs, so we do our best to support filtering them as well.

---

I only managed to test this outside the context of a must-gather, since the must-gather will use the `oc` binary present in its image. Follow these steps to test it:

```
go run ./cmd/oc adm inspect --rotated-pod-logs --since-time=2024-01-10T10:00:00Z ns/openshift-cluster-version
ll inspect.local.4575015416930189447/namespaces/openshift-cluster-version/pods/cluster-version-operator-6d799d6cc6-zwbcb/cluster-version-operator/cluster-version-operator/logs/rotated
# inspect files and make sure they do not include dates before the provided --since-time.
# repeat for --since
```

Here's an example from one of my runs on openshift local running with `--since-time=2024-01-10T10:00:00Z`
```
inspect.local.2059246489365228252/namespaces/openshift-cluster-version/pods/cluster-version-operator-6d799d6cc6-zwbcb/cluster-version-operator/cluster-version-operator/logs/
├── current.log
├── previous.insecure.log
├── previous.log
└── rotated
    ├── 4.log.20240112-031433.gz
    ├── 4.log.20240112-065352.gz
    ├── 4.log.20240112-103153.gz
    ├── 4.log.20240112-141052
    ├── 5.log.20240115-235129.gz
    ├── 5.log.20240116-021747.gz
    ├── 5.log.20240116-044405.gz
    └── 5.log.20240116-071013

2 directories, 11 files
```
And without `--since-time`:
```
tree inspect.local.2335573221674443086/namespaces/openshift-cluster-version/pods/cluster-version-operator-6d799d6cc6-zwbcb/cluster-version-operator/cluster-version-operator/logs
inspect.local.2335573221674443086/namespaces/openshift-cluster-version/pods/cluster-version-operator-6d799d6cc6-zwbcb/cluster-version-operator/cluster-version-operator/logs
├── current.log
├── previous.insecure.log
├── previous.log
└── rotated
    ├── 3.log.20240109-035321.gz     <-- skipped when filtered by the 10th!
    ├── 3.log.20240109-073120.gz     <-- skipped when filtered by the 10th!
    ├── 3.log.20240109-110927.gz     <-- skipped when filtered by the 10th!
    ├── 3.log.20240109-144731        <-- skipped when filtered by the 10th!
    ├── 4.log.20240112-031433.gz
    ├── 4.log.20240112-065352.gz
    ├── 4.log.20240112-103153.gz
    ├── 4.log.20240112-141052
    ├── 5.log.20240115-235129.gz
    ├── 5.log.20240116-021747.gz
    ├── 5.log.20240116-044405.gz
    └── 5.log.20240116-071013

2 directories, 15 files
```

### DIFF
--- a/pkg/cli/admin/inspect/pod.go
+++ b/pkg/cli/admin/inspect/pod.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"time"
 
 	"golang.org/x/net/html"
 
@@ -126,6 +127,18 @@ func (o *InspectOptions) gatherContainerRotatedLogFiles(destDir string, pod *cor
 		return err
 	}
 
+	// when sinceTime is given we use that to compare the log file name with
+	// the provided date.
+	// when since is given we subtract the given duration from time.Now(),
+	// then use that as the requested time.
+	var requestedTime time.Time
+	if len(o.sinceTime) > 0 {
+		requestedTime = o.sinceTimestamp.Time
+	}
+	if o.since != 0 {
+		requestedTime = time.Now().Add(-o.since)
+	}
+
 	// rotated log files have a suffix added at the end of the file name
 	// e.g: 0.log.20211027-082023, 0.log.20211027-082023.gz
 	reRotatedLog := regexp.MustCompile(`[0-9]+\.log\..+`)
@@ -140,6 +153,24 @@ func (o *InspectOptions) gatherContainerRotatedLogFiles(destDir string, pod *cor
 			}
 			if !reRotatedLog.MatchString(fileName) {
 				return
+			}
+
+			if !requestedTime.IsZero() {
+				// make sure rotated logs honor --since and --since-time
+				// when set by the user.
+				if parts := strings.Split(fileName, "."); len(parts) >= 3 {
+					if rotatedTime, err := time.Parse("20060102-150405", parts[2]); err == nil {
+						if rotatedTime.Before(requestedTime) {
+							klog.V(4).Infof(
+								"skipping rotated logs for %q because it's outside of user-provided time constraints",
+								fileName,
+							)
+							return
+						}
+					} else {
+						innerErrs = append(innerErrs, err)
+					}
+				}
 			}
 
 			// ensure destination dir exists
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -117,7 +117,7 @@ func NewCmdInspect(streams genericiooptions.IOStreams) *cobra.Command {
 	cmd.Flags().BoolVarP(&o.allNamespaces, "all-namespaces", "A", o.allNamespaces, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
 	cmd.Flags().StringVar(&o.sinceTime, "since-time", o.sinceTime, "Only return logs after a specific date (RFC3339). Defaults to all logs. Only one of since-time / since may be used.")
 	cmd.Flags().DurationVar(&o.since, "since", o.since, "Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs. Only one of since-time / since may be used.")
-	cmd.Flags().BoolVar(&o.rotatedPodLogs, "rotated-pod-logs", o.rotatedPodLogs, "Experimental: If present, retrieve rotated log files that are available for selected pods. This can significantly increase the collected logs size. since/since-time is ignored for rotated logs.")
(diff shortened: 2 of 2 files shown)


pkg/cli/admin/inspect/pod.go
```diff
@@ -126,6 +127,18 @@ func (o *InspectOptions) gatherContainerRotatedLogFiles(destDir string, pod *cor
 		return err
 	}
 
+	// when sinceTime is given we use that to compare the log file name with
+	// the provided date.
+	// when since is given we subtract the given duration from time.Now(),
+	// then use that as the requested time.
+	requestedTime := time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)
```
ardaguclu, line 134: nit: What about? and check that `requestedTime.IsZero()` in rotated log filtering below?

```suggestion
	var requestedTime time.Time
	if len(o.sinceTime) > 0 {
	       // when sinceTime is given we use that to compare the log file name with
	        // the provided date.
		requestedTime = o.sinceTimestamp.Time
	}
	if o.since != 0 {
		// when since is given we subtract the given duration from time.Now(),
	        // then use that as the requested time.
		requestedTime = time.Now().Add(-o.since)
	}
```

pkg/cli/admin/inspect/pod.go
```diff
@@ -142,6 +155,19 @@ func (o *InspectOptions) gatherContainerRotatedLogFiles(destDir string, pod *cor
 				return
 			}
 
+			// make sure rotated logs honor --since and --since-time
+			// when set by the user.
+			if parts := strings.Split(fileName, "."); len(parts) >= 3 {
```
ardaguclu, line 160: I wonder about can we extract the dates via plain regex instead of `Split`s?

pkg/cli/admin/inspect/pod.go
```diff
@@ -142,6 +155,19 @@ func (o *InspectOptions) gatherContainerRotatedLogFiles(destDir string, pod *cor
...
+				datePart := strings.Split(parts[2], "-")[0]
+				if rotatedTime, err := time.Parse("20060102", datePart); err == nil {
+					if rotatedTime.Before(requestedTime) {
+						return
```
ardaguclu, line 164: Do we need a high verbosity log about why users can not find some rotated logs, when they run must-gather?

pkg/cli/admin/inspect/pod.go
flavianmissi: that is indeed more elegant, thanks - updated.

pkg/cli/admin/inspect/pod.go
flavianmissi: that was my initial approach but I could not easily produce a regex that matched only the date part in the filename (there is more than one numeric part to it and I'm not very good with regex). 
if you're willing to help me out we can definitely change it.

pkg/cli/admin/inspect/pod.go
flavianmissi: Good idea, added.

pkg/cli/admin/inspect/pod.go
ardaguclu: What about something like this?;
```go
re := regexp.MustCompile(`\.(\d+-\d+)\.gz$`)
match := re.FindStringSubmatch(filename)

if len(match) != 2 {
	return "", fmt.Errorf("No match found")
}

return match[1], nil
```

flavianmissi: Thanks for the regex! nice to know how it can be done.
I agree with you about keeping the `Split`, it kinda looks dumb but that can be a strength (easy to understand).

pkg/cli/admin/inspect/pod.go
```diff
@@ -142,6 +155,25 @@ func (o *InspectOptions) gatherContainerRotatedLogFiles(destDir string, pod *cor
 				return
 			}
 
+			if !requestedTime.IsZero() {
+				// make sure rotated logs honor --since and --since-time
+				// when set by the user.
+				if parts := strings.Split(fileName, "."); len(parts) >= 3 {
+					datePart := strings.Split(parts[2], "-")[0]
+					if rotatedTime, err := time.Parse("20060102", datePart); err == nil {
+						if rotatedTime.Before(requestedTime) {
```
ardaguclu, line 164: Last question, we only filter based on dates, not time in `strings.Split(parts[2], "-")[1]`?

flavianmissi: good point - I didn't realize it before but now that you mention it the part after the `"-"` is the time right? I can change the parse layout so that it takes that into consideration as well.

flavianmissi: fixed

ardaguclu:
Seems unrealated
/hold cancel

PR #1654, opened 2024-01-16
NO-JIRA: adm update status: add TODOs to remind us to revise early decisions

petr-muller:
This client-side code is early stage experimental work we do mostly to obtain more knowledge about how exactly an eventual API should look like. Some decisions are made without the scrutiny we would give to a proper API. Take notes, in the form of TODOs, of these decisions, so that we remember to revise them when we start working server-side.


wking (approved):
/lgtm

PR #1659, opened 2024-01-23
OCPBUGS-27489: Use correct hash sum of extracted tarball after re-signing with codesign

ardaguclu:
When binaries are extracted via `oc adm release extract` command, we have to re-sign these binaries if the operating system is `darwin` and architecture is `arm64`. This manual re-signing process seems to be working well, at least we haven't gotten any issues so far. However, since we make modifications to tarball during the re-signing process, this changes the hash sum value of the tarball, whereas the value in sha256sum file refers to the hash sum of the tarball prior to re-sign process creates disperancy.

This PR writes modified tarball's hash sum value in sha256sum file to prevent mismatches.

This PR can be considered as a continuation of https://github.com/openshift/oc/pull/1242


ardaguclu:
/cc @ingvagabund 
Could you please take a look at, when you are available?. Thanks.

ardaguclu:
not related;
/retest-required

PR #1656, opened 2024-01-19
NO-JIRA: Consider kubectl version checks only from cli and cli-artifact images

ardaguclu:
Currently, when we bump oc to newer k8s versions, our CI jobs are failing due to kubectl version mismatch error. The reason is that when we bump oc to newer kubectl(i.e. 1.29.0), cli and cli-artifact images manifest this version. On the other hand, other images in other operators have not-bumped kubectl version(i.e. 1.28.2).

This PR only populates kubectl versions in cli and cli-artifact images(because they are the source of truth) but in addition to that this PR checks the other images as well to have a consistency of a rule that all images should match either kubectl version in cli, cli-artifact or match with each other (all versions in other images at least should equal to 1.28.2).

<details>
<summary>valid image sets</summary>

```
cli 1.29.0
cli-artifact 1.29.0
others all 1.28.2
```
```
cli 1.29.0
cli-artifact 1.29.0
others all 1.29.0
```
```
cli 1.29.0
cli-artifact 1.29.0
others 1.28.2
another one  1.29.0
```
</details>

<details>
<summary>invalid image sets</summary>

```
cli 1.29.0
cli-artifact 1.29.0
others 1.28.2
another one 1.27.3
```
```
cli 1.29.0
cli-artifact 1.28.2
others all 1.28.2
```
</details>

This PR is prerequisite of https://github.com/openshift/oc/pull/1655

PR #1661, opened 2024-01-24
WRKLDS-1016: Bump kube dependencies to 1.29.0

ardaguclu:
This PR bumps k8s dependencies to 1.29.0 for OCP 4.16.

PR #1662, opened 2024-01-24
NO-JIRA: Simplified kubectl component version check

ardaguclu:
With this PR https://github.com/openshift/oc/pull/1656, we relaxed the kubectl component version
check by only allowing kubectl version in cli and cli-artifacts images and kubectl version between 
other images. However, as seen in this revert https://github.com/openshift/oc/pull/1660, there are other oc commands
having this strict component version check which caused failures in HCP.

This PR adopts simpler approach for kubectl specific component version checking. It simply
populates all the kubectl versions in a set and if the length is greater than 2(actual kubectl version 1.29.0 and the current ones 1.28.2) which means that this is indeed kubectl version mismatch.

### DIFF
--- a/pkg/cli/admin/release/image_mapper.go
+++ b/pkg/cli/admin/release/image_mapper.go
@@ -10,11 +10,13 @@ import (
 	"sort"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+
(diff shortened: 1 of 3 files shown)


ardaguclu:
For more justification: https://github.com/openshift/oc/pull/1661#issuecomment-1907650866

soltysh (approved):
/lgtm

PR #1663, opened 2024-01-25
NO-JIRA: pkg/cli/admin/inspect: format RFC3339 parse error

flavianmissi:
Ideally this would be done in the `Validate` method, but `Complete` is invoked before `Validate`.

--
For reference, this is what the code on master return:
```
oc adm inspect --since-time=2024-01-25 ns/openshift-cluster-version
error: parsing time "2024-01-25" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "T"
```
And this is what the new code in this PR return (it also matches the error for an invalid `--since-time` value from `oc adm must-gather`):
```
go run ./cmd/oc adm inspect --since-time=2024-01-25 ns/openshift-cluster-version
error: --since-time only accepts times matching RFC3339, eg '2006-01-02T15:04:05Z'
```

### DIFF
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -169,7 +169,7 @@ func (o *InspectOptions) Complete(args []string) error {
 	if len(o.sinceTime) > 0 {
 		o.sinceTimestamp, err = util.ParseRFC3339(o.sinceTime, metav1.Now)
 		if err != nil {
-			return err
+			return fmt.Errorf("--since-time only accepts times matching RFC3339, eg '2006-01-02T15:04:05Z'")
 		}
 	}
 


pkg/cli/admin/inspect/inspect.go
```diff
@@ -169,7 +169,8 @@ func (o *InspectOptions) Complete(args []string) error {
 	if len(o.sinceTime) > 0 {
 		o.sinceTimestamp, err = util.ParseRFC3339(o.sinceTime, metav1.Now)
 		if err != nil {
-			return err
+			klog.V(3).Infof("error parsing --since-time: %s", err)
```
ardaguclu, line 172: Why we don't simply `return fmt.Errorf("--since-time only accepts times matching RFC3339, eg '2006-01-02T15:04:05Z err: %w'", err)`?

flavianmissi: Yeah I considered that. What made me not do it is that the actual error doesn't add much to the overall message. This is what it would look like:
```
error: --since-time only accepts times matching RFC3339, eg '2006-01-02T15:04:05Z' err: parsing time "2024-01-25" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "T"
```

flavianmissi: Let me know if you think that's useful and I'll add it :)

ardaguclu: ah, it makes sense. Then we can also remove this `klog` printing :)

flavianmissi: I don't love the idea of swallowing errors but maybe this one can be okay. removed.

ardaguclu: > I don't love the idea of swallowing errors but maybe this one can be okay. removed.

Agreed. But it seems that there is no useful error come along from
https://github.com/golang/go/blob/c51beb4342ac0ee2fe9c1a774eac15679a447122/src/time/format.go#L1032, we can simply warn user about our error.

ardaguclu:
/lgtm

Interesting failures;
/retest

pkg/cli/admin/inspect/inspect.go
flavianmissi: Makes sense!

PR #1647, opened 2024-01-04
OCPBUGS-25461: Add RHEL9 and RHEL8 based oc as new targets in command extraction

ardaguclu:
As explained in the referenced bug, oc compiled in RHEL8 does not work
on FIPS enabled RHEL9 environment. However, we need to provide a way to
users to enable extraction of oc compiled in RHEL9 to work on FIPS enabled clusters.

This PR adds new command targets in `oc adm release extract` command that
users can extract oc specific to their RHEL versions as a continuation of 
https://github.com/openshift/oc/pull/1632

<details>
<summary>Single command extraction</summary>

```sh
$ ./oc adm release extract registry.ci.openshift.org/ocp/release:4.16.0-0.nightly-2024-01-03-193825 --to=/tmp/test_extract --command=oc.rhel9
# Downloads oc.rhel9
$ ./oc adm release extract registry.ci.openshift.org/ocp/release:4.16.0-0.nightly-2024-01-03-193825 --to=/tmp/test_extract --command=oc.rhel8
# Downloads oc.rhel8
$ ./oc adm release extract registry.ci.openshift.org/ocp/release:4.16.0-0.nightly-2024-01-03-193825 --to=/tmp/test_extract --command=oc --command-os=linux/arm64
# Downloads oc compiled in linux/arm64 with whichever RHEL based image is set(currently RHEL8)
$ ./oc adm release extract registry.ci.openshift.org/ocp/release:4.16.0-0.nightly-2024-01-03-193825 --to=/tmp/test_extract --command=oc.rhel9 --command-os=linux/arm64
# Downloads oc.rhel9 compiled in RHEL9 arm64 architecture
```
</details>

<details>
<summary>All tools extraction</summary>

```sh
$ ./oc adm release extract registry.ci.openshift.org/ocp/release:4.16.0-0.nightly-2024-01-03-193825 --to=/tmp/test_extract --tools
ccoctl-linux-4.16.0-0.nightly-2024-01-03-193825.tar.gz
openshift-client-linux-4.16.0-0.nightly-2024-01-03-193825.tar.gz
openshift-client-linux-amd64-rhel9-4.16.0-0.nightly-2024-01-03-193825.tar.gz
# There should also be openshift-client-linux-amd64-rhel8-4.16.0-0.nightly-2024-01-03-193825.tar.gz but we haven't added that yet
openshift-install-linux-4.16.0-0.nightly-2024-01-03-193825.tar.gz
```
</details>

<details>
[...]

PR #1665, opened 2024-01-25
OCPBUGS-28204: Add RHEL8 based oc as new targets in command extraction

soltysh:
Manual backport of #1647 
/assign @ardaguclu


ardaguclu:
I carefully looked at the commits and it looks good to me. There is no rhel9 compilation and correctly there isn't any rhel9 based binary in extraction targets.
/lgtm

ardaguclu:
`rpm-build` and `verify` job failures are fixed in hypershift by bumping the rhel version https://github.com/openshift/hypershift/pull/3466. I'm worried about we are closing a gigantic deadlock because we are already trying to bump oc to rhel9

ardaguclu:
There is a bug which should be fixed by https://github.com/openshift/oc/pull/1669;

/hold

ardaguclu:
It seems that we needed oc rhel9 extraction 4.15 as well and I proposed this https://github.com/openshift/oc/pull/1672 including the bug fix https://github.com/openshift/oc/pull/1669

PR #1669, opened 2024-01-29
OCPBUGS-28540: Copy oc.rhel8 instead of symlink

ardaguclu:
With this https://github.com/openshift/oc/pull/1647, we started allowing extraction of `oc.rhel8`.
However, it seems that symlinked binaries(i.e. `oc.rhel8`) are simply ignored during file path
traversal in tarballs. That results in the error message `oc.rhel8` is not found being raised. 

To solve the problem, in this PR we are now copying oc.rhel8 binaries from already compiled oc instead of
symlinking.


ardaguclu:
@ingvagabund could you PTAL, when you are available?. Thanks.

ardaguclu:
Without this PR, users can't extract oc.rhel8 binary and I think, it is critical and it seems that changes have no risk;

ardaguclu:
/hold
as we discussed with @ingvagabund. Copying 3 new binaries take ~450 mb of storage which is not feasible.

ardaguclu:
According to our latest discussions, it seems that these binaries wouldn't take too much storage as expected, since they are archived. Additionally, after doing some investigation, I concluded that there is no elegant and simple way of symlink extraction.

/hold cancel

PR #1674, opened 2024-01-31
OTA-1080: pkg/cli/admin/inspectalerts: New tech-preview inspect-alerts subcommand

wking:
The new subcommand is gated by `OC_ENABLE_CMD_INSPECT_ALERTS` to avoid customers accidentally using it while it's tech-preview:

```console
$ OC_ENABLE_CMD_INSPECT_ALERTS=true oc adm inspect-alerts | jq -r '[.[] | .startsAt + " " + .endsAt + " " + .status.state + " " + (.labels | .severity + " " + .alertname) + " " + .annotations.summary] | sort[]'
2024-01-19T20:13:28.083Z 2024-01-31T05:23:28.083Z suppressed none Watchdog An alert that should always be firing to certify that Alertmanager is working properly.
2024-01-19T20:14:13.174Z 2024-01-31T05:24:13.174Z suppressed info ClusterNotUpgradeable One or more cluster operators have been blocking minor version cluster upgrades for at least an hour.
2024-01-19T20:14:20.600Z 2024-01-31T05:24:20.600Z suppressed warning TechPreviewNoUpgrade Cluster has enabled tech preview features that will prevent upgrades.
2024-01-22T15:56:45.824Z 2024-01-31T05:23:15.824Z suppressed info PodStartupStorageOperationsFailing Pods can't start because volume_mount of volume plugin kubernetes.io/configmap is permanently failing on node build0-gstfj-w-b-d6xzf.c.openshift-ci-build-farm.internal.
2024-01-22T15:56:45.824Z 2024-01-31T05:23:15.824Z suppressed info PodStartupStorageOperationsFailing Pods can't start because volume_mount of volume plugin kubernetes.io/configmap is permanently failing on node build0-gstfj-w-c-q85c8.c.openshift-ci-build-farm.internal.
2024-01-30T06:02:23.913Z 2024-01-31T05:24:23.913Z active warning KubeJobFailed Job failed to complete.
```

This provides a more convenient syntax based on [the existing docs'][1]:

```console
$ oc login -u <username> -p <password>
$ host=$(oc -n openshift-monitoring get route alertmanager-main -ojsonpath={.spec.host})
$ token=$(oc whoami -t)
$ curl -H "Authorization: Bearer $token" -k "https://$host/api/v2/receivers"
```

But I'm using [the `/alerts` path][2] instead of the `/receivers` path given in the docs example.  This provides convenient access when the console isn't available (e.g. because the `Console` ClusterVersion capability is not enabled) or for use in other tooling like the tech-preview `oc adm upgrade status`, and has [a long-standing (but currently deferred) RFE][3].

I'm also dropping the `-k`/`--insecure` from the doc'ed curl in my Go translation, because:

* I don't want to leak a sensitive token to an untrusted server.
* Clusters are supposed to have [set up a well-known cert/CA for their ingress][4]:

    > The Ingress Operator generates a default certificate for an Ingress Controller to serve as a placeholder until you configure a custom default certificate. Do not use Operator-generated default certificates in production clusters.

    so our client will likely trust the cluster.

and failing on "I don't trust that server" seems fine for this tech-preview proof-of-concept, vs. trying to use the kubeconfig-protected connections to automatically figure out which CA I need to trust.

I'm iterating over `status.ingress[].host` entries instead of using the doc'ed `spec.host`, because [the `spec` property is `optional`][5], while [the `status` property is `required`][6].  It also seems safer to consume the Route controller's output `status` instead of wiring up directly to Route-admin written `spec` input.

I'm not currently filtering on `RouteIngress.Conditions` including an `Admitted=True` entry, although something like that might be reasonable to avoid wasting time poking known-to-be-broken hosts.

I haven't dug into `Bearer` token RFCs or the REST config structure to know if I need to worry about any encoding issues (base64?) in this handoff, but in testing it works for me.  Please feel free to stiffen, or at least research, that handoff before making this generally available :).

If I get a failure from one `status.ingress[].host`, I don't bother attempting the others.  Wiring up that kind of fallback with error aggregation when all hosts fail would be a useful future refinement.

We seem to have types in the vendored `github.com/prometheus/common/model` that could be used to deserialize this output (e.g. for table printing), but for now I'll just dump the JSON to output.

[1]: https://docs.openshift.com/container-platform/4.14/monitoring/accessing-third-party-monitoring-apis.html
[2]: https://github.com/prometheus/alertmanager/blob/v0.26.0/api/v2/openapi.yaml#L134
[3]: https://issues.redhat.com/browse/RFE-928
[4]: https://docs.openshift.com/container-platform/4.14/security/certificate_types_descriptions/ingress-certificates.html#location
[5]: https://github.com/openshift/api/blob/4f00b4f16b634830bb16b432eec91f6e514f2981/route/v1/types.go#L95
[6]: https://github.com/openshift/api/blob/4f00b4f16b634830bb16b432eec91f6e514f2981/route/v1/types.go#L352-L353

### DIFF
--- a/pkg/cli/admin/admin.go
+++ b/pkg/cli/admin/admin.go
@@ -21,6 +21,7 @@ import (
 	"github.com/openshift/oc/pkg/cli/admin/createproviderselectiontemplate"
 	"github.com/openshift/oc/pkg/cli/admin/groups"
 	"github.com/openshift/oc/pkg/cli/admin/inspect"
+	"github.com/openshift/oc/pkg/cli/admin/inspectalerts"
 	"github.com/openshift/oc/pkg/cli/admin/migrate"
 	migrateteicsp "github.com/openshift/oc/pkg/cli/admin/migrate/icsp"
 	migratetemplateinstances "github.com/openshift/oc/pkg/cli/admin/migrate/templateinstances"
@@ -47,6 +48,9 @@ var adminLong = ktemplates.LongDesc(`
 
 	Actions for administering an OpenShift cluster are exposed here.`)
 
+// inspectAlertsFeatureGate is an environment variable used to gate the inclusion of the inspect-alerts subcommand.
+const inspectAlertsFeatureGate = "OC_ENABLE_CMD_INSPECT_ALERTS"
+
 func NewCommandAdmin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
 	// Main command
 	cmds := &cobra.Command{
@@ -56,17 +60,23 @@ func NewCommandAdmin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *co
 		Run:   kcmdutil.DefaultSubCommandRun(streams.ErrOut),
 	}
 
+	clusterManagement := []*cobra.Command{
+		upgrade.New(f, streams),
+		top.NewCommandTop(f, streams),
+		mustgather.NewMustGatherCommand(f, streams),
+		ocpcertificates.NewCommandOCPCertificates(f, streams),
+		waitforstable.NewCmdWaitForStableClusterOperators(f, streams),
+		inspect.NewCmdInspect(streams),
+	}
+
+	if kcmdutil.FeatureGate(inspectAlertsFeatureGate).IsEnabled() {
+		clusterManagement = append(clusterManagement, inspectalerts.New(f, streams))
+	}
+
 	groups := ktemplates.CommandGroups{
 		{
-			Message: "Cluster Management:",
-			Commands: []*cobra.Command{
-				upgrade.New(f, streams),
-				top.NewCommandTop(f, streams),
-				mustgather.NewMustGatherCommand(f, streams),
-				inspect.NewCmdInspect(streams),
-				ocpcertificates.NewCommandOCPCertificates(f, streams),
-				waitforstable.NewCmdWaitForStableClusterOperators(f, streams),
-			},
+			Message:  "Cluster Management:",
+			Commands: clusterManagement,
 		},
 		{
 			Message: "Node Management:",
--- a/pkg/cli/admin/inspectalerts/inspectalerts.go
+++ b/pkg/cli/admin/inspectalerts/inspectalerts.go
@@ -1,0 +1,129 @@
+// Package inspectalerts provides access to in-cluster alerts.
+package inspectalerts
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	"k8s.io/client-go/rest"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+
+	routev1 "github.com/openshift/api/route/v1"
+	routev1client "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+)
+
+// RouteGetter is a function that gets a Route.
+type RouteGetter func(ctx context.Context, namespace string, name string, opts metav1.GetOptions) (*routev1.Route, error)
+
+type Alert struct {
+}
+
+type options struct {
+	genericiooptions.IOStreams
+	RESTConfig *rest.Config
+	getRoute   RouteGetter
+}
+
+func newOptions(streams genericiooptions.IOStreams) *options {
+	return &options{
+		IOStreams: streams,
+	}
+}
+
+func New(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	o := newOptions(streams)
+	cmd := &cobra.Command{
+		Use:   "inspect-alerts",
+		Short: "Collect information about running alerts.",
+		Run: func(cmd *cobra.Command, args []string) {
+			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			kcmdutil.CheckErr(o.Run(cmd.Context()))
+		},
+	}
+
+	return cmd
+}
+
+func (o *options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
+	kcmdutil.RequireNoArguments(cmd, args)
+
+	cfg, err := f.ToRESTConfig()
+	if err != nil {
+		return err
+	}
+
+	o.RESTConfig = cfg
+
+	routeClient, err := routev1client.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
+	o.getRoute = func(ctx context.Context, namespace string, name string, opts metav1.GetOptions) (*routev1.Route, error) {
+		return routeClient.Routes(namespace).Get(ctx, name, opts)
+	}
+
+	return nil
+}
+
+func (o *options) Run(ctx context.Context) error {
+	alertBytes, err := GetWithBearer(ctx, o.getRoute, "openshift-monitoring", "alertmanager-main", "/api/v2/alerts", o.RESTConfig.BearerToken)
+	if err != nil {
+		return err
+	}
+
+	_, err = o.Out.Write(alertBytes)
+	return nil
+}
+
+// GetWithBearer gets a Route by namespace/name, contructs a URI using
+// status.ingress[].host and the path argument, and performs GETs on that
+// URI using Bearer authentication with the token argument.
+func GetWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name, path, bearerToken string) ([]byte, error) {
+	if len(bearerToken) == 0 {
+		return nil, fmt.Errorf("no token is currently in use for this session")
+	}
+
+	route, err := getRoute(ctx, namespace, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{}
+	uris := make([]string, 0, len(route.Status.Ingress))
+	for _, ingress := range route.Status.Ingress {
+		uri := &url.URL{
+			Scheme: "https",
+			Host:   ingress.Host,
+			Path:   path,
+		}
+		uris = append(uris, uri.String())
+		req, err := http.NewRequest("GET", uri.String(), nil)
+		if err != nil {
+			return nil, err
+		}
+
+		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", bearerToken))
+
+		resp, err := client.Do(req)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		return body, err
+	}
+
+	return nil, fmt.Errorf("unable to get %s from any of %d URIs in the %s Route in the %s namespace: %s", path, len(uris), name, namespace, strings.Join(uris, ", "))
+}
--- a/pkg/cli/admin/inspectalerts/inspectalerts.go
+++ b/pkg/cli/admin/inspectalerts/inspectalerts.go
@@ -73,19 +73,37 @@ func (o *options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string
 }
 
 func (o *options) Run(ctx context.Context) error {
-	alertBytes, err := GetWithBearer(ctx, o.getRoute, "openshift-monitoring", "alertmanager-main", "/api/v2/alerts", o.RESTConfig.BearerToken)
+	alertBytes, err := GetAlerts(ctx, o.getRoute, o.RESTConfig.BearerToken)
 	if err != nil {
 		return err
 	}
 
 	_, err = o.Out.Write(alertBytes)
-	return nil
+	return err
 }
 
-// GetWithBearer gets a Route by namespace/name, contructs a URI using
+// GetAlerts gets alerts (both firing and pending) from openshift-monitoring Thanos.
+func GetAlerts(ctx context.Context, getRoute RouteGetter, bearerToken string) ([]byte, error) {
+	uri := &url.URL{ // configure everything except Host, which will come from the Route
+		Scheme: "https",
+		Path:   "/api/v1/alerts",
+	}
+
+	// if we end up going this way, probably port to github.com/prometheus/client_golang/api/prometheus/v1 NewAPI
+	alertBytes, err := getWithBearer(ctx, getRoute, "openshift-monitoring", "thanos-querier", uri, bearerToken)
+	if err != nil {
+		return alertBytes, err
+	}
+
+	// if we end up going this way, probably check and error on 'result' being an empty set (it should at least contain Watchdog)
+
+	return alertBytes, nil
+}
+
+// getWithBearer gets a Route by namespace/name, contructs a URI using
 // status.ingress[].host and the path argument, and performs GETs on that
 // URI using Bearer authentication with the token argument.
-func GetWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name, path, bearerToken string) ([]byte, error) {
+func getWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name string, baseURI *url.URL, bearerToken string) ([]byte, error) {
 	if len(bearerToken) == 0 {
 		return nil, fmt.Errorf("no token is currently in use for this session")
 	}
@@ -98,11 +116,8 @@ func GetWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name, p
 	client := &http.Client{}
 	uris := make([]string, 0, len(route.Status.Ingress))
 	for _, ingress := range route.Status.Ingress {
-		uri := &url.URL{
-			Scheme: "https",
-			Host:   ingress.Host,
-			Path:   path,
-		}
+		uri := *baseURI
+		uri.Host = ingress.Host
 		uris = append(uris, uri.String())
 		req, err := http.NewRequest("GET", uri.String(), nil)
 		if err != nil {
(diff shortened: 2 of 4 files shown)


pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -0,0 +1,133 @@
...
+			Host:   ingress.Host,
+			Path:   path,
+		}
+		uris = append(uris, uri.String())
+		req, err := http.NewRequest("GET", uri.String(), nil)
+		if err != nil {
+			return nil, err
+		}
+
+		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", bearerToken))
```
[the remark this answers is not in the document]
ardaguclu: If `/api/v2/alerts` is registered to API server, why we don't use the builtin functionality, instead of manually adding bearer token in header?

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -0,0 +1,133 @@
...
+	route, err := getRoute(ctx, namespace, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{}
+	uris := make([]string, 0, len(route.Status.Ingress))
+	for _, ingress := range route.Status.Ingress {
+		uri := &url.URL{
+			Scheme: "https",
```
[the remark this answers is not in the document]
wking: Nope, there's no CA handling at all, I'm just using a stock Go `http.Client{}`.  Is there some kind of tooling/precedent for using the kubeconfig to figure out the CAs that should be trusted to connect to a particular Route?

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	route, err := getRoute(ctx, namespace, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{}
+	uris := make([]string, 0, len(route.Status.Ingress))
+	for _, ingress := range route.Status.Ingress {
```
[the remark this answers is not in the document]
wking: From my PR/commit message:

> If I get a failure from one `status.ingress[].host`, I don't bother attempting the others. Wiring up that kind of fallback with error aggregation when all hosts fail would be a useful future refinement.

If you have a preference on that or the `Admitted=True` filtering I floated in the PR/commit message, let me know about that too :)

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -0,0 +1,133 @@
...
+
+// GetWithBearer gets a Route by namespace/name, contructs a URI using
+// status.ingress[].host and the path argument, and performs GETs on that
+// URI using Bearer authentication with the token argument.
+func GetWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name, path, bearerToken string) ([]byte, error) {
+	if len(bearerToken) == 0 {
+		return nil, fmt.Errorf("no token is currently in use for this session")
+	}
+
+	route, err := getRoute(ctx, namespace, name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
wking: We're expecting to consume this in future `oc adm upgrade status` work, and I wanted to bundle as much as possible into the shared helper, just to make the call-site simpler on the consumer side.  But having two separate call-and-catch-`err` in the `status` call-site would just be a bit less compact, and not hard to do, so I don't mind pivoting to passing `route` if you aren't  buying my "easier on Go consumers" pitch here ;)

pkg/cli/admin/admin.go
```diff
@@ -56,17 +57,23 @@ func NewCommandAdmin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *co
...
+	clusterManagement := []*cobra.Command{
+		upgrade.New(f, streams),
+		top.NewCommandTop(f, streams),
+		mustgather.NewMustGatherCommand(f, streams),
+		ocpcertificates.NewCommandOCPCertificates(f, streams),
+		waitforstable.NewCmdWaitForStableClusterOperators(f, streams),
+		inspect.NewCmdInspect(streams),
+	}
+
+	if kcmdutil.FeatureGate("OC_ENABLE_CMD_INSPECT_ALERTS").IsEnabled() {
```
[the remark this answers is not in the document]
wking: Moved to a new `inspectAlertFeatureGate` `const` with ab49d5c3f -> a26bea5d2.

pkg/cli/admin/inspectalerts/inspectalerts.go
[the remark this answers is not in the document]
wking: This is a non-Kube API, e.g. in a 4.16.0-ec.1 cluster, the only alert API resources are:

```console
$ oc api-resources | grep -i alert
alertmanagerconfigs                   amcfg                                                                                  monitoring.coreos.com/v1beta1                   true         AlertmanagerConfig
alertmanagers                         am                                                                                     monitoring.coreos.com/v1                        true         Alertmanager
alertingrules                                                                                                                monitoring.openshift.io/v1                      true         AlertingRule
alertrelabelconfigs                                                                                                          monitoring.openshift.io/v1                      true         AlertRelabelConfig
```

And none of those is this list of Alertmanager alerts.  An alternative access mechanism would be for the Alertmanager to register its alert type with Kube, e.g. as [an aggregated API server][1].  But that sort of pivot would be large, and in the meantime, we want some kind of convenient access to firing alerts for the `status` subcommand.

But narrowly, if there is Go/Kube/other tooling for `Bearer` injection on a shelf somewhere, I'm happy to pivot to using that when constructing these requests.  I was unable to find it on a brief search yesterday, but I didn't sink too much time into hunting, because I wanted to limit the time invested in this tech-preview proof-of-concept. 

[1]: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/

LalatenduMohanty:
Can we add an option to filter the alters based on severity e.g. `"severity": "warning",` ?

wking:
> Can we add an option to filter the alerts based on the severity e.g. `"severity": "warning"`, ?

All things are possible, but I was trying to avoid sinking time into aspects that `oc adm upgrade status` wouldn't need.  It will presumably deserialize the JSON into  [`github.com/prometheus/common/model`'s `Alerts`][1], and then use structured filtering.  While folks using the JSON dump I've implemented so far can do that with `jq`, e.g.:

```console
$ OC_ENABLE_CMD_INSPECT_ALERTS=true ./oc adm inspect-alerts | jq -r '[.[] | select(.labels.severity == "warning") | .startsAt + " " + .endsAt + " " + .status.state + " " + (.labels | .severity + " " + .alertname) + " " + .annotations.summa
ry] | sort[]'
2024-01-19T20:14:20.600Z 2024-02-01T00:31:50.600Z suppressed warning TechPreviewNoUpgrade Cluster has enabled tech preview features that will prevent upgrades.
```

And since the main issue I personally have consuming the raw JSON is excluding all the properties I don't care about, I'm already using `jq`  (or we'd have to discuss and come to  some kind of agreement about which properties were important enough to show by default, and what the command line API looks like to adjust that if a user wants something custom).

Anyhow, things in this space seem like they'd be nice-to-have, but I don't see them as proof-of-concept blockers.

[1]: https://pkg.go.dev/github.com/prometheus/common/model#Alerts

machine424:
> Can we add an option to filter the alerts based on the severity e.g. "severity": "warning", ?

In future iterations, we could consider wiring up [amtool](https://github.com/prometheus/alertmanager?tab=readme-ov-file#amtool) to avoid implementing another client. 
It offers many useful features and it'll be handy to have it in `oc`  (one less binary to worry about), we could even allow customizing the alertmanager URL, the platform one being the default. (we'll need to think about compatibilities though)

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -0,0 +1,133 @@
...
+	_, err = o.Out.Write(alertBytes)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GetWithBearer gets a Route by namespace/name, contructs a URI using
+// status.ingress[].host and the path argument, and performs GETs on that
```
machine424, line 90: Maybe https://github.com/openshift/library-go/blob/2ad786549f07f264339f63b12b71a0fe5e77e832/pkg/route/routeapihelpers/routeapihelpers.go#L18 could help here, it also checks the conditions ;)

simonpasquier:
I'd rather use the Thanos Querier API endpoints than Alertmanager:
* Customers may disable the in-cluster Alertmanager (especially true with ACM satellite clusters).
* It would be possible to retrieve grab pending alerts.
* Thanos aggregates both platform and user-defined alerts (in case this is useful).

The downside is that you won't see which alerts are being silenced since it's only applicable to alerts from Alertmanager.

pkg/cli/admin/inspectalerts/inspectalerts.go
wking: `oc` vendors library-go, but not yet that sub-package^, and I wanted to avoid mucking with `vendor/` and introduce thinking about iteration over `ingress[]` (`IngressURI` only returns the first match).  But yes, we could pivot to using that.  Or extend library-go to grow a callback-consuming function that supports iteration over multiiple matches, etc.  Thoughts for this pull?

wking:
> Customers may disable the in-cluster Alertmanager...

Ah, docs around that [here][1].  I'd mistakenly thought it was Alertmanager converting Prometheus metrics into alerts, but looks like Prometheus calculates the alerts, and Alertmanager layers on silences and push notification.  So yeah, I'll pivot to Thanos.

[1]: https://docs.openshift.com/container-platform/4.14/monitoring/configuring-the-monitoring-stack.html#monitoring-disabling-the-local-alertmanager_configuring-the-monitoring-stack

petr-muller:
> The downside is that you won't see which alerts are being silenced since it's only applicable to alerts from Alertmanager.

I'm a bit worried about confused users who would expect to *not* see a silenced alert when they do have an alertmanager, but for the purpose of `oc adm upgrade status` I think we're probably good with documenting that little caveat (over doing something more complicated like trying alertmanager and then falling back to thanos or something else...).

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = o.Out.Write(alertBytes)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
petr-muller, line 86: nit: this is `return err`

wking:
> > Customers may disable the in-cluster Alertmanager...
>
> ... So yeah, I'll pivot to Thanos.

Ok, poking at this locally, looks like Thanos is only giving me some metric labels:

```console
$ curl -sH "Authorization: Bearer $(oc whoami -t)" "https://$(oc -n openshift-monitoring get route thanos-querier -o jsonpath='{.status.ingress[0].host}')/api/v1/query?query=ALERTS" | jq '.data.result[0]'
{
  "metric": {
    "__name__": "ALERTS",
    "alertname": "ClusterAutoscalerUnschedulablePods",
    "alertstate": "pending",
    "container": "cluster-autoscaler",
    "endpoint": "metrics",
    "instance": "10.129.0.57:8085",
    "job": "cluster-autoscaler-default",
    "namespace": "openshift-machine-api",
    "pod": "cluster-autoscaler-default-f8dd547c7-dg9t5",
    "prometheus": "openshift-monitoring/k8s",
    "service": "cluster-autoscaler-default",
    "severity": "info"
  },
  "value": [
    1706914546.182,
    "1"
  ]
}
```

While Alertmanager gives me `.annotations.summary` and other context that is needed to act on the alert.  I guess it's theoretically possible for `oc` to head back to the cluster and pull PrometheusRules and... no, we still wouldn't be able to do the dynamic template stuff that happens somewhere in the monitoring component today.

But anyway, are folks ok if I stick with using Alertmanager for this initial proof-of-concept?

pkg/cli/admin/inspectalerts/inspectalerts.go
wking: Thanks; fixed with 675fdb4ad -> 41e5e82a8.

pkg/cli/admin/inspectalerts/inspectalerts.go
machine424: We can leave that for future iterations for sure.

simonpasquier:
Sorry I was a bit dense: Thanos also exposes a /api/v1/alerts endpoint which returns all active alerts including the annotations.

wking:
Ok, I've pushed 41e5e82a8 -> da5669fc9 with a pivot to Thanos and `/api/v1/alerts`:

```console
$ OC_ENABLE_CMD_INSPECT_ALERTS=true oc adm inspect-alerts | jq -r '[.data.alerts[] | .activeAt + " " + .state + " " + (.labels | .severity + " " + .alertname) + " " + .annotations.summary] | sort[]'
2024-01-18T15:42:50Z firing warning TechPreviewNoUpgrade Cluster has enabled tech preview features that will prevent upgrades.
2024-01-18T15:45:13Z firing info ClusterNotUpgradeable One or more cluster operators have been blocking minor version cluster upgrades for at least an hour.
2024-01-19T20:20:58.083333925Z firing none Watchdog An alert that should always be firing to certify that Alertmanager is working properly.
2024-01-22T15:51:45.824704657Z firing info PodStartupStorageOperationsFailing Pods can't start because volume_mount of volume plugin kubernetes.io/configmap is permanently failing on node build0-gstfj-w-b-d6xzf.c.openshift-ci-build-farm.internal.
2024-01-22T15:51:45.824704657Z firing info PodStartupStorageOperationsFailing Pods can't start because volume_mount of volume plugin kubernetes.io/configmap is permanently failing on node build0-gstfj-w-c-q85c8.c.openshift-ci-build-farm.internal.
2024-02-05T20:09:16.894430475Z pending warning PodDisruptionBudgetAtLimit The pod disruption budget is preventing further disruption to pods.
2024-02-05T20:12:56.428212511Z pending info ClusterAutoscalerUnschedulablePods Cluster Autoscaler has 13 unschedulable pods
2024-02-05T20:15:43.59211931Z pending warning MachineNotYetDeleted machine build0-gstfj-ci-tests-worker-d-rrdl9 has been in Deleting phase for more than 6 hours
...
```

wking:
[images hit build04 node capacity][1].

/retest-required

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/1674/pull-ci-openshift-oc-master-images/1754603983437041664

jan--f:
:+1: This looks good to me.

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	o.getRoute = func(ctx context.Context, namespace string, name string, opts metav1.GetOptions) (*routev1.Route, error) {
+		return routeClient.Routes(namespace).Get(ctx, name, opts)
+	}
+
+	return nil
+}
+
+func (o *options) Run(ctx context.Context) error {
+	alertBytes, err := GetAlerts(ctx, o.getRoute, o.RESTConfig.BearerToken)
```
ardaguclu, line 76: There are many ways to send authenticated request to API server in RESTConfig https://github.com/openshift/oc/blob/4fe33f165fbda5d9c3814dc4258859a502f43eb5/vendor/k8s.io/client-go/rest/config.go#L69-L89 and as far as I know, bearerToken is not mandatory. For example, if user may use credentials exec plugin which skips bearer token step.

pkg/cli/admin/inspectalerts/inspectalerts.go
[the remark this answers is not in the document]
simonpasquier: Yes the request will be rejected. I'm not sure what you're advising for though?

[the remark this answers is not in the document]
ardaguclu: > As you referred above in the source code, we could leverage github.com/prometheus/client_golang/v1 + github.com/prometheus/common for token injection but it can be deferred to a following PR.

I'm advising this ^ but it seems that it is not possible soon. I'd suggest to mention this case in the description of the command to warn user about this.

ardaguclu:
/approve
for adding monitoring-approvers in OWNER_ALIASES

pkg/cli/admin/inspectalerts/inspectalerts.go
wking: Bearer token is [what we doc][1], so it should be good enough for a env-var-gated first attempt, right?

[1]: https://docs.openshift.com/container-platform/4.14/monitoring/accessing-third-party-monitoring-apis.html#accessing-third-party-monitoring-web-service-apis_accessing-third-party-monitoring-apis

ardaguclu: Yes, I think so

wking: must-gather is [using a ServiceAccount token][1] and [collecting the default ingress CA][2].  Both of which we could use.  But :shrug:, also work to implement, and not clear to me that we need to sink that time in before landing a subcommand behind a `OC_ENABLE_CMD_INSPECT_ALERTS=true` gate.  Is "if there isn't a bearer token or the request is rejected with an HTTP auth code, go see if we have the power to borrow a must-gather ServiceAccount token" the sort of thing we need to land the initial implementation?  There's always work we could invest to make any implementation better, but where are the goal-posts for merging an initial, env-var-gated command?

[1]: https://github.com/openshift/must-gather/blob/c8fec85caeab4757d1c67daf38620fc89aabca69/collection-scripts/gather_monitoring#L27-L29
[2]: https://github.com/openshift/must-gather/blob/c8fec85caeab4757d1c67daf38620fc89aabca69/collection-scripts/gather_monitoring#L31-L34

LalatenduMohanty (approved):
/lgtm

pkg/cli/admin/inspectalerts/inspectalerts.go
LalatenduMohanty: RE: stable API , we can take up the work post the merge. This should not block the merge as this PR is behind  OC_ENABLE_CMD_INSPECT_ALERTS=true flag.

simonpasquier: > Isn't there a stable API we could access? How our console is accessing that data? Can we re-use that same API?

The console uses the same API service to query and display the cluster alerts.

PR #1685, opened 2024-02-14
NO-JIRA: Print log rather than returning an error when kubectl version mistmaches

ardaguclu:
There are images that are based on `cli` image in the release payload and those images also manifest their kubectl versions (i.e. [cluster-samples-operator](https://github.com/openshift/cluster-samples-operator/blob/master/manifests/image-references) [1]). It is reasonable to assume that all the kubectl versions should match, because base image is the same for all of them. However, rebuilt strategy of some images simply rely on a new commit in their repository. That means if there is no new commit merged for a repository that the image is residing in, this image will not be rebuilt and still manifests an old kubectl version. That causes a new release payload generation mechanism a failure (even a minor kubectl version bump in oc would cause a org-wide failures).

Due to this uncontrolled environment for kubectl versions, this PR switches the kubectl version check from error to information log. Users will still see mismatches but this just won't cause any failure. This shouldn't create a problem because our source of truths are still `cli` and `cli-artifacts` images.

[1] _For example must-gather image manifests kubectl 1.28.2 https://quay.io/repository/openshift/origin-must-gather/manifest/sha256:b36b9314580632d7eb4f4d5d71b1297eb637b91f84f0ab7df51137524f704228 and refers to an old commit because there is no changes in https://github.com/openshift/must-gather within this time period._

### DIFF
--- a/pkg/cli/admin/release/image_mapper.go
+++ b/pkg/cli/admin/release/image_mapper.go
@@ -3,6 +3,7 @@ package release
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


ardaguclu:
/hold 
until verification

ardaguclu:
unrelated failures;
/retest
/hold
because somehow images in cluster-samples-operator are replaced with correct kubectl versions in cluster-bot. I couldn't test this. Since the changes are straight-forward, I'm unholding this PR.

pkg/cli/admin/release/image_mapper.go
```diff
@@ -213,7 +213,7 @@ func loadImageStreamTransforms(input, local *imageapi.ImageStream, allowMissingI
 	}
 
 	if kubectlVersions.Len() > 2 {
-		return nil, nil, nil, fmt.Errorf("input image stream has multiple versions defined for version kubectl for all the tags %s", strings.Join(kubectlVersions.UnsortedList(), ","))
+		klog.Infof("input image stream has multiple versions defined for version kubectl for all the tags %s", strings.Join(kubectlVersions.UnsortedList(), ","))
```
soltysh, line 216: How about doing `fmt.Fprintf(o.ErrOut, "warning:...` instead? To make the message stand out a bit more.

ardaguclu: Updated.

soltysh (approved):
/lgtm

PR #1686, opened 2024-02-14
OCPBUGS-29210: Update some commands incorrect examples

ardaguclu:
This PR updates the incorrect examples. Changes in here aligns with the incorrect example findings in https://github.com/openshift/openshift-docs/pull/71213

### DIFF
--- a/pkg/cli/admin/copytonode/command.go
+++ b/pkg/cli/admin/copytonode/command.go
@@ -20,11 +20,8 @@ var (
 	`)
 
 	copyToNodeExample = templates.Examples(`
-		# copy a new bootstrap kubeconfig file to every node
-		oc adm copy-to-node --copy=new-bootstrap-kubeconfig=/etc/kubernetes/kubeconfig
-
-		# copy a new bootstrap kubeconfig file to masters
-		oc adm copy-to-node --copy=new-bootstrap-kubeconfig=/etc/kubernetes/kubeconfig -l node-role.kubernetes.io/master`)
+		# copy a new bootstrap kubeconfig file to node-0
+		oc adm copy-to-node --copy=new-bootstrap-kubeconfig=/etc/kubernetes/kubeconfig node/node-0`)
 )
 
 type CopyToNodeOptions struct {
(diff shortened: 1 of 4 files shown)


pkg/cli/admin/copytonode/command.go
```diff
@@ -21,10 +21,7 @@ var (
 
 	copyToNodeExample = templates.Examples(`
 		# copy a new bootstrap kubeconfig file to every node
```
zhouying7780, line 23: we should also update the annotation here.

pkg/cli/admin/ocpcertificates/certregen/command.go
```diff
@@ -25,11 +25,8 @@ var (
 	`)
 
 	regenerateSignersExample = templates.Examples(`
-		# Regenerate all known signing certificates on the cluster.
-		oc adm certificates regenerate-signers -A --all
-
 		# Regenerate the signing certificate contained in a particular secret.
-		oc adm certificates regenerate-signers -n openshift-kube-apiserver-operator loadbalancer-serving-signer
+		oc adm ocp-certificates regenerate-top-level -n openshift-kube-apiserver-operator loadbalancer-serving-signer
```
zhouying7780, line 29: here should be `oc adm ocp-certificates regenerate-top-level -n openshift-kube-apiserver-operator secret/loadbalancer-serving-signer`

ardaguclu: Updated

pkg/cli/admin/copytonode/command.go
ardaguclu: Example with the annotation removed altogether.

pkg/cli/admin/ocpcertificates/certregen/command.go
zhouying7780: @ardaguclu   please remove the `line breaks` in the example.

ardaguclu: @zhouying7780 thanks. I think, no need to run all the CI jobs just for this newline. Because also other examples have it as well.

ardaguclu:
This is just example changes and no risk;

PR #1689, opened 2024-02-19
OTA-1165: Add Worker Node Status

Davoska:


pkg/cli/admin/upgrade/status/workerpool.go
```diff
@@ -0,0 +1,522 @@
...
+}
+
+const workerPoolStatusTemplateRaw = `= Worker Pool =
+Worker Pool:     {{ .Name }}
+Assessment:      {{ .Assessment }}
+Completion:      {{ printf "%.0f" .Completion }}%
+Worker Status: {{ .NodesOverview.Total }} Total, {{ .NodesOverview.Available }} Available, {{ .NodesOverview.Progressing }} Progressing, {{ .NodesOverview.Outdated }} Outdated, {{ .NodesOverview.Draining }} Draining, {{ .NodesOverview.Excluded }} Excluded, {{ .NodesOverview.Degraded }} Degraded
+`
+
+func (pool *poolDisplayData) WriteNodes(w io.Writer) {
```
petr-muller, line 477: pls unify receiver names (`pool` here vs `d` on L466)

petr-muller:
/retitle OTA-1165: `adm upgrade status`: Add Worker Node Status

pkg/cli/admin/upgrade/status/workerpool.go
```diff
@@ -0,0 +1,506 @@
...
+	"strings"
+	"text/tabwriter"
+	"text/template"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	mcfgv1 "github.com/openshift/api/machineconfiguration/v1"
+	mcfgv1client "github.com/openshift/client-go/machineconfiguration/clientset/versioned"
+	mcfgcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
+	mcfgconst "github.com/openshift/machine-config-operator/pkg/daemon/constants"
```
petr-muller, line 17: The size of the vendored changes was bugging me so I followed that rabbit hole and the result is https://github.com/openshift/oc/pull/1697, check it out.

I discovered that we are vendoring o/m-c-o for basically two purposes, nothing else:
1. Constants
2. `LayeredNodeState` we use for `IsUnavailable`

We can inline the constants (or better, copy the files that contain them into a little `pkg/cli/admin/upgrade/status/mco/constants.go` ). The `LayeredNodeState` code is a little harder, but still fairly self-contained, check out https://github.com/openshift/oc/pull/1697/commits/1a61719ca0adba097b571f20d1c73e408a20045e. We can hardcopy two files (say, again, into `pkg/cli/admin/upgrade/status/mco/...`) and delete the bits we do not use, resulting in very small part of the o/m-c-o code tht we actually use, and *then* all we need for deps is `go mod vendor` which simply brings the missing parts of client-go and o/api - without modifying go.sum/go.mod at all.

This would be a little heavy-handed, but remember all this client-side code is experimental and will be eventually removed, and as long as we comment the extracted MCO code with prominent `TODO: replace this with proper MCO code when going server-side`, I think it is worth it (avoids bringing a ton of new deps for code that will never reach customers).

WDYT?

Davoska: Seems very reasonable. I will proceed with this suggestion. Thank you very much, I appreciate the extensive explanation.

pkg/cli/admin/upgrade/status/examples/4.14.1-control-plane-updated-workers-progressing.output
```diff
@@ -0,0 +1,25 @@
+An update is in progress for 2m58s: Cluster version is 4.14.1
+
+= Control Plane =
+Assessment:      Completed
+Completion:      100%
+Duration:        32m0s
```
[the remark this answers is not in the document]
petr-muller: Yep, this looks good. I remember we fixed this inconsistency in our pair session, IIRC we showed "time since progressing changed" in SINCE, which in this phase is "when control plane finished updating".

petr-muller (approved):
:tada: 

Really nice job, thanks :+1:

petr-muller:
/cc @ardaguclu 

Arda, can we have an `/approve` for the `vendor/*` changes? The 092156a7c53c7ef3388056ea2c51b0a45d5f3112 commit just runs `go mod vendor` which vendors more pieces of o/api and o/client-go. No new dependencies are actually added in `go.mod`.

pkg/cli/admin/upgrade/status/examples/4.14.1-degraded-mcp.yaml
```diff
@@ -0,0 +1,162 @@
+apiVersion: v1
```
ardaguclu, line 1: nit: Do we really need to all these examples that are unnecessarily increasing repo size?

ardaguclu (commented):
Vendor changes look good to me;
/approve

pkg/cli/admin/upgrade/status/examples/4.14.1-degraded-mcp.yaml
petr-muller: I agree the inputs grew a bit :( I do not expect us to see more testcases but I think we'll eventually need more inputs; I'll make sure we come up with some solution until then, maybe write a script that prunes the mock data from unnecessary members or something similar.

Davoska:
Thank you, @petr-muller, for all the suggestions, recommendations, and tips on and off GitHub!

It made this pull request much better 🎉

PR #1672, opened 2024-01-30
[release-4.15] OCPBUGS-28204: Add RHEL9 and RHEL8 based oc as new targets in command extraction

ardaguclu:
This PR is manual backport of a group of PRs to enable the extraction of rhel9 and rhel8 based oc binaries.

* https://github.com/openshift/oc/pull/1632
* https://github.com/openshift/oc/pull/1647
* https://github.com/openshift/oc/pull/1669

This PR also covers https://github.com/openshift/oc/pull/1665

These PRs are prerequisite of this PR;
* https://github.com/openshift-eng/ocp-build-data/pull/4303
* https://github.com/openshift/oc/pull/1669

PR #1691, opened 2024-02-26
OCPBUGS-23386: Extract oc also for linux/ppc64le

ardaguclu:
As with https://github.com/openshift/oc/pull/1672 in 4.16, we started compiling oc in linux/ppc64le architecture separately for rhel8 and rhel9. But users can not extract this binary from `oc adm release extract`. This PR
adds oc `linux/ppc64le` in extraction target list to let users can get the correct oc in this architecture.


ardaguclu:
/cc @ingvagabund 
what is your opinion about this?.

ardaguclu:
According to [here](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.15.0-rc.8/), it will add 48 MB * 3. Would it be reasonable?.

ingvagabund:
The image size should be less. https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.15.0-rc.8/ are archives produced from the image.

PR #1692, opened 2024-02-27
WRKLDS-1076: oc idle: Remove redundant endpoint resource modification

ardaguclu:
In the past, oc idle command annotates the **endpoint** resource by modifying it after marking the deployments (or deploymentconfigs or replicasets) as idled.

However, same behavior can be satisfied by annotating the **service** resource (that was added by https://github.com/openshift/oc/pull/720) and this PR removes the endpoint resource modification which is already an unwanted behavior.

### DIFF
--- a/pkg/cli/idle/idle.go
+++ b/pkg/cli/idle/idle.go
@@ -233,7 +233,6 @@ func scanLinesFromFile(filename string) ([]string, error) {
 // idleUpdateInfo contains the required info to annotate an endpoints object
 // with the scalable resources that it should unidle
 type idleUpdateInfo struct {
-	obj       *corev1.Endpoints
 	service   *corev1.Service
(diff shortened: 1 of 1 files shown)


ardaguclu:
It is expected that some tests will fail because they are checking the annotation in endpoints not services https://github.com/openshift/origin/blob/master/test/extended/cli/idle.go

ardaguclu:
This PR is blocked by https://github.com/openshift/origin/pull/28619

ardaguclu:
It seems that combination of this PR and the origin one https://github.com/openshift/origin/pull/28619 succeeded https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/1762437001774108672

soltysh:
/hold
until https://github.com/openshift/origin/pull/28619 merges

PR #1706, opened 2024-03-11
cli/admin/release: add github method for changelogs

AlexNPavel:
This PR adds flags and logic for generating changelogs using the github API. When using an OAuth token, this method is considerably faster than using the `git` method without an existing `git` directory and requires no disk space. It is considerably slower than using a warmed up `git` directory, but this may be reduced by using a github cache server such as the one used by prow.  This new changelog generation method is intended to be used mainly by the `release-controller`, which will have access to Openshift Prow's github cache.

### DIFF
--- a/pkg/cli/admin/release/github.go
+++ b/pkg/cli/admin/release/github.go
@@ -1,0 +1,455 @@
+package release
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/openshift/oc/pkg/version"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+)
+
+type github struct {
+	url   string
+	token string
+	http  *http.Client
+}
+
+func NewGitHubClient(url, tokenPath string) (*github, error) {
+	httpClient := &http.Client{
+		Timeout: 5 * time.Minute,
+	}
+	var token []byte
+	if tokenPath != "" {
+		var err error
+		token, err = os.ReadFile(tokenPath)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to read github token: %w", err)
+		}
+	}
+	return &github{
+		url:   url,
+		token: string(bytes.TrimSpace(token)),
+		http:  httpClient,
+	}, nil
+}
+
+func GitHubCommitToMergeCommitSingle(rCommit RepositoryCommit, squash bool) (*MergeCommit, error) {
+	// check if merge commit is squash is false
+	if !squash && len(rCommit.Parents) == 1 {
+		return nil, nil
+	}
+	mCommit := &MergeCommit{
+		CommitDate: rCommit.Commit.Committer.Date,
+		Commit:     rCommit.SHA,
+	}
+	// github only sends the parents in the full RepositoryCommit
+	for _, parent := range rCommit.Parents {
+		mCommit.ParentCommits = append(mCommit.ParentCommits, parent.SHA)
+	}
+
+	fullMsg := rCommit.Commit.Message
+	var mergeMsg, msg string
+	if !squash {
+		splitMessage := strings.Split(fullMsg, "\n\n")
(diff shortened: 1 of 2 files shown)


AlexNPavel:
This PR is currently able to produce a near identical changelog as using the standard `git` method for changelog generation. 2 issues remain to be resolved:

1. GitHub claims that some `csi-operator` commits do not exist, but `git log` has no problem with them. This may be an issue on GitHub's side.
2. The GitHub method includes `cloud-provider-openstack` commits for 4.14.0 to 4.15.0 (and possibly other ranges) when the normal `git` method includes none. This is a pretty minor issue but it would be nice to resolve it before this PR merges.

AlexNPavel:
In regards to point 2 above, `git` says that the `from` commit doesn't exist. According to GitHub, the commit does exist, but not in any of the branch of the repository we're interested in and may belong to a fork outside of the repository. It still returns the actual commit, but the changelog generated for the repo does not contain any actual merge commits for the repo itself, only rebases. Here is an example of one of those commits when doing a changelog between 4.16.0-ec.2 and ec.3:
https://github.com/openshift/cluster-api-provider-openstack/commit/20e33ac0d8042d69f9369203e4caff238e37e13d

AlexNPavel:
The first issue listed above has been resolved by adding the option for a `git` fallback. This is used in cases where an image has more than one git source listed, which can happen an image starts being built from a different repo (for example, `aws-ebs-csi-driver-operator` started being built from `csi-operator` between 4.14 and 4.15). In these situations, github cannot be used to compare the differences between commits, so we can fall back to git. This does result in some disk space being used for the git operation, but only a few hundred megabytes for one repo vs gigabytes and tens to hundreds of repos for a standard git-based changelog.

For the second issue, I am still investigating. In most situations, the output is incorrect, but in some situations, the output is correct, whereas the usual git flow simply returns no output for the repo due to a missing `from` commit.

AlexNPavel:
I've found that the second issue is actually a preexisting problem with how we identify repos that use squash merges vs merge commits. I can't find an easy way to improve our detection of that, so that issue will not be resolved by this PR.
This PR should now be ready for review.

bradmwilliams (commented):
/lgtm

ardaguclu:
Thanks for the work and detailed analysis, it is really helpful to track.

What is the main motivation behind these changes?;
* To be faster than the git command execution?. I saw that it is only faster during the cold start and the rest it is slower.
* To get rid of disk space usage?. I saw that we still require disk space for warmed directory already.

Now, there are 2 competing same functionalities and there is no sufficient reasoning about why we are trying to drop the current working git command execution and instead migrating to Github API. If we plan to live with both of them, what problem the second one solves that `git` command execution does not?.

AlexNPavel:
There were 2 motivations for this, both related to disk space usage:
1. Due to the huge disk space requirements, the release-controller uses 1 git-cache pod per architecture (though I believe we could just use 1 instead of multiple), each with its own 200GB PVC (about 55GB are actually in use). This pod would sometimes take a long time to come back up if restarted due to issues reattaching the PVC and changelogs were completely unavailable until the pod finished restarting and building its cache. Relying on the github-api would mean we could run the command locally since we would depend on the 50GB of git cache, which would make the changelog generation more reliable.
2. Even with a PVC, initializing the git-cache pod was slow. On start, we have an init container to perform some maintenance by running  `git gc` and `git pull` on every repo in the PVC and then uses `oc` to generate a changelog between the latest stable and latest preview in order to retrieve any repos that may be missing. This would take 15-20 minutes, during which changelog generation would be completely unavailable in the release-controller's web pages.

That being said, I've been benchmarking the github-api changes in this PR within the app.ci cluster and using prow's github-api cache, and it is still too slow at about 2 minutes to generate the 4.14.0->4.15.0 changelog. After doing some more research into `git`, I discovered we can reduce the size of the cache to ~1GB by using `--filter=blob:none --bare` in the `git clone` command, which makes it small enough that we can run it locally on our pods without exhausting the disk space of the underlying nodes when using `emptyDir` for cache storage instead of a PVC. Warming up a cache with those flags only takes ~1 minutes on the cluster as well.

I will close this PR and make a new one with the optimized `git clone` flags.

/close

ardaguclu:
Thanks. I experience same problem every time running this command on local and it is really annoying.

AlexNPavel:
Related to WRKLDS-1128.

PR #1707, opened 2024-03-13
[must-gather] Add support for tilde in dest dir path

black-dragon74:
This patch adds the functionality to correctly parse tilde (~) to user's home directory.

Earlier a directory with name (~) was created in the CWD.

Ex: `oc adm must-gather --image= --dest-dir=~/mg` will create a directory named 'mg' in user's home directory.

Regards


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -191,6 +191,13 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 	if len(o.DestDir) == 0 {
 		o.DestDir = fmt.Sprintf("must-gather.local.%06d", rand.Int63())
 	}
+	if strings.HasPrefix(o.DestDir, "~/") {
```
ardaguclu, line 194: This seems to be specific to Unix and does not work on Windows. Besides, nearly all commands accept full path and does not accept `~/`. I'd prefer leave this as is.

PR #1708, opened 2024-03-13
WRKLDS-1128: cli/admin/release/git: use optimized git flags

AlexNPavel:
This PR adds the `--filter=blob:none --bare` flags to the `git clone` commands, which significantly reduces disk space usage (for example, for 4.14.0->4.15.0, reduction from 31GB to 900MB).

### DIFF
--- a/pkg/cli/admin/release/git.go
+++ b/pkg/cli/admin/release/git.go
@@ -68,7 +68,7 @@ func (g *git) ChangeContext(path string) (*git, error) {
 }
 
 func (g *git) Clone(repository string, out, errOut io.Writer) error {
-	cmd := exec.Command("git", "clone", repository, g.path)
+	cmd := exec.Command("git", "clone", "--filter=blob:none", "--bare", repository, g.path)
 	cmd.Stdout = out
 	cmd.Stderr = errOut
 	return cmd.Run()
--- a/pkg/cli/admin/release/git.go
+++ b/pkg/cli/admin/release/git.go
@@ -68,7 +68,7 @@ func (g *git) ChangeContext(path string) (*git, error) {
 }
 
 func (g *git) Clone(repository string, out, errOut io.Writer) error {
-	cmd := exec.Command("git", "clone", "--filter=blob:none", "--bare", repository, g.path)
+	cmd := exec.Command("git", "clone", "--filter=blob:none", "--bare", "--origin="+remoteNameForRepo(repository), repository, g.path)
 	cmd.Stdout = out
 	cmd.Stderr = errOut
 	return cmd.Run()
@@ -303,7 +303,7 @@ func ensureCloneForRepo(dir string, repo string, alternateRepos []string, out, e
 			return nil, err
 		}
 	} else {
-		if err := ensureRemoteForRepo(extractedRepo, repo); err != nil {
+		if err := ensureFetchedRemoteForRepo(extractedRepo, repo); err != nil {
 			return nil, err
 		}
 	}
@@ -312,7 +312,7 @@ func ensureCloneForRepo(dir string, repo string, alternateRepos []string, out, e
 		if altRepo == repo {
 			continue
 		}
-		if err := ensureRemoteForRepo(extractedRepo, altRepo); err != nil {
+		if err := ensureFetchedRemoteForRepo(extractedRepo, altRepo); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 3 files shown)


ardaguclu:
Could you please create a Jira ticket and link this PR and https://github.com/openshift/oc/pull/1706 to have a place containing historical context?

pkg/cli/admin/release/git.go
```diff
@@ -68,7 +68,7 @@ func (g *git) ChangeContext(path string) (*git, error) {
 }
 
 func (g *git) Clone(repository string, out, errOut io.Writer) error {
-	cmd := exec.Command("git", "clone", repository, g.path)
+	cmd := exec.Command("git", "clone", "--filter=blob:none", "--bare", repository, g.path)
```
ardaguclu, line 71: This change looks reasonable and it will indeed reduce the disk usage size substantially but we need to ensure that this won't break anything. For example, when git command fails, command executes `git fetch` https://github.com/openshift/oc/blob/0d093afbb0d5b706be03a16d79e8b9b436453df7/pkg/cli/admin/release/git.go#L183, what will happen in that case?

ardaguclu: Additionally, do we test this against squash commits https://github.com/openshift/oc/blob/0d093afbb0d5b706be03a16d79e8b9b436453df7/pkg/cli/admin/release/git.go#L204

AlexNPavel: In regards to squash commits, I have tested changelogs for 4.14.0->4.15.0, which includes repos that use squash commits. The changelogs produced are identical.

AlexNPavel:
/hold

These changes break `oc adm release extract --git=...`. I'll need to make some changes to allow the `CheckoutCommit` function to get a full clone instead of the stripped down one we're using for logs.

AlexNPavel:
The `oc adm release extract --git...` command has been fixed by creating a function that converts bare repositories into standard ones. This does not disable the partial clone mode for the remotes, so the amount of space used is significantly less than before (31GB->22GB for 4.15.0). As a result, the command is now significantly faster on first run, but is somewhat slower when checking out previous releases (for instance, checking out 4.14.0 after 4.15.0) as is needs to download the missing objects.

pkg/cli/admin/release/git.go
```diff
@@ -97,7 +98,13 @@ func (g *git) VerifyCommit(repo, commit string) (bool, error) {
 	return err == nil, nil
 }
 
-func (g *git) CheckoutCommit(repo, commit string) error {
+func (g *git) CheckoutCommit(repo, commit string, out, errOut io.Writer) error {
+	// to reduce size requirements, clones are normally bare; to checkout a git commit, we must convert it to a normal
+	// git directory
+	if err := g.ensureFullClone(out, errOut); err != nil {
```
ardaguclu, line 104: If we have to perform full clone eventually, how do we reduce the disk usage size?. It seems to me that we are reducing disk size insignificant comparing to the complexity we are adding according to the latest status of this PR.

What was the problem in Github API side, rate limit errors from Github, different outputs, etc.?

AlexNPavel: The `CheckoutCommit` command is only executed if a user runs the `oc adm release extract --git=...` command, and `CheckoutCommit` is the only function using `ensureFullClone`. It does not get executed by `oc release adm info --changelog/--bugs`, so the changelog generation remains unchanged, only requiring ~1GB of space.

The problem on the Github API side was speed. Running it locally with a github token would take about 5 minutes, and even running in the app.ci cluster using Prow's Github API cache, the changelog generation still took 2 minutes due to to the speed at which the API calls were returned (uncached in app.ci took ~3 minutes). I tested running all the API calls in parallel as well, but that only reduced it to ~1.5 minutes due to rate limiting. If you don't use an access token, you would also immediately hit the hourly rate limit for changelogs that weren't small nightly changes. With this PR, the generation of the first changelog takes about 2-3 minutes locally and 1 minute in the app.ci cluster. Future runs also only take a couple seconds, whereas the github method doesn't improve for future runs.

ardaguclu (commented):
I haven't tested code but bare cloning first and only full cloning when it is inevitable sounds good to me. Left only minor comment.

pkg/cli/admin/release/info.go
```diff
@@ -91,9 +91,8 @@ func NewInfo(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Comm
 			a tag when verifying an image is recommended since it ensures an attacker cannot trick you
 			into installing an older, potentially vulnerable version.
 
-			The --bugs and --changelog flags will use git to clone the source of the release and display
-			the code changes that occurred between the two release arguments. This operation is slow
-			and requires sufficient disk space on the selected drive to clone all repositories.
+			The --bugs and --changelog flags will use git to clone the git history of the release and display
+			the code changes that occurred between the two release arguments.
```
ardaguclu, line 95: nit: What about still mentioning sufficient disk space as current state, since we may still perform full clone. WDYT?

AlexNPavel: I've added back the note on speed and disk space.

ardaguclu:
Thanks;
/approve
I'll defer @bradmwilliams for last tag

bradmwilliams (approved):
/lgtm

PR #1714, opened 2024-03-22
NO-JIRA: update featuregate diff for clusterprofiles

deads2k:
FeatureGates can vary by clusterprofile, this adds handling for both new and old style and updates the format to be  table.


deads2k:
/hold cancel

includes revert of https://github.com/openshift/oc/pull/1708 so we can use the newly built version while the image bits are worked out.

deads2k:
manually tested by Brad.

PR #1715, opened 2024-03-27
cli/admin/release/git: rename origin after clone

AlexNPavel:
This PR removes the `--origin` flag from the initial git clone as it is incompatible when used with the `--bare` flag in the version of git in RHEL 8 and thus the current 4.16 cli and tools images. Instead, we now rename the `origin` remote to the proper hashed name after the initial clone finishes.


AlexNPavel:
/hold

Seeing a different weird error from git within the container that I did not see locally...

AlexNPavel:
I was able to fix the issue I was seeing. However, due to some changes attempting to move the images to RHEL 9, this PR may not be needed, so I'll leave this PR on hold.

ardaguclu:
> I was able to fix the issue I was seeing. However, due to some changes attempting to move the images to RHEL 9, this PR may not be needed, so I'll leave this PR on hold.

@AlexNPavel we are trying to move to RHEL9. However, it will not happen soon. I think, we shouldn't leave RHEL 8 based oc with a bug and the current solution seems to me like a workaround (and we should make additional changes when we move to RHEL 9). 

I'm not sure about the exact reasons but this PR https://github.com/openshift/oc/pull/1714 also tries to revert bare git config changes. What do you think, if we revert this PR https://github.com/openshift/oc/pull/1708 and reintroduce it when default oc is based on RHEL9?.

bradmwilliams:
Hey @ardaguclu,
Some background...
Due to some technical issues found after #1708 merged, we've had to drop our version of the `oc` binary from `4.16` back to `4.15` while the issues get sorted.  I believe that @AlexNPavel has a potential solution, but has been plagued with the failing test issues related to library-go (or so it seems).  At the same time @deads2k has been working on #1714 and would like to get it out ASAP.  Unfortunately, for us to take that fix, we'll need to bump back up to the `4.16` version of `oc`.  So, he included the revert in his PR to facilitate that happening.
What's the status on the RHEL9 switch? 
I saw https://github.com/openshift/release/pull/49421 the other day and secretly hoped we could avoid all this mess, but fear that's not going to be the case...

ardaguclu:
> Hey @ardaguclu, Some background... Due to some technical issues found after #1708 merged, we've had to drop our version of the `oc` binary from `4.16` back to `4.15` while the issues get sorted. I believe that @AlexNPavel has a potential solution, but has been plagued with the failing test issues related to library-go (or so it seems). At the same time @deads2k has been working on #1714 and would like to get it out ASAP. Unfortunately, for us to take that fix, we'll need to bump back up to the `4.16` version of `oc`. So, he included the revert in his PR to facilitate that happening. What's the status on the RHEL9 switch? I saw [openshift/release#49421](https://github.com/openshift/release/pull/49421) the other day and secretly hoped we could avoid all this mess, but fear that's not going to be the case...

Thanks for the historical context. Since oc (and `cli` image) is used extensively, we are very cautiously making changes and fixing bits (e.g. https://github.com/openshift-metal3/dev-scripts/pull/1646), but it will take time and will not happen soon. I think, best route is to revert this (we can live without slow but working oc adm release info) and once oc is based on RHEL9, we can reintroduce it.

bradmwilliams:
> I think, best route is to revert this (we can live slow but working oc adm release info) and once oc is based on RHEL9, we can reintroduce it.

Given the current situation, I agree.

PR #1717, opened 2024-03-28
OCPBUGS-31484: Remove some of newapp unit tests relying on external deprecated images

ardaguclu:
We invested a good amount of effort to maintain the unit tests for `oc newapp` (aka `TestNewApp*`) by migrating to the equivalents that are alive, when the images they use were removed. However, after all the images have been removed from `quay.io/centos7` (e.g. `quay.io/centos7/ruby-27-centos7`, `quay.io/centos7/redis-5-centos7`), it seems that there is nowhere we can go.

Creating new images and maintaining them separately just for these tests would be an option but it doesn't seems to be the reasonable one. That's why, this PR removes the tests depending on these images. Because we don't/won't make changes to this command (`oc newapp`).

### DIFF
--- a/pkg/helpers/newapp/newapptest/newapp_test.go
+++ b/pkg/helpers/newapp/newapptest/newapp_test.go
@@ -661,49 +661,6 @@ func TestNewAppRunAll(t *testing.T) {
 			expectInsecure:  sets.NewString("example"),
 		},
 		{
-			name: "emptyDir volumes",
-			config: &cmd.AppConfig{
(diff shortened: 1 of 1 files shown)


ardaguclu:
@ingvagabund could you PTAL?. Thanks.

ardaguclu:
This PR is blocked by https://github.com/openshift/origin/pull/28675

soltysh (approved):
/lgtm
/hold
for https://github.com/openshift/origin/pull/28675

PR #1719, opened 2024-03-29
make output table slightly easier to read

deads2k:
This hides the "new disabled field", shortens some text, adds hard line breaks to make a prettier table.

/assign @bradmwilliams


bradmwilliams (approved):
/lgtm

PR #1721, opened 2024-04-03
OCPBUGS-31615: Add help text the behavior of filter-by-os against single image

ardaguclu:
`--filter-by-os` flag is ignored (even with invalid value `--filter-by-os=invalid`), when it is used against an single image. This PR adds this into the long text.


ardaguclu:
@ingvagabund what is your opinion about this?. Thanks.
/cc @kasturinarra @zhouying7780

PR #1725, opened 2024-04-04
WRKLDS-1171: Add openshift-install-fips target to oc release extract

zaneb:
Allow the openshift-baremetal-install binary to also be extracted using the openshift-install-fips command. This is clearer for users who need this binary only for FIPS compliance reasons.

### DIFF
--- a/pkg/cli/admin/release/extract_tools.go
+++ b/pkg/cli/admin/release/extract_tools.go
@@ -449,6 +449,18 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			ArchiveFormat:        "openshift-baremetal-install-linux-%s.tar.gz",
 		},
 		{
+			OS:       "linux",
+			Arch:     targetReleaseArch,
+			Command:  "openshift-install-fips",
+			Optional: true,
+			Mapping:  extract.Mapping{Image: "baremetal-installer", From: "usr/bin/openshift-install"},
+
+			Readme:               readmeInstallUnix,
+			InjectReleaseImage:   true,
+			InjectReleaseVersion: true,
+			ArchiveFormat:        "openshift-install-rhel-%s.tar.gz",
+		},
+		{
 			OS:      "linux",
 			Arch:    targetReleaseArch,
 			Command: "ccoctl",


pkg/cli/admin/release/extract_tools.go
```diff
@@ -448,6 +448,18 @@ func (o *ExtractOptions) extractCommand(command string) error {
...
+			OS:       "linux",
+			Arch:     targetReleaseArch,
+			Command:  "openshift-install-fips",
+			Optional: true,
+			Mapping:  extract.Mapping{Image: "baremetal-installer", From: "usr/bin/openshift-install"},
+
+			Readme:               readmeInstallUnix,
+			InjectReleaseImage:   true,
+			InjectReleaseVersion: true,
+			ArchiveFormat:        "openshift-baremetal-install-linux-%s.tar.gz",
```
ardaguclu, line 461: Since this target's tarball name is same with the above one, there might be unexpected binary overrides. But I verified that `--tools` flag ignores the targets that are marked as Optional, so that these ArchiveFormats are not used at all and we are good.

pkg/cli/admin/release/extract_tools.go
```diff
@@ -448,6 +448,18 @@ func (o *ExtractOptions) extractCommand(command string) error {
 			InjectReleaseVersion: true,
 			ArchiveFormat:        "openshift-baremetal-install-linux-%s.tar.gz",
 		},
+		{
+			OS:       "linux",
+			Arch:     targetReleaseArch,
+			Command:  "openshift-install-fips",
```
ardaguclu, line 454: There is an option to change the resulted binary name by using `TargetCommandName` as in https://github.com/openshift/oc/blob/425af75f33def2d4dfedbc6ca8a0465be84e343f/pkg/cli/admin/release/extract_tools.go#L279, if you'd want to. Currently, it extracts `openshift-install-fips`.

ardaguclu (commented):
/lgtm
/hold
in case you'd want to add `TargetCommandName` field and also it is better to see a green IPv6.

ardaguclu:
Just for the historical context, this will only work against 4.16 release payloads? or is this binary generated in 4.15 (probably not in the olders)?

pkg/cli/admin/release/extract_tools.go
zaneb: Thanks, I updated this one anyway.

pkg/cli/admin/release/extract_tools.go
zaneb: I think that's what we want to minimise confusion, so I will leave it as is.

zaneb:
> Just for the historical context, this will only work against 4.16 release payloads? or is this binary generated in 4.15 (probably not in the olders)?

It's the same binary as openshift-baremetal-installer, so it has been there since 4.4 or something.

PR #1690, opened 2024-02-22
OTA-1174: `upgrade status`: --details shows descriptions and references

petr-muller:
With --details, every insight will have two more fields - description and remediation URL (runbook, documentation...). Output is not a table, but a `oc describe`-like tree output. Move the condition message from the insight summary to its description (which is only shown with --details). Remediation links point to o/runbooks runbook for the matching alert. Condition messages are propagated via an insight description, which is only shown with `--details` in a tree-like output. This allows us to keep linebreaks originally present in the message, we just need to pad each line enough to align it with the indentation of the message start. Incomplete insights are not not actionable enough to show them to the user, so we filter them out while processing.

PR #1736, opened 2024-04-18
OTA-1174: `upgrade status`: Expand health insights with `--detailed=health`

petr-muller:
### Review notes:
Please focus on the structure, not on the actual content of the expaned insights. Insights are data, we can easily improve them. Especially the content of the node-related insights can be improved (and generating single insight per node is probably overwhelming) and https://github.com/openshift/oc/pull/1690 goes further in that direction, but  **this** PR concerns mostly about the new `--detailed=health` formatting, not with existing insight content.

Some smaller changes are separated into small commits, I recommend reviewing by commit.

---

`--details=health` shows reference and description

Every insight will have two more fields:
 - description: human-oriented, potentially longer form message that should explain what is going on and should indicate a direction for fixing the problem
 - remediation URL: link to a helpful runbook or documentation location

In detailed mode the output is arranged not in a table, but as a `oc describe`-like tree.

---

populate descriptions and refs for insights

Descriptions were populated with `message` strings coming from respective sources: either a `ClusterOperator` condition or MCO annotation.

References were populated with existing runbooks for matching `ClusterOperator` alerts, or with links to existing MCO documentation for the given domain.

---

pad linebreaks in descriptions

Descriptions (with linebreaks) are no longer shown in tabular output, just in detailed tree one. Hence, we may keep them, but should pad them respectively so the blocks of text are properly indented.

---

ignore incomplete insights

Incomplete insights are not not actionable enough to show them to the user, so we filter them out while processing.

---

always tabulate 'All is well'

"All is well" is an artificial insight so that we can show *something* when everything goes fine. There is no point in showing details for it because all detailed fields are empty.


wking (approved):
/lgtm

wking:
[`The openshift-console downloads pods ... should be scheduled on different nodes` business][1] is orthogonal to this change.

/retest-required

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/1736/pull-ci-openshift-oc-master-e2e-aws-ovn-serial/1781097386169339904

PR #1740, opened 2024-04-22
OCPBUGS-33896: add alert data to upgrade health in oc adm upgrade status

PratikMahajan:
adds alerts that fire during the upgrade to
`upgrade health` section.

by default all the alerts that started firing after initiating the
upgrade will appear in the upgrade health section

we also have allowed alerts that will show alerts that started 
firing before the upgrade was started.

### DIFF
--- a/pkg/cli/admin/upgrade/status/alerts.go
+++ b/pkg/cli/admin/upgrade/status/alerts.go
@@ -30,6 +30,7 @@ type AlertLabels struct {
 type AlertAnnotations struct {
 	Description string `json:"description,omitempty"`
 	Summary     string `json:"summary,omitempty"`
+	Runbook     string `json:"runbook_url,omitempty"`
 }
 
 type Alert struct {
@@ -59,13 +60,18 @@ func parseAlertDataToInsights(alertData AlertData, startedAt time.Time) []update
 		if startedAt.After(alert.ActiveAt) && !allowedAlerts.Contains(alert.Labels.AlertName) {
 			continue
 		}
+		if alert.State == "pending" {
+			continue
+		}
 		updateInsights = append(updateInsights, updateInsight{
 			startedAt: alert.ActiveAt,
 			impact: updateInsightImpact{
-				level:      alertImpactLevel(alert.Labels.Severity),
-				impactType: unknownImpactType,
-				summary:    "Alert: " + alert.Annotations.Summary,
+				level:       alertImpactLevel(alert.Labels.Severity),
+				impactType:  unknownImpactType,
+				summary:     "Alert: " + alert.Annotations.Summary,
+				description: alert.Annotations.Description,
 			},
+			remediation: updateInsightRemediation{reference: alert.Annotations.Runbook},
 			scope: updateInsightScope{
 				scopeType: scopeTypeCluster,
 				resources: []scopeResource{{
--- a/pkg/cli/admin/upgrade/status/alerts.go
+++ b/pkg/cli/admin/upgrade/status/alerts.go
@@ -1,0 +1,93 @@
+package status
+
+import (
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// Alerts that will be included in the health upgrade evaluation, even if they were triggered before the upgrade began.
+type AllowedAlerts map[string]struct{}
+
+var allowedAlerts AllowedAlerts = map[string]struct{}{
+	"PodDisruptionBudgetLimit":   {},
+	"PodDisruptionBudgetAtLimit": {},
+}
+
+func (al AllowedAlerts) Contains(alert string) bool {
+	_, exists := al[alert]
+	return exists
+}
+
+type AlertLabels struct {
+	AlertName string `json:"alertname,omitempty"`
+	Name      string `json:"name,omitempty"`
+	Namespace string `json:"namespace,omitempty"`
+	Reason    string `json:"reason,omitempty"`
+	Severity  string `json:"severity,omitempty"`
+}
+
+type AlertAnnotations struct {
+	Description string `json:"description,omitempty"`
+	Summary     string `json:"summary,omitempty"`
+}
+
+type Alert struct {
+	Labels                  AlertLabels      `json:"labels,omitempty"`
+	Annotations             AlertAnnotations `json:"annotations,omitempty"`
+	State                   string           `json:"state,omitempty"`
+	Value                   string           `json:"value,omitempty"`
+	ActiveAt                time.Time        `json:"activeAt,omitempty"`
+	PartialResponseStrategy string           `json:"partialResponseStrategy,omitempty"`
+}
+
+// Stores alert data returned by thanos
+type AlertData struct {
+	Status string `json:"status"`
+	Data   Data   `json:"data"`
+}
+
+type Data struct {
+	Alerts []Alert `json:"alerts"`
+}
+
+func parseAlertDataToInsights(alertData AlertData, startedAt time.Time) []updateInsight {
+	var alerts []Alert = alertData.Data.Alerts
+	var updateInsights []updateInsight = []updateInsight{}
+
+	for _, alert := range alerts {
+		if startedAt.After(alert.ActiveAt) && !allowedAlerts.Contains(alert.Labels.AlertName) {
+			continue
+		}
+		updateInsights = append(updateInsights, updateInsight{
+			startedAt: alert.ActiveAt,
+			impact: updateInsightImpact{
+				level:      alertImpactLevel(alert.Labels.Severity),
+				impactType: unknownImpactType,
+				summary:    "Alert: " + alert.Annotations.Summary,
+			},
+			scope: updateInsightScope{
+				scopeType: scopeTypeCluster,
+				resources: []scopeResource{{
+					kind:      scopeGroupKind{group: configv1.GroupName, kind: "Alert"},
+					namespace: alert.Labels.Namespace,
+					name:      alert.Labels.AlertName,
+				}},
+			},
+		})
+	}
+	return updateInsights
+}
+
+func alertImpactLevel(ail string) impactLevel {
+	switch ail {
+	case "warning":
+		return warningImpactLevel
+	case "critical":
+		return criticalInfoLevel
+	case "info":
+		return infoImpactLevel
+	default:
+		return infoImpactLevel
+	}
+}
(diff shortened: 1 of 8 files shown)


pkg/cli/admin/upgrade/status/status.go
```diff
@@ -240,6 +264,20 @@ func (o *options) Run(ctx context.Context) error {
...
+	var alertData AlertData
+
+	alertBytes, err := inspectalerts.GetAlerts(ctx, o.getRoute, o.RESTConfig.BearerToken)
+	if err != nil {
+		return err
+	}
+
+	// Unmarshal the JSON data into the struct
+	if err := json.Unmarshal(alertBytes, &alertData); err != nil {
+		fmt.Println("Error parsing JSON: ", err)
```
wking, line 276: `options.Run` seems to prefer returning the error, like:

```go
return fmt.Errorf("unmarshaling alerts: %w", err)
```

and that's what you're doing a few lines up with the `GetAlerts` error (although there you're assuming that `GetAlerts` fails with sufficient context to not need additional wrapping).  I could see us wanting to continue processing if we failed to `GetAlerts`; maybe Thanos on the cluster is having some kind of problem.  But at this point, when we've successfully retrieved JSON from `GetAlerts`, I'd expect us to fail out if we failed to unmarshal.

pkg/cli/admin/upgrade/status/alerts.go
```diff
@@ -0,0 +1,73 @@
...
+	Alerts []Alert `json:"alerts"`
+}
+
+func parseAlertDataToInsights(alertData AlertData, startedAt time.Time) []updateInsight {
+	var alerts []Alert = alertData.Data.Alerts
+	var updateInsights []updateInsight
+
+	for _, alert := range alerts {
+		if startedAt.After(alert.ActiveAt) {
+			continue
```
wking, line 44: No worries if you want to push back, but [OTA-1157](https://issues.redhat.com/browse/OTA-1157)'s current asks include:

> Alerts that started firing before the upgrade but are present on an allowlist (hardcoded for now) of alerts relevant for update

so we can pass along `PodDisruptionBudgetLimit` and `PodDisruptionBudgetAtLimit` and such, even if they'd started firing before this update was initiated.

pkg/cli/admin/upgrade/status/alerts.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		updateInsights = append(updateInsights, updateInsight{
+			startedAt: alert.ActiveAt,
+			impact: updateInsightImpact{
+				level:      alertResponseStrategyToImpact(alert.Level),
```
wking, line 49: I'm not sure what `partialResponseStrategy` is about (and that's what you're currently mapping to `Level`).  But for `impactLevel` here, I think you want to consume `alert.Labels.Severity`:

```console
$ OC_ENABLE_CMD_INSPECT_ALERTS=true oc adm inspect-alerts | jq '.data.alerts[].labels.severity' | sort | uniq -c
      9 "critical"
      2 "info"
      1 "none"
     26 "warning"
```

Without teaching this function alert-specific issues, I'd expect something like the following mapping:

* `none` (this is `Watchdog`, and we don't need to pass that along, although we might want to warn when it isn't present?).
* `info`, map to `infoImpactLevel`.
* `warning`, map to `warningImpactLevel`.
* `critical`, map to `errorImpactLevel`.  Seems unlikely that anything that's only coming in through alerts deserves `criticalInfoLevel`, and if we decide it does, we probably want to hand-pick those alert names.  For example:

```console
$ OC_ENABLE_CMD_INSPECT_ALERTS=true oc adm inspect-alerts | jq -c '.data.alerts[] | select(.labels.severity == "critical") | {alertname: .labels.alertname, annotations}' | sort | uniq -c
      9 {"alertname":"promtail-dropping-data","annotations":{"message":"Promtail is unable to ship logs and dropping data."}}
```

is currently firing in build02, and while that is claiming it is `critical`, it is unlikely to actually matter vs. cluster updates.

pkg/cli/admin/upgrade/status/status.go
```diff
@@ -76,6 +83,23 @@ func (o *options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string
...
+
+	o.RESTConfig = cfg
+
+	routeClient, err := routev1client.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
+	o.getRoute = func(ctx context.Context, namespace string, name string, opts metav1.GetOptions) (*routev1.Route, error) {
+		return routeClient.Routes(namespace).Get(ctx, name, opts)
+	}
```
petr-muller, line 101: Seems like we only need `BearerToken` from the `RESTConfig` so we may as well only save that and not the full config. 

But looking at how `getRoute` and the token is actually used, I think we can build an alert-fetching closure function here, like this:

```go
routeGeter := func(ctx context.Context, namespace string, name string, opts metav1.GetOptions) (*routev1.Route, error) {
		return routeClient.Routes(namespace).Get(ctx, name, opts)
}
o.getAlerts = func(ctx context.Context) ([]byte, error) {
  return inspectalerts.GetAlerts(ctx, routeGetter, cfg.BearerToken
}
```

and then the callsite in `Run()` can just call it easily with just `ctx`, the internal details like routeGetter and token may not be passed around

pkg/cli/admin/upgrade/status/status.go
```diff
@@ -240,6 +264,20 @@ func (o *options) Run(ctx context.Context) error {
...
+	if err != nil {
+		fmt.Println("Unable to fetch alerts from thanos, ignoring alerts in 'Update Health': ", err)
+	}
+
+	// Unmarshal the JSON data into the struct
+	if err := json.Unmarshal(alertBytes, &alertData); err != nil {
+		return fmt.Errorf("unmarshaling alerts: %w", err)
+	}
+
+	updateInsights = append(updateInsights, parseAlertDataToInsights(alertData, startedAt)...)
```
petr-muller, line 279: These two should be conditional only when GetAlerts was successful.

`parseAlertDataToInsights` should be unit-tested

pkg/cli/admin/upgrade/status/health.go
```diff
@@ -13,6 +13,7 @@ type scopeType string
 const (
 	scopeTypeControlPlane scopeType = "ControlPlane"
 	scopeTypeWorkerPool   scopeType = "WorkerPool"
+	scopeTypeAlert        scopeType = "Alert"
```
petr-muller, line 16: I don't think we should have a new scope type - I guess assigning alerts to either control plane or worker pool domain is not easy because the alerts do not carry that metadata?

I think that for now we could have a scope type `Cluster` which we could use for items that we cannot pin to a more precise scope.

pkg/cli/admin/upgrade/status/health.go
```diff
@@ -21,6 +22,7 @@ const (
 	scopeKindClusterOperator   allowedScopeKind = "ClusterOperator"
 	scopeKindNode              allowedScopeKind = "Node"
 	scopeKindMachineConfigPool allowedScopeKind = "MachineConfigPool"
+	scopeKindAlert             allowedScopeKind = "Alert"
```
petr-muller, line 25: I do not think we need this. This is used to tie insights to resources in the cluster - either we can do that and we will know the exact kind, or we can and we can have the list of related resources to be empty.

pkg/cli/admin/upgrade/status/health.go
```diff
@@ -78,6 +80,7 @@ type impactType string
 // considered whether these are exactly the ones that we need.
 const (
 	noneImpactType                    impactType = "None"
+	alertImpactType                   impactType = "Alert"
```
petr-muller, line 83: "Alert" is not an impact on the cluster though. This is a manifestation of a similar problem like https://issues.redhat.com/browse/OTA-1173 : we see an alert but do not know what it actually means.

Maybe we should have an impact type `Unknown` for now, now sure :(

pkg/cli/admin/upgrade/status/status.go
```diff
@@ -76,6 +83,23 @@ func (o *options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string
 		o.mockData.nodesPath = strings.Replace(o.mockData.cvPath, cvSuffix, "-node.yaml", 1)
 	}
 
+	kcmdutil.RequireNoArguments(cmd, args)
```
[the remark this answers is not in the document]
wking: We already cover this [here][1].  Using [`RequireNoArguments`][2] there would be better for consistency, but is orthogonal to the rest of this pull request.

[1]: https://github.com/openshift/oc/blame/5ffe8230ec8aa94dc37dfa589c108402d12f8615/pkg/cli/admin/upgrade/status/status.go#L83-L85
[2]: https://pkg.go.dev/k8s.io/kubectl/pkg/cmd/util#RequireNoArguments

pkg/cli/admin/upgrade/status/status.go
```diff
@@ -260,6 +279,21 @@ func (o *options) Run(ctx context.Context) error {
 		fmt.Fprintf(o.ErrOut, "warning: No current %s info, see `oc describe clusterversion` for more details.\n", clusterStatusFailing)
 	}
 
+	// get the alerts for the cluster. if we're unable to fetch the alerts, we'll let the user know that alerts
+	// are not being fetched, but rest of the command should work.
+	var alertData AlertData
+	alertBytes, err := o.getAlerts(ctx)
+	if err != nil {
+		fmt.Println("Unable to fetch alerts from thanos, ignoring alerts in 'Update Health': ", err)
```
wking, line 287: nit: "thanos" -> "Thanos", to match the casing preferred by the Thanos folks, for [example][1]:

> Join users and companies that are using Thanos in production.

[1]: https://thanos.io/

wking: although actually, this call-site has no idea that we're reaching out to Thanos; that's an implementation detail.  I'd expect `inspectalerts` to be providing sufficient context for debugging failed requests if we fail to fetch alerts, and here we can probably just do:

```golang
fmt.Fprintf(o.ErrOut, "warning: Unable to fetch alerts, ignoring alerts in 'Update Health': %v\n", err)
```

which will also get us logging this warning to `ErrOut`, as we already do [here][1].

[1]: https://github.com/openshift/oc/blob/5ffe8230ec8aa94dc37dfa589c108402d12f8615/pkg/cli/admin/upgrade/status/status.go#L260

petr-muller:
Neat work, looking forward to testing this on b02  :D 

We need to pass tests and if Trevor's review is addressed, we can merge this :+1:

pkg/cli/admin/upgrade/status/examples/4.15.0-ec2-unavailable-mco-20m.detailed-output
```diff
@@ -25,6 +25,15 @@ ip-10-0-4-159.us-east-2.compute.internal    Outdated     Pending   4.14.0-rc.3
 ip-10-0-99-40.us-east-2.compute.internal    Outdated     Pending   4.14.0-rc.3   ?     
 
 = Update Health =
+Message: Alert: Target disappeared from Prometheus target discovery.
```
wking, line 28: This one is `pending` in your `4.15.0-ec2-unavailable-mco-20m-alerts.json`.  We probably want to filter to just `.State == "firing"`?

wking (approved):
/lgtm

pkg/cli/admin/upgrade/status/status.go
```diff
@@ -245,6 +267,42 @@ func (o *options) Run(ctx context.Context) error {
...
+			}
+			alertTimesLen := len(startTimes)
+			if alertTimesLen != 0 {
+				// Sort the slice of time.Time values
+				sort.Slice(startTimes, func(i, j int) bool {
+					return startTimes[i].Before(startTimes[j])
+				})
+				startedAt = startTimes[alertTimesLen%3]
+			}
+		}
```
petr-muller, line 301: /lgtm cancel

I don't think this is needed at all now with doctored data - `startAt` is when the tool think the update started, which is driven by the CV. If I remove this block and rerun the test, I get a simple change in the expected output, which actually looks correct - the alertmanager alert started before the update, so we do *not* want  to see it, it is not on the allowlist.

```diff
--- a/pkg/cli/admin/upgrade/status/examples/4.15.0-ec2-unavailable-mco-20m.output
+++ b/pkg/cli/admin/upgrade/status/examples/4.15.0-ec2-unavailable-mco-20m.output
@@ -28,7 +28,6 @@ ip-10-0-99-40.us-east-2.compute.internal    Outdated     Pending   4.14.0-rc.3
 SINCE      LEVEL     IMPACT             MESSAGE
 20m24s     Error     API Availability   Cluster Operator machine-config is unavailable (MachineConfigControllerFailed)
 6m35s      Warning   Unknown            Alert: Pod has been in a non-ready state for more than 15 minutes.
-24h0m46s   Warning   Unknown            Alert: Receivers (notification integrations) are not configured on Alertmanager
 24h8m54s   Warning   Unknown            Alert: PodDisruptionBudgetAtLimit for pods <>
```

petr-muller (approved):

petr-muller (approved):
/lgtm
/approve

petr-muller (approved):

petr-muller:

This is a part of a TechPreview feature, all functionality gated behind an opt-in envvar

petr-muller:
OCPBUGS-33896: add alert data to upgrade health in oc adm upgrade status

petr-muller:
/retitle OCPBUGS-33896: add alert data to upgrade health in oc adm upgrade status

PR #1742, opened 2024-04-24
OCPBUGS-33897: `upgrade status`: control plane completion estimate

petr-muller:
Works like this:

- The procedure tries to estimate a model duration of the whole control plane update based on the available information. The estimate is treated as a mode or perhaps a target duration, not as a maximum tolerated value. Longer durations are tolerated, up to a limit (more on that later)
- Baseline duration of the control plane is 60m
- If there is a completed update in the CV history other than the latest (current update) or the first (which is usually an install), its duration is treated as a baseline instead of the 60m
- Because completion in the domain of updated COs is non linear (typically the first ~40% takes about a third of the time, updating KAS & etcd, then the highly parallelized runlevels around 50 update most of the COs except the last three very quickly, then the daemonset COs take some time and finally, the MCO takes about the third of the time, rebooting masters while ~97% COs are updated), CO completion is turned into "timewise" completion, using a curve-fitted update progress I measured on b01.
- Using the timewise completion, the elapsed time is projected into an estimate duration
- "Elapsed time" is a source of imprecision - for the projection to work, the elapsed time needs to be the time when the input completion was achieved. If we used current time instead of that, we would never converge because the projection would be always shooting further to the future for the given completion. Client does not have that information, so we use the most recent updated CO with Progressing=False (assumed to finished updating) or outdated CO with Progressing=True (assumed to started updating). It is not ideal and in autoscaling clusters, COs go Progressing=True/False for other reasons than updating, so this makes estimation at the end of the process, when only MCO is updating, more conservative than necessary. I was not able to solve this in a better way.
- We do not attempt to project from small elapsed times - we need at least five minutes of elapsed time (to last observed progress). Until that, we use the baseline.
- Using the current duration of the update and the projected duration, we compute the remaining time, and overestimate this by a certain amount (add 20%) - this is one way to tolerate longer durations than the estimate
- The estimate will never be precise enough, so when we estimate less than 10m, we stop showing the precise value and show just `<10m` . We do this even 10m _over_ the estimation (another toleration for slightly longer updates).
- When update for longer than we estimate for, we turn the assessment to `Progressing - Slow`
- After 10m *after* the estimated finish (including the 20% toleration) we stop trying to estimate and say `N/A; estimated duration was XXX`, and turn the assessment into stalled.


petr-muller:
Recording of a cluster bot Azure cluster updated two times in a row: https://asciinema.org/a/uCBDVPkufJGAui36N4FNoCooa

Results are fairly good actually, but I spotted several issues that the new commits should address:
- There are short overoptimistic periods caused when a CO 1) updates 2) hence increases completion 3) is still Progressing=True so its transition time is not recorded, resulting in higher completion in identical time as before. I changed this by treating most recent Progressing change as last observed change, no matter the state of the CO - this will cause the algorithm to create overpessimistic estimates that correct themselves when new data points arrive (be conservative and create positive suprises rather than negative is IMO better UX)
- The algorithm did not converge at the end of the process, when MCO takes a lot of time and daemonset-owning CO polluted the last observed progress by going progressing true/false on node reboots. I added a hack to treat MCO transition time as last observed progress when it is the only outdated CO and progressing=True. Ugly hack but can do the trick.

evakhoni:
pre-merge verified successfully (while accepting the known limitations)
/label qe-approved

pkg/cli/admin/upgrade/status/controlplane.go
```diff
@@ -288,6 +427,6 @@ const controlPlaneStatusTemplateRaw = `= Control Plane =
 Assessment:      {{ .Assessment }}
 Target Version:  {{ .TargetVersion }}
 Completion:      {{ printf "%.0f" .Completion }}%
-Duration:        {{ .Duration }}
+Duration:        {{ shortDuration .Duration }}{{ if .EstTimeToComplete }} (Est Time Remaining: {{ vagueUnder .EstTimeToComplete .EstDuration }}){{ end }}
```
wking, line 430: nit: can we use "Est. Time Remaining" or "estimated time remaining" or something that either has a `.` to say "I'm abbreviating!" or doesn't abbreviate?

petr-muller: This comes from Justin's [mockup](https://docs.google.com/presentation/d/1BtVSGWaG6LD0qye9BHPU5ssnEYTGPOlhHeogFfECUnk/edit#slide=id.g25513f80e3d_0_514)... 

I don't mind adding a dot there too much but I liked the output without it more actually. The output is quite technical and reads like a dashboard, abbreviations without a dot are not that uncommon in structured technical content.

petr-muller: Addressed in [503b0f5](https://github.com/openshift/oc/pull/1742/commits/503b0f5f9d7b20ecc871b87a16fdafe13338fd04)

wking (approved):
/lgtm

PR #1744, opened 2024-04-24
OTA-1279: pkg/cli/admin/upgrade/status: Move Failing from free-form to updateInsight

wking:
Structure this output, so it gets all the usual pretty-printing, detail, etc. that the sad-ClusterOperator conditions are getting. Sometimes `Failing` will complain about sad ClusterOperators, and in that case we'll double up on that messaging.  But we're punting on "consolidate when multiple `updateInsights` complain about the same root cause" for now.  And sometimes `Failing` will complain about other resources, such as [`ProgressDeadlineExceeded` operator Deployments][1], and in that case the information is only flowing out through ClusterVersion, and not via the other resources we check when rendering status.

[1]: https://github.com/openshift/cluster-version-operator/blob/1acac06742fb0e3e49ffe2294864007f26a7799d/lib/resourcebuilder/apps.go#L122C124-L122C148

### DIFF
--- a/pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
+++ b/pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
@@ -1,11 +1,5 @@
 An update is in progress for 4h3m46s: Error while reconciling 4.16.0-ec.3: the cluster operator machine-config is degraded
 
-Failing=True:
-
-  Reason: ClusterOperatorDegraded
-  Message: Cluster operator machine-config is degraded
-
-
 = Control Plane =
 Assessment:      Completed
 Completion:      100%
@@ -151,6 +145,15 @@ Message: Node build0-gstfj-ci-tests-worker-c-dcz9p is degraded
   Resources:
     nodes: build0-gstfj-ci-tests-worker-c-dcz9p
   Description: failed to drain node: build0-gstfj-ci-tests-worker-c-dcz9p after 1 hour. Please see machine-config-controller logs for more information
+
+Message: Cluster Version version is failing to proceed with the update (ClusterOperatorDegraded)
+  Since:       0s
+  Level:       Warning
+  Impact:      Update Stalled
+  Reference:   https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/ClusterOperatorDegraded.md
+  Resources:
+    clusterversions.config.openshift.io: version
+  Description: Cluster operator machine-config is degraded
 
 Message: Node build0-gstfj-ci-tests-worker-c-jq5rk is unavailable
   Since:       -
--- a/pkg/cli/admin/upgrade/status/controlplane.go
+++ b/pkg/cli/admin/upgrade/status/controlplane.go
@@ -18,6 +18,10 @@ const (
 	assessmentStatePending     assessmentState = "Pending"
 	assessmentStateExcluded    assessmentState = "Excluded"
 	assessmentStateDegraded    assessmentState = "Degraded"
+
+	// clusterStatusFailing is set on the ClusterVersion status when a cluster
+	// cannot reach the desired state.
+	clusterStatusFailing = v1.ClusterStatusConditionType("Failing")
 )
 
 type operators struct {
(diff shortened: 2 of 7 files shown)


wking:
[`OVNKubernetesResourceRetryFailure` firing][1] is unrelated to my change:

/retest-required

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/1744/pull-ci-openshift-oc-master-e2e-aws-ovn-serial/1783263632440692736

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
```diff
@@ -152,6 +146,15 @@ Message: Node build0-gstfj-ci-tests-worker-c-dcz9p is degraded
     nodes: build0-gstfj-ci-tests-worker-c-dcz9p
   Description: failed to drain node: build0-gstfj-ci-tests-worker-c-dcz9p after 1 hour. Please see machine-config-controller logs for more information
 
+Message: Cluster Version version is failing to reconcile (ClusterOperatorDegraded)
```
petr-muller, line 149: The message reads a bit cryptic to me, let's figure out how to make it more informative to users.

Thinking loud: CV has a `Failing=True` condition (during upgrade) when it is failing to proceed with the update process for some reason. Can we  just say that? I do not think we want to mention "Cluster Version version" at all the CV resource is just a whiteboard CVO uses to write things on.

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
```diff
@@ -152,6 +146,15 @@ Message: Node build0-gstfj-ci-tests-worker-c-dcz9p is degraded
...
+  Since:       0s
```
petr-muller, line 150: Is this just an artifact of how the mock execution calculates "now" (=last transition of any kind)? If would expect this to be a time since the transition to Failing=True happened.

petr-muller, line 150: Also, do we want to throttle when this shows up to the user? For CO conditions we wait for the condition to persist for some time before we start showing insights for them.

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
```diff
@@ -152,6 +146,15 @@ Message: Node build0-gstfj-ci-tests-worker-c-dcz9p is degraded
...
+  Level:       Warning
+  Impact:      Update Stalled
+  Reference:   https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/ClusterOperatorDegraded.md
```
petr-muller, line 153: This would be a great location for the generic troubleshooting document. Degraded CO happens to be good for this case but not all CV Failing=True is degraded operators.

wking: Not all `Failing=True` is because of ClusterOperators, but since openshift/cluster-version-operator#746, `Failing=True` ClusterVersion will trigger a `ClusterOperatorDegraded` alert with `name="version"` labels.  And the runbook will point to this same location.  So I think this is fine here, but we should extend the runbook text with a section on `Failing=True`.

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
wking: We throttle ClusterOperator conditions because we don't maintain those components, and sometimes it takes us a few years to talk their maintainers into denoising.  We maintain the cluster-version operator, so if `Failing` is noisy, we can denoise server-side as quickly as we like.

wking: yup, this is because test-now is "the largest `lastTransitionTime` on ClusterVersion conditions", and in the test fixtures, `Failing` is the one with the most recent change-time.  We could add 10m or something when generating "now" if you wanted to see something non-zero here?

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
wking: I was pattern-matching [fmt.Sprintf("Cluster Operator %s is degraded (%s)", name, degraded.Reason)][1].  I could see dropping the `name` `%s` based on reasoning like "odds that the cluster has multiple ClusterVersions is very low, and it's not worth those low-signal characters in the vast majority of clusters where there is one just to decrease ambiguity in the zero-to-few clusters where there are multiple ClusterVersions".

I'm agnostic on "failing to proceed with the update" instead of "failing to reconcile".  To me those mean the same thing, but possibly one or the other is more likely to be clear to more admins?

[1]: https://github.com/openshift/oc/blob/958116d19854df93308103afaae8a3da8af7d31a/pkg/cli/admin/upgrade/status/controlplane.go#L86

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
petr-muller: Makes sense :+1:

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
petr-muller: OK, potentially throttling some `Failing` causes is better to do in CVO, I agree. Good point.

petr-muller: It's fine, I saw the code doing the right thing after I made this comment and forgot to delete it.

pkg/cli/admin/upgrade/status/examples/4.16.0-ec2-control-plane-updated-pdb-prohibits-draining.detailed-output
petr-muller: Remembering my time at TP I think I would understand "failing to proceed with the update" much better than "[CVO] is failing to reconcile [a ClusterVersion]", because it does not require me to understand that "cluster upgrade" is actually CVO reconciling individual manifests (we did not have great docs about that until recently) and using CV resource (with a specific name, knowing it is a singleton) as spec/status API for that.

So I'd vote for changing this message to something that clearly says "cluster update is stuck because of this".

petr-muller: Remembering my time at TP I think I would understand "failing to proceed with the update" much better than "[CVO] is failing to reconcile [a ClusterVersion]", because it does not require me to understand that "cluster upgrade" is actually CVO reconciling individual manifests (we did not have great docs about that until recently) and using CV resource (with a specific name, knowing it is a singleton) as spec/status API for that (I think we also discouraged folks from touching `ClusterVersion` directly and rely on `oc` interpreting it, so we have some prior art in thinking that CV is closer to an internal detail than something admin should know about.

So I'd vote for changing this message to something that clearly says "cluster update is stuck because of this". I think the "failing to proceed with the update" is actually pretty good, maybe "CVO is failing to proceed with the update". I would keep the CV in the `Resources` field though, would just change the message only.

wking: Pivoted to failing to `proceed with the update` with 50e76ea90 -> 7977ec94e.

petr-muller (approved):
Thanks!
/lgtm
/approve

PR #1745, opened 2024-04-25
Update mustgather.go

danielyeap:
(1) ability to send local files into the mustgather pods (needed for custom mustgather tool) 
(2) ability to run the mustgather pod using a user-defined user (instead of fixed to the "default" user/serviceaccount)

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"bytes"
 
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


danielyeap:
Hi @ardaguclu @deads2k 
Anything pending on my side for this review?
Sorry...this is my first time.
Thanks.

ardaguclu:
@danielyeap thank you for spending time on this PR. However, I don't think we can merge the changes in here.
The reasons are:
* Adding a new flag would require strong justification and even the ones that are marked as hidden is there for a reason. Especially, there are multiple images exist for must-gather, we can't add new flags without notifying the all image owners. 
* We have to be backwards compatible. All the changes are only be applied if they are working against older images, OCP version, k8s version etc. I saw a couple of changes are not backwards compatible.

I hope it makes sense. Thanks again.

danielyeap:
Hi @ardaguclu 
Thanks for spending time to respond.

I hope you can help with a few questions:
(1) Can you point me to all the must-gather images?
(2) What are the changes that are not backward-compatible?

Lastly, while there are 2 features added through this code change, the truly critical one is the ability to send files into the mustgather pod. 

Any idea whether that is a planned feature for the future please?

Thanks again!

PR #1746, opened 2024-04-25
OCPBUGS-32786: Improve error message when binary is not in target list

ardaguclu:
SSIA


ardaguclu:
@ingvagabund could you PTAL, when you have a chance?. Thanks.

PR #1756, opened 2024-05-04
OCPNODE-2282: Support multiple ImageDigestMirrorSet objects in the th…

QiWang19:
…e single file

### DIFF
--- a/pkg/cli/image/strategy/onerror.go
+++ b/pkg/cli/image/strategy/onerror.go
@@ -1,8 +1,11 @@
 package strategy
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/cli/image/strategy/onerror.go
```diff
@@ -323,15 +327,27 @@ func readIDMSsFromFile(idmsFile string) ([]apicfgv1.ImageDigestMirrorSet, error)
...
 		return nil, fmt.Errorf("no data found in ImageDigestMirrorSet %s", idmsFile)
 	}
-	idmsObj, err := runtime.Decode(apicfgv1scheme.Codecs.UniversalDeserializer(), idmsData)
-	if err != nil {
-		return nil, fmt.Errorf("error decoding ImageDigestMirrorSet from %s: %v", idmsFile, err)
-	}
-	idms, ok := idmsObj.(*apicfgv1.ImageDigestMirrorSet)
-	if !ok {
-		return nil, fmt.Errorf("could not decode ImageDigestMirrorSet from %s", idmsFile)
+	reader := utilyaml.NewYAMLReader(bufio.NewReader(bytes.NewReader(idmsData)))
```
ardaguclu, line 330: How do we know that `idmsData` is not in json format and we are making a move that is not backwards-compatible?

QiWang19: it is compatible if idmsData is in json format,json is a subset of yaml. If the yaml sepatator is not found in the document the original byte can be unpacked to the go struct.

PR #1760, opened 2024-05-07
OCPBUGS-31086: Add checksum verification of mirrored image against the actual image

ardaguclu:
Network hiccups (and the other reasons) may corrupt the mirrored image and thus, it's sha256 checksum value may be different than the actual image's checksum value. However, that also leads that image integrity would not be satisfied. This PR adds checksum verification of the mirrored image by comparing the digest with the actual image's digest. If there is any mismatch found, it errors out to the user to force retry.

This change affects all the commands which are related to image mirroring (i.e. `adm release mirror`, `image mirror`, etc.).

### DIFF
--- a/pkg/cli/image/imagesource/file.go
+++ b/pkg/cli/image/imagesource/file.go
@@ -412,6 +412,16 @@ func (w *fileWriter) ReadFrom(r io.Reader) (int64, error) {
 		}
 		w.size = n
 		w.digest = blobDigest
+
+		if read, err := os.ReadFile(path); err != nil {
+			return 0, err
+		} else {
+			actualDigest := godigest.FromBytes(read)
+			if actualDigest.String() != w.digest.String() {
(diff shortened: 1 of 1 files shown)


ardaguclu:
/hold
until pre-merge tests prove that this change does not cause any regression.

ardaguclu:
/cc @lmzuccarelli  @sherine-k  @aguidirh 
Since this change also affects oc-mirror and I'd like to get your opinions about this. If that may create any regression, we can close.

lmzuccarelli:
@ardaguclu - Ok thanks for submitting this PR. We'll get QE to ensure their are no regressions.

ardaguclu:
> @ardaguclu - Ok thanks for the head up for oc-mirror. We'll get QE to ensure their are no regressions. 

@lmzuccarelli after careful thinking on this PR, it may bring less benefits than the risk of regression. Besides, current sanity checks in the library seems sufficient and thereby, I'm closing this PR.

Thanks for your comment.

/close

PR #1763, opened 2024-05-13
AUTH-482: set required-scc annotation to privileged for node debug pods

liouk:


ardaguclu:
/lgtm
/hold
@liouk in case you think pre-merge tests requiring for this change. Otherwise, feel free to cancel hold.

liouk:
@ardaguclu in non-runlevel namespaces, where SCC admission is enabled, node debug pods are getting anyway admitted with the `privileged` SCC; in runlevel namespaces, admission is disabled so this change isn't effective.

This means that overall we're not changing any behavior here, just making the required SCC explicit. Therefore I don't think pre-merge tests are required for this.

/hold cancel

PR #1765, opened 2024-05-14
OCPBUGS-32682: Fix the retrieval of API objects with an empty name

Davoska:
The main goal of this pull request is to fix an occurring error that resulted in the error
message: `error: resource name may not be empty`

This was caused by not checking the value of the
machineconfiguration.openshift.io/currentConfig and
machineconfiguration.openshift.io/desiredConfig annotations of nodes.

In the event, the annotation was not set or was empty, the command would
try to get a MachineConfig with an empty name from the API server.

The commit also adds additional functionalities and testing for cases
when different MCD annotations are missing.

The general strategy is to declare the states of nodes only when they
are provably in the specific states.

This pull request references https://issues.redhat.com//browse/OCPBUGS-32682


petr-muller (approved):

petr-muller:
/lgtm

I always forget :sweat:

PR #1766, opened 2024-05-15
OCPBUGS-33651: Add s390x based oc into extraction target

ardaguclu:
Default oc compiled on s390x is only working on RHEL9 (after we have changed the base image to RHEL9). Therefore, users still on s390x RHEL8 are getting failures. 

This PR adds s390x based oc (both oc.rhel8 and oc.rhel9) into the extraction list. So that, users can explicitly use s390x based oc.rhel8.


joepvd:
/hold
unsure if this fixes the actual report.

joepvd:
False alarm:
/unhold
/lgtm

ardaguclu:
https://github.com/openshift/origin/pull/28808 will fix the issue

PR #1773, opened 2024-05-18
OCPBUGS-33762: Hardcode resource groups/kinds for now

petr-muller:
Previous code used `GroupVersionKind` method on actual resources to determine group/kind strings for them, but client-go unfortunately drops `TypeMeta` on resources processed by typed client (https://github.com/kubernetes/client-go/issues/1328, https://github.com/kubernetes/client-go/issues/541). Fortunately we know what types we build insights for, so we can use appropriate strings from client code for groups and hardcode kinds ourselves.


wking (approved):
/lgtm

petr-muller:
Hm, will needs alerts to land first :/

PR #1774, opened 2024-05-18
OCPBUGS-33903: Simplify output when not updating

petr-muller:
Specific condition details are unlikely to be useful:

```console
$ oc adm upgrade status

  Reason: <none>
  Message: Cluster version is 4.14.1
```

Sometimes they can get even confusing, especially when the command picks up CVO `Failing=True` condition present before the update attempt:

```console
$ oc adm upgrade status --details=none

  Reason: MultipleErrors
  Message: Error while reconciling 4.16.0-0.nightly-2024-05-07-025557: an unknown error has occurred: MultipleErrors
```

I also changed one user-facing occurrence of "upgrade" to "update" as a rider in this PR


petr-muller:
I realized previous output could have been useful when the user would be expecting an update but it does not start for some reason: prechecks, download, verification. Maybe it would be useful if we conditionally emited something if spec desired differs from status desired?

pkg/cli/admin/upgrade/status/status.go
```diff
@@ -261,7 +261,7 @@ func (o *options) Run(ctx context.Context) error {
 		if message = progressing.Message; message == "" {
 			message = "<none>"
 		}
-		fmt.Fprintf(o.Out, "The cluster version is not updating (%s=%s).\n\n  Reason: %s\n  Message: %s\n", progressing.Type, progressing.Status, reason, strings.ReplaceAll(message, "\n", "\n  "))
+		fmt.Fprintf(o.Out, "The cluster is not updating.")
```
Davoska, line 264: We should include a newline in the output IMO.

```suggestion
		fmt.Fprintf(o.Out, "The cluster is not updating.\n")
```

Davoska, line 264: We can remove the logic regarding the `reason` and `message` variables in this condition as they are not used.

```go
	if progressing.Status != configv1.ConditionTrue && !isWorkerPoolOutdated {
		fmt.Fprintf(o.Out, "The cluster is not updating.\n")
		return nil
	}
```

Davoska:
> I realized previous output could have been useful when the user would be expecting an update but it does not start for some reason: prechecks, download, verification. 

That is true. 

> Maybe it would be useful if we conditionally emited something if spec desired differs from status desired?

I do like the idea of this. We could work on this in the future. If you think implementation-wise this is easy to implement and you are interested in this little improvement, feel free to include it in the PR. Otherwise, we can follow up on this.

pkg/cli/admin/upgrade/status/status.go
evakhoni: correct. just verified it prints % in stdout

evakhoni:
> > I realized previous output could have been useful when the user would be expecting an update but it does not start for some reason: prechecks, download, verification.
> 
> That is true.
> 
> > Maybe it would be useful if we conditionally emited something if spec desired differs from status desired?
> 
> I do like the idea of this. We could work on this in the future. If you think implementation-wise this is easy to implement and you are interested in this little improvement, feel free to include it in the PR. Otherwise, we can follow up on this.

imo can be a separate card in the future, maybe something a bit more detailed on why do we're unable to start the upgrade, and ideally what can we do about it :thinking:

pkg/cli/admin/upgrade/status/status.go
petr-muller: Addressed :+1:

wking (approved):
/lgtm

PR #1775, opened 2024-05-20
OCPBUGS-32108: stop loading legacy schemes on global scheme 

flavianmissi:
for some reason, when loading ungroupified legacy schemes `oc set env` overrides `apiVersion` values of openshift objects, removing the api group, which makes them invalid in recent versions of openshift.

### what changed
* remove global loading of legacy schemes
* create custom schemes for create and expose commands
* override upstream expose print flag type setter to use the custom expose scheme
 
### testing
* regression tests for `oc`
* ensure that `oc set env` no longer overrides the `apiVersion` of routes and deployment configs
* ensure we don't regress https://issues.redhat.com/browse/OCPBUGS-15772; in other words, make sure `oc set route-backends` and `oc edit route` work in microshift
* ?

### DIFF
--- a/pkg/cli/create/scheme.go
+++ b/pkg/cli/create/scheme.go
@@ -1,0 +1,37 @@
+package create
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	"github.com/openshift/api"
+	"github.com/openshift/api/quota"
+	"github.com/openshift/api/route"
+
+	schemehelper "github.com/openshift/oc/pkg/helpers/scheme"
+)
+
+var (
+	createCmdScheme = runtime.NewScheme()
+	createCmdCodecs = serializer.NewCodecFactory(createCmdScheme)
+)
+
+func init() {
+	utilruntime.Must(api.InstallKube(createCmdScheme))
+	schemehelper.InstallSchemes(createCmdScheme)
+	// All the other commands can use route object
+	// as CRD and there is no benefit installing route
+	// as native object which is normally managed
+	// by openshift-apiserver(microshift has no openshift-apiserver).
+	// However, the create command requires routes for PrintObj,
+	// so we define a custom scheme and install routes here.
+	// see https://github.com/openshift/oc/pull/1534 for background.
+	utilruntime.Must(route.Install(createCmdScheme))
+	utilruntime.Must(quota.Install(createCmdScheme))
+}
+
+func createCmdJSONEncoder() runtime.Encoder {
+	return unstructured.NewJSONFallbackEncoder(createCmdCodecs.LegacyCodec(createCmdScheme.PrioritizedVersionsAllGroups()...))
+}
--- a/pkg/cli/create/user_identity_mapping.go
+++ b/pkg/cli/create/user_identity_mapping.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
-	"k8s.io/kubectl/pkg/scheme"
 	"k8s.io/kubectl/pkg/util"
 	"k8s.io/kubectl/pkg/util/templates"
 
@@ -126,7 +125,7 @@ func (o *CreateUserIdentityMappingOptions) Run() error {
 		User:     corev1.ObjectReference{Name: o.User},
 	}
 
-	if err := util.CreateOrUpdateAnnotation(o.CreateSubcommandOptions.CreateAnnotation, mapping, scheme.DefaultJSONEncoder()); err != nil {
+	if err := util.CreateOrUpdateAnnotation(o.CreateSubcommandOptions.CreateAnnotation, mapping, createCmdJSONEncoder()); err != nil {
 		return err
 	}
 
--- a/pkg/cli/expose/expose.go
+++ b/pkg/cli/expose/expose.go
@@ -13,7 +13,6 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/kubectl/pkg/cmd/expose"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
-	"k8s.io/kubectl/pkg/scheme"
 	"k8s.io/kubectl/pkg/util"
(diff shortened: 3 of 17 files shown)


flavianmissi:
errors seem legit: 
```
error: missing apiVersion or kind and cannot assign it; no kind is registered for the type v1.ClusterResourceQuota in scheme \"k8s.io/kubectl/pkg/scheme/scheme.go:28\"
...
error: missing apiVersion or kind and cannot assign it; no kind is registered for the type v1.Route in scheme "k8s.io/kubectl/pkg/scheme/scheme.go:28"
```
will look into it later.

flavianmissi:
test failures seem unrelated
/retest

pkg/helpers/scheme/scheme.go
```diff
@@ -31,7 +29,6 @@ func InstallSchemes(scheme *apimachineryruntime.Scheme) {
 	utilruntime.Must(InstallNonCRDSecurity(scheme))
 	utilruntime.Must(template.Install(scheme))
 	utilruntime.Must(user.Install(scheme))
-	legacy.InstallExternalLegacyAll(scheme)
```
ardaguclu, line 34: You can remove all codes (basically everything under `legacy` folder) in a new commit as they will not be used anymore and finally we get rid of them.

pkg/cli/create/user_identity_mapping.go
```diff
@@ -126,7 +125,7 @@ func (o *CreateUserIdentityMappingOptions) Run() error {
 		User:     corev1.ObjectReference{Name: o.User},
 	}
 
-	if err := util.CreateOrUpdateAnnotation(o.CreateSubcommandOptions.CreateAnnotation, mapping, scheme.DefaultJSONEncoder()); err != nil {
+	if err := util.CreateOrUpdateAnnotation(o.CreateSubcommandOptions.CreateAnnotation, mapping, createCmdJSONEncoder()); err != nil {
```
ardaguclu, line 128: If this is possible, I think it is better to use `scheme.DefaultJSONEncoder` for the ones that don't require custom scheme?

pkg/helpers/scheme/scheme.go
flavianmissi: `legacy.InstallExternalLegacyAll` is still used in https://github.com/openshift/oc/blob/master/pkg/helpers/graph/genericgraph/test/runtimeobject_nodebuilder.go?plain=1#L136.
 I see that's a test helper of some sort. Do you think it's safe to remove the call from there as well?

pkg/cli/create/user_identity_mapping.go
flavianmissi: It should be possible. Do you think it could be a little confusing though? Imagine looking at the directory tree for `pkg/cli/create`, and seeing a `scheme.go` file there. You would instinctively know that this package uses its own scheme instead of relying on the global one. Maybe you need to change something in it, and when you do, it doesn't do what you expected. I don't know if this is a real possibility with the json encoder, but it's worth considering. 
WDYT?

pkg/helpers/scheme/scheme.go
flavianmissi: I gave a stab at removing the call from the graph builder and got some unit test failures. Having a look so that I can estimate whether it's best to do this here or as a follow-up PR.

flavianmissi: this change does indeed introduce a lot more changes... there are several references to the `legacy` package in `pkg/helpers`. I think it's best to open a separate PR. WDYT?

ardaguclu: I'm shockingly surprised to see the usage in https://github.com/openshift/oc/blob/master/pkg/helpers/graph/genericgraph/test/runtimeobject_nodebuilder.go?plain=1#L136. You are right, it is better to spend time on this in a follow up not in here.

pkg/cli/create/user_identity_mapping.go
ardaguclu: yeah, I agree with you. That creates confusion.

pkg/cli/expose/expose.go
```diff
@@ -78,8 +78,10 @@ type ExposeOptions struct {
 }
 
 func NewExposeFlags(streams genericiooptions.IOStreams) *ExposeFlags {
+	flags := expose.NewExposeFlags(streams)
+	flags.PrintFlags = genericclioptions.NewPrintFlags("exposed").WithTypeSetter(exposeCmdScheme)
```
flavianmissi, line 82: Curiously enough, this is the only place that needed to use `exposeCmdScheme` for `oc expose` to work in the simple local tests I ran. However, for consistency sake I kept the use of `exposeCmdScheme` throughout this file.

pkg/cli/expose/scheme.go
```diff
@@ -0,0 +1,38 @@
...
+	schemehelper "github.com/openshift/oc/pkg/helpers/scheme"
+)
+
+var (
+	exposeCmdScheme = runtime.NewScheme()
+	exposeCmdCodecs = serializer.NewCodecFactory(exposeCmdScheme)
+)
+
+func init() {
+	utilruntime.Must(api.InstallKube(exposeCmdScheme))
```
flavianmissi, line 21: @ardaguclu as you suggested, I added this here to make sure all the upstream types work as before, but I didn't note any difference with or without it when exposing a service. I'm fine with keeping this if we think some other uses of this command might break without it though.

pkg/cli/create/scheme.go
```diff
@@ -0,0 +1,38 @@
...
+	schemehelper "github.com/openshift/oc/pkg/helpers/scheme"
+)
+
+var (
+	createCmdScheme = runtime.NewScheme()
+	createCmdCodecs = serializer.NewCodecFactory(createCmdScheme)
+)
+
+func init() {
+	schemehelper.InstallSchemes(createCmdScheme)
```
ardaguclu, line 21: Are we sure that we don't need to also install kube resources just you did in expose?

pkg/cli/expose/expose.go
ardaguclu: That is exactly what I suggested. Thank you.

pkg/cli/expose/scheme.go
ardaguclu: yes, I agree with you

pkg/cli/create/scheme.go
flavianmissi: I will do it just in case but everything seems work just fine.

flavianmissi: done

flavianmissi:
latest force push only rewords a commit to better describe its changes.

ardaguclu:
Thank you
/lgtm

flavianmissi:
thanks for the approval @ardaguclu ! as promised, here's a follow-up PR removing the `legacy` pkg: https://github.com/openshift/oc/pull/1791

PR #1782, opened 2024-05-23
OCPBUGS-33896: `status/inspect-alerts`: handle non-200 by Thanos

petr-muller:
We have seen instances where Thanos responses contained html, apparently by because it is briefly down. We can be slightly more robust to that by detecting non-OK status code right away, instead of passing content body out:

```
Ignoring alerts in 'Update Health'. Error unmarshaling alerts: %w invalid character '<' looking for beginning of value
```

Also improve debugging by allowing to emit http call details and response body on higher verbosity settings, consistent with `client-go` calls to apiserver.

Additionally, improve the error handling by refactoring the `getWithBearer`  method, Previously, it iterated over URIs from a route but instead of searching for success it actually searched until first failure, which is against the point of iterating over possible URIs in the first place. Refactor the method so that it does not immediately return on error, and return on success instead. Only return with an error if all URIs failed to yield a workable result. Slightly optimize the error for the common case where there is only a single URI to try, and shorten the string by using a namespace/name notation.

### DIFF
--- a/pkg/cli/admin/inspectalerts/inspectalerts.go
+++ b/pkg/cli/admin/inspectalerts/inspectalerts.go
@@ -2,7 +2,9 @@
 package inspectalerts
 
 import (
+	"bytes"
 	"context"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"net/http"
@@ -13,6 +15,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+	"k8s.io/klog/v2"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 
 	routev1 "github.com/openshift/api/route/v1"
@@ -100,7 +104,7 @@ func GetAlerts(ctx context.Context, getRoute RouteGetter, bearerToken string) ([
 	return alertBytes, nil
 }
 
-// getWithBearer gets a Route by namespace/name, contructs a URI using
+// getWithBearer gets a Route by namespace/name, constructs a URI using
 // status.ingress[].host and the path argument, and performs GETs on that
 // URI using Bearer authentication with the token argument.
 func getWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name string, baseURI *url.URL, bearerToken string) ([]byte, error) {
@@ -113,7 +117,18 @@ func getWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name st
 		return nil, err
 	}
 
-	client := &http.Client{}
+	withDebugWrappers, err := transport.HTTPWrappersForConfig(
+		&transport.Config{
+			UserAgent:   rest.DefaultKubernetesUserAgent() + "(inspect-alerts)",
+			BearerToken: bearerToken,
+		},
+		http.DefaultTransport,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{Transport: withDebugWrappers}
 	uris := make([]string, 0, len(route.Status.Ingress))
 	for _, ingress := range route.Status.Ingress {
 		uri := *baseURI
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -121,24 +124,24 @@ func getWithBearer(ctx context.Context, getRoute RouteGetter, namespace, name st
...
 		defer resp.Body.Close()
 
 		body, err := io.ReadAll(resp.Body)
 		if err != nil {
-			return nil, err
+			return nil, 0, err
 		}
 
-		return body, err
+		return body, resp.StatusCode, err
```
wking, line 143: Can we put the `!= http.StatusOK` check here, and return an `error` in the case where we failed?  Having some details on why would also be interesting, but dumping HTML to the terminal might be too chatty.  I wonder if we could set `Accept` to say "we want JSON back, but if you can't do that, `text/plain`" to get something more terminal-compatible?  Or maybe Thanos isn't that media-type aware?  Or maybe it's the router that's giving us an error (like "none of the Pods backing this Service are `Ready=True`", and not Thanos?)?

petr-muller: I did it this way because `getWithBearer` sounded like a low-level HTTP method which I would not expect to have opinions on status codes (maybe something would want to call it and handle retries or something).

Router's "service is currently down" page is what I assume was happening.

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -94,7 +98,9 @@ func GetAlerts(ctx context.Context, getRoute RouteGetter, bearerToken string) ([
 	if err != nil {
 		return alertBytes, err
 	}
+	glogBody("Request Body", alertBytes)
 	if statusCode != http.StatusOK {
+
```
wking, line 103: drop this blank line?

wking (approved):
/lgtm

pkg/cli/admin/inspectalerts/inspectalerts.go
wking: I'm not entirely convinced about high-level vs. low-level, but the 8fd03afed69225b logging gives folks calling with `-v8` or higher access to the body, and that's the main think I was hoping for.  And `getWithBearer` is internal, so we can shift things around later if we change our minds.  Feel free to mark this thread resolved :)

jan--f: I would second putting the check here. No reason really to extend the interface, but its really a nit, feel free to ignore.

As for thanos responses, it implements the [Prometheus query api](https://prometheus.io/docs/prometheus/latest/querying/api/).  There it says `The API response format is JSON`

pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -90,10 +94,14 @@ func GetAlerts(ctx context.Context, getRoute RouteGetter, bearerToken string) ([
 	}
 
 	// if we end up going this way, probably port to github.com/prometheus/client_golang/api/prometheus/v1 NewAPI
-	alertBytes, err := getWithBearer(ctx, getRoute, "openshift-monitoring", "thanos-querier", uri, bearerToken)
+	alertBytes, statusCode, err := getWithBearer(ctx, getRoute, "openshift-monitoring", "thanos-querier", uri, bearerToken)
 	if err != nil {
 		return alertBytes, err
 	}
+	glogBody("Response Body", alertBytes)
```
jan--f, line 101: Should this log the response even for a `StatusOK` code?

jan--f:
/lgtm
/hold
just adding a hold in case you want to address any nits. Feel free to unhold any time.

pkg/cli/admin/inspectalerts/inspectalerts.go
petr-muller: I have put the check to `getWithBeared` as you advise, but while I did it, I noticed that the iteration over ingress hosts from a route does not make much sense the way it was done - the iteration aborted on failure and continued on success, so it actually required all options to succeed and then returned the last one. I guess this would not be an issue normally because there would be just a single option, but still changed the method to search for a first success, and only return with an error if all options failed with an error.

jan--f: Yeah good catch :+1:

PR #1787, opened 2024-05-27
OCPBUGS-33896: `upgrade status`: polish alert insights

petr-muller:
- skip alerts without required labels
- add context on why we show the insight (started firing during update or is known to affect updates)
- skip alerts with info level
- explicitly mention the alert does not have a runbook

### DIFF
--- a/pkg/cli/admin/upgrade/status/alerts.go
+++ b/pkg/cli/admin/upgrade/status/alerts.go
@@ -31,6 +31,7 @@ type AlertAnnotations struct {
 	Description string `json:"description,omitempty"`
 	Summary     string `json:"summary,omitempty"`
 	Runbook     string `json:"runbook_url,omitempty"`
+	Message     string `json:"message,omitempty"`
 }
 
 type Alert struct {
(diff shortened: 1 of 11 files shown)


pkg/cli/admin/upgrade/status/alerts.go
```diff
@@ -57,27 +57,62 @@ func parseAlertDataToInsights(alertData AlertData, startedAt time.Time) []update
...
-		if startedAt.After(alert.ActiveAt) && !allowedAlerts.Contains(alert.Labels.AlertName) {
+		var alertName string
+		if alertName = alert.Labels.AlertName; alertName == "" {
 			continue
 		}
+
+		var description string
+		switch {
+		case startedAt.After(alert.ActiveAt) && !allowedAlerts.Contains(alertName):
+			continue
```
wking, line 68: nit: possibly shift this to an `else: continue` at the end, so we don't have to wonder if we've missed a case and left an empty `description` without a `continue`?

petr-muller: I refactored this a bit, introduced named conditions which hopefully help with readability

pkg/cli/admin/upgrade/status/alerts.go
```diff
@@ -57,27 +57,62 @@ func parseAlertDataToInsights(alertData AlertData, startedAt time.Time) []update
...
+
+		var runbook string
+		if runbook = alert.Annotations.Runbook; runbook == "" {
+			runbook = "<alert does not have a runbook_url annotation>"
+		}
+
+		if alert.Annotations.Description == "" {
+			description += " The alert has no description."
+		} else {
+			description += " The alert description is: " + alert.Annotations.Description
```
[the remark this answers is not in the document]
petr-muller: Good catch! 4.16.0-rc.2 payload shows we still have a few manifests that use `message` so I added handling this in the code. Some of our alerts use all fields (message, description, summary) so I try to include them all, when present. Also added fixtures to exercise this case.

```consome
$ rg 'message: ' *prom*yaml
0000_90_cluster-authentication-operator_03_prometheusrule.yaml
17:            message: >-

0000_90_ingress-operator_03_prometheusrules.yaml
25:          message: "HAProxy reloads are failing on {{ $labels.pod }}. Router is not respecting recently created or modified routes"
34:          message: "HAProxy metrics are reporting that HAProxy is down on pod {{ $labels.namespace }} / {{ $labels.pod }}"
43:          message: |
54:          message: |
78:            message: "Ingress {{ $labels.namespace }}/{{ $labels.name }} is missing the IngressClassName for 1 day."
87:            message: "Route {{ $labels.namespace }}/{{ $labels.name }} is owned by an unmanaged Ingress."

0000_50_cluster-storage-operator_12_prometheusrules.yaml
28:          message: "StorageClass count check is failing (there should not be more than one default StorageClass)"
55:            Events of the Pods should contain exact error message: "oc describe pod -n <pod namespace> <pod name>".
```

wking (approved):
/lgtm

PR #1791, opened 2024-05-31
NO-JIRA: pkg/helpers: remove legacy pkg

flavianmissi:
we haven't supported "ungroupified" APIs for a couple of years. this cleans up the legacy package, effectively completely dropping support for them.

--
this is a follow up of https://github.com/openshift/oc/pull/1775.


ardaguclu:
This looks good to me. But before moving forward, it is better to wait the new nightly including the previous PR and check that everything is ok. After that, we can merge this.

ardaguclu:
Thank you
/lgtm
/retest

PR #1793, opened 2024-05-31
NO-JIRA: Update comment about using OpenFileSequential

rhmdnd:
A previous commit refactored OpenFileSequential since it was deprecated, and replaced it with OpenFile.

  https://github.com/openshift/oc/commit/7a8c458fbbf06a1c8411e4199d61e6d43bf67334

This commit just updates the comment to be accurate, since I stumbled on it walking through the code.


ardaguclu:
/retitle NO-JIRA: Update comment about using OpenFileSequential
/lgtm

PR #1795, opened 2024-06-03
OCPBUGS-16418: Show more clear info about the DeploymentConfig is deprecated

sanchezl:
Suppress the `DeploymentConfig` deprecation warning from `oc status` and `oc get all`.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -3,13 +3,15 @@ package cli
 import (
 	"flag"
 	"fmt"
+	"log/slog"
 	"os"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/MakeNowJust/heredoc"
+	"github.com/openshift/oc/pkg/helpers/apps"
 	"github.com/spf13/cobra"
-
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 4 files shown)


pkg/cli/cli.go
```diff
@@ -176,8 +178,19 @@ func NewOcCommand(o kubecmd.KubectlOptions) *cobra.Command {
...
 		Run:   runHelp,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			rest.SetDefaultWarningHandler(warningHandler)
-
+			switch cmd.Name() {
+			case "status":
+				rest.SetDefaultWarningHandler(apps.NewIgnoreDeploymentConfigWarningHandler(warningHandler))
+			case "get":
+				if slices.ContainsFunc(args, func(s string) bool { return slices.Contains(strings.Split(s, ","), "all") }) {
+					rest.SetDefaultWarningHandler(apps.NewIgnoreDeploymentConfigWarningHandler(warningHandler))
```
ardaguclu, line 186: I'm not sure this will work because as we concluded that builder does not take custom warning handler into account https://github.com/kubernetes/kubectl/issues/1586#issuecomment-2102450494 and `get` uses builder or our conclusion is wrong?

pkg/cli/cli.go
```diff
@@ -176,8 +178,19 @@ func NewOcCommand(o kubecmd.KubectlOptions) *cobra.Command {
 		Long:  cliLong,
 		Run:   runHelp,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			rest.SetDefaultWarningHandler(warningHandler)
-
+			switch cmd.Name() {
```
ardaguclu, line 181: I'd not prefer modifying this core cli entrance point just to update the warning text.

PR #1797, opened 2024-06-05
OCPBUGS-34877: Ignore pruning when deployment points to replicationcontroller

ardaguclu:
Due to the ongoing migration processes from deploymentconfig to deployment, people may fall into the situation where deployment object in reality points to replicationcontroller instead of replicaset. `oc adm prune` command panicks in that cases during the object casting. This PR handles this gracefully by ignoring these resources for prune.


ardaguclu:
/cc @ingvagabund 
PTAL, when you have a chance?. Thank you.

ardaguclu:
Although this is not critical fix, it is not possible that this patch can cause any problems and no point of wasting our CI resources just for this patch for retests (in case of commit sha changes when the ack label requirement is lifted);

PR #1764, opened 2024-05-14
OCPBUGS-33905: pkg/cli/admin/upgrade/rollback: Gate on OC_ENABLE_CMD_UPGRADE_ROLLBACK

wking:
In case the hidden-ness from 73074c32ba (#1642) is not sufficient to scare users away, add a gating environment variable.  We also use a gating env. var. for `status` (701c2eee92, #1554) and `inspect-alerts` (229c2a8460, #1674).

Also reword the LongDesc to avoid saying "supported" (which readers might misconstrue as "what Red Hat support will do if we open a case").  I'm just trying to talk about "what the cluster-version operator will do if you request a rollback", and hopefully the new wording makes this more clear.

PR #1806, opened 2024-06-24
OCPBUGS-35994: pkg/cli/admin/upgrade/rollback: Drop this command

wking:
Current CI/QE capacity doesn't have space to support rollbacks, which require basically the same level of effort as supporting roll-forward updates, but which would never be used by as many clusters as the number of clusters that use roll-forward updates.  In addition, paused MachineConfigPools and HyperShift's NodePools and similar allow for decoupling compute updates from control-plane updates.  And control-plane updates have PodDisruptionBudgets and such to wedge gracefully if an issue comes up during updates that destabilizes the control plane.  And if a control-plane update goes smoothly, but its behavior changes destabilize cluster workloads, ClusterVersion overrides and similar give an emergency safety valve to patch things up while admins wait for new releases that address the newly-discovered issues, while [conditional updates][1] allow the risk to be declared so additional clusters are warned before attempting an exposed update.

With the moderate operational benefit weighted against the substantial cost of CI/QE test coverage, this commit reverts the subcommand which I'd initially added in 73074c32ba (#1642) and gated in 9b2842a0c9 (#1764).

[1]: https://github.com/openshift/enhancements/blob/0977611b030fcb667b22cdf36a86e3852ee54e84/enhancements/update/targeted-update-edge-blocking.md


petr-muller (approved):
Code change LGTM, holding so that you can finish [talking to SD](https://redhat-internal.slack.com/archives/CJ1J9C3V4/p1719252728248179?thread_ts=1719251074.026659&cid=CJ1J9C3V4)

/hold

wking:
Service Delivery isn't very chatty about this, and re-reverts aren't hard if it comes to that.

/hold cancel

evakhoni:
pre-merge verified
/label qe-approved

PR #1808, opened 2024-06-27
WRKLDS-1368: define node-image commands structure

andfasano:
This initial PR defines just the initial high level structure for the new add nodes commands (see epic https://issues.redhat.com/browse/WRKLDS-937 for more details), with no flags and no implementation.
The main goal of this patch is to converge on the commands code organizations and naming (flags and implementation will followup in subsequent PRs).

Layout will look like:

```
$ oc adm node-image
The subcommands allow you to create an ISO image to be used for adding the desired nodes to an OpenShift cluster, and
also to monitor the process.

Available Commands:
  create        Create an ISO image for booting the nodes to be added to the target cluster
  monitor       Monitor the process of adding new nodes to an OpenShift cluster

Use "oc adm node-image <command> --help" for more information about a given command.
```

### DIFF
--- a/pkg/cli/admin/addnodes/addnodes.go
+++ b/pkg/cli/admin/addnodes/addnodes.go
@@ -1,0 +1,23 @@
+package addnodes
+
+import (
+	"github.com/spf13/cobra"
+
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+func NewCmdAddNodes(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add-nodes",
+		Short: "Commands for adding new nodes to an OpenShift cluster",
+		Long: templates.LongDesc(`
+			The subcommands allow you to create an ISO image to be used for adding the desired
+			nodes to an OpenShift cluster, and also to monitor the process.
+			`),
+	}
+	cmd.AddCommand(NewCreate(f, streams))
+	cmd.AddCommand(NewMonitor(f, streams))
+	return cmd
+}
--- a/pkg/cli/admin/addnodes/create.go
+++ b/pkg/cli/admin/addnodes/create.go
@@ -1,0 +1,54 @@
+package addnodes
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+func NewCreate(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	o := NewCreateOptions(streams)
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create an ISO image for booting the nodes to be added to the target cluster",
+		Long: templates.LongDesc(`
+			<TODO>
+		`),
+		Example: templates.Examples(`
+			<TODO>
+		`),
+		Run: func(cmd *cobra.Command, args []string) {
+			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			kcmdutil.CheckErr(o.Validate())
+			kcmdutil.CheckErr(o.Run(cmd.Context()))
+		},
+	}
+	_ = cmd.Flags()
+	return cmd
+
+}
+
+func NewCreateOptions(streams genericiooptions.IOStreams) *CreateOptions {
+	return &CreateOptions{
+		IOStreams: streams,
+	}
+}
+
+type CreateOptions struct {
+	genericiooptions.IOStreams
+}
+
+func (o *CreateOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
+	return nil
+}
+
+func (o *CreateOptions) Validate() error {
+	return nil
+}
+
+func (o *CreateOptions) Run(ctx context.Context) error {
+	return nil
+}
--- a/pkg/cli/admin/admin.go
+++ b/pkg/cli/admin/admin.go
@@ -12,7 +12,6 @@ import (
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 	ktemplates "k8s.io/kubectl/pkg/util/templates"
 
-	"github.com/openshift/oc/pkg/cli/admin/addnodes"
 	"github.com/openshift/oc/pkg/cli/admin/buildchain"
 	"github.com/openshift/oc/pkg/cli/admin/catalog"
 	"github.com/openshift/oc/pkg/cli/admin/copytonode"
@@ -91,7 +90,8 @@ func NewCommandAdmin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *co
 				copytonode.NewCmdCopyToNode(f, streams),
 				rebootmachineconfigpool.NewCmdRebootMachineConfigPool(f, streams),
 				waitfornodereboot.NewCmdWaitForNodeReboot(f, streams),
-				addnodes.NewCmdAddNodes(f, streams),
+				// Will uncomment when the new command implementation will be ready
+				// addnodes.NewCmdAddNodes(f, streams),
 			},
 		},
 		{
--- a/pkg/cli/admin/admin.go
+++ b/pkg/cli/admin/admin.go
@@ -12,6 +12,7 @@ import (
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 	ktemplates "k8s.io/kubectl/pkg/util/templates"
 
+	"github.com/openshift/oc/pkg/cli/admin/addnodes"
 	"github.com/openshift/oc/pkg/cli/admin/buildchain"
 	"github.com/openshift/oc/pkg/cli/admin/catalog"
 	"github.com/openshift/oc/pkg/cli/admin/copytonode"
@@ -90,6 +91,7 @@ func NewCommandAdmin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *co
 				copytonode.NewCmdCopyToNode(f, streams),
 				rebootmachineconfigpool.NewCmdRebootMachineConfigPool(f, streams),
 				waitfornodereboot.NewCmdWaitForNodeReboot(f, streams),
+				addnodes.NewCmdAddNodes(f, streams),
 			},
 		},
 		{
--- a/pkg/cli/admin/admin.go
+++ b/pkg/cli/admin/admin.go
@@ -90,8 +90,6 @@ func NewCommandAdmin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *co
 				copytonode.NewCmdCopyToNode(f, streams),
 				rebootmachineconfigpool.NewCmdRebootMachineConfigPool(f, streams),
 				waitfornodereboot.NewCmdWaitForNodeReboot(f, streams),
-				// Will uncomment when the new command implementation will be ready
-				// addnodes.NewCmdAddNodes(f, streams),
 			},
 		},
 		{
--- a/pkg/cli/admin/nodeimage/OWNERS
+++ b/pkg/cli/admin/nodeimage/OWNERS
@@ -12,3 +12,4 @@ reviewers:
   - pawanpinjarkar
   - rwsu
   - zaneb
+  
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -1,4 +1,4 @@
-package addnodes
+package nodeimage
 
 import (
 	"context"
@@ -9,11 +9,11 @@ import (
 	"k8s.io/kubectl/pkg/util/templates"
 )
 
-func NewMonitor(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
-	o := NewMonitorOptions(streams)
+func NewCreate(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	o := NewCreateOptions(streams)
 	cmd := &cobra.Command{
-		Use:   "monitor",
-		Short: "Monitor the process of adding new nodes to an OpenShift cluster",
+		Use:   "create",
+		Short: "Create an ISO image for booting the nodes to be added to the target cluster",
 		Long: templates.LongDesc(`
 			<TODO>
 		`),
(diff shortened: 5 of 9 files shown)


pkg/cli/admin/addnodes/addnodes.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+func NewCmdAddNodes(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add-nodes",
+		Short: "Commands for adding new nodes to an OpenShift cluster",
```
pawanpinjarkar, line 14: Nit: `Command`

pkg/cli/admin/addnodes/addnodes.go
```diff
@@ -0,0 +1,23 @@
...
+		Long: templates.LongDesc(`
+			The subcommands allow you to create an ISO image to be used for adding the desired
```
pawanpinjarkar, line 16: Nit: `The subcommand`

pkg/cli/admin/addnodes/create.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+func NewCreate(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	o := NewCreateOptions(streams)
+	cmd := &cobra.Command{
+		Use:   "create",
```
rwsu, line 15:
```suggestion
		Use:   "create-image",
```
Later on there would be a create-pxe-files command. This would also line up better with existing openshift-install "agent create image" and "agent create pxe-files" commands.

pkg/cli/admin/addnodes/addnodes.go
andfasano: This is the parent that groups a set of different (sub)commands - maybe we could use a description more similar to the `oc adm release` one (`This tool is used to...`)

pkg/cli/admin/addnodes/addnodes.go
andfasano: There are more than one sub-command

pkg/cli/admin/addnodes/create.go
andfasano: That seems a fair point, I'd like to hear also other comments/suggestion from the others cc @zaneb @pawanpinjarkar @bfournie

zaneb: I was picturing a `--pxe` flag rather than a separate command.
In the installer, `agent create image` and `agent create pxe-files` are separate due to the architecture of the installer preventing us from passing arguments; we had to have separate commands with separate targets. We don't have that restriction here and I think I prefer the flag - it will be easier to document.

zaneb: That said, in the context of the full command `add-nodes create` is a little weird (we're creating... an add-nodes?).
Something like `oc adm new-nodes image` and `oc adm new-nodes add` would make sense to me.

andfasano: > I was picturing a --pxe flag rather than a separate command.

I will also prefer a separate flag

> Something like oc adm new-nodes image and oc adm new-nodes add would make sense to me.

I think it's important to keep the keyword `add` in the command name, so possible options could be:
* `oc adm add-nodes image`
* `oc adm new-nodes add`
(with a slightly stronger preference on the first one, as it seems to describe better what the command does)

bfournie: Looking at the existing` oc adm` commands, I think it does make sense to make the new commands a verb for consistency, i.e. something like `oc adm add-nodes `or `oc adm create-nodes`.  Most (but not all) of the existing commands fit this pattern, e.g
`copy-to-node`
`cordon`
`create-bootstrap-project-template`
`drain`
etc.

Perhaps for the monitor could use `oc adm wait-for-add-nodes` to be consistent with
`wait-for-node-reboot`
`wait-for-stable-cluster`

pawanpinjarkar: - `oc adm add-nodes-create-image`
- `oc adm add-nodes-create-pxe-files` ( consistent with agent installer)
- `oc adm wait-for-add-nodes` ( consistent with what @bfournie said above and agent installer, day 1 wait-for) , this will monitor the day2 progress

If we plan with `oc adm wait-for-add-nodes` which is actually monitor-add nodes, I wonder if we can rename `monitor-add-nodes` to `wait-for-add-nodes`. Day2 functionality is a DP so maybe we could still make improvements? 

Overall, I would like to have similar names and invocations for the things we are already doing for agent installer , i.e. using a seperate command for pxe like @rwsu said instead of flags, using the same name for monitoring day 2 progress, adding nodes etc.

andfasano: @pawanpinjarkar @bfournie not sure if you missed it but it was intentional to have a common command name (currently `add-nodes`) that will act as a parent for a set of sub-commands (currently `create` and `monitor`). As suggested by the workload team duringt the first meeting, it would be also easier to maintain the code and define the commands ownership (especially for troubleshooting). So the suggested layout should match the form `oc adm <command> <sub-commands>`. For example, based on the previous discussion, a possible layout for the two commands could be:

* `oc adm add-nodes image` - The command that generates the ISO (or pxe) artifacts
* `oc adm add-nodes monitor` - The command that monitors the adding activity

> I wonder if we can rename monitor-add-nodes to wait-for-add-nodes
Sincerely I'm not worried too much about the `node-joiner` tool interface look'n'feel, given that it's an internal tool only

rwsu: > - oc adm add-nodes image - The command that generates the ISO (or pxe) artifacts
> - oc adm add-nodes monitor - The command that monitors the adding activity

This sounds fine to me.

pawanpinjarkar: Thank you @andfasano. Do we need to stick to `add-nodes`? `oc adm add-nodes image` isn't very readable. We are adding nodes ( and not adding an image)  to expand the cluster so how about 

- `oc adm expand-cluster create-image` - The command that generates the ISO artifacts
- `oc adm expand-cluster create-pxe-files` - The command that generates the PXE artifacts
-  `oc adm expand-cluster monitor-nodes`  or `oc adm expand-cluster wait-for-nodes` - The command that monitors adding nodes to an existing cluster

I still  think `add-nodes` could give users a impression that the command is adding a node to the cluster where as in reality it would be only creating an ISO or PXE files and users need to boot up the nodes by themselves. Happy to talk on 1:1 or team call to discuss further.

andfasano: As a further discussion, I think it would be better to keep the commands in a form that is closer to what they do (they do not directly expand the cluster, rather creating a image that will be used for enlarging it). Considering that `oc` already has a `node-logs` command, considering also the previous bits mentioned by @bfournie, and another [comment](https://issues.redhat.com/browse/WRKLDS-1314) from @zaneb, a reasonable form could be:

* `oc adm node-image create` (with an optional flag for generating the pxe artifacts) for creating the image
* `oc adm node-image monitor` for the monitoring activity

andfasano: (meanwhile, renamed to `node-image` to see how it looks like)

andfasano: Looks like there is a consensus for the latest proposal, resolving the discussion

rwsu (commented):
/lgtm

pkg/cli/admin/nodeimage/create.go
```diff
@@ -0,0 +1,54 @@
...
+		Long: templates.LongDesc(`
+			<TODO>
+		`),
+		Example: templates.Examples(`
+			<TODO>
+		`),
+		Run: func(cmd *cobra.Command, args []string) {
+			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			kcmdutil.CheckErr(o.Validate())
+			kcmdutil.CheckErr(o.Run(cmd.Context()))
```
ardaguclu, line 26: Most of the time `cmd.Context()` is nil and unfortunately people tend to use it, if it passes (assuming that it is always initialized). I'd recommend removing and once there is a need of a context, it is better to initialize it in `Run` function.

pkg/cli/admin/nodeimage/create.go
```diff
@@ -0,0 +1,54 @@
...
+		},
+	}
+	_ = cmd.Flags()
```
ardaguclu, line 29: Better to be in a separate function `addFlags`

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -0,0 +1,54 @@
...
+		Long: templates.LongDesc(`
+			<TODO>
+		`),
+		Example: templates.Examples(`
+			<TODO>
+		`),
+		Run: func(cmd *cobra.Command, args []string) {
+			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			kcmdutil.CheckErr(o.Validate())
+			kcmdutil.CheckErr(o.Run(cmd.Context()))
```
ardaguclu, line 26: Same

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -0,0 +1,54 @@
...
+		},
+	}
+	_ = cmd.Flags()
```
ardaguclu, line 29: `addFlags`

pkg/cli/admin/nodeimage/nodeimage.go
```diff
@@ -0,0 +1,23 @@
...
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+func NewCmdNodeImage(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "node-image",
+		Short: "Add nodes to an existing cluster.",
+		Long: templates.LongDesc(`
```
ardaguclu, line 15: I'd recommend to define these in consts like https://github.com/openshift/oc/blob/a2e5b09d7d14b77c14f1b32550f8c5a0328e73b3/pkg/cli/admin/mustgather/mustgather.go#L56-L82 otherwise, it is difficult to manage raw strings in here.

pkg/cli/admin/nodeimage/OWNERS
```diff
@@ -0,0 +1,14 @@
...
+  - bfournie
+  - pawanpinjarkar
+  - rwsu
+  - zaneb
+reviewers:
+  - andfasano
+  - bfournie
+  - pawanpinjarkar
+  - rwsu
+  - zaneb
```
ardaguclu, line 14: nit: new line

pkg/cli/admin/admin.go
```diff
@@ -90,6 +90,8 @@ func NewCommandAdmin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *co
 				copytonode.NewCmdCopyToNode(f, streams),
 				rebootmachineconfigpool.NewCmdRebootMachineConfigPool(f, streams),
 				waitfornodereboot.NewCmdWaitForNodeReboot(f, streams),
+				// Will uncomment when the new command implementation will be ready
+				// nodeimage.NewCmdNodeImage(f, streams),
```
ardaguclu, line 94: No need to modify this file for this. We can add `nodeimage.NewCmdNodeImage(f, streams),` in next PR?

ardaguclu:
I've dropped some comments but overall looks good to me. Thank you

pkg/cli/admin/nodeimage/nodeimage.go
ardaguclu: Sounds good. Thank you.

pkg/cli/admin/nodeimage/create.go
```diff
@@ -0,0 +1,54 @@
...
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+func NewCreate(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	o := NewCreateOptions(streams)
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create an ISO image for booting the nodes to be added to the target cluster",
```
ardaguclu, line 16:
```suggestion
		Short: "Experimental: Create an ISO image for booting the nodes to be added to the target cluster",
```

Convention is firstly releasing commands as `Experimental` and subsequent releases marking them as GA'd by removing the `Experimental` prefixes.

andfasano:
> /hold
> 
> until at least one (preferably both) of the subcommands is/are introduced. We don't wanna merge empty skeletons.

The work for the create command is currently in progress here https://github.com/openshift/oc/pull/1819. The current PR does not link the skeleton commands to `oc adm`, so the current code will not be visible

pkg/cli/admin/nodeimage/create.go
andfasano: Not sure if it applies to this specific case - where the DevPreview feature has been released in a different form - but https://issues.redhat.com/browse/OCPSTRAT-1316 it's directly a GA feature. cc @racedo

ardaguclu: This may tie to GA feature but these commands are released now and command itself is marked as experimental. Isn't this too strong justification for this newly implemented command is ready for GA?

zaneb: Perhaps we could mark as Experimental for now, but aim to remove that before 4.17 branches?

ardaguclu: This is not a blocker for me and I'm fine with your any decision. But conventionally I'd recommend to have it experimental in 4.17 and promote it to GA in 4.18.

andfasano: @ardaguclu your point looks reasonable to me since from the point of view of `oc` it's a freshly new command. Could be acceptable to mark the command as `Experimental` in 4.17 (where the feature is marked as GA) and remove it in 4.18?

racedo:
Re experimental vs GA, we are now in a developer preview in OpenShift 4.16, which can be considered experimental. The feature is expected as a GA after this dev preview, and we aim to support it as such if all goes to plan during this dev cycle. I'd suggest to please remove the "experimental" note in the description to avoid confusion. Many thanks!

pkg/cli/admin/nodeimage/create.go
andfasano: See comment https://github.com/openshift/oc/pull/1808#issuecomment-2222418180

andfasano:
> > /hold
> > until at least one (preferably both) of the subcommands is/are introduced. We don't wanna merge empty skeletons.
> 
> The work for the create command is currently in progress here #1819. The current PR does not link the skeleton commands to `oc adm`, so the current code will not be visible

Hi @ingvagabund , the first subcommand PR is now introduced https://github.com/openshift/oc/pull/1819, can we proceed with this PR (since it's the base for 1819)? Thanks

rwsu (commented):
/lgtm

andfasano:
Included in https://github.com/openshift/oc/pull/1819

PR #1817, opened 2024-07-05
Potential fix for openshift/oc/issues/1814.

germanparente:
oc adm groups sync is comparing dn's / basedn's case sensitive when it should not be in ldap.

### DIFF
--- a/vendor/github.com/openshift/library-go/pkg/security/ldapquery/query.go
+++ b/vendor/github.com/openshift/library-go/pkg/security/ldapquery/query.go
@@ -112,9 +112,9 @@ func (o *LDAPQueryOnAttribute) NewSearchRequest(attributeValue string, attribute
 		if err != nil {
 			return nil, fmt.Errorf("could not search by dn, invalid dn value: %v", err)
 		}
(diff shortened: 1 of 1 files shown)


vendor/github.com/openshift/library-go/pkg/security/ldapquery/query.go
```diff
@@ -112,9 +112,9 @@ func (o *LDAPQueryOnAttribute) NewSearchRequest(attributeValue string, attribute
 		if err != nil {
 			return nil, fmt.Errorf("could not search by dn, invalid dn value: %v", err)
 		}
-		if !baseDN.AncestorOf(dn) && !baseDN.Equal(dn) {
-			return nil, NewQueryOutOfBoundsError(attributeValue, o.BaseDN)
-		}
+                if !baseDN.AncestorOfFold(dn) && !baseDN.EqualFold(dn) {
```
ardaguclu, line 115: We don't directly modify the files under the `vendor` directory.

/hold

germanparente:
Apologies for the noise. I will close this and do my PR in https://github.com/openshift/library-go

PR #1823, opened 2024-07-12
 WRKLDS-1316: oc adm node-image monitor command

rwsu:
Adds the implementation and unit test for the new `oc adm node-image monitor` command.


pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -1,52 +1,368 @@
...
+const (
+	nodeJoinerMonitorContainer = "node-joiner-monitor"
 )
 
+var (
+	monitorLong = templates.LongDesc(`
+		Monitor nodes being added to a cluster using an image generated from
+		the "oc adm node-image create" command.
+
+		Each node being added to the cluster has two certificate signing requests
```
andfasano, line 41: Before citing this paragraph, I think that it could be useful to describe the fact that the monitor command will allow the user to observe the whole workflow. In particular, during the initial phase, when the pre-flight validations are being applied (and in case, if something went wrong, reported to the user)

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -1,52 +1,368 @@
...
-	genericiooptions.IOStreams
+	CommonOptions
+
+	IPAddressesToMonitor string
 }
 
-func (o *MonitorOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string) error {
+// Complete completes the required options for the monitor command.
+func (o *MonitorOptions) Complete(f genericclioptions.RESTClientGetter, cmd *cobra.Command, args []string) error {
+	o.RESTClientGetter = f
```
andfasano, line 107: Mentioned also in another comment, this portion seems that could be refactored out in common

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -1,52 +1,368 @@
...
 }
 
-func (o *MonitorOptions) Run() error {
+// Run creates a temporary namespace to kick-off a pod for running the node-joiner
+// monitor cli tool. Logs from node-joiner monitor are streamed from the pod
+// to stdout.
+func (o *MonitorOptions) Run(ctx context.Context) error {
+	defer o.cleanup(ctx)
+
+	err := o.createNamespace(ctx)
```
andfasano, line 147: couldn't this be moved into the `runNodeJoinerPod`?

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -1,52 +1,368 @@
...
+		}
+		return true, &utilsexec.CodeExitError{
+			Err:  fmt.Errorf("%s/%s unexpectedly terminated: exit code: %v, reason: %s, message: %s", o.nodeJoinerNamespace.Name, o.nodeJoinerPod.Name, state.Terminated.ExitCode, state.Terminated.Reason, state.Terminated.Message),
+			Code: int(state.Terminated.ExitCode),
+		}
+	}
+	return false, nil
+}
+
+func (o *MonitorOptions) runNodeJoinerPod(ctx context.Context) error {
```
andfasano, line 240: This method (except the `tasks` list`) could be probably be defined in the base class

pkg/cli/admin/nodeimage/monitor_test.go
```diff
@@ -0,0 +1,58 @@
+package nodeimage
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestValidateMonitor(t *testing.T) {
```
andfasano, line 8: Do you maybe plan to add a more complete test on the monitor command later?

pkg/cli/admin/nodeimage/common.go
```diff
@@ -0,0 +1,198 @@
...
+	nodeJoinerImage          string
+	nodeJoinerNamespace      *corev1.Namespace
+	nodeJoinerServiceAccount *corev1.ServiceAccount
+	nodeJoinerRole           *rbacv1.ClusterRole
+	RESTClientGetter         genericclioptions.RESTClientGetter
+	nodeJoinerPod            *corev1.Pod
+	command                  string
+}
+
+func (o *CommonOptions) getNodeJoinerPullSpec(ctx context.Context) error {
```
ardaguclu, line 42: Please remove this common package entirely. This is not the preferred way and also I checked `getNodeJoinerPullSpec` (also `fetchClusterReleaseImage`) and monitor.go does not use this function. If the intend is `CommonOptions` will be a base for both commands, monitor command can use anything in create command as they are under the same `nodeimage` package. If there is strong necessity to have a base struct for both commands, as @andfasano suggested `BaseNodeImageCommand` can be used and it can be stored in nodeimage/create command rather than a new `common.go`

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -1,52 +1,368 @@
...
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Complete(f, cmd, args))
 			kcmdutil.CheckErr(o.Validate())
-			kcmdutil.CheckErr(o.Run())
+			kcmdutil.CheckErr(o.Run(cmd.Context()))
 		},
 	}
-	_ = cmd.Flags()
-	return cmd
+	flags := cmd.Flags()
```
ardaguclu, line 82: Would be better if flags are added in `o.AddFlags` function

pkg/cli/admin/nodeimage/common.go
andfasano: I think that the main goal is to remove as much as possible the code duplication - which is a good thing - given the fact that the two command are almost very similar (as they use almost the same behavioral pattern). AFAICS though the package is still the `nodeimage` one, I'm more relaxed on having the common code in a separate file (or one of the command), as long as the naming is consistent (and the package remains the same ofc).

ardaguclu: I'm fine centralizing some of the components to prevent code duplication. I think, this can be achieved within `create` or `monitor` command instead of `common.go`.

pkg/cli/admin/nodeimage/monitor_test.go
rwsu: Yes, a test for Run will be added shortly.

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -1,52 +1,299 @@
...
 
+var (
+	monitorLong = templates.LongDesc(`
+		Monitor nodes being added to a cluster using an image generated from
+		the "oc adm node-image create" command.
+
+		After the node image ISO has been booted on the host, the monitor command
+		reports any pre-flight validations that may have failed impeding the
+		host from being added to the cluster. If validations are successful, the
+		node installation starts. The node's boot disk is reimaged and the node
```
andfasano, line 39:
```suggestion
		node installation starts. 
```

pkg/cli/admin/nodeimage/monitor.go
```diff
@@ -1,52 +1,299 @@
...
+		reboots.
```
andfasano, line 40:
```suggestion
		.
```

pkg/cli/admin/nodeimage/create.go
```diff
@@ -500,6 +623,19 @@ func (o *CreateOptions) createRolesAndBindings(ctx context.Context) error {
...
+				},
+				Resources: []string{
+					"certificatesigningrequests",
+					"clusterversions",
+				},
+				Verbs: []string{
+					"get",
+					"list",
+				},
+			},
```
andfasano, line 638: @rwsu can you please also add another rule specifically for the secrets, so that they could also be created? ie

```
...
{
			APIGroups: []string{
				"",
			},
			Resources: []string{
				"secrets",
			},
			Verbs: []string{
				"get",
				"list",
				"create",
			},
		},
...
```

Otherwise [this incoming change](https://github.com/openshift/installer/pull/8717/files#diff-174c7f0a2bed95144baa2b38b752cb3a48aab7bc4380c16028faed7dd97924f9R269)  from @pawanpinjarkar  PR will break the commands

rwsu: Thanks for the heads up. We will also need to [include update](https://github.com/openshift/installer/pull/8717/files#diff-174c7f0a2bed95144baa2b38b752cb3a48aab7bc4380c16028faed7dd97924f9R284).

PR #1833, opened 2024-07-30
[WIP] WRKLDS-1318: flags for single node interface

andfasano:
This patch introduces a new set of flags to simplify simpler scenarios based on adding a single node. Instead of defining a complete yaml file (which could be more suitable and handy when handling multiples nodes), the user now can define a number of command line flags to finely configure the required parts of the single node to be added.
When single node flags will be used:
* The only mandatory flag is `mac-address`, to identify the host to be added. All the others will be optional
* The configuration file, if present in the working dir, will be ignored

### DIFF
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -44,6 +45,15 @@ const (
 	nodeJoinerContainer         = "node-joiner"
 )
 
+const (
+	snFlagMacAddress     = "mac-address"
+	snFlagCpuArch        = "cpu-architecture"
+	snFlagSshKey         = "ssh-key"
+	snFlagHostname       = "hostname"
+	snFlagRootDeviceHint = "root-device-hint"
+	snFlagNetworkConfig  = "network-config-file"
+)
+
 var (
 	createLong = templates.LongDesc(`
 		Create an ISO image from an initial configuration for a given set of nodes,
@@ -54,12 +64,17 @@ var (
 		The downloaded ISO image could then be used to boot a previously selected
 		set of nodes, and add them to the target cluster in a fully automated way.
 
-		A nodes-config.yaml config file must be created to provide the required
-		initial configuration for the selected nodes.
-
 		The command also requires a connection to the target cluster, and a valid
 		registry credentials to retrieve the required information from the target
 		cluster release.
+
+		A nodes-config.yaml config file must be created to provide the required
+		initial configuration for the selected nodes. 
+		Alternatively, to support simpler configurations for adding just a single
+		node, it's also possible to use a set of flags to configure the host. In
+		such case the '--mac-address' is the only mandatory flag - while all the
+		others will be optional (note: any eventual configuration file present 
+		will be ignored).
 	`)
 
 	createExample = templates.Examples(`
@@ -71,6 +86,13 @@ var (
 
 		# Specify a custom image name
 		  oc adm node-image create --o=my-node.iso
+
+		# Create an ISO to add a single node without using the configuration file
+		  oc adm node-image create --mac-address=00:d8:e7:c7:4b:bb
+
+		# Create an ISO to add a single node with a root device hint and without
+		# using the configuration file
+		  oc adm node-image create --mac-address=00:d8:e7:c7:4b:bb --root-device-hint=deviceName:/dev/sda
 	`)
 
 	createCommand = "oc adm node-image create"
@@ -116,9 +138,20 @@ type CreateOptions struct {
 	AssetsDir string
 	// OutputName allows the user to specify the name of the generated image.
 	OutputName string
+	// Simpler interface for creating a single node
+	SingleNodeOpts *singleNodeCreateOptions
 
 	nodeJoinerExitCode int
 	rsyncRshCmd        string
+}
+
+type singleNodeCreateOptions struct {
+	MacAddress      string
+	CPUArchitecture string
+	SSHKey          string
+	Hostname        string
+	RootDeviceHints string
+	NetworkConfig   string
 }
 
 // AddFlags defined the required command flags.
@@ -127,6 +160,14 @@ func (o *CreateOptions) AddFlags(cmd *cobra.Command) {
 
 	flags.StringVar(&o.AssetsDir, "dir", o.AssetsDir, "The path containing the configuration file, used also to store the generated artifacts.")
 	flags.StringVarP(&o.OutputName, "output-name", "o", "node.iso", "The name of the output image.")
+
+	flags.StringP(snFlagMacAddress, "m", "", "Single node flag. MAC address used to identify the host to apply the configuration. If specified, the nodes-config.yaml config file will not be used.")
+	usageFmt := "Single node flag. %s. Valid only when `mac-address` is defined."
+	flags.StringP(snFlagCpuArch, "c", "", fmt.Sprintf(usageFmt, "The CPU architecture to be used to install the node"))
+	flags.StringP(snFlagSshKey, "k", "", fmt.Sprintf(usageFmt, "The SSH key used to access the node"))
+	flags.String(snFlagHostname, "", fmt.Sprintf(usageFmt, "The hostname to be set for the node"))
+	flags.String(snFlagRootDeviceHint, "", fmt.Sprintf(usageFmt, "Hint for specifying the storage location for the image root filesystem. Format accepted is <hint name>:<value>."))
+	flags.String(snFlagNetworkConfig, "", fmt.Sprintf(usageFmt, "YAML file containing the NMState configuration to be applied for the node"))
 }
 
 // Complete completes the required options for the create command.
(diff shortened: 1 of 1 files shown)


andfasano:
cc @zaneb @rwsu

pkg/cli/admin/nodeimage/create.go
```diff
@@ -54,12 +64,17 @@ var (
...
-
 		The command also requires a connection to the target cluster, and a valid
 		registry credentials to retrieve the required information from the target
 		cluster release.
+
+		A nodes-config.yaml config file must be created to provide the required
+		initial configuration for the selected nodes. 
+		Alternatively, to support simpler configurations for adding just a single
+		node, it's also possible to use a set of flags to configure the host. In
+		such case the '--mac-address' it's the only mandatory flag - while all the
```
pawanpinjarkar, line 75: nit: `'--mac-address' is the only mandatory`

pkg/cli/admin/nodeimage/create.go
```diff
@@ -71,6 +86,13 @@ var (
 
 		# Specify a custom image name
 		  oc adm node-image create --o=my-node.iso
+
+		# Create an ISO to add a single node without using the configuration file
+		  oc adm node-image create --mac-address=00:d8:e7:c7:4b:bb
+
+		# Create an ISO to add a single node without using the configuration file
```
pawanpinjarkar, line 93: nit: `Create an ISO to add a single node with a root device hint and without using the configuration file`

pkg/cli/admin/nodeimage/create.go
```diff
@@ -116,17 +138,36 @@ type CreateOptions struct {
...
 // AddFlags defined the required command flags.
 func (o *CreateOptions) AddFlags(cmd *cobra.Command) {
 	flags := o.addBaseFlags(cmd)
 
 	flags.StringVar(&o.AssetsDir, "dir", o.AssetsDir, "The path containing the configuration file, used also to store the generated artifacts.")
 	flags.StringVarP(&o.OutputName, "output-name", "o", "node.iso", "The name of the output image.")
+
+	flags.StringP(snFlagMacAddress, "m", "", "Single node flag. MAC address used to identify the host to apply the configuration. If specified, the nodes-config.yaml config file will not be used.")
+	usageFmt := "Single node flag. %s. Valid only when `mac-address` is defined."
+	flags.StringP(snFlagCpuArch, "c", "", fmt.Sprintf(usageFmt, "The CPU architecture to be use to install the node"))
```
pawanpinjarkar, line 166: nit: `to be used`

pkg/cli/admin/nodeimage/create.go
```diff
@@ -116,17 +138,36 @@ type CreateOptions struct {
...
+	flags.StringP(snFlagSshKey, "k", "", fmt.Sprintf(usageFmt, "The SSH key used to access the node"))
+	flags.String(snFlagHostname, "", fmt.Sprintf(usageFmt, "The hostname to be set for the node"))
+	flags.String(snFlagRootDeviceHint, "", fmt.Sprintf(usageFmt, "Hint for specifying the storage location for the image root filesystem"))
```
pawanpinjarkar, line 169: nit: `Root device hint to specify the storage location for the image root filesystem` Similar verbiage to other flags.

pkg/cli/admin/nodeimage/create.go
```diff
@@ -116,17 +138,36 @@ type CreateOptions struct {
...
+	flags.String(snFlagNetworkConfig, "", fmt.Sprintf(usageFmt, "File containing the NMState configuration to be applied for the node"))
```
pawanpinjarkar, line 170: nit: `File to specify the NMState configuration to be applied for the node`

rwsu: I suggest stating the file format is yaml.

pkg/cli/admin/nodeimage/create.go
```diff
@@ -116,17 +138,36 @@ type CreateOptions struct {
...
 func (o *CreateOptions) AddFlags(cmd *cobra.Command) {
 	flags := o.addBaseFlags(cmd)
 
 	flags.StringVar(&o.AssetsDir, "dir", o.AssetsDir, "The path containing the configuration file, used also to store the generated artifacts.")
 	flags.StringVarP(&o.OutputName, "output-name", "o", "node.iso", "The name of the output image.")
+
+	flags.StringP(snFlagMacAddress, "m", "", "Single node flag. MAC address used to identify the host to apply the configuration. If specified, the nodes-config.yaml config file will not be used.")
+	usageFmt := "Single node flag. %s. Valid only when `mac-address` is defined."
+	flags.StringP(snFlagCpuArch, "c", "", fmt.Sprintf(usageFmt, "The CPU architecture to be use to install the node"))
+	flags.StringP(snFlagSshKey, "k", "", fmt.Sprintf(usageFmt, "The SSH key used to access the node"))
```
rwsu, line 167: Does the key have to be encoded or formatted in a certain way to be specified as a single line string?

pkg/cli/admin/nodeimage/create.go
rwsu: It would be helpful to mention what is the root device hint format. Can the help refer to a url that describes the format?

pkg/cli/admin/nodeimage/create.go
pawanpinjarkar: My assumption was this will be the filepath instead of putting the actual key). If its a filepath, please update the message text.

andfasano: It's exactly the same of https://github.com/openshift/installer/blob/a651e1a49ceafcb5cbe6f6244e315a2381013aab/data/data/install.openshift.io_installconfigs.yaml#L4677-L4680 (note: it's not a path).
The node-joiner applies the standard ssh key validation, a format example it's shown in this [test](https://github.com/openshift/installer/blob/a651e1a49ceafcb5cbe6f6244e315a2381013aab/pkg/asset/agent/joiner/addnodesconfig_test.go#L39)

pkg/cli/admin/nodeimage/create.go
andfasano: I though to delegate more of this stuff to the documentation, but probably a direct mention here could be useful

pkg/cli/admin/nodeimage/create.go
```diff
@@ -71,6 +86,13 @@ var (
 
 		# Specify a custom image name
 		  oc adm node-image create --o=my-node.iso
+
+		# Create an ISO to add a single node without using the configuration file
+		  oc adm node-image create --mac-address=00:d8:e7:c7:4b:bb
```
zaneb, line 91: What effect does this actually have?

pkg/cli/admin/nodeimage/create.go
```diff
@@ -116,17 +138,36 @@ type CreateOptions struct {
...
+}
+
 // AddFlags defined the required command flags.
 func (o *CreateOptions) AddFlags(cmd *cobra.Command) {
 	flags := o.addBaseFlags(cmd)
 
 	flags.StringVar(&o.AssetsDir, "dir", o.AssetsDir, "The path containing the configuration file, used also to store the generated artifacts.")
 	flags.StringVarP(&o.OutputName, "output-name", "o", "node.iso", "The name of the output image.")
+
+	flags.StringP(snFlagMacAddress, "m", "", "Single node flag. MAC address used to identify the host to apply the configuration. If specified, the nodes-config.yaml config file will not be used.")
```
zaneb, line 164: I think we should omit this flag and apply any config given on the command line unconditionally to whichever host gets booted with the ISO. (I recognise that this requires changes to the installer and likely our client for assisted-service - perhaps we can take the opportunity to make the same improvement for SNO on day 1.) This would relieve the user of the need to find out the MAC address of the host, which can be a giant pain, without any loss of functionality.

pkg/cli/admin/nodeimage/create.go
zaneb: Let's read the key from a file path passed as an arg here.
Everybody already has their SSH public key in a file.
Nobody wants to deal with a 400 character-long field in the middle of a CLI command.
Picking a sensible default will require us to read it from the file anyway, so this makes it simpler as well.

pkg/cli/admin/nodeimage/create.go
rwsu: @andfasano indicated that --mac-address is supposed to be the only flag that is required if the nodes-config.yaml file is not specified. But after testing this patch, I can enter any valid mac-address and the node still is added to the cluster if the node has an IP address configured on the DHCP server.

pkg/cli/admin/nodeimage/create.go
rwsu: Removing the mac address flag will require changes to the installer. There is [a validation](https://github.com/openshift/installer/blob/fe126cafe857fa5e4e7f88a2134a19ae2f089b85/pkg/asset/agent/agentconfig/agenthosts.go#L156) that checks that at least one MAC address is specified for each host. I like the idea of making the MAC address optional, but it would require more planning.

rwsu:
Since Andrea is out for a few weeks, I've created another PR to move this forward: https://github.com/openshift/oc/pull/1841
The new PR makes the ssh-key flag a path to file containing the key and adds a unit test to exercise createConfigFileFromFlags.

PR #1836, opened 2024-08-01
WRKLDS-1316: Retry and do not fail when logs cannot be updated

rwsu:
The monitoring command can often times fail prematurely if it has trouble reading the node-joiner-monitor container logs.

Example:
E0801 11:18:52.520562 3198099 monitor.go:169] monitoring never finished: http2: server sent GOAWAY and closed the connection; LastStreamID=23, ErrCode=NO_ERROR, debug=""

In such cases, the command should not exit but retry.

Also added retry for initial Pod fetch in
waitForMonitoringToComplete and additional logging to aid in debugging.

### DIFF
--- a/pkg/cli/admin/nodeimage/monitor.go
+++ b/pkg/cli/admin/nodeimage/monitor.go
@@ -9,9 +9,9 @@ import (
 
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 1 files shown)


pawanpinjarkar (commented):
/lgtm

PR #1837, opened 2024-08-01
WRKLDS-1429: Adds a guardrail for node-image commands which is supported only for Openshift version 4.17+

pawanpinjarkar:
Adds a guardrail for node-image commands which is supported only for Openshift version 4.17+

### DIFF
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -9,6 +9,7 @@ import (
 	"io/fs"
 	"os"
 	"path"
+	"regexp"
 	"strconv"
 	"time"
 
@@ -40,8 +41,9 @@ import (
 )
 
 const (
-	nodeJoinerConfigurationFile = "nodes-config.yaml"
-	nodeJoinerContainer         = "node-joiner"
+	nodeJoinerConfigurationFile       = "nodes-config.yaml"
+	nodeJoinerContainer               = "node-joiner"
+	nodeJoinerMinimumSupportedVersion = "4.17"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/nodeimage/create.go
```diff
@@ -471,6 +472,10 @@ func (c *BaseNodeImageCommand) fetchClusterReleaseImage(ctx context.Context) (st
 			return "", fmt.Errorf("command expects a connection to an OpenShift 4.x server")
 		}
 	}
+	// Adds a guardrail for node-image commands which is supported only for Openshift version 4.17 and later
+	if isVersionBefore(cv.Status.Desired.Version, 4, 17) {
```
rwsu, line 476: Let's use an existing library to check versions. Here's an example: https://github.com/openshift/assisted-installer/blob/37fd43a46b3743208c1ddea98cf3b98faf0792ad/src/common/common.go#L223

rwsu (changes requested):

pkg/cli/admin/nodeimage/create_test.go
```diff
@@ -133,6 +133,14 @@ func TestRun(t *testing.T) {
 		}
 	}
 
+	ClusterVersion_4_16_ObjectFn := func(repo string, manifestDigest string) []runtime.Object {
```
rwsu, line 136: Unit tests are failing because defaultClusterVersionObjectFn's Status.Desired.Version needs to be set. The same should be applied to monitor_test.go.

rwsu (changes requested):

pkg/cli/admin/nodeimage/create_test.go
pawanpinjarkar: Tests are now failing, added a check in the code

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/containers/image/v5 v5.30.1
 	github.com/coreos/go-oidc/v3 v3.10.0
+	github.com/coreos/go-semver v0.3.1
```
ardaguclu, line 19: We prefer not to add any new dependencies, unless there is strong justification as it brings about it's own maintenance burden. So, could you please remove this dependency and rely only simple regexes, e.g. https://github.com/openshift/oc/blob/e77658e7eaf7dfe310e779ad9b6a74e581a3715c/pkg/cli/admin/release/extract_tools.go#L1246

pkg/cli/admin/nodeimage/create.go
```diff
@@ -471,6 +472,14 @@ func (c *BaseNodeImageCommand) fetchClusterReleaseImage(ctx context.Context) (st
 			return "", fmt.Errorf("command expects a connection to an OpenShift 4.x server")
 		}
 	}
+	// Adds a guardrail for node-image commands which is supported only for Openshift version 4.17 and later
+	currentVersion := cv.Status.Desired.Version
+	matches := regexp.MustCompile(`^(\d+[.]\d+)[.].*`).FindStringSubmatch(currentVersion)
+	if len(matches) < 2 {
+		return "", fmt.Errorf("failed to parse major.minor version from ClusterVersion status.desired.version %q", currentVersion)
+	} else if matches[1] != "4.17" {
```
zaneb, line 480: won't this fail on 4.18?

pawanpinjarkar: right, it should be `<4.17`

rwsu (commented):
/lgtm

PR #1842, opened 2024-08-08
Include must-gather image name in ENV

black-dragon74:
The request for such feature came from a Bugzilla
for odf-must-gather.

We want the must-gather containers to be able to check the image being used. This is useful in cases where we want to deduce if the customer is using the image we ask them to.

While it is possible to call the k8s API internally (this is how it is implemented currently for odf-mg), it would be much efficient if the value is present inside the container as an ENV.

This patch sets an ENV var `MUST_GATHER_IMAGE` to reflect the image source of the pods.

Regards


black-dragon74:
PTAL @ardaguclu @soltysh

PR #1855, opened 2024-08-27
OCPBUGS-38450:  Fix copy artifacts for all CPU architectures

rwsu:
The "oc adm node-image create" command currently expects the filename to be node.x86_64.iso which doesn't work for non x86_64 architectures. Updated RsyncOptions to copy any iso file to include files for any CPU architecture.

Authored by @andfasano

### DIFF
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"io/fs"
 	"os"
-	"path"
 	"regexp"
 	"strconv"
 	"strings"
@@ -331,15 +330,17 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
 	klog.V(2).Infof("Copying artifacts from %s", o.nodeJoinerPod.GetName())
 	rsyncOptions := &rsync.RsyncOptions{
 		Namespace:     o.nodeJoinerNamespace.GetName(),
-		Source:        &rsync.PathSpec{PodName: o.nodeJoinerPod.GetName(), Path: path.Join("/assets", "node.x86_64.iso")},
+		Source:        &rsync.PathSpec{PodName: o.nodeJoinerPod.GetName(), Path: "/assets/"},
 		ContainerName: nodeJoinerContainer,
-		Destination:   &rsync.PathSpec{PodName: "", Path: path.Join(o.AssetsDir, o.OutputName)},
+		Destination:   &rsync.PathSpec{PodName: "", Path: o.AssetsDir},
 		Client:        o.Client,
 		Config:        o.Config,
 		Compress:      true,
 		RshCmd:        fmt.Sprintf("%s --namespace=%s -c %s", o.rsyncRshCmd, o.nodeJoinerNamespace.GetName(), nodeJoinerContainer),
 		IOStreams:     o.IOStreams,
 		Quiet:         true,
+		RsyncInclude:  []string{"*.iso"},
+		RsyncExclude:  []string{"*"},
 	}
 	rsyncOptions.Strategy = o.copyStrategy(rsyncOptions)
 	return rsyncOptions.RunRsync()
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -307,6 +308,12 @@ func (o *CreateOptions) Run() error {
 	if err != nil {
 		return err
 	}
+
+	err = o.renameImageIfOutputNameIsSpecified()
+	if err != nil {
+		return err
+	}
+
 	klog.V(1).Info("Command successfully completed")
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/nodeimage/create.go
```diff
@@ -345,6 +375,22 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
...
+	case string(machineconfig.Amd64):
+		architectureFileName = string(machineconfig.X86_64)
+	case "":
+		architectureFileName = string(machineconfig.X86_64) // default to X86_64 if unspecified
+	default:
+		architectureFileName = o.CPUArchitecture
+	}
+	return fmt.Sprintf(isoNameFormat, architectureFileName)
+}
+
```
andfasano, line 393: Also this block should be removed, as it requires a knowledge at oc level that it'd be better to avoid. The only safe assumption is that the node-joiner will create an ISO file in case of success. I spent some time experimenting with the `rsync` option, and by using the following code it will be possible to copy the generated ISO with wildcards:

```
	rsyncOptions := &rsync.RsyncOptions{
		Namespace:     o.nodeJoinerNamespace.GetName(),
		Source:        &rsync.PathSpec{PodName: o.nodeJoinerPod.GetName(), Path: "/assets/"},
		ContainerName: nodeJoinerContainer,
		Destination:   &rsync.PathSpec{PodName: "", Path: o.AssetsDir},
		Client:        o.Client,
		Config:        o.Config,
		Compress:      true,
		RshCmd:        fmt.Sprintf("%s --namespace=%s -c %s", o.rsyncRshCmd, o.nodeJoinerNamespace.GetName(), nodeJoinerContainer),
		IOStreams:     o.IOStreams,
		Quiet:         true,
		RsyncInclude:  []string{"*.iso"},
		RsyncExclude:  []string{"*"},
	}
```

rwsu: Thanks for finding a better and more simpler way to fix the problem. I'll update the PR as the rest of the changes aren't needed.

pkg/cli/admin/nodeimage/create.go
```diff
@@ -331,20 +337,64 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
...
 	}
 	rsyncOptions.Strategy = o.copyStrategy(rsyncOptions)
 	return rsyncOptions.RunRsync()
 }
 
+func (o *CreateOptions) renameImageIfOutputNameIsSpecified() error {
+	if o.OutputName == "" {
+		return nil
+	}
+	// AssetDir doesn't exist in unit test fake filesystem
```
andfasano, line 360:
```suggestion
	// AssetDir doesn't exist in unit test fake filesystem
```

pkg/cli/admin/nodeimage/create.go
```diff
@@ -331,20 +337,64 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
...
+	err = filepath.Walk(o.AssetsDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info.IsDir() && info.Name() != o.OutputName && strings.HasSuffix(info.Name(), ".iso") {
+			newPath := filepath.Join(filepath.Dir(path), o.OutputName)
+
+			// Check if another file has the same name
+			if _, err := os.Stat(newPath); err == nil {
```
andfasano, line 379: Since `newPath` is invariant, couldn't be moved out this check before starting looking in the folder?

pkg/cli/admin/nodeimage/create.go
```diff
@@ -331,20 +337,64 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
...
+				return fmt.Errorf("file already exists: %s", newPath)
+			}
+
+			err := os.Rename(path, newPath)
```
andfasano, line 383: We do not expect more than one file to be renamed, isn't it?

pkg/cli/admin/nodeimage/create.go
```diff
@@ -331,20 +337,64 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
...
+	_, err := os.Stat(o.AssetsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		} else {
+			return err
+		}
+	}
+
+	err = filepath.Walk(o.AssetsDir, func(path string, info os.FileInfo, err error) error {
```
andfasano, line 370: wondering if in this case using [os.ReadDir](https://pkg.go.dev/os#ReadDir) could lead to a simpler implementation

pkg/cli/admin/nodeimage/create.go
rwsu: That would assume the ISO isn't written to a subdirectory under o.AssetDir. Which is the case now. But by not assuming it is invariant, makes it less coupled to the node-joiner implementation.

pkg/cli/admin/nodeimage/create.go
rwsu: Yes, we can end the filepath.Walk if a rename is successful.

pkg/cli/admin/nodeimage/create.go
rwsu: I tried os.ReadDir. It is pretty much the same.

pkg/cli/admin/nodeimage/create.go
rwsu: I don't see a difference here.

andfasano: My suggestion was to remove the comment, given that it's related to a test

pkg/cli/admin/nodeimage/create.go
```diff
@@ -331,20 +337,66 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
...
+func (o *CreateOptions) renameImageIfOutputNameIsSpecified() error {
+	if o.OutputName == "" {
+		return nil
+	}
+	// AssetDir doesn't exist in unit test fake filesystem
+	_, err := os.Stat(o.AssetsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		} else {
```
andfasano, line 365: For code readability, I'd prefer to use more happy path here

PR #1857, opened 2024-08-27
OCPBUGS-38990: use proxy settings when required

andfasano:
Currently the `oc adm node-image create` command ignored the cluster-wide proxy settings when running the node-joiner pod, thus the command will fail in a scenario where a proxy is defined but the cluster does not have internet access (to fetch / inspect the release image).
This patch fixes the issue by configuring properly the node-joiner pod when the cluster-wide proxy `cluster` resource was found in the target cluster.


pawanpinjarkar (commented):
/lgtm

PR #1863, opened 2024-09-02
OTA-1271: Create the `oc adm upgrade recommend` subcommand

Davoska:
This PR will add the backbones for the new `oc adm upgrade recommend` subcommand.

The logic is copied over from an existing `github.com/openshift/oc/pkg/cli/admin/upgrade` package. The logic is to be changed over the upcoming iterations of development.

For the initial commit, the new `recommend` package contains the logic from the `github.com/openshift/oc/pkg/cli/admin/upgrade` package as of the 93286c9260285c104259f0c50704c4b880c4fec3 commit. Only the logic regarding the default output of the `oc adm upgrade` command and outputs regarding available or available but not recommended updates was copied over.

This logic was placed behind a feature gate environment variable named `OC_ENABLE_CMD_UPGRADE_RECOMMEND`. The new subcommand can be tested by running:

```
$ OC_ENABLE_CMD_UPGRADE_RECOMMEND=true ./oc adm upgrade recommend
```

As most of the internal functions are not exported, the existing tests were copied over as well for the existing used functions in the new package `recommend`.

This pull request references https://issues.redhat.com/browse/OTA-1271

### DIFF
--- a/pkg/cli/admin/upgrade/recommend/recommend.go
+++ b/pkg/cli/admin/upgrade/recommend/recommend.go
@@ -1,0 +1,243 @@
+// Package recommend displays recommended update information.
+package recommend
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/blang/semver"
+	"github.com/spf13/cobra"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+)
+
+const (
+	// clusterStatusFailing is set on the ClusterVersion status when a cluster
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/upgrade/recommend/recommend.go
```diff
@@ -0,0 +1,236 @@
...
+	return &options{
+		IOStreams: streams,
+	}
+}
+
+func New(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command {
+	o := newOptions(streams)
+	cmd := &cobra.Command{
+		Use:   "recommend",
+		Short: "Recommends cluster updates.",
```
wking, line 40: nit: Maybe `Displays cluster update recommendations.`  or something else that includes "display" or other wording to underline the read-only-ness?  Maybe also additional sentences like:

> This subcommand is read-only, and does not affect the state of the cluster.  To initiate an update, use the 'oc adm upgrade ...' subcommand.

wking (approved):
/lgtm

In case you want to address [this nit][1]:

/hold

But feel free to lift the hold without changing anything if you want to punt on expanding that string.

[1]: https://github.com/openshift/oc/pull/1863#discussion_r1742458994

pkg/cli/admin/upgrade/recommend/recommend.go
Davoska: Thank you for the suggestion. `Displays cluster update recommendations.` looks better IMO.

I have also added a longer description.

PR #1864, opened 2024-09-03
OCPBUGS-39375: Update push targets of digest with new appended tags

ardaguclu:
Code actually has an ability to handle different tags with same layers. However, it is forgotten to update the tag list with the new list. This PR fixes this.


dgoodwin:
We discussed on TRT standup, the issue affecting the agnostic job is a known issue outside this PR, and the coverage via other jobs is reasonable to ensure this change should be good. We're ok with merging this to help the kube rebase out.

/override ci/prow/e2e-agnostic-ovn-cmd

PR #1869, opened 2024-09-05
WIP: CLID-160: upgrade distribution/distribution to the latest released version

aguidirh:
Currently in oc-mirror it is not possible to upgrade containers/image dependency because of the indirect dependency on oc. The upgrade of containers/image is required to fix CVE-2024-3727

In order to upgrade the containers/image it is required to upgrade the distribution/distribution version in oc. Before this PR the oc was depending on a non released version of distribution/distribution `v3.0.0-20230519140516-983358f8e250`. This PR upgrades the oc to the [latest v3 release version](https://github.com/distribution/distribution/releases/tag/v3.0.0-beta.1).

Since the latest version of distribution removed the docker schema1 (which were already deprecated a long time ago), it was also removed from oc code the code that was still supporting docker image schema1. There is also this [deprecation notice from RedHat](https://access.redhat.com/solutions/4346011).

@ardaguclu please give an special attention to the files mirror.go and manifest.go where I removed the support for docker image schema1 and please let me know if you think it could bring us problems. I only removed because there is no place to refer schema1 anymore (actually there is one but downgrading the distribution would be a step back).

NOTE: unit tests are still failing and I am trying to find out why.


pkg/cli/image/imagesource/dryrun.go
```diff
@@ -5,8 +5,8 @@ import (
 	"net/http"
 
 	"github.com/distribution/distribution/v3"
-	"github.com/distribution/distribution/v3/reference"
-	registryclient "github.com/distribution/distribution/v3/registry/client"
+	"github.com/distribution/reference"
+	registryclient "github.com/openshift/library-go/pkg/registry/client"
```
ardaguclu, line 9: Our main motivation is to bump the distribution in this PR?. Why we put ourselves under the pressure of moving to library-go as well?. I'd prefer first we have a plain distribution bump in one PR and only after we merge it, we can switch to library-go in a followup PR?.

pkg/cli/image/manifest/manifest.go
```diff
@@ -558,155 +521,21 @@ func PutManifestInCompatibleSchema(
 	} else {
 		klog.V(5).Infof("Put manifest %s", ref)
 	}
-	switch t := srcManifest.(type) {
-	case *schema1.SignedManifest:
```
ardaguclu, line 562: We have to ensure that these types are not used anywhere, otherwise that would break scripts. We'd need to follow deprecation policy. Notify that we deprecate this in this release and remove in next release after that. Or we have already deprecate it?

pkg/helpers/image/test/util.go
```diff
@@ -21,6 +20,10 @@ import (
 	imagev1 "github.com/openshift/api/image/v1"
 )
 
+// Deprecated: Docker Image Manifest v2, Schema 1 is deprecated since 2015.
```
ardaguclu, line 23: hmm, this explains the above. Is there any reference for this?

ardaguclu (commented):
I know that this is WIP but dropped a few comments that we need to think on them anyway. Golang changes will disappear after the rebase to 1.31.

pkg/cli/image/manifest/manifest.go
ardaguclu: It looks like this is the link https://access.redhat.com/solutions/4346011 given in PR description.

pkg/cli/image/imagesource/dryrun.go
aguidirh: oc-mirror needs to upgrade distribution/distribution, but oc-mirror has indirect dependency of library-go and oc. So in order to upgrade distribution/distribution in oc-mirror we need to upgrade library-go and oc as well. Since oc depends on libray-go that's why I opened the PR to library-go first and added the removed packages of distribution/distribution there (since it is a common package for several components).

If we try to bump distribution/distribution in oc first, it will break saying that the new distribution/distribution versions does not have the code required by library-go.

pkg/cli/image/manifest/manifest.go
aguidirh: Yes, and it was deprecated by docker back in 2015

pkg/helpers/image/test/util.go
aguidirh: Yes, I found two references from official websites:

[CNCF distribution - current maintainer of distribution](https://distribution.github.io/distribution/spec/deprecated-schema-v1/)
[Docker - previous maintaner of distribution](https://docs.docker.com/engine/deprecated/#pushing-and-pulling-with-image-manifest-v2-schema-1)

PR #1874, opened 2024-09-11
Fixing -o flag in command example

bergerhoffer:
And also removing punctuation (short descriptions use no ending period).

No jira.


ardaguclu:
Sounds good to me. Thank you
/lgtm

PR #1875, opened 2024-09-12
OCPBUGS-41111: vendor: Update openshift/api to pick up v4.17 and v4.18 capability sets

wking:
Catching up with openshift/api#2022.  Generated with:

```console
$ go get github.com/openshift/api@master
$ go mod tidy
$ go mod vendor
$ git add -A go.* vendor
```

all using:

```console
$ go version
go version go1.22.2 linux/amd64
```

This addresses [OCPBUGS-41111][2]:

```console
$ cat v4.17-basecap.yaml
---
apiVersion: v1
platform:
  gcp:
    foo: bar
capabilities:
  baselineCapabilitySet: v4.17
$ ./oc adm release extract --install-config v4.17-basecap.yaml --included --credentials-requests --from quay.io/openshift-release-dev/ocp-release:4.17.0-rc.1-x86_64 --to /tmp/test
  error: unrecognized baselineCapabilitySet "v4.17"
```

because [`pkg/cli/admin/release/extract_tools.go` uses the vendored `openshift/api/config/v1` to unpack capabilities][3].

[1]: https://github.com/
[2]: https://issues.redhat.com/browse/OCPBUGS-41111
[3]: https://github.com/openshift/oc/blob/e3584b965787e221a84b81d53107c41d20250047/pkg/cli/admin/release/extract_tools.go#L77

PR #1876, opened 2024-09-12
OCPBUGS-41642: vendor: Update openshift/api to pick up the v4.17 capability set

wking:
Catching up with openshift/api#2023, and a backport-ish of 45ab914729 (#1875), which may or may not land [due to conflicts with the Kubernetes rebase][2].  Generated with:

```console
$ GOPROXY=direct go get github.com/openshift/api@release-4.17
$ go mod tidy
$ go mod vendor
$ git add -A go.* vendor
```

all using:

```console
$ go version
go version go1.22.2 linux/amd64
```

where `GOPROXY=direct` avoids [a caching delay][3]:

> *I committed a new change (or released a new version) to a repository, why isn't it showing up when I run `go get -u` or `go list -m --versions`?*
>
> In order to improve our services' caching and serving latencies, new versions may not show up right away. If you want new code to be immediately available in the mirror, then first make sure there is a semantically versioned tag for this revision in the underlying source repository. Then explicitly request that version via go get module@version. The new version should be available within one minute. Note that if someone requested the version before the tag was pushed, it may take up to 30 minutes for the mirror's cache to expire and fresh data about the version to become available. If the version is still not available after 30 minutes, please [file an issue](https://golang.org/issue/new?title=proxy.golang.org%3A+).

This addresses [OCPBUGS-41111][4]:

```console
$ cat v4.17-basecap.yaml
---
apiVersion: v1
platform:
  gcp:
    foo: bar
capabilities:
  baselineCapabilitySet: v4.17
$ ./oc adm release extract --install-config v4.17-basecap.yaml --included --credentials-requests --from quay.io/openshift-release-dev/ocp-release:4.17.0-rc.1-x86_64 --to /tmp/test
error: unrecognized baselineCapabilitySet "v4.17"
```

because [`pkg/cli/admin/release/extract_tools.go` uses the vendored `openshift/api/config/v1` to unpack capabilities][5].

[2]: https://github.com/openshift/oc/pull/1875#issuecomment-2345246831
[3]: https://proxy.golang.org/
[4]: https://issues.redhat.com/browse/OCPBUGS-41111
[5]: https://github.com/openshift/oc/blob/e3584b965787e221a84b81d53107c41d20250047/pkg/cli/admin/release/extract_tools.go#L77


jiajliu:
pre-merge verify pass

/label qe-approved

PR #1881, opened 2024-09-20
NO-JIRA: Update root-approvers of oc

ardaguclu:
This PR updates the root-aproovers of oc.


atiratree:
Thanks Arda! lgtm, but let's also wait for @ingvagabund :)

PR #1883, opened 2024-09-24
OCPBUGS-37344: Skip inspection of the daemonset pods running on data plane

ardaguclu:
Daemonset pods running on worker nodes generate high volume of log which increase storage size of must-gather output drastically in addition to the longer collection duration. On the other hand, our primary focus is to collect logs from control  plane rather than data plane. So that it looks reasonable to **not** collect daemonset pods running on data plane if the cluster's worker node count is greater than 18 (still collect logs of the pods whose are running on control plane of given daemonset).

This PR also introduces node-selector flag into the oc adm inspect command. This will work as a workaround, if someone want to collect logs of those skipped pods like this;

```
oc adm inspect namespace openshift-multus --node-selector='node-role.kubernetes.io/worker'
```

### DIFF
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -75,6 +75,7 @@ type InspectOptions struct {
 	rotatedPodLogs bool
 	sinceInt       int64
 	sinceTimestamp metav1.Time
+	nodeSelector   string
 
 	// directory where all gathered data will be stored
 	DestDir string
@@ -91,10 +92,11 @@ func NewInspectOptions(streams genericiooptions.IOStreams) *InspectOptions {
 		printFlags.JSONYamlPrintFlags.ShowManagedFields = true
 	}
 	return &InspectOptions{
-		printFlags:  printFlags,
-		configFlags: genericclioptions.NewConfigFlags(true),
-		overwrite:   true,
-		IOStreams:   streams,
+		printFlags:   printFlags,
+		configFlags:  genericclioptions.NewConfigFlags(true),
+		nodeSelector: "node-role.kubernetes.io/control-plane",
+		overwrite:    true,
+		IOStreams:    streams,
 	}
 }
 
@@ -117,6 +119,7 @@ func NewCmdInspect(streams genericiooptions.IOStreams) *cobra.Command {
 	cmd.Flags().BoolVarP(&o.allNamespaces, "all-namespaces", "A", o.allNamespaces, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
 	cmd.Flags().StringVar(&o.sinceTime, "since-time", o.sinceTime, "Only return logs after a specific date (RFC3339). Defaults to all logs. Only one of since-time / since may be used.")
 	cmd.Flags().DurationVar(&o.since, "since", o.since, "Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs. Only one of since-time / since may be used.")
+	cmd.Flags().StringVar(&o.nodeSelector, "node-selector", o.nodeSelector, "Set a specific node selector to use - only relevant for the daemonset pods. If the cluster's worker node count is less than or equal to 18, this flag will be ignored and every daemonset pod in all nodes will be collected.")
 	cmd.Flags().BoolVar(&o.rotatedPodLogs, "rotated-pod-logs", o.rotatedPodLogs, "Experimental: If present, retrieve rotated log files that are available for selected pods. This can significantly increase the collected logs size. since/since-time will be matched against the date in the log file name.")
 
 	// The rotated-pod-logs option should be removed once support for retrieving rotated logs is added to kubelet
--- a/pkg/cli/admin/inspect/namespace.go
+++ b/pkg/cli/admin/inspect/namespace.go
@@ -9,6 +9,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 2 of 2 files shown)


pkg/cli/admin/inspect/namespace.go
```diff
@@ -92,3 +101,65 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
...
+	}
+	if !isDaemonSet {
+		return true, nil
+	}
+
+	nodes, err := o.kubeClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return true, err
+	}
+	const daemonsetCollectionMaxNodeSize = 20
```
ardaguclu, line 128: Would it be 10 better?

pkg/cli/admin/inspect/inspect.go
```diff
@@ -117,6 +118,7 @@ func NewCmdInspect(streams genericiooptions.IOStreams) *cobra.Command {
 	cmd.Flags().BoolVarP(&o.allNamespaces, "all-namespaces", "A", o.allNamespaces, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
 	cmd.Flags().StringVar(&o.sinceTime, "since-time", o.sinceTime, "Only return logs after a specific date (RFC3339). Defaults to all logs. Only one of since-time / since may be used.")
 	cmd.Flags().DurationVar(&o.since, "since", o.since, "Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs. Only one of since-time / since may be used.")
+	cmd.Flags().StringVar(&o.nodeSelector, "node-selector", o.nodeSelector, "Experimental: Set a specific node selector to use - only relevant for the daemonset pods running on data plane nodes, since the default may discard them from inspection.")
```
ingvagabund, line 121: As a user I'd expect to collect every pod when `--node-selector` is not specified. What about to set the node selector to `node-role.kubernetes.io/control-plane` by default? Presumably any node with `node-role.kubernetes.io/control-plane` is also labeled with `node-role.kubernetes.io/master` in our installations. So only the former is needed.

Also, this new experimental flag already changes the default behavior. So it can not be marked as experimental. Let's promote the flag to a GA flag and mark this as a bug fix as collecting the DS pods from data nodes has never been an intention to perform by default.

pkg/cli/admin/inspect/namespace.go
ingvagabund: Note: it's ok to keep this as a non-configurable limit until we get complains/requests to change the value. At which point we can turn it into a new flag.

pkg/cli/admin/inspect/namespace.go
```diff
@@ -92,3 +101,65 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
...
+	if nodes.Items == nil || len(nodes.Items) < daemonsetCollectionMaxNodeSize {
```
ingvagabund, line 129: Given this is mainly for the data nodes what about to extract the number of control plane nodes from len(nodes.Items) first and then check for the limit?

ardaguclu: 20 is for total number of nodes in the cluster regardless of data plane or control plane. Do you suggest using a threshold based on the number of data plane nodes?. It makes sense. Do you have any threshold suggestion?. What about 15?

pkg/cli/admin/inspect/inspect.go
ardaguclu: Thank you for dropping your comment. I've updated the PR with respect to your changes. Could you PTAL?.

pkg/cli/admin/inspect/namespace.go
ingvagabund:
```
$ echo $((10 + $RANDOM % 10))
18
```

pkg/cli/admin/inspect/inspect.go
```diff
@@ -117,6 +118,7 @@ func NewCmdInspect(streams genericiooptions.IOStreams) *cobra.Command {
 	cmd.Flags().BoolVarP(&o.allNamespaces, "all-namespaces", "A", o.allNamespaces, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
 	cmd.Flags().StringVar(&o.sinceTime, "since-time", o.sinceTime, "Only return logs after a specific date (RFC3339). Defaults to all logs. Only one of since-time / since may be used.")
 	cmd.Flags().DurationVar(&o.since, "since", o.since, "Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs. Only one of since-time / since may be used.")
+	cmd.Flags().StringVar(&o.nodeSelector, "node-selector", o.nodeSelector, "Set a specific node selector to use - only relevant for the daemonset pods. If the cluster's node count is less than or equal to 20, default is empty. If the cluster's node count is greater than 20, default is set to 'node-role.kubernetes.io/control-plane'")
```
ingvagabund, line 121: Can you set the default value in `NewInspectOptions` as part of `return &InspectOptions{`?

ardaguclu: Default value is dynamic based on the node count. Since there is no kubeconfig in `NewInspectOptions` (because initializations happen in `Complete` func), we show default as empty which is the default of "node count is less than threshold".

ardaguclu: After thinking on this, I'll apply the suggested change.

ardaguclu: Updated.

pkg/cli/admin/inspect/namespace.go
```diff
@@ -92,3 +101,67 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
...
+				isDaemonSet = true
+				break
+			}
+		}
+	}
+	if !isDaemonSet {
+		return true, nil
+	}
+
+	nodes, err := o.kubeClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
```
deads2k, line 124: do this once before iterating the pods.  Then use the list of nodes.  If you don't find a node in the list, assume you should collect the pod logs.

ardaguclu: Updated.

pkg/cli/admin/inspect/namespace.go
```diff
@@ -74,12 +75,28 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
 	}
 
 	klog.V(1).Infof("    Gathering pod data for namespace %q...\n", namespace)
+	nodeList, err := o.kubeClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
```
ingvagabund, line 78: Just a nit: You could put the `nodeList` code under `if pods := resourcesToStore[corev1.SchemeGroupVersion.WithResource("pods")]; pods != nil {` condition as that's where it's used. Just in case the inspect command will not inspect pods resource to avoid listing the nodes unnecessarly.

ardaguclu: Good idea. Updated. Thank you for careful review :).

pkg/cli/admin/inspect/namespace.go
```diff
@@ -92,3 +109,58 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
...
+		workerNodeCount++
+	}
+	const daemonsetCollectionMaxNodeSize = 18
+	if workerNodeCount <= daemonsetCollectionMaxNodeSize {
+		return true, nil
+	}
+
+	var podNode *corev1.Node
+	for _, node := range nodes {
+		if node.Name != pod.Spec.NodeName {
```
ingvagabund, line 146: This tests whether a DS pod is scheduled to a non-control-plane node. Would it make more sense to do the opposite and check whether a DS pod is "not" scheduled to a control plane node? Usually the number of control-plane nodes is much lower than the non-control-plane node.

ingvagabund: Gotcha. Then the podEligibleForCollection's description needs to be updated to better phrase the "running on data planes" since the default value of the node selector is not known in advance.

ardaguclu: I've updated the description. Please let me know what you think. Thank you.

ingvagabund:
/lgtm
Feel free to unhold

pkg/cli/admin/inspect/namespace.go
rob-stl: Thank you for this, friends.  However, the customer from which this originates has impacted clusters with as few as 12 total nodes.  As such, please consider lowering the threshold to 8 data plane node (to trigger skipping daemonset pods.)

oarribas: If that is mainly impacting one customer at this moment, and due to the missing information reducing that value more could cause for lots of other cases (in whose it will be needed to ask for the DS pod inspection again), maybe it would be better to allow a parameter to configure that value, or even better, a parameter to disable the full inspection of DS pods, and ask that specific customer to run all their must-gathers with that specific parameter.

PR #1887, opened 2024-09-27
OCPBUGS-42528: fix outputname flag for node-image create command

andfasano:
By default the name of the output image is the one naturally provided by the node-joiner - since it includes also the target architecture prefix. The renaming must be applied only if specified by the user.


pawanpinjarkar (commented):
/lgtm

rwsu:
We will need to update the  dev-scripts agent_add_node target as it expects the filename to be node.iso: https://github.com/openshift-metal3/dev-scripts/pull/1703

PR #1898, opened 2024-10-14
WRKLDS-1320: Support PXE files with "node-image create"

rwsu:
Adds --pxe flag to "oc adm node-image create".

This causes the command to generate PXE files in place of the ISO.


andfasano:
I think it could be safe to override the aws job, if it keeps failing

PR #1901, opened 2024-10-17
API-1835: gather discovery when using inspect

deads2k:
Required for reading the content back out.

### DIFF
--- a/pkg/cli/admin/inspect/discovery.go
+++ b/pkg/cli/admin/inspect/discovery.go
@@ -1,0 +1,62 @@
+package inspect
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/client-go/discovery"
+	"path/filepath"
+	"sigs.k8s.io/yaml"
+)
+
+func inspectDiscovery(ctx context.Context, destDir string, discoveryClient discovery.CachedDiscoveryInterface) error {
+	if err := writeDiscovery(ctx, filepath.Join(destDir, "aggregated-discovery-apis.yaml"), discoveryClient, "/apis"); err != nil {
+		return err
+	}
+	if err := writeDiscovery(ctx, filepath.Join(destDir, "aggregated-discovery-api.yaml"), discoveryClient, "/api"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+var v2GVK = schema.GroupVersionKind{Group: "apidiscovery.k8s.io", Version: "v2", Kind: "APIGroupDiscoveryList"}
+
+func writeDiscovery(ctx context.Context, destFile string, discoveryClient discovery.CachedDiscoveryInterface, url string) error {
+	var responseContentType string
+	var statusCode int
+	apiBytes, err := discoveryClient.RESTClient().Get().
+		AbsPath(url).
+		SetHeader("Accept", discovery.AcceptV2).
+		Do(ctx).
+		ContentType(&responseContentType).
+		StatusCode(&statusCode).
+		Raw()
+	if err != nil {
+		return fmt.Errorf("discovery %q failed: %w", url, err)
+	}
+
+	if statusCode < 200 && statusCode > 299 {
+		return fmt.Errorf("discovery %q returned with status %v", url, statusCode)
+	}
+	if ok, _ := discovery.ContentTypeIsGVK(responseContentType, v2GVK); !ok {
+		return fmt.Errorf("discovery %q returned with contentType %v", url, responseContentType)
+	}
+
+	apiMap := map[string]interface{}{}
+	if err := json.Unmarshal(apiBytes, &apiMap); err != nil {
+		return fmt.Errorf("discovery %q unmarshal failed: %w", url, err)
+	}
+	apiYAML, err := yaml.Marshal(apiMap)
+	if err != nil {
+		return fmt.Errorf("discovery %q marshal failed: %w", url, err)
+	}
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/inspect/discovery.go
```diff
@@ -0,0 +1,56 @@
+package inspect
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/apimachinery/pkg/util/json"
```
ardaguclu, line 6: nit: go import

pkg/cli/admin/inspect/discovery.go
```diff
@@ -0,0 +1,56 @@
...
+		SetHeader("Accept", discovery.AcceptV2).
+		Do(ctx).
+		ContentType(&responseContentType).
+		StatusCode(&statusCode).
+		Raw()
+	if err != nil {
+		return fmt.Errorf("discovery %q failed: %w", url, err)
+	}
+
+	if statusCode < 200 && statusCode > 299 {
```
ardaguclu, line 39: Why we are not checking only `statusCode != http.StatusOK`. Since we are not sending any etags (may result in 304), the only expected status should be 200?

pkg/cli/admin/inspect/discovery.go
```diff
@@ -0,0 +1,56 @@
...
+		return fmt.Errorf("discovery %q returned with status %v", url, statusCode)
+	}
```
ardaguclu, line 41: I think we can also ensure that the content type is discovery/v2 as in https://github.com/kubernetes/kubernetes/blob/462dd326c2e98d937a96d49002883000efe4b2d6/staging/src/k8s.io/client-go/discovery/discovery_client.go#L330 (in case library-go changes rely on discovery/v2)

```go
if isGVK, err := discovery.ContentTypeIsGVK(responseContentType, schema.GroupVersionKind{Group: "apidiscovery.k8s.io", Version: "v2", Kind: "APIGroupDiscoveryList"}); !isGVK || err != nil {
return fmt.Errorf("unexpected")
}
```

pkg/cli/admin/inspect/inspect.go
```diff
@@ -243,6 +243,10 @@ func (o *InspectOptions) Run() error {
 		return err
 	}
 
+	if err := InspectDiscovery(context.Background(), o.DestDir, discoveryClient); err != nil {
+		return fmt.Errorf("failed inspecting discovery: %w", err)
```
ardaguclu, line 247: In my opinion, we can continue inspection after accumulating this error to allErrs. Otherwise, any failure would stop the entire inspection.

ardaguclu (commented):
the resulted files sizes are negligible and almost fixed;

```
117K Oct 18 08:59 discovery-apis.yaml
9.5K Oct 18 08:59 discovery-api.yaml
```

pkg/cli/admin/inspect/discovery.go
```diff
@@ -0,0 +1,56 @@
...
+	"k8s.io/apimachinery/pkg/util/json"
+	"os"
+	"path/filepath"
+	"sigs.k8s.io/yaml"
+
+	"k8s.io/client-go/discovery"
+)
+
+func InspectDiscovery(ctx context.Context, destDir string, discoveryClient discovery.CachedDiscoveryInterface) error {
+	if err := writeDiscovery(ctx, filepath.Join(destDir, "discovery-apis.yaml"), discoveryClient, "/apis"); err != nil {
```
p0lyn0mial, line 15: should we name the file as `aggregated-discovery-apis.yaml` to highlight it contains the aggregated discovery data?

pkg/cli/admin/inspect/discovery.go
deads2k: > Why we are not checking only `statusCode != http.StatusOK`. Since we are not sending any etags (may result in 304), the only expected status should be 200?

I'm not completely sure how the server responds and figured anything in the "ok" range was probably ok

pkg/cli/admin/inspect/discovery.go
deads2k: > I think we can also ensure that the content type is discovery/v2 as in

good point.  will update

pkg/cli/admin/inspect/inspect.go
deads2k: > In my opinion, we can continue inspection after accumulating this error to allErrs. Otherwise, any failure would stop the entire inspection

Agreed.  I thought of this in the shower this morning.

pkg/cli/admin/inspect/discovery.go
```diff
@@ -0,0 +1,56 @@
...
+		return err
+	}
+	if err := writeDiscovery(ctx, filepath.Join(destDir, "discovery-api.yaml"), discoveryClient, "/api"); err != nil {
```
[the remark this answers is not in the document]
deads2k: > just `return writeDiscovery(ctx, filepath.Join(destDir, "discovery-api.yaml"), discoveryClient, "/api")`

I personally dislike that structure.  @ardaguclu your preference?

pkg/cli/admin/inspect/discovery.go
deads2k: > should we name the file as `aggregated-discovery-apis.yaml` to highlight it contains the aggregated discovery data?

Up to @ardaguclu

ardaguclu: There is no other discovery anymore but `aggregated-discovery-apis.yaml` seems to be better reflect the feature that apis are retrieved from.

ardaguclu:
/lgtm
/hold
in case you want to see the CI's reaction

ardaguclu:
Discovery files were created successfully in the failed jobs;
/retest

PR #1904, opened 2024-10-29
[wip] provide way to determine transitive closure of permissions on node

deads2k:
excepting the kubelet permissions itself.  Might be useful to include.

/hold


deads2k:
use rbac-police

PR #1906, opened 2024-11-06
OCPBUGS-8259: pkg/cli/admin/upgrade: Propagate --force for --to-multi-arch

wking:
Folks might want to break through cluster-version-operator-side guards to test `--to-multi-arch` updates as well.

Also error on `--to-multi-arch` with `--to-latest`.   `--to-multi-arch` is about transitioning from single-arch to multi-arch without changing the version, so all of the "pick a new target version/release" options are incompatible.

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -240,8 +240,17 @@ func (o *Options) Run() error {
 			fmt.Fprintf(o.ErrOut, "warning: --allow-upgrade-with-warnings is bypassing: %s\n", err)
 		}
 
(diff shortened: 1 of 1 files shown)


evakhoni:
nice to see this one addressed! btw as for to-multi with to-latest, we may allow that as well :thinking: updating for example from "ec.0 amd64" to "ec.1 multi"

wking:
> ...updating for example from "ec.0 amd64" to "ec.1 multi"

I expect those updates will generally work, but the need for them is low, with `--to-multi-arch` being a one-time thing for existing clusters, so not something that will get lots of use.  To safe effort for more commonly-used things, we're only testing and supporting single-arch-4.y.z -> multi-arch-4.y.z that isolates the multi-arch transition from version bumps.

PR #1907, opened 2024-11-06
OTA-902: pkg/cli/admin/upgrade/recommend: Treat Upgradeable=False as a conditional update risk

wking:
We've [had `Upgradeable` since 2019][1], but it is a confusing condition, because [the "between minor versions" wording][2] in the `message` isn't obvious to users who have not yet internalized [SemVer's `MAJOR.MINOR.PATCH` terminology][3].  [Conditional update risks landed in 2021][4] and give us a clear API for declaring exactly which update targets have the exposure.  This commit adds client-side code to the tech-preview `recommend` subcommand, so folks can get a feel for that user experience.  If it goes over well, we can shift the logic to the cluster-version operator so all clients can benefit.  The `alreadyHasUpgradeableRisk` check ensures we don't double up if the CVO eventually picks up this pivot.

[1]: https://github.com/openshift/api/pull/206
[2]: https://github.com/openshift/cluster-version-operator/blob/c4b8362d8acd08d63a600b4d53c33e8737ed7a53/pkg/cvo/upgradeable.go#L218-L228
[3]: https://semver.org/spec/v2.0.0.html#summary
[4]: https://github.com/openshift/api/pull/1011

### DIFF
--- a/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.output
+++ b/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.output
@@ -1,8 +1,3 @@
-Upgradeable=False
-
-  Reason: AdminAckRequired
-  Message: Kubernetes 1.26 and therefore OpenShift 4.13 remove several APIs which require admin consideration. Please see the knowledge article https://access.redhat.com/articles/6958394 for details and instructions.
-
 Upstream: https://api.integration.openshift.com/api/upgrades_info/graph
 Channel: stable-4.13 (available channels: candidate-4.12, candidate-4.13, eus-4.12, eus-4.14, fast-4.12, fast-4.13, stable-4.12, stable-4.13)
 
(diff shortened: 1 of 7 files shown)


pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.output
```diff
@@ -1,23 +1,22 @@
...
 Updates to 4.13:
 
   Version: 4.13.50
   Image: quay.io/openshift-release-dev/ocp-release@sha256:6afb11e1cac46fd26476ca134072937115256b9c6360f7a1cd1812992c065f02
-  Reason: EvaluationFailed
-  Message: Exposure to ARODNSWrongBootSequence is unknown due to an evaluation failure: client-side throttling: only 8m30.897651224s has elapsed since the last match call completed for this cluster condition backend; this cached cluster condition request has been queued for later execution
+  Reason: MultipleReasons
+  Message: Kubernetes 1.26 and therefore OpenShift 4.13 remove several APIs which require admin consideration. Please see the knowledge article https://access.redhat.com/articles/6958394 for details and instructions.
+  
+  Exposure to ARODNSWrongBootSequence is unknown due to an evaluation failure: client-side throttling: only 8m30.897651224s has elapsed since the last match call completed for this cluster condition backend; this cached cluster condition request has been queued for later execution
```
petr-muller, line 11: Probably not something to be addressed in this PR, just a thought. If we're going to rely on conditional risks more we'll definitely need to improve the UX for `MultipleReasons`, the individual risks get really lost in the `Message` concatenation. How many risks are there? What names do they have?

pkg/cli/admin/upgrade/recommend/recommend.go
```diff
@@ -360,3 +362,63 @@ func findClusterOperatorStatusCondition(conditions []configv1.ClusterOperatorSta
...
+							target.Conditions[i].Reason = condition.Reason
+							target.Conditions[i].Message = condition.Message
+						} else {
+							target.Conditions[i].Reason = "MultipleReasons"
+							target.Conditions[i].Message = fmt.Sprintf("%s\n\n%s", condition.Message, c.Message)
+						}
+						target.Conditions[i].Status = metav1.ConditionFalse
+						break
+					}
+				}
```
petr-muller, line 418: Four levels of `for` depth makes this a bit hard to track at first, could be helpful to extract the inner block to something like:

```
if hasUpgradeableRisk(risks, message) {
	continue
}

addUpgradeableRisk(target, condition)
```

petr-muller (approved):
/lgtm

Left two thoughts. One's a minor refactor proposal, feel free to followup or ignore. I'm a bit concerned about the `MultipleReasons` case UX. AdminAckNeeded will make any known issue a `MultipleReasons` which is imo quite hard to work to work with from the user PoV. This is not something to address in this PR but I think the GA `oc adm upgrade recommend` will need handle this case much better (and CVO will likely need to cooperate).

petr-muller:
/hold

Just want to give Trevor to chance to see the commends and unhold as needed.

pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.output
wking: one issue with trying to make `reason` a clear summary on its own is whether we want to distinguish "risk matches" vs. this^ kind of "risk matching unknown".  I'm not too worried, because the entry you comment on here has a `message` covering the distinction, and I want customers to wade through that whole `message`.  I'm not expecting the tabular output to be actionable without using `--version ...` or some other step to get more details on a 4.y.z that you're thinking about from the table output.  Folks might be tempted to go deep in history to pick out a release that isn't exposed to some risk that modern releases aren't exposed to, but that's going to leave the cluster exposed to all sorts of bugs since fixed in modern releases, and I dunno if we want to make it especially convenient.  That said, any update is going to fix bugs, so if you go from 4.y.ancient to 4.y.moderatelyOld, that's still better than being spooked by 4.y.new regressions and staying on 4.y.ancient.

pkg/cli/admin/upgrade/recommend/recommend.go
wking: refactored with 604071f5 -> 22cf5c6c, which also grows the ability to handle `target` that somehow have no `Recommended` condition at all.

/hold cancel

petr-muller (approved):
/retest

LGTM thanks!

Unit tests fail on unrelated code, not sure if flake or if HEAD is broken:

```
INFO[2024-11-08T21:15:09Z] vendor/github.com/openshift/build-machinery-go/make/targets/golang/test-unit.mk:12: warning: undefined variable 'GO_TEST_ARGS'
set -o pipefail; go test  -race -json ./...  | gotest2junit > /logs/artifacts/junit_report.xml
FAIL: github.com/openshift/oc/pkg/helpers/newapp/newapptest TestNewAppRunBuilds/successful_build_from_dockerfile 230ms
FAIL: github.com/openshift/oc/pkg/helpers/newapp/newapptest TestNewAppRunBuilds 2.43s
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/golang/test-unit.mk:12: test-unit] Error 1
```

wking:
Ah, [unit failures][1] are getting worked in openshift/oc#1910.

```
=== RUN   TestNewAppRunBuilds/successful_build_from_dockerfile
    newapp_test.go:1259: successful build from dockerfile: unexpected error: unable to locate any images in image streams, images in the container image registry with name "openshift/origin:v1.0.6"
--- FAIL: TestNewAppRunBuilds/successful_build_from_dockerfile (0.31s)
```

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/1907/pull-ci-openshift-oc-master-unit/1856674293862109184

wking:
#1910 just merged.

/test unit

PR #1917, opened 2024-11-19
NO-JIRA: adm release: simplify manifest iteration

petr-muller:
Full `imageData` was passed to the callback but existing callbacks only needed the containing directory. Therefore the callback signature can be simplified to `func(directory, content, operator)`

/cc @wking @DavidHurta


DavidHurta (approved):

petr-muller:
b06 issues :(

petr-muller:
I think we are hitting [TRT-1900](https://issues.redhat.com/browse/TRT-1900) now

PR #1925, opened 2024-11-26
WORK IN PROGRESS: Add a new command on oc adm to generate the communication matrix

aabughosh:


ardaguclu:
@aabughosh thanks for the PR. Our policy is to deliver these kind of commands in separate plugins rather than adding them into oc as builtin commands. You can have a look at other plugins in https://krew.sigs.k8s.io/plugins/.

PR #1931, opened 2024-12-03
OCPBUGS-45311: fix oc adm node-image create --pxe command to copy the correct artifacts

andfasano:
This patch takes care of copying only the required artifacts from the node-joiner pod  when the --pxe command flag is used (instead the copy the whole folder)


rwsu (commented):
/lgtm

PR #1936, opened 2024-12-04
NO-JIRA: adm release: clarify manifest processing

petr-muller:
Rename variables to use `src` and `dst` more clearly

/cc @DavidHurta @wking @hongkailiu


wking (approved):
/lgtm

PR #1944, opened 2024-12-05
NO-JIRA: adm release: tighten regex to save code

petr-muller:
Part of `NewComponentVersionsMapper` code dealt with imprecise matches with byte slice operations, we can avoid that by tightening the regex used to find matches.


petr-muller:
Does not seem like a significant simplification, just code churn

PR #1947, opened 2024-12-09
NO-JIRA: make 'oc status' compatible with bare kube

deads2k:
Fixes an annoyance that I don't think has a purpose for remaining.


pkg/helpers/describe/projectstatus.go
```diff
@@ -265,7 +265,7 @@ func (d *ProjectStatusDescriber) Describe(namespace, name string) (string, error
 		if allNamespaces {
 			fmt.Fprintf(out, describeAllProjectsOnServer(f, d.Server))
 		} else {
-			fmt.Fprintf(out, describeProjectAndServer(f, project, d.Server))
```
ardaguclu, line 268: `describeProjectAndServer` is not used by anyone. Can we also remove it as well?.

ardaguclu (commented):
Overall looks good to me and https://github.com/openshift/origin/blob/master/test/extended/cli/status.go likely catches any regression. Just dropped a comment.

pkg/helpers/describe/projectstatus.go
deads2k: > `describeProjectAndServer` is not used by anyone. Can we also remove it as well?.

Tidied up along with the project client.

pkg/helpers/describe/projectstatus_test.go
```diff
@@ -8,36 +8,33 @@ import (
...
 	fakebuildclient "github.com/openshift/client-go/build/clientset/versioned/fake"
 	fakebuildv1client "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1/fake"
 	fakeimageclient "github.com/openshift/client-go/image/clientset/versioned/fake"
 	fakeimagev1client "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1/fake"
-	fakeprojectclient "github.com/openshift/client-go/project/clientset/versioned/fake"
-	fakeprojectv1client "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1/fake"
 	fakerouteclient "github.com/openshift/client-go/route/clientset/versioned/fake"
 	fakeroutev1client "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1/fake"
 	osgraph "github.com/openshift/oc/pkg/helpers/graph/genericgraph"
+	corev1 "k8s.io/api/core/v1"
```
ingvagabund, line 26: Nit: there's still a visible value in keeping k8s.io and github.com package imports separated

ardaguclu:
It is nice to see that we are dropping a dependency (i.e. `vendor/github.com/openshift/client-go/project/`) and verify-deps is failing due to this. I think, we need go mod tidy/vendor.

ardaguclu:
Assuming that CI will be green
/lgtm

ardaguclu:
/retitle NO-JIRA: make 'oc status' compatible with bare kube
/retest-required

PR #1949, opened 2024-12-10
OCPBUGS-38975: monitor should not exit for temp api disconnect

rwsu:
This commit updates the isMonitoringDone function to retry and not exit if the error from reading the node-joiner pod is

Get "https://api.ostest.test.metalkube.org:6443/api/v1/namespaces/openshift-node-joiner-zfbp9/pods/node-joiner-monitor-4lx8l": EOF

This happens if the oc client is temporarily disconnected from the api server. The error returned is type net/url/Error.

### DIFF
--- a/pkg/cli/admin/nodeimage/monitor.go
+++ b/pkg/cli/admin/nodeimage/monitor.go
@@ -2,8 +2,10 @@ package nodeimage
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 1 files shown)


rwsu:
This change only affects the "oc adm node-image" command. It fixes an issue that we see quite often in CI and testing, so we'd like to have it merged.

PR #1952, opened 2024-12-12
feat: allow the service account to be in a different namespace than t…

machine424:
…he rolebinding for 'policy {add,remove}-[cluster]-role-to-user'

- [ ] add tests

### DIFF
--- a/pkg/cli/admin/policy/modify_roles.go
+++ b/pkg/cli/admin/policy/modify_roles.go
@@ -109,6 +109,7 @@ type RoleModificationOptions struct {
 	RoleBindingNamespace string
 	RbacClient           rbacv1client.RbacV1Interface
 	SANames              []string
+	SANamespace         string
 
 	UserClient           userv1client.UserV1Interface
 	ServiceAccountClient corev1client.ServiceAccountsGetter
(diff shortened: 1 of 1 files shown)


pkg/cli/admin/policy/modify_roles.go
```diff
@@ -178,6 +179,7 @@ func NewCmdAddRoleToUser(f kcmdutil.Factory, streams genericiooptions.IOStreams,
 	cmd.Flags().StringVar(&o.RoleBindingName, "rolebinding-name", o.RoleBindingName, "Name of the rolebinding to modify or create. If left empty creates a new rolebinding with a default name")
 	cmd.Flags().StringVar(&o.RoleNamespace, "role-namespace", o.RoleNamespace, "namespace where the role is located: empty means a role defined in cluster policy")
 	cmd.Flags().StringSliceVarP(&o.SANames, "serviceaccount", "z", o.SANames, "service account in the current namespace to use as a user")
+	cmd.Flags().StringVar(&o.SANamespace, "serviceaccount-namespace", o.SANamespace, "namespace where the service account is located: empty means the default namespace")
```
ardaguclu, line 182: From user point of view, it is already very hard for users to use these commands (e.g. `adm policy *`) as the current flag set (and their interactions between each other) is already hard to understand. I'd not prefer adding a new flag that makes the case worse.
$ oc project test-proj
$ oc policy add-role-to-user view user1
can still be used to modify the default namespace as a workaround.

machine424:
Forgot about `oc create rolebinding` that seems to allow having the role/binding and the SA in different namespaces, I'll give that a shot.

PR #1955, opened 2024-12-13
OCPBUGS-46361: Mark adm pod-network command as hidden and depcated

ardaguclu:
`oc adm pod-network` command only works on OpenShift SDN which has been deprecated long time ago. So still advertising this command in help output is misleading. 

This PR marks this command as hidden, so `oc adm --help` will not print it. Besides, this PR marks this command as deprecated, so there will be a deprecation message during its usage.

```sh
$ ./oc adm pod-network
Command "pod-network" is deprecated, pod-network command only works on OpenShift SDN which has been deprecated.
Manage pod network in the cluster
```

Other than that, users still rely on the functionality of this command will continue using it as is (just an additional deprecation message that should not cause any issues).


ardaguclu:
@danwinship does this deprecation message make sense?.

danwinship:
sure, lgtm

PR #1959, opened 2025-01-14
NO-ISSUE: admin/upgrade/status/examples/README: Document *-alerts.json

wking:
Catching the docs up with 10c41f5262 (#1740).

Also drop the enumeration that started when the file was created in 57e984491c (#1595).  Folks who care how many inputs and outputs there currently are can count for themselves, and this content is easier to maintain without having to bump later entry numbers when inserting an earlier entry.


petr-muller (approved):
Thanks!

petr-muller:

This is a docs-only change

wking:
Doc-only change:

/retest-required

PR #1965, opened 2025-01-23
OCPBUGS-44637: Mount proxy certificate to node-joiner pod

rwsu:
When a proxy is configured with a self-signed certificate, the certificate needs to be made available to the node-joiner pod to allow it to communicate with the proxy.

It is assumed that the proxy certificate is configured in the cluster proxy spec as
    
    spec:
      trustedCA:
        name: user-ca-bundle

The certificate is stored in a config map named "user-ca-bundle" in the "openshift-config" namespace in a file named ca-bundle.crt.
    
    apiVersion: v1
    data:
      ca-bundle.crt: |
        -----BEGIN CERTIFICATE-----
        ** redacted **
        -----END CERTIFICATE-----
    kind: ConfigMap
    metadata:
      name: user-ca-bundle
      namespace: openshift-config
    
If the certificate is included in the additionalTrustBundle field in install-config.yaml prior to cluster installation, the proxy and user-ca-bundle is automatically configured as illustrated
above.

Previously, the certificate wasn't mounted to the node-joiner pod so that when the pod attempted to pull images through the proxy, it failed.

Now the user-ca-bundle config map is copied to the node-joiner pod's namespace and the certificate is mounted and made available as /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem.

### DIFF
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -798,5 +798,76 @@ func (o *CreateOptions) configurePodProxySetting(ctx context.Context, pod *corev
 	for i := range pod.Spec.Containers {
 		pod.Spec.Containers[i].Env = append(pod.Spec.Containers[i].Env, proxyVars...)
 	}
+
+	// An external proxy may be configured with a self-signed certificate.
+	// In such cases, the cluster's proxy would be configured to contain
+	// the certificate by including it in the additionalTrustBundle in the
+	// install-config.
+	//
+	// If the certificate is configured, it is saved
+	// to the "user-ca-bundle" config map in the "openshift-config"
+	// namespace.
+	//
+	// The certificate must be made available to the node-joiner pod
+	// and in its namespace so that the certificate can be read
+	// when the proxy is accessed by node-joiner to pull container
+	// images.
+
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/nodeimage/create.go
```diff
@@ -798,5 +798,76 @@ func (o *CreateOptions) configurePodProxySetting(ctx context.Context, pod *corev
...
+				},
+			},
+		},
+	})
+
+	pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts, corev1.VolumeMount{
+		Name:      "user-ca-bundle",
+		MountPath: "/etc/pki/ca-trust/extracted/pem",
+	})
+
```
andfasano, line 871: Given that it seems the reasonable motivation for the issue, there is one point not clear to me: why the oc command must be responsible for extracting a certificate, and injecting it into the node-joiner container?
Couldn't be something directly managed by the node-joiner tool itself?
Personally I'd prefer to avoid having additional logic in the oc wrapper layer - if not the one just required the proper execution of the container (for the rest, the node-joiner must be able to work also on its own)

rwsu:
Replaced by https://github.com/openshift/installer/pull/10532

PR #1966, opened 2025-01-28
COS-3023: pkg/cli/admin/release/info: support generating RPM diffs

jlebon:
It's often useful when looking up release images to know the list of RPM packages that shipped in the node image. Add new switches for this:
- `oc adm release info --rpmdb $IMG` will list all the packages in the node image for the given release image payload
- `oc adm release info --rpmdb-diff $IMG1 $IMG2` will diff the set of packages in the node image for the given release image payloads

The code is generic over the actual target image. By default, the node image is used, but `--rpmdb-image` can be used to select a different one.

The primary motivation for this is
https://github.com/openshift/enhancements/pull/1637, in which the node image will no longer be built within the CoreOS pipeline as a base image. Instead, it will be a layered image built in OpenShift CI/Konflux. As a result, all layered packages will not show up in the CoreOS release browser differ.

With this functionality, the release controller will be able to render RPM diffs in the web UI, greatly de-emphasize the CoreOS differ and effectively dropping the requirement for having VPN access.

Some notes on the implementation:
- The rpmdb for a given image is cached, keyed by the image digest.
- We don't try to be smart here and e.g. only download some layers. There are some issues with doing that. We literally do download the full image, _but_ we only cache the rpmdb content and throw away the rest. That said, the high cost isn't an issue in practice because the release controller can nicely represent operations which take time so it didn't feel worth the effort of trying to optimize this further.

Once we have SBOMs available for all our images, this should be a much cheaper way to query its RPM contents. Additionally/alternatively, for the node image specifically, if we ever end up with lockfiles in the git repo, this would effectively mean that the git changelog _is_ the RPM changelog also, meshing nicely with the existing infrastructure around that.

### DIFF
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -716,6 +716,7 @@ type ReleaseInfo struct {
 	// ComponentVersions.
 	DeprecatedComponentVersions map[string]string `json:"versions"`
 	ComponentVersions           ComponentVersions `json:"displayVersions"`
+	ComponentTags               map[string]string `json:"versionsTags"`
 
 	Images map[string]*Image `json:"images"`
 
@@ -862,7 +863,7 @@ func (o *InfoOptions) LoadReleaseInfo(image string, retrieveImages bool) (*Relea
 		return nil, fmt.Errorf("release image did not contain an image-references file")
 	}
 
-	release.ComponentVersions, errs = readComponentVersions(release.References, o.ErrOut)
+	release.ComponentVersions, release.ComponentTags, errs = readComponentVersions(release.References, o.ErrOut)
 	for _, err := range errs {
 		release.Warnings = append(release.Warnings, err.Error())
 	}
@@ -938,6 +939,7 @@ func readComponentVersions(is *imageapi.ImageStream, errOut io.Writer) (Componen
 	var errs []error
 	combined := make(map[string]sets.String)
 	combinedDisplayNames := make(map[string]sets.String)
+	componentTags := make(map[string]string)
 	kubectlVersions := make(map[string]sets.Set[string])
 	for _, tag := range is.Spec.Tags {
 		versions, ok := tag.Annotations[annotationBuildVersions]
@@ -949,6 +951,8 @@ func readComponentVersions(is *imageapi.ImageStream, errOut io.Writer) (Componen
 			errs = append(errs, fmt.Errorf("the referenced image %s had an invalid version annotation: %v", tag.Name, err))
 		}
 		for k, v := range all {
+			// just let the last one win in case a component is defined multiple times
+			componentTags[k] = tag.Name
 			if k == "kubectl" {
 				if val, exist := kubectlVersions[v.Version]; exist {
 					kubectlVersions[v.Version] = val.Insert(tag.Name)
@@ -1025,7 +1029,7 @@ func readComponentVersions(is *imageapi.ImageStream, errOut io.Writer) (Componen
 	if len(multiples) > 0 {
 		errs = append(errs, fmt.Errorf("multiple versions or display names reported for the following component(s): %v", strings.Join(multiples.List(), ",  ")))
 	}
-	return out, errs
+	return out, componentTags, errs
 }
 
 func errorList(errs []error) string {
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -2,15 +2,19 @@ package release
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"path"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -97,6 +101,13 @@ func NewInfo(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Comm
 			the code changes that occurred between the two release arguments. This operation is slow
 			and requires sufficient disk space on the selected drive to clone all repositories.
 
+			Additionally, the --rpmdb-cache flag will cache rpmdb content for images in the release. Then,
+			the --rpmdb flag will display the RPM contents of an image. The --rpmdb-diff flag will display
+			RPM changes that occurred between the two release arguments. For images optimized for this
+			operation (e.g. recent enough machine-os images), this is quite fast and efficient. Otherwise,
+			it is a slow operation that requires sufficient disk space. By default, the image containing the
+			machine-os component is targeted for RPM queries. Other images can be targeted with --rpmdb-image.
+
 			If the specified image supports multiple operating systems, the image that matches the
 			current operating system will be chosen. Otherwise you must pass --filter-by-os to
 			select the desired image.
@@ -147,6 +158,10 @@ func NewInfo(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Comm
 	flags.StringVar(&o.ImageFor, "image-for", o.ImageFor, "Print the pull spec of the specified image or an error if it does not exist.")
 	flags.StringVarP(&o.Output, "output", "o", o.Output, "Display the release info in an alternative format: digest|json|name|pullspec|template|jsonpath.")
 	flags.StringVar(&o.ChangelogDir, "changelog", o.ChangelogDir, "Generate changelog output from the git directories extracted to this path.")
+	flags.StringVar(&o.RpmdbCacheDir, "rpmdb-cache", o.RpmdbCacheDir, "Cache rpmdb content in this directory.")
+	flags.BoolVar(&o.RpmdbList, "rpmdb", o.RpmdbList, "List RPM packages in image.")
+	flags.BoolVar(&o.RpmdbDiff, "rpmdb-diff", o.RpmdbDiff, "Generate RPM package diff.")
+	flags.StringVar(&o.RpmdbImage, "rpmdb-image", "", "The image to use for RPM queries.")
 	flags.StringVar(&o.BugsDir, "bugs", o.BugsDir, "Generate bug listings from the changelogs in the git repositories extracted to this path.")
 	flags.BoolVar(&o.IncludeImages, "include-images", o.IncludeImages, "When displaying JSON output of a release output the images the release references.")
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be copied under.")
@@ -174,9 +189,13 @@ type InfoOptions struct {
 	ICSPFile      string
 	IDMSFile      string
 
-	ChangelogDir string
-	BugsDir      string
-	SkipBugCheck bool
+	ChangelogDir  string
+	RpmdbCacheDir string
+	RpmdbList     bool
+	RpmdbDiff     bool
+	RpmdbImage    string
+	BugsDir       string
+	SkipBugCheck  bool
 
 	ParallelOptions imagemanifest.ParallelOptions
 	SecurityOptions imagemanifest.SecurityOptions
@@ -222,6 +241,8 @@ type versionGraph struct {
 }
 
 const defaultGraphURL = "https://api.openshift.com/api/upgrades_info/v1/graph"
+
+const metalayerMetadata = "usr/share/openshift/base/meta.json"
 
 // replaceStableSemanticArgs attempts to look up known major versions in existing public stable
 // channels.
@@ -424,13 +445,22 @@ func (o *InfoOptions) Validate() error {
 	if o.SkipBugCheck && o.Output != "name" && o.Output != "json" {
 		return fmt.Errorf("--skip-bug-check requires --output to be set to 'name' or 'json'")
 	}
-	if len(o.ChangelogDir) > 0 || len(o.BugsDir) > 0 {
+	if len(o.ChangelogDir) > 0 || len(o.BugsDir) > 0 || o.RpmdbDiff {
 		if len(o.From) == 0 {
-			return fmt.Errorf("--changelog/--bugs require --changes-from")
+			return fmt.Errorf("--changelog/--bugs/--rpmdb-diff require --changes-from")
 		}
 	}
-	if len(o.ChangelogDir) > 0 && len(o.BugsDir) > 0 {
-		return fmt.Errorf("--changelog and --bugs may not both be specified")
+	if o.RpmdbList && o.RpmdbDiff {
+		return fmt.Errorf("--rpmdb/--rpmdb-diff are mutually exclusive")
+	}
+	if o.RpmdbList || o.RpmdbDiff {
+		if o.RpmdbCacheDir == "" {
+			return fmt.Errorf("--rpmdb/--rpmdb-diff require --rpmdb-cache")
+		}
+	}
+	exclusiveOps := boolToInt(len(o.ChangelogDir) > 0) + boolToInt(len(o.BugsDir) > 0) + boolToInt(o.RpmdbDiff)
+	if exclusiveOps > 1 {
+		return fmt.Errorf("--changelog/--bugs/--rpmdb-diff are mutually exclusive")
 	}
 	switch {
 	case len(o.BugsDir) > 0:
@@ -444,6 +474,12 @@ func (o *InfoOptions) Validate() error {
 		case "", "json":
 		default:
 			return fmt.Errorf("--output only supports 'json' for --changelog")
+		}
+	case o.RpmdbList || o.RpmdbDiff:
+		switch o.Output {
+		case "", "json":
+		default:
+			return fmt.Errorf("--output only supports 'json' for --rpmdb/--rpmdb-diff")
 		}
 	default:
 		output := strings.SplitN(o.Output, "=", 2)[0]
@@ -464,6 +500,14 @@ func (o *InfoOptions) Validate() error {
 	}
 
 	return o.FilterOptions.Validate()
+}
+
+func boolToInt(b bool) int {
+	if b {
+		return 1
+	} else {
+		return 0
+	}
 }
 
 func (o *InfoOptions) Run() error {
@@ -502,6 +546,9 @@ func (o *InfoOptions) Run() error {
 		if len(o.ChangelogDir) > 0 {
 			return describeChangelog(o.Out, o.ErrOut, release, diff, o.ChangelogDir, o.Output)
 		}
+		if o.RpmdbDiff {
+			return o.describeRpmDiff(release, diff, o.RpmdbCacheDir, o.Output, o.RpmdbImage)
+		}
 		return describeReleaseDiff(o.Out, diff, o.ShowCommit, o.Output)
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,298 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+	}
+
+	// not sure this can ever happen, but just sanity-check anyway
+	if ref.ID == "" {
+		return fmt.Errorf("image not digest-based: %s", ref)
+	}
+
+	dbpath := filepath.Join(dir, ref.ID)
+	if _, err = os.Stat(dbpath); os.IsNotExist(err) {
+		err = o.downloadRpmdb(ref.Name, dbpath)
```
Prashanth684, line 2072: shouldn't this just use `targetTag.From.Name` instead? ref.Name only points to the image name without the repo or the tag. i get an `unable to connect to image repository : invalid reference format` when ref.Name is used

jlebon: Fixed, thanks! I hadn't retested a change I did in that area correctly because I was relying on cache.

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
 	return nil
 }
 
+func (o *InfoOptions) listRpmdb(releaseInfo *ReleaseInfo, dir, format, targetImage string) error {
+	if targetImage == "" {
+		// find image for machine-os component
+		if tag, ok := releaseInfo.ComponentTags["machine-os"]; !ok {
```
sdodson, line 2041: Is this the image both before and after openshift/enhancements#1637 or when would `rhel-coreos` be used?

sdodson:
/uncc @deads2k @atiratree @Prashanth684 
/cc @petr-muller @wking 
Since it seems like most of this is covered by update-approvers

pkg/cli/admin/release/info.go
jlebon: Yeah, nothing changes in that respect with https://github.com/openshift/enhancements/pull/1637; the node image (`rhel-coreos`/`stream-coreos`) still declares the `machine-os` component and version.

This code works identically before and after https://github.com/openshift/enhancements/pull/1637.

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+	if err != nil {
+		return err
+	}
+
+	// not sure this can ever happen, but just sanity-check anyway
+	if ref.ID == "" {
+		return fmt.Errorf("image not digest-based: %s", ref)
+	}
+
+	dbpath := filepath.Join(dir, ref.ID)
```
wking, line 2070: nit: `ID` is a digest, and [all digests will contain `:`][1].  Non-POSIX systems (OS X?  Windows?) might have trouble with that, and things like [the OCI image-layout format][1] suggest splitting the digest, using the algorithm as a directory, and the encoded portion as the child filename.  But will anyone be running this with caching outside of POSIX?  Maybe we don't care enough to be worth coding up that kind of hoop-jumping?

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0/descriptor.md#digests
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.0/image-layout.md#blobs

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+			fmt.Fprintf(o.Out, "  %s-%s\n", pkg, db[pkg])
+		}
+	}
+
+	return nil
+}
+
+func (o *InfoOptions) describeRpmDiff(releaseInfo *ReleaseInfo, diff *ReleaseDiff, dir, format, targetImage string) error {
+	if diff.To.Digest == diff.From.Digest {
+		return fmt.Errorf("releases are identical")
```
wking, line 2106: Different functions in this file currently have different opinions on whether "you asked for `diff(A, A)`" is a problem or not:

```console
$ git blame pkg/cli/admin/release/info.go | grep -hni1 'func \|releases are identical' | grep -iB4 -A1 'releases are identical'
1052:a3fbf02c3f pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-04-30 10:41:42 -0400 1052) func describeReleaseDiff(out io.Writer, diff *ReleaseDiff, showCommit bool, outputMode string) error {
1053-a3fbf02c3f pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-04-30 10:41:42 -0400 1053)      switch outputMode {
--
1066-77c04436ce pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2018-11-05 23:32:23 -0500 1066)      if diff.To.Digest == diff.From.Digest {
1067:77c04436ce pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2018-11-05 23:32:23 -0500 1067)              fmt.Fprintf(out, "Releases are identical\n")
1068-77c04436ce pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2018-11-05 23:32:23 -0500 1068)              return nil
--
--
1648-f052a0ffb7 pkg/cli/admin/release/info.go                                     (Clayton Coleman 2020-03-28 14:08:02 -0400 1648) 
1649:e465790921 pkg/cli/admin/release/info.go                                     (David Eads      2024-04-05 15:21:48 -0400 1649) func describeChangelog(out, errOut io.Writer, releaseInfo *ReleaseInfo, diff *ReleaseDiff, dir, format string) error {
1650-6e37bfd7d8 pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-02-04 13:50:19 -0500 1650)      if diff.To.Digest == diff.From.Digest {
1651:6e37bfd7d8 pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-02-04 13:50:19 -0500 1651)              return fmt.Errorf("releases are identical")
1652-6e37bfd7d8 pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-02-04 13:50:19 -0500 1652)      }
--
1881-6e37bfd7d8 pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-02-04 13:50:19 -0500 1881) 
1882:41caf81d0b pkg/cli/admin/release/info.go                                     (Alex Pavel      2020-05-28 14:54:58 -0700 1882) func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format string, skipBugCheck bool) error {
1883-6e37bfd7d8 pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-02-04 13:50:19 -0500 1883)      if diff.To.Digest == diff.From.Digest {
1884:6e37bfd7d8 pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-02-04 13:50:19 -0500 1884)              return fmt.Errorf("releases are identical")
1885-6e37bfd7d8 pkg/oc/cli/admin/release/info.go                                  (Clayton Coleman 2019-02-04 13:50:19 -0500 1885)      }
```

Personally, non-fatal seems reasonable to me (we _can_ evaluate `diff(A, A)`, even if the results aren't very interesting), and I don't see anything in any of those commit messages trying to motivate either approach.

No worries if this seems like something you don't want to get into, you have precedent for taking this in either direction 😅

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+			targetImage = tag
+		}
+	}
+
+	osDiff, ok := diff.ChangedImages[targetImage]
+	if !ok {
+		// image didn't change; we're done!
+		if format == "json" {
+			fmt.Println("{}")
+		}
```
wking, line 2123:
```golang
} else {
  fmt.Printf("Releases have the same %s image.\n", targetImage)
}
```

or something to handle the non-JSON text case?

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+	if toRef, err = imagereference.Parse(osDiff.To.From.Name); err != nil {
+		return err
+	}
+
+	// not sure this can ever happen, but just sanity-check anyway
+	if fromRef.ID == "" {
+		return fmt.Errorf("image not digest-based: %s", fromRef)
+	}
+	if toRef.ID == "" {
+		return fmt.Errorf("image not digest-based: %s", toRef)
```
wking, line 2141: nit: to make debugging slightly easier, can we include some distinguishing string in these context wrappers?  `initial machine-os image not digest-based` and `new machine-os image not digest-based` or some such?  There's an earlier `image not digest-based` error around line 2067 that could also be uniquified somehow.  I'm just trying to set things up so that if the command exits with these errors, the way the offending ref was extracted from the input arguments is somewhat clear, and which of these several image-lookup paths lead to the surprising data.

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+			}
+			delete(toDb, pkg)
+		} else {
+			removed[pkg] = v
+		}
+	}
+	rpmDiff := RpmDiff{
+		Changed: changed,
+		Added:   toDb,
+		Removed: removed,
```
wking, line 2188: nit: you might be able to save yourself a bit of space by shifting this up, and avoiding the unattached local variables.  Something like:

```golang
rpmDiff := RpmDiff{
	Changed: make(map[string]RpmChangedDiff),
	Added:   toDb,
	Removed: make(map[string]string),
}
... rpmDiff.Changed[pkg] = RpmChangedDiff{...}
```

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+	Old string `json:"old,omitempty"`
+	New string `json:"new,omitempty"`
+}
+
+func (o *InfoOptions) downloadRpmdb(image imagereference.DockerImageReference, dir string) error {
+	// NOTE: in the future, should check if there's an SBOM and get the rpmlist from that instead
+
+	ref := imagesource.TypedImageReference{Type: imagesource.DestinationRegistry, Ref: image}
+
+	tmpDir := dir + ".work"
```
wking, line 2248: Can you use [`os.MkdirTemp`][1] here?  Or are you worried about leaking temporary directories if the process is terminated mid-download?  There's some tension here between "the release-controller use-case is one long-running process, and we'd rather not leak directories there" and "`oc adm release info ...` can be used in other places, with multiple invocations in parallel, and if we hard-code this kind of scratch directory, we could have two parallel runs corrupting the dir they didn't realize they were sharing".  I guess also theoretically on the table would be lock files or some such so at least parallel invocations could notice overlap and exit out?  With some way to have the lock expire to recover from died-mid-download?  Lots of corner cases...

[1]: https://pkg.go.dev/os#MkdirTemp

pkg/cli/admin/release/info.go
jlebon: Yeah, IIRC I basically just did this to match `describeBugs()` here. No strong opinion one way or the other though I guess we should try to be consistent unless there's a good reason to differ.

pkg/cli/admin/release/info.go
jlebon: Yeah, IIRC I basically just did this to match `describeBugs()` here. No strong opinion one way or the other though I guess we should try to be consistent unless there's a good reason to differ.

pkg/cli/admin/release/info.go
jlebon: Good idea, I think that makes sense to do regardless.

pkg/cli/admin/release/info.go
jlebon: Yes, I _usually_ do this too, for the same reason. :)

pkg/cli/admin/release/info.go
jlebon: Hmm, yes good point. I need to check what the release controller does if multiple requests come in for the same thing. This must be a solved case at least for the git-based stuff which already exists. Overall, it feels like locking/"request joining" (for lack of a better term; i.e. having a secondary request join an earlier matching request already ongoing) should live at the release controller level (or similarly anything driving potentially multiple invocations of `oc`) rather than within `oc` using file locks?

wking: release-controller isn't the only `oc` caller though.  It's theoretically possible that a user like me with my own `oc` copy runs this command twice in the same directory at the same time

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+		} else {
+			continue
+		}
+	}
+
+	if rpmdbPath == "" {
+		return fmt.Errorf("no rpmdb found in %s", image)
+	}
+
+	if err := os.Rename(rpmdbPath, dir); err != nil {
```
wking, line 2300: are hardlinks the only indirection?  E.g. if `usr/share/rpm` was a symlink, this `Rename` wouldn't be sufficient, right?  Do we want to use [`os.CopyFS`][1] or something, to at least fail fast if we hit a symlink?

[1]: https://pkg.go.dev/os#CopyFS

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+		return err
+	}
+
+	if err := os.RemoveAll(tmpDir); err != nil {
```
wking, line 2304: you probably want this in a `defer` block just after creating the temporary directory to avoid leaking on the earlier `return` branches (e.g. `Lstat` failing on something other than non-existance).

wking (approved):
I left a bunch of thoughts inline, and this kind of thing is complicated enough that there's probably quite a bit of room for polish and hardening and all of that.  But having this functionality land at all, even if it has some rough edges, would provide a big release-controller UX win.  And it seems niche enough that I'm not expecting customers to be poking around with it and complaining about rough edges (but if they do, that's at least giving us trackers to work on polishing).  So I'm approving as you have it, and also holding in case you want to do any more polishing now.  Feel free to lift the hold when you're ready and ping me for a fresh LGTM if necessary.

/lgtm
/hold

pkg/cli/admin/release/info.go
```diff
@@ -1987,6 +2035,306 @@ func describeBugs(out, errOut io.Writer, diff *ReleaseDiff, dir string, format s
...
+	return nil
+}
+
+func loadRpmdb(dir string) (map[string]string, error) {
+	var out bytes.Buffer
+	// A pitfall here is that we're using the `rpm` from the release controller
+	// environment to query the rpmdb of the target image. they may not be
+	// compatible, but normally should be. A better approach would be to run the rpm
+	// query from within a container with the target rootfs.
+	cmd := exec.Command("rpm", "-qa", "--qf", "%{name}\\t%{evr}.%{arch}\\n", "--dbpath", dir)
```
wking, line 2317: The release controller is running `oc` commands like this one inside a container.  _It_ could launch a new container in the same cluster to collect this information.  But `oc adm release info ...` isn't necessarily running in a container system, and there might not be a runtime or whatever available to execute RPM gathering in a separate container than the `oc` invocation.  E.g. I can run `oc adm release info ...` on my laptop with no kubeconfig to connect me to a cluster, and no Podman/etc. to run containers locally, and still expect this command to work.  Hah, and actually for me:

```console
$ command -V rpm
-bash: command: rpm: not found
```

so I guess this is going to fail for me locally.  But the point about run-a-container access not being a guarantee still holds.

pkg/cli/admin/release/info.go
bradmwilliams: The release-controller runs its `oc adm release info` commands in memory now (not out on its own pod).  It will cache results based on the `from` and `to` versions being compared.

pkg/cli/admin/release/info.go
jlebon: Given that (as you found out), this currently requires `rpm` anyway, I think for now let's confine these commands to Linux. As mentioned in the commit message, there's potential with SBOM that this becomes much more lightweight and as a result also cross-platform.

pkg/cli/admin/release/info.go
jlebon: > release-controller isn't the only `oc` caller though. It's theoretically possible that a user like me with my own `oc` copy runs this command twice in the same directory at the same time

I'm not too worried about the user case TBH, though OTOH you're right that if locking needs to be added somewhere, it might as well be here to cover all possible callers. Added basic locking for now! We can make it fancier down the road if needed.

pkg/cli/admin/release/info.go
jlebon: We use `Lstat` above and only accept directories, so this shouldn't be a concern.

pkg/cli/admin/release/info.go
jlebon: It'd definitely be awkward to try to reach out to Podman/Docker even if it exists. By "container" here I meant more that we assemble a minimal container environment ourselves. Possibly there are helpers already for that we vendor in somewhere, but I didn't check. Or a simple chroot might even work. But the problem is that it assumes we have enough privs to set it up.

> ```
> -bash: command: rpm: not found
> ```

Hmm, yeah I'm just assuming here that rpm is in the environment. @bradmwilliams can you confirm whether where `oc` runs for the release-controller-api case, it has access to `rpm`?

jlebon: > Might be worth while to check, earlier, in the oc invocation that the machine, where the command is running, actually supports rpm.

Yeah, that's fair. Will add.

bradmwilliams: I can confirm that the container, where the release-controller's `oc` commands are executed, is RHEL 9 and does indeed have `rpm`.

jlebon: > > Might be worth while to check, earlier, in the oc invocation that the machine, where the command is running, actually supports rpm.
> 
> Yeah, that's fair. Will add.

Done!

jlebon:
OK updated this now!

The biggest change is that we now detect if the image implements [the new "metalayer semantic" now in use in layered node images](https://github.com/openshift/os/pull/1761) (though the mechanism is generic and could be implemented in other images). This makes retrieving the package list for a  node image _vastly_ cheaper _and_ in a cross-platform way (i.e. without an RPM dependency).

That said, we still also keep the rpmdb fallback to handle the case where a node image had more layers added and also to allow the release-controller to provide diffs between older releases and 4.19. (But also because I still believe in having this code work across all container images, and not just the node image.)

pkg/cli/admin/release/info.go
jlebon: As mentioned in https://github.com/openshift/oc/pull/1966#issuecomment-2730483954, this now does work on systems without rpm _if_ the images involved all have metalayers. Otherwise, we fallback to rpm querying again (checking first if rpm is installed).

pkg/cli/admin/release/info.go
jlebon: Tweaked this now to just substitute `:` for `_` to be compliant.

petr-muller (approved):
Let's move this forward, Trevor reviewed and was fine to merge this, Brad tested, if we subtly break something it's more likely to be discovered after merge than by further human scrutiny before that, and we can always revert or fix forward.

/lgtm

PR #1980, opened 2025-02-20
NO-ISSUE: pkg/cli/admin/upgrade/recommend: "Upstream" -> "Upstream update service"

wking:
The ClusterVersion spec property is `spec.upstream`, and that's a v1 API we cannot easily change.  But in freeform text here, we can get closer to [the "OpenShift Update Service" wording we use in docs][1].  I haven't included "OpenShift", because hopefully that context is obvious to folks running `oc ....`.  I haven't Title Cased, because I'm not listing a particular product/implementation; it's possible that the customer has pointed their cluster at a custom update service like [a raw GitHub file][3].  I've kept "Upstream" to make it easier for folks to associate with spec.upstream.  And I've left the GA `oc adm upgrade` alone on this, because it's been using "Upstream:" for years now, and this seems minor enough that I don't think it's worth disrupting things there.

[1]: https://docs.openshift.com/container-platform/4.17/updating/understanding_updates/intro-to-updates.html
[3]: https://issues.redhat.com/browse/OTA-520

### DIFF
--- a/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.output
+++ b/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.output
@@ -5,7 +5,7 @@ recommended/Alert=Unknown:
   Reason: NoTestData
   Message: This --mock-clusterversion run did not have alert data available at examples/4.12.16-longest-not-recommended-alerts.json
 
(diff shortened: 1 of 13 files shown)


petr-muller (approved):
I'm not _that_ sold on keeping "Upstream" because I think it makes the line harder for most people while conveying information useful for not that many people, but that's just an opinion and we can always act on feedback if people keep asking us what this line means.

LGTM

/retest-required

PR #1981, opened 2025-02-20
NO-ISSUE: pkg/cli/admin/upgrade/recommend: Configureable version for test fixtures

wking:
Always testing 4.12.51 was easy, but is less interesting once we have one `no updates to 4.12 available` example.  Make the target version for each ClusterVersion dump configurable, so we can see what `--version <VERSION>` with working versions expands too on more recent dumps as well.


petr-muller (approved):
LGMT

/retest-required

wking:
Rebased around #1980 with 719fbb353 -> cbecb375c.  No conflicts per local Git; GitHub was just being twitchy.

PR #1990, opened 2025-03-11
[OCPBUGS-50992]: Filter out unreachable taints from tolerations 

midu16:
[OCPBUGS-50992]: Filter out unreachable taints from tolerations:

- Define a list of excluded taints (`unreachableTaintKey`) with effects `NoExecute` and `NoSchedule`.
- Add tolerations for `node-role.kubernetes.io/master` (if applicable) and `node.kubernetes.io/not-ready`.
- Filter out tolerations that match the excluded taints to prevent unintended scheduling behavior.
- Implement a loop to iterate over tolerations and exclude ones that tolerate `unreachableTaintKey`.

This change ensures that workloads do not unintentionally tolerate unreachable nodes while still allowing necessary tolerations.


midu16:
@ingvagabund , @atiratree i am trying to get some sense on the failed checks, but i cannot find the corelation between my proposed changed patch and the errors. What am I missing here from your point of view? 

Much appreciated,
M

ingvagabund:
Looks like all three e2e tests are flaking (unrelated to this PR) and the verify needs to run "make updade-gofmt".

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -941,6 +943,38 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool) *corev1.P
...
+	
+	excludedTaints := []corev1.Taint{
+		{Key: unreachableTaintKey, Effect: corev1.TaintEffectNoExecute},
+		{Key: unreachableTaintKey, Effect: corev1.TaintEffectNoSchedule},
+	}
+
+	var tolerations []corev1.Toleration
+	if node == "" && hasMaster {
+		tolerations = append(tolerations, corev1.Toleration{
+			Key:      "node-role.kubernetes.io/master",
```
ingvagabund, line 956: Could be turned into another const

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -941,6 +943,38 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool) *corev1.P
...
+			Operator: corev1.TolerationOpExists,
+			Effect:   corev1.TaintEffectNoSchedule,
+		})
+	}
+
+	tolerations = append(tolerations, corev1.Toleration{
+		Key:      "node.kubernetes.io/not-ready",
```
ingvagabund, line 963: Could be turned into another const

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -941,6 +943,38 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool) *corev1.P
...
 	volumeUsageChecker := fmt.Sprintf(volumeUsageCheckerScript, cleanedSourceDir, cleanedSourceDir, o.VolumePercentage, o.VolumePercentage, executedCommand)
+	//fmt.Sprintf(volumeUsageCheckerScript,cleanedSourceDir,o.VolumePercentage,executedCommand,)
+	
+	excludedTaints := []corev1.Taint{
+		{Key: unreachableTaintKey, Effect: corev1.TaintEffectNoExecute},
+		{Key: unreachableTaintKey, Effect: corev1.TaintEffectNoSchedule},
+	}
+
+	var tolerations []corev1.Toleration
+	if node == "" && hasMaster {
```
ingvagabund, line 954: There's already the same condition in the method. Moving the code a bit up might make the code easier to read to avoid thinking about "does this cluster has master nodes" condition again.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -941,6 +943,38 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool) *corev1.P
...
+		tolerations = append(tolerations, corev1.Toleration{
+			Key:      "node-role.kubernetes.io/master",
+			Operator: corev1.TolerationOpExists,
+			Effect:   corev1.TaintEffectNoSchedule,
+		})
+	}
+
+	tolerations = append(tolerations, corev1.Toleration{
```
ingvagabund, line 962: You could replace the first `var tolerations []corev1.Toleration` with:
```go
tolerations := []corev1.Toleration{
	{
		Key:      "node.kubernetes.io/not-ready",
		Operator: corev1.TolerationOpExists,
		Effect:   corev1.TaintEffectNoSchedule,
	},
}
```

directly.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -941,6 +943,38 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool) *corev1.P
...
+		Key:      "node.kubernetes.io/not-ready",
+		Operator: corev1.TolerationOpExists,
+		Effect:   corev1.TaintEffectNoSchedule,
+	})
+
+	filteredTolerations := make([]corev1.Toleration, 0)
+TolerationLoop:
+	for _, tol := range tolerations {
+		for _, excluded := range excludedTaints {
+			if tol.ToleratesTaint(&excluded) {
```
ingvagabund, line 972: So you are excluding taints that are added in the same method. What about not adding the excluded taints instead of adding them and then excluding them? Or, is this code to be extended later with a list of user provided excluded taints?

Also, none of those two tolerations tolerate the taint as the `Key` fields are always different:
- `"node-role.kubernetes.io/master"` != `"node.kubernetes.io/unreachable"`
- `"node.kubernetes.io/not-ready"` != `"node.kubernetes.io/unreachable"`

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -1016,14 +1050,7 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool) *corev1.P
...
 			TerminationGracePeriodSeconds: &zero,
-			Tolerations: []corev1.Toleration{
-				{
-					// An empty key with operator Exists matches all keys,
-					// values and effects which means this will tolerate everything.
-					// As noted in https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
-					Operator: "Exists",
-				},
-			},
+			Tolerations:                   filteredTolerations,
```
ingvagabund, line 1053: There are two case to consider. Once for when `node == "" && hasMaster` is true. And the other case. So either the list of tolerations has both tolerations or just the one with `"node.kubernetes.io/not-ready"`.

ingvagabund:
What about extend the `oc adm must-gather` code with a warning saying "hey, the must-gather pod got scheduled to THIS node that is tainted with `node.kubernetes.io/unreachable`. Maybe you'd like to schedule it to a different node to avoid the pod getting stuck. Here are some suggestions: NODES". This will help with a user initiated case. Not much when running the must-gather through an automation.

ingvagabund:
What if all control plane nodes are "node.kubernetes.io/unreachable" tainted? What if all nodes are "node.kubernetes.io/unreachable" tainted? Some nodes might be still eligible for running must-gather. Presence of the taint does not necessarily means the node is "gone". From https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions:
```
node.kubernetes.io/unreachable: Node is unreachable from the node controller. This corresponds to the NodeCondition Ready being "Unknown".
```

Maybe it's just the node controller that's temporarily separated from the rest of the control plane?

midu16:
> What if all control plane nodes are "node.kubernetes.io/unreachable" tainted? What if all nodes are "node.kubernetes.io/unreachable" tainted? Some nodes might be still eligible for running must-gather. Presence of the taint does not necessarily means the node is "gone". From https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions:
> 
> ```
> node.kubernetes.io/unreachable: Node is unreachable from the node controller. This corresponds to the NodeCondition Ready being "Unknown".
> ```
> 
> Maybe it's just the node controller that's temporarily separated from the rest of the control plane?


I can expand this topic as follows: 

- Don’t outright exclude tainted nodes.

- Prioritize nodes without the taints.
 
- Fallback to unreachable or not-ready nodes if no other healthy candidates exist.
 
- Optionally, add logic to:
 
   - Check node heartbeat age (status.conditions.LastHeartbeatTime)

   - Or try lightweight probing before deploying the workload.

pkg/cli/admin/mustgather/mustgather.go
midu16: @ingvagabund added these suggestions to https://github.com/openshift/oc/pull/1990/commits/365fb3a6f3ca97cdfa687da01a28121c19b11589

pkg/cli/admin/mustgather/mustgather.go
midu16: @ingvagabund added these suggestions to https://github.com/openshift/oc/pull/1990/commits/365fb3a6f3ca97cdfa687da01a28121c19b11589

pkg/cli/admin/mustgather/mustgather.go
midu16: @ingvagabund added these suggestions to https://github.com/openshift/oc/pull/1990/commits/365fb3a6f3ca97cdfa687da01a28121c19b11589

midu16:
> What about to extend the `oc adm must-gather` code with a warning saying "hey, the must-gather pod got scheduled to THIS node that is tainted with `node.kubernetes.io/unreachable`. Maybe you'd like to schedule it to a different node to avoid the pod getting stuck. Here are some suggestions: NODES". This will help with a user initiated case. Not much when running the must-gather through an automation. On the other hand the automation might set a special flag to tell `oc adm must-gather` to be more strict in selecting which nodes are still "safe" to tolerate. So, as we probably suggested in one of the previous bug reports it's preferable to have user/automation provide a list of explicit tolerations if the current "tolerate everything" does not work well.

This been implemented here: https://github.com/openshift/oc/pull/1990/commits/93d98b5e9b2733b137ef371de5fbefd03eb8353e

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: Still unanswered

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -403,6 +418,42 @@ func (o *MustGatherOptions) Validate() error {
...
 
+// prioritizeHealthyNodes returns a preferred node to run the must-gather pod on, and a fallback node if no preferred node is found.
+func prioritizeHealthyNodes(nodes *corev1.NodeList) (preferred *corev1.Node, fallback *corev1.Node) {
+	var fallbackNode *corev1.Node
+	var latestHeartbeat time.Time
+
+	for _, node := range nodes.Items {
+		var hasUnhealthyTaint bool
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == "node.kubernetes.io/unreachable" || taint.Key == "node.kubernetes.io/not-ready" {
```
ingvagabund: "node.kubernetes.io/unreachable" can be replaced with "unreachableTaintKey". The same for other two occurrences.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -464,6 +515,33 @@ func (o *MustGatherOptions) Run() error {
...
 		LabelSelector: o.NodeSelector,
 	})
+	if err != nil {
+		return err
+	}
+
+	if o.NodeName == "" {
+		preferred, fallback := prioritizeHealthyNodes(nodes)
+		if preferred != nil {
+			o.NodeName = preferred.Name
```
ingvagabund: This changes the default. This might lead to unexpected behavior. If there's any such mechanism to be implemented it needs to be explicitly enabled. E.g. through an experimental flag.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -403,6 +418,42 @@ func (o *MustGatherOptions) Validate() error {
 	return nil
 }
 
+// prioritizeHealthyNodes returns a preferred node to run the must-gather pod on, and a fallback node if no preferred node is found.
+func prioritizeHealthyNodes(nodes *corev1.NodeList) (preferred *corev1.Node, fallback *corev1.Node) {
```
ingvagabund, line 448: This looks good. Thank you.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -51,6 +51,7 @@ import (
 
 const (
 	gatherContainerName = "gather"
+	unreachableTaintKey = "node.kubernetes.io/unreachable"
```
ingvagabund, line 55: Could be extended with `notReadyTaintKey = "node.kubernetes.io/not-ready"`

pkg/cli/admin/mustgather/mustgather.go
midu16: intention was not to include a new flag. If this is the case, i can revert this change. What do you think ?

ingvagabund: We can't change defaults. So if a new functionality is introduced (for non-bug/security fixes) changing the defaults a new flag (or a different mechanism) for enabling the functionality needs to be introduced. I am not against the new functionality. It looks helpful.

midu16: If I include a new flag, could this potentially affect the merging process?

I am prepared to address the necessary updates; however, I have some reservations about introducing the new flag at this stage. It might be more appropriate to defer this addition to a future pull request.

ingvagabund: > If I include a new flag, could this potentially affect the merging process?

Smaller PRs are better to review and revert if needed.

> It might be more appropriate to defer this addition to a future pull request.

Up to you. Though, preferable.

pkg/cli/admin/mustgather/mustgather.go
midu16: @ingvagabund with respect to this, i have added a comment that i am intending to support a user-defined exclusion tains in the hope to preserve this logic, should we define a unit test which is validating this logic further ?

ingvagabund: How far in the future do you expect the user-defined exclusion taints support added? If the current logic has no use it's better to introduce it in the PR that introduces the user-defined exclusion taints. So all the relevant changes are in the same PR to preserve the context. This way the code changes will be apart. Maybe never extended/finished.

midu16: @ingvagabund I expect to expand it in the next PR with the experimental flag from here: https://github.com/openshift/oc/pull/1990#discussion_r2144830092

ingvagabund: Gotcha. In that case it's better to introduce the loop as part of the next PR.

pkg/cli/admin/mustgather/mustgather.go
midu16: @ingvagabund this has been removed from the new push.

pkg/cli/admin/mustgather/mustgather.go
midu16: @ingvagabund unfortunetly this cannot be removed for next PR, main feature its to manage the tolerations for the must-gather pod and avoid a random alocation to a pod which cannot be scheduled. Initial implementation was allowing the must-gather pod to tolerate everything, hence this section was change to exclude specific taints.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -401,6 +417,42 @@ func (o *MustGatherOptions) Validate() error {
...
+				hasUnhealthyTaint = true
+				break
+			}
+		}
+
+		// Look at heartbeat time for readiness hint
+		for _, cond := range node.Status.Conditions {
+			if cond.Type == corev1.NodeReady && cond.Status == corev1.ConditionTrue {
+				// Check if heartbeat is recent (less than 2m old)
+				if time.Since(cond.LastHeartbeatTime.Time) < 2*time.Minute {
```
ingvagabund, line 465: Note: This assumes the clock running on a host where `oc adm must-gather` runs is in sync with the nodes. Which might not be always the case.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -462,6 +514,10 @@ func (o *MustGatherOptions) Run() error {
 	nodes, err := o.Client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
 		LabelSelector: o.NodeSelector,
 	})
+	if err != nil {
```
ingvagabund, line 540: Duplicated code

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -401,6 +417,42 @@ func (o *MustGatherOptions) Validate() error {
 	return nil
 }
 
+// prioritizeHealthyNodes returns a preferred node to run the must-gather pod on, and a fallback node if no preferred node is found.
+func prioritizeHealthyNodes(nodes *corev1.NodeList) (preferred *corev1.Node, fallback *corev1.Node) {
```
ingvagabund, line 448: `prioritizeHealthyNodes` is currently unused in the code

ingvagabund:
Running the added code locally for filtering the taints (with `hasMaster==false`):
```
func TestExcludeTolerations(t *testing.T) {
	excludedTaints := []corev1.Taint{
		{Key: unreachableTaintKey, Effect: corev1.TaintEffectNoExecute},
		{Key: notReadyTaintKey, Effect: corev1.TaintEffectNoSchedule},
	}

	candidateTolerations := []corev1.Toleration{tolerationNotReady}

	filteredTolerations := make([]corev1.Toleration, 0, len(candidateTolerations))
TolerationLoop:
	for _, tol := range candidateTolerations {
		for _, excluded := range excludedTaints {
			if tol.ToleratesTaint(&excluded) {
				// Skip this toleration if it tolerates an excluded taint
				continue TolerationLoop
			}
		}
		filteredTolerations = append(filteredTolerations, tol)
	}
	fmt.Printf("filteredTolerations: %#v\n", filteredTolerations)
}
```
In case a cluster has no masters the list of filteredTolerations will be empty. Which is/might be the case on hypershift clusters (https://github.com/openshift/oc/pull/1347). Which makes the must-gather less resilient than now.

> This change ensures that workloads do not unintentionally tolerate unreachable nodes while still allowing necessary tolerations.

Which invalidates the original intention to still allow necessary tolerations.

ingvagabund:
@midu16 there's no simple solution for [OCPBUGS-50992](https://issues.redhat.com/browse/OCPBUGS-50992). Even the current code adjustments changes the must-gather pod defaults. We need to keep the current behavior preserved even for control plane nodes that are NotReady. Since even a NotReady control plane may provide useful debugging data. Inspection of a cluster with NotReady nodes usually requires to make an informed decision and choose the right target node to get the best results.

Given the main motivation for [OCPBUGS-50992](https://issues.redhat.com/browse/OCPBUGS-50992) is automation I suggest to create a new `oc adm must-gather` option that will specify a scheduling profile. The current scheduling constrains are set for best-effort. In the automation case the choice of scheduling constrains can be more focused on finding more suitable nodes. E.g. your code changes suggesting the most suitable control plane node for scheduling. Given it's currently unknown which scheduling constrains are sufficiently suitable for automation I suggest to create an experimental profile (that will need to be enabled by default) with your code improvements so we can collect feedback from customers and see what else might be a suitable guiding principle. E.g.
```
$ oc adm must-gather --scheduling-profile=PROFILENAME
```

With the following profiles:
- "" (empty) or "BEST_EFFORT": the current (default) choice of tolerating any taint
- "HEALTHY_NODES" (or a different name): focused on selecting the most suitable node. E.g. setting a node anti-affinity to avoid scheduling to not ready nodes. E.g.
  ```
  spec:
    affinity:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
          - matchExpressions:
            - key: kubernetes.io/hostname
              operator: NotIn
              values:
              - control-plane-123.example.com
              - ...
  ```

MarSik:
@ingvagabund But this is not about collecting data from NotReady nodes, this is about starting the main must gather process (dispatcher if you want) iirc. When scheduled to a NotReady node, it will not start and no collection happens at all, right?

Surely you have at least one node in the cluster that can start workloads. If not, you are in much worse trouble.

ingvagabund:
`oc adm must-gather` (run locally from CLI) is the main dispatcher. Creating must-gather pods. The CLI part can detect which nodes are more suitable for scheduling.

MarSik:
Isn't it creating just one MG Pod?

The pod that then starts the main https://github.com/openshift/must-gather/blob/main/collection-scripts/gather which schedules the other pods and daemonsets?

Anyway, the issue here is that the `oc must-gather` tool execution fails when the main pod gets scheduled to a NotReady node. Even when there are other nodes that would run the pod given the chance. This is very user unfriendly, you basically need to try multiple times and hope for the best.

ingvagabund:
> Isn't it creating just one MG Pod?

Many must-gather pods can be created (depending on how many must-gather images get passed in).

> The pod that then starts the main https://github.com/openshift/must-gather/blob/main/collection-scripts/gather which schedules the other pods and daemonsets?

the `gather` script is ran within a container. There's no more pod creation/scheduling involved here.

> Anyway, the issue here is that the oc must-gather tool execution fails when the main pod gets scheduled to a NotReady node. 

There's no main pod. All the orchestration is perform through `oc adm must-gather` subcommand. The command first creates must-gather pods and waits until all the pods complete. Then starts pulling the data from each container under the same host file system where `oc adm must-gather` is running. There's currently no other than the pod scheduling involved.

> Even when there are other nodes that would run the pod given the chance. This is very user unfriendly, you basically need to try multiple times and hope for the best.

I can understand the user frustration here. There's currently no "smart" scheduling involved. Also, there's no simple way to know which tolerations are the right ones to tolerate.

MarSik:
> There's no main pod. All the orchestration is perform through oc adm must-gather subcommand. The command first creates must-gather pods and waits until all the pods complete.

Ok, when I simplify the case to no extra images then just one pod is started. That is what I called the main pod. Because the default gather pod actually starts couple of others (typically daemonsets to collect node level data).

However nothing happens except an error, when the node selected by oc adm must-gather is NotReady.

> I can understand the user frustration here. There's currently no "smart" scheduling involved. Also, there's no simple way to know which tolerations are the right ones to tolerate.

There is a simple way to know which tolerations are not however. Any taint that prevents the gathering pod from starting at all. Like NotReady or Unschedulable.

The non-determinism is what prevents automation and causes the frustration.

Why pick an unhealthy node, when a healthy node is available? The healthy one should be selected when possible, don't you think?

ingvagabund:
> Ok, when I simplify the case to no extra images then just one pod is started. That is what I called the main pod. Because the default gather pod actually starts couple of others (typically daemonsets to collect node level data).

Gotcha. Then this steps is performed within a running must-gather image. Which is outside of control of `oc adm must-gather`. When the must-gather pod gets scheduled to a NotReady control plane node and fails to reach to kubeapiserver no additional pods (e.g. the DS pods) get created.

> However nothing happens except an error, when the node selected by oc adm must-gather is NotReady.

Currently, `oc adm must-gather` does not selector any nodes. Unless `--node-name` option is used. Though, point taken.

> There is a simple way to know which tolerations are not however. Any taint that prevents the gathering pod from starting at all. Like NotReady or Unschedulable.

This is the part where we don't have a full list of all tolerations that are ok to tolerate. Some of the important ones (I don't know which ones these are) might be omitted. Which might cause more trouble than now or maybe not. This is where my knowledge is limited. For that I suggest to take a slower path and allow user to explicitly specify a list of such tolerations (previously discussed in https://issues.redhat.com/browse/OCPBUGS-29940).

> Why pick an unhealthy node, when a healthy node is available? The healthy one should be selected when possible, don't you think?

The current `oc adm must-gather` does not perform any scheduling or node classification atm. This part is quite limited.

MarSik:
Think about this:

> This is where my knowledge is limited. For that I suggest to take a slower path and allow user to explicitly specify a list of such tolerations (previously discussed in https://issues.redhat.com/browse/OCPBUGS-29940).

How should the user know which tolerations should be used when you don't? Especially when integrating must gather into an automated solution?

The customers are interested in a fire and forget solution, their operations people are not going to do the diagnostics and are not experts. And my experience tells me that people simply forget to use extra arguments.

> This is the part where we don't have a full list of all tolerations that are ok to tolerate.

Ok, so it only covers the majority and some rare cases might still require a manual intervention. Sounds good enough to me.

MarSik:
After an offline discussion with @ingvagabund I understand his objections a bit better.

The general concept of skipping unhealthy nodes seems to be OK, but must gather should prefer control plane nodes over random workers, because the data collection is API call heavy. Ignoring taints like this would not guarantee that.

Also we need to keep Hypershift in mind as that is an environment where the control plane is not visible / not able to run pods.

For those reasons he proposed we check whether there are unhealthy control plane nodes as well as healthy one(s) and use explicit nodeAffinity instead of taints/tolerations to force the scheduler to pick the healthy nodes.

I suppose the same could be applied to workers as a fallback in case there is no healthy control plane node available like on hypershift.

@ingvagabund Is this a good summary of what needs to be changed?

ingvagabund:
This aligns what we discussed offline. Thank you for summarizing the main points here.

> use explicit nodeAffinity instead of taints/tolerations

Just a nit. The "tolerate everything" toleration can stay as it is. With node affinity in place pointing to ready nodes will direct the scheduler to ignore any NotReady nodes. In case the number of Ready control plane nodes is low, the list of node affinity nodes can be extended with unschedulable nodes as well. So the following steps will work:
- if there are sufficient number of Ready and schedulable nodes set the node affinity to include only these nodes during scheduling
- if there are not, extend the list with unschedulable nodes as well to improve the odds
- fall back to the current behavior

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -99,6 +101,44 @@ sleep 5
...
+
+	tolerationMasterNoSchedule = corev1.Toleration{
+		Key:      "node-role.kubernetes.io/master",
+		Operator: corev1.TolerationOpExists,
+		Effect:   corev1.TaintEffectNoSchedule,
+	}
+)
+
+// buildNodeSelector builds a node selector from the provided node names.
+func buildNodeAffinity(nodeNames []string) *corev1.Affinity {
```
ricardomaraschini, line 119: It can't be both `buildNodeSelector` and `buildNodeAffinity`. Fix the comment or fix the function name.

midu16:
merged under https://github.com/openshift/oc/pull/2060

PR #1992, opened 2025-03-20
Use fedora image in unit tests instead of centos

ardaguclu:
It seems that centos is entirely removed and unit tests in oc are permanently failing. This PR uses `fedora` image instead of removed `centos` image to fix this failure.


ardaguclu:
@ingvagabund would you mind tagging this, when you have a chance?. Thank you.

PR #1997, opened 2025-03-24
NO-JIRA: `adm release new`: simplify `writePayload`

petr-muller:
There was code that was supposed to handle potentially nested path, but this path (parts) is hardcoded to be a single top-level directory, so none of the complexity is necessary and can be removed, including the special `writeNestedTarHeader` method.

Some declarations can be moved closer to where they are used, improving readability.

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -1311,23 +1311,6 @@ type nopCloser struct {
 
 func (_ nopCloser) Close() error { return nil }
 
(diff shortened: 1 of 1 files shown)


hongkailiu (commented):
/lgtm

nit: I would avoid using same title for two consecutive commits. Maybe rename the 2nd to `adm release new: remove writeNestedTarHeader` or something?

in case you want to rename ^^^
/hold

petr-muller:

This is only a refactoring, no functionality change

PR #1998, opened 2025-03-25
OCPBUGS-30080: `adm release new`: PoC: replace version placeholders in CVO manifests

petr-muller:
This a proof-of-concept of a fix for OCPBUGS-30080, allowing the `0.0.0-snapshot` placeholders to be replaced in CVO manifests, just like it is in the manifests extracted from the operator images.

The CVO image, typically identified through the `--to-base-image-ref` (with a default of `cluster-version-operator`) should be processed and its manifests extracted for later inclusion in the payload image. CVO image supplied this way is always included as an input to the payload:

```golang
// the base tag should always be included in a release payload
if len(o.ToImageBaseTag) > 0 {
  o.AlwaysInclude = append(o.AlwaysInclude, o.ToImageBaseTag)
}
```

But its manifests are not extracted, because CVO image is not labeled with `io.openshift.release.operator` (if it was, its manifests would be extracted and included in `release-manifests` directory, like all other). The PoC relaxes this condition by allowing the tag specified in `o.ToImageBaseTag` to be extracted even if it is not labeled this way. The extracted manifests are later processed for release payload inclusion.

Second part of the PoC change is to recognize the CVO manifests when individual manifests are processed, transformed and considered for inclusion in the payload. The CVO manifests are recognized by passing `o.ToImageBaseTag` into the responsible function, and placing these manifests into `manifests` directory instead, but this allows them to be transformed with the simple versions mapper, which replaces the `0.0.1-snapshot` placeholder in them.

This is a tricky change because of an unknown interaction with `--to-base-image`. That image is not used as an input to the release payload, but simply as a destination, and the CVO binary and `manifests` directory contents presence is simply assumed. The PoC would change this behavior, but it is unclear what use case explicit `--to-base-image` not corresponding to the `--to-base-image-tag` has.

### DIFF
--- a/pkg/cli/admin/release/new.go
+++ b/pkg/cli/admin/release/new.go
@@ -769,7 +769,7 @@ func (o *NewOptions) Run(ctx context.Context) error {
 	pr, pw := io.Pipe()
 	go func() {
 		var err error
-		operators, err = writePayload(pw, is, cm, ordered, metadata, o.AllowMissingImages, verifiers, o.ErrOut)
+		operators, err = writePayload(pw, is, cm, ordered, metadata, o.AllowMissingImages, verifiers, o.ErrOut, o.ToImageBaseTag)
 		pw.CloseWithError(err)
 	}()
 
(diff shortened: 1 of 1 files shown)


petr-muller:
LOL am I breaking release builds somehow :D

petr-muller:
/close

Explanation in the JIRA [comment](https://issues.redhat.com/browse/OCPBUGS-30080?focusedId=27114264&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27114264), effort superseded by OTA-1531

PR #1999, opened 2025-03-25
NO-JIRA: `adm release new`: simplify no operator error condition

petr-muller:
The actual `operators` content is never used, so it does not need to be sorted, and treating `nil` as a silent good case and `len()==0` as an error case does not look right (`writePayload` never produces an empty non-nil `operators`).


petr-muller:

This is only a refactoring, no functionality change

PR #2001, opened 2025-04-02
CNTRLPLANE-371: oc idle: Use endpointslice instead of deprecated endpoints

ardaguclu:
`endpoints` resource has been deprecated and `endpointslices` is the recommended resource instead. `oc idle` command heavily relies on endpoints and this PR switches it to endpointslices to unblock 1.33 rebasing process.

This PR should not have any functional changes.

### DIFF
--- a/pkg/cli/idle/idle.go
+++ b/pkg/cli/idle/idle.go
@@ -14,6 +14,7 @@ import (
 	kappsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/discovery/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -45,7 +46,7 @@ var (
 		Idle scalable resources.
 
 		Idling discovers the scalable resources (such as deployment configs and replication controllers)
-		associated with a series of services by examining the endpoints of the service.
+		associated with a series of services by examining the endpointslices of the service.
 		Each service is then marked as idled, the associated resources are recorded, and the resources
 		are scaled down to zero replicas.
 
@@ -230,7 +231,7 @@ func scanLinesFromFile(filename string) ([]string, error) {
 	return lines, nil
 }
 
-// idleUpdateInfo contains the required info to annotate an endpoints object
+// idleUpdateInfo contains the required info to annotate a service object
 // with the scalable resources that it should unidle
 type idleUpdateInfo struct {
 	service   *corev1.Service
@@ -238,8 +239,8 @@ type idleUpdateInfo struct {
 }
 
 // calculateIdlableAnnotationsByService calculates the list of objects involved in the idling process from a list of services in a file.
-// Using the list of services, it figures out the associated scalable objects, and returns a map from the endpoints object for the services to
-// the list of scalable resources associated with that endpoints object, as well as a map from CrossGroupObjectReferences to scale to 0 to the
+// Using the list of services, it figures out the associated scalable objects, and returns a map from the endpointslices object for the services to
+// the list of scalable resources associated with that endpointslices object, as well as a map from CrossGroupObjectReferences to scale to 0 to the
 // name of the associated service.
 func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(resource.VisitorFunc) error) (map[types.NamespacedName]idleUpdateInfo, map[namespacedCrossGroupObjectReference]types.NamespacedName, error) {
 	podsLoaded := make(map[corev1.ObjectReference]*corev1.Pod)
@@ -312,14 +313,19 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
 			return err
 		}
 
-		endpoints, isEndpoints := info.Object.(*corev1.Endpoints)
+		endpoints, isEndpoints := info.Object.(*v1.EndpointSlice)
 		if !isEndpoints {
-			return fmt.Errorf("you must specify endpoints, not %v (view available endpoints with \"oc get endpoints\").", info.Mapping.Resource)
+			return fmt.Errorf("you must specify endpointslices, not %v (view available endpointslices with \"oc get endpointslices\").", info.Mapping.Resource)
+		}
+
+		serviceName, ok := endpoints.Labels[v1.LabelServiceName]
+		if !ok {
+			serviceName = endpoints.Name
 		}
 
 		endpointsName := types.NamespacedName{
 			Namespace: endpoints.Namespace,
-			Name:      endpoints.Name,
+			Name:      serviceName,
 		}
 		scaleRefs, err := findScalableResourcesForEndpoints(endpoints, getPod, getController)
 		if err != nil {
@@ -333,9 +339,9 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
 			targetScaleRefs[ref] = endpointsName
 		}
 
-		svc, err := o.ClientSet.CoreV1().Services(endpoints.Namespace).Get(context.TODO(), endpoints.Name, metav1.GetOptions{})
+		svc, err := o.ClientSet.CoreV1().Services(endpoints.Namespace).Get(context.TODO(), serviceName, metav1.GetOptions{})
 		if err != nil {
-			return fmt.Errorf("unable to get service %s/%s: %v", endpoints.Namespace, endpoints.Name, err)
+			return fmt.Errorf("unable to get service %s/%s: %v", endpoints.Namespace, serviceName, err)
 		}
 
 		idleInfo := idleUpdateInfo{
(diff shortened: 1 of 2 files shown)


ardaguclu:
@ingvagabund @atiratree could you ptal, when you have a chance?

ardaguclu:
This PR is complementary with https://github.com/openshift/origin/pull/29647 but not directly depend on each other.

pkg/helpers/scheme/scheme.go
```diff
@@ -16,6 +11,10 @@ import (
 	securityv1 "github.com/openshift/api/security/v1"
 	"github.com/openshift/api/template"
 	"github.com/openshift/api/user"
+	corev1 "k8s.io/api/core/v1"
```
ingvagabund, line 14: Worth moving this back where it was to undo the no-op changes to this file

ardaguclu: Thank you for review. Nice catch. Updated.

ingvagabund:
Good to invite some from the networking team for the final conformation.

ingvagabund:
From chatGPT:

If you have manually created Endpoints, you need to create corresponding EndpointSlices. Example:

Example of a Kubernetes Endpoints object:
```yaml
apiVersion: v1
kind: Endpoints
metadata:
  name: my-service
subsets:
  - addresses:
      - ip: 192.168.1.1
      - ip: 192.168.1.2
    ports:
      - port: 80
```
Equivalent EndpointSlice:
```yaml
apiVersion: discovery.k8s.io/v1
kind: EndpointSlice
metadata:
  name: my-service-slice
  labels:
    kubernetes.io/service-name: my-service
addressType: IPv4
endpoints:
  - addresses:
      - "192.168.1.1"
  - addresses:
      - "192.168.1.2"
ports:
  - name: http
    port: 80
    protocol: TCP
```

ardaguclu:
@danwinship would you mind taking a look at this change?.

pkg/cli/idle/idle.go
```diff
@@ -230,16 +231,16 @@ func scanLinesFromFile(filename string) ([]string, error) {
...
 type idleUpdateInfo struct {
 	service   *corev1.Service
 	scaleRefs map[unidlingapi.CrossGroupObjectReference]struct{}
 }
 
 // calculateIdlableAnnotationsByService calculates the list of objects involved in the idling process from a list of services in a file.
-// Using the list of services, it figures out the associated scalable objects, and returns a map from the endpoints object for the services to
-// the list of scalable resources associated with that endpoints object, as well as a map from CrossGroupObjectReferences to scale to 0 to the
+// Using the list of services, it figures out the associated scalable objects, and returns a map from the endpointslices object for the services to
+// the list of scalable resources associated with that endpointslices object, as well as a map from CrossGroupObjectReferences to scale to 0 to the
```
danwinship, line 243: With Endpoints, the object name is plural, and there's guaranteed to be only one of them (hence, "the endpoints object"). With EndpointSlice, the object name is singular, but there may be more than one of them if the service has lots of endpoints or is dual-stack, so it should be "the endpointslice objects".

However, this also means the new code is slightly wrong; if there are multiple EndpointSlices for a Service, then you need to merge together the results of `findScalableResourcesForEndpoints` for each one together in the returned `endpointsInfo`, whereas the current version of this PR will overwrite the info calculated for the first EndpointSlice when it processes the second EndpointSlice.

pkg/cli/idle/idle.go
```diff
@@ -312,14 +313,19 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
 			return err
 		}
 
-		endpoints, isEndpoints := info.Object.(*corev1.Endpoints)
+		endpoints, isEndpoints := info.Object.(*v1.EndpointSlice)
```
danwinship, line 316:
```suggestion
		slice, isEndpointSlice := info.Object.(*v1.EndpointSlice)
```

?

Yeah, that will require more rewriting, but it makes it clearer

pkg/cli/idle/idle.go
```diff
@@ -312,14 +313,19 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
...
 		if !isEndpoints {
-			return fmt.Errorf("you must specify endpoints, not %v (view available endpoints with \"oc get endpoints\").", info.Mapping.Resource)
+			return fmt.Errorf("you must specify endpointslices, not %v (view available endpointslices with \"oc get endpointslices\").", info.Mapping.Resource)
+		}
+
+		serviceName, ok := endpoints.Labels[v1.LabelServiceName]
+		if !ok {
+			serviceName = endpoints.Name
```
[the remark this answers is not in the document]
danwinship: The label should always be present. If it's not, then the EndpointSlice does not represent a Service in any normal way, so you should ignore it.

pkg/cli/idle/idle.go
```diff
@@ -312,14 +313,19 @@ func (o *IdleOptions) calculateIdlableAnnotationsByService(infoVisitor func(reso
...
 		}
 
 		endpointsName := types.NamespacedName{
 			Namespace: endpoints.Namespace,
-			Name:      endpoints.Name,
+			Name:      serviceName,
```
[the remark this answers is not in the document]
danwinship: given this, `endpointsName` is no longer a good name for the variable ( `serviceName` would work except it's already taken...)

ardaguclu:
> I can't confirm whether this is a compatible change. https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/ seems to support a conversion path from endpoints to endpointslices.
> 
> https://medium.com/@kedarnath93/kubernetes-endpoints-vs-endpointslices-with-an-example-a3b4f48ce60d gives some examples so change in this PR looks valid.

We have a couple of tests for this command. So that if they pass, it can be said that we cover most part. In any case, it is better to ask from an expert.

pkg/cli/idle/idle.go
```diff
@@ -393,24 +399,22 @@ func normalizedNSOwnerRef(namespace string, ownerRef *metav1.OwnerReference) nam
...
 }
 
-// findScalableResourcesForEndpoints takes an Endpoints object and looks for the associated
+// findScalableResourcesForEndpoints takes an EndpointSlice object and looks for the associated
 // scalable objects by checking each address in each subset to see if it has a pod
 // reference, and the following that pod reference to find the owning controller,
 // and returning the unique set of controllers found this way.
-func findScalableResourcesForEndpoints(endpoints *corev1.Endpoints, getPod func(corev1.ObjectReference) (*corev1.Pod, error), getController func(namespacedOwnerReference) (metav1.Object, error)) (map[namespacedCrossGroupObjectReference]struct{}, error) {
-	// To find all RCs and DCs for an endpoint, we first figure out which pods are pointed to by that endpoint...
+func findScalableResourcesForEndpoints(endpoints *v1.EndpointSlice, getPod func(corev1.ObjectReference) (*corev1.Pod, error), getController func(namespacedOwnerReference) (metav1.Object, error)) (map[namespacedCrossGroupObjectReference]struct{}, error) {
```
danwinship, line 406:
```suggestion
func findScalableResourcesForEndpoints(slice *v1.EndpointSlice, getPod func(corev1.ObjectReference) (*corev1.Pod, error), getController func(namespacedOwnerReference) (metav1.Object, error)) (map[namespacedCrossGroupObjectReference]struct{}, error) {
```

pkg/cli/idle/idle.go
```diff
@@ -480,7 +484,7 @@ func pairScalesWithScaleRefs(serviceName types.NamespacedName, annotations map[s
 		var oldTargets []unidlingapi.RecordedScaleReference
 		oldTargetsSet := make(map[unidlingapi.CrossGroupObjectReference]int)
 		if err := json.Unmarshal([]byte(oldTargetsRaw), &oldTargets); err != nil {
-			return nil, fmt.Errorf("unable to extract existing scale information from endpoints %s: %v", serviceName.String(), err)
+			return nil, fmt.Errorf("unable to extract existing scale information from endpointslices %s: %v", serviceName.String(), err)
```
danwinship, line 487:
```suggestion
			return nil, fmt.Errorf("unable to extract existing scale information from endpointslice %s/%s: %v", slice.Namespace, slice.Name, err)
```

pkg/cli/idle/idle.go
```diff
@@ -593,17 +597,44 @@ func (o *IdleOptions) RunIdle() error {
...
+			})
+			if err != nil {
+				return err
+			}
+			for _, slice := range endpointSlices.Items {
+				endpointSliceNames = append(endpointSliceNames, slice.Name)
+			}
+		}
+
+		b.ResourceNames("endpointslices", endpointSliceNames...)
```
danwinship, line 614: (I'm not convinced that this whole "visitor pattern" thing is really a simplification rather than a complication at this point..)

pkg/cli/idle/idle.go
```diff
@@ -693,7 +724,7 @@ func (o *IdleOptions) RunIdle() error {
 		return err
 	}
 
-	// annotate the endpoints objects to indicate which scalable resources need to be unidled on traffic
+	// annotate the service objects to indicate which scalable resources need to be unidled on traffic
```
danwinship, line 727: (might be worth splitting all of the "endpoints"->"service" comment fixes into a separate commit, since that's a belated fix to match how the code already was, rather than being associated with the new patch.)

ardaguclu:
This change will take longer than I anticipated. We need to plan this work in a sprint. Besides https://github.com/openshift/origin/pull/29647 has unblocked the rebasing process.
/close

PR #2003, opened 2025-04-09
WRKLDS-1103: Adding sos.conf file for default sos config into the tools image

oarribas:
Adding `sos.conf` file with default configuration for `sos` as discussed in [WRKLDS-1103](https://issues.redhat.com//browse/WRKLDS-1103).

cc @ingvagabund

### DIFF
--- a/images/tools/Dockerfile
+++ b/images/tools/Dockerfile
@@ -5,6 +5,7 @@ RUN make build --warn-undefined-variables
 
 FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
 COPY --from=builder /go/src/github.com/openshift/oc/oc /usr/bin/
+COPY --from=builder /go/src/github.com/openshift/oc/images/tools/sos.conf /etc/sos/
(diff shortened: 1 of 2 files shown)


ingvagabund:
@oarribas I don't think we have any e2e test to validate this change. @zhouying7780 would you please sync with @oarribas to discuss how this change will get validated? E.g. starting a container from the tools, copy pasting the new sos.conf and testing the current configuration is sufficient for the expected use cases.

oarribas:
@zhouying7780 , I tested this creating the container with the modified files and starting it. When in the container, I checked that the `sos.conf` file is in place, and then run a `sos report` and checked the logs to see all the parameters are in place:
```
# cat /etc/sos/sos.conf

# sos report
# cd /var/tmp/
# tar -xf sosreport-xxx
# head -5 sosreport-xxx/sos_logs/sos.log 
```

oarribas:
it's not clear the test is really failing because the changes in this PR

/retest

ardaguclu:
This change looks good to me. If you agree upon the configurations in sos.conf, we can merge this
/approve

PR #2005, opened 2025-04-09
OCPBUGS-53106: Get installer image from configMap for node-image command

bfournie:
In order to handle disconnected environments for the node-image command, get the installer image from the configMap created by the installer instead of extracting it from the releaseImage.

### DIFF
--- a/pkg/cli/admin/nodeimage/basenodeimagecommand.go
+++ b/pkg/cli/admin/nodeimage/basenodeimagecommand.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"regexp"
 	"strings"
 	"time"
@@ -107,16 +108,24 @@ func (c *BaseNodeImageCommand) getNodeJoinerPullSpec(ctx context.Context) error 
 				}
 			}
 		}
-	} else if !kapierrors.IsNotFound(err) {
-		c.log("failure retrieving installer-images configMap %s", err.Error())
-	} else {
-		c.log("configMap containing installer-images is not available, trying to get image from registry")
 	}
 
-	// If cannot obtain installer image from configMap, get it from the releaseImage
-	// Note that after OCP 4.20 this should not be necessary since the configMap was added to installer in 4.19
+	// If cannot obtain installer image from configMap, get it from the releaseImage.
+	// Note that after OCP 4.20, this should not be necessary since the configMap was added to installer in 4.19.
+	c.log("configMap containing installer-images is not available, trying to get image from registry")
 	opts := ocrelease.NewInfoOptions(c.IOStreams)
 	opts.SecurityOptions = c.SecurityOptions
+	idmsFile, err := c.getIdmsFile(ctx)
+	if err != nil {
+		return err
+	}
+	opts.IDMSFile = idmsFile
+	defer func() {
+		if opts.IDMSFile != "" {
+			os.Remove(opts.IDMSFile)
+		}
+	}()
+
 	release, err := opts.LoadReleaseInfo(releaseImage, false)
 	if err != nil {
 		return err
@@ -435,4 +444,46 @@ func (o *BaseNodeImageCommand) runNodeJoinerPod(ctx context.Context, tasks []fun
 		}
 	}
 	return nil
+}
+
+func (c *BaseNodeImageCommand) getIdmsFile(ctx context.Context) (string, error) {
+	imageDigestMirrorSets, idmsErr := c.ConfigClient.ConfigV1().ImageDigestMirrorSets().List(ctx, metav1.ListOptions{})
+	if idmsErr != nil && !kapierrors.IsNotFound(idmsErr) {
+		return "", idmsErr
+	}
+	imageContentPolicies, icspErr := c.ConfigClient.ConfigV1().ImageContentPolicies().List(ctx, metav1.ListOptions{})
+	if icspErr != nil && !kapierrors.IsNotFound(icspErr) {
+		return "", icspErr
+	}
+	if idmsErr != nil || len(imageDigestMirrorSets.Items) == 0 {
+		imageDigestMirrorSets = nil // handle empty idms
+	}
+	if icspErr != nil || len(imageContentPolicies.Items) == 0 {
+		imageContentPolicies = nil // handle empty icsp
+	}
+	if imageDigestMirrorSets == nil && imageContentPolicies == nil {
+		return "", nil
+	}
+
+	// Build IDMS file from contents of the IDMS and ICSP mirror digests
+	contents, err := getIdmsContents(c.LogOut, imageDigestMirrorSets, imageContentPolicies)
+	if err != nil {
+		c.log("failure parsing imageDigestMirrorSet %s", err.Error())
+		return "", err
+	}
+
+	// create temp file which will be removed by caller
+	idmsFile, err := os.CreateTemp("", "idms-file")
+	if err != nil {
+		return "", err
+	}
+
+	c.log("Building IDMS file with contents %s", contents)
+	if _, err := idmsFile.Write(contents); err != nil {
+		idmsFile.Close()
+		os.Remove(idmsFile.Name())
+		return "", err
+	}
+	idmsFile.Close()
+	return idmsFile.Name(), nil
 }
--- a/pkg/cli/admin/nodeimage/basenodeimagecommand.go
+++ b/pkg/cli/admin/nodeimage/basenodeimagecommand.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/kubectl/pkg/cmd/exec"
+	"sigs.k8s.io/yaml"
 
 	ocpv1 "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/nodeimage/basenodeimagecommand.go
```diff
@@ -87,7 +88,28 @@ func (c *BaseNodeImageCommand) getNodeJoinerPullSpec(ctx context.Context) error
...
+					// Found pullSpec from configMap
+					c.log("installer pullspec obtained from installer-images configMap %s", installer)
+					c.nodeJoinerImage = installer
+					return nil
+				}
+			}
+		}
+	}
+
+	// If cannot obtain installer image from configMap, get it from the releaseImage
```
andfasano, line 112: Hi @bfournie that part looks fine to me, for when the configmap is available.

But for backward compatibility (oc has also a skew support of -1/+1), or in general for when it's not available, the below fallback mechanism should be fixed to use the mirror info if present in the cluster.

This means that `opts.IDMSFile` (or `opts.ICSPFile`) should be set in case such info were found. For that the existing node-joiner could be reused https://github.com/openshift/installer/blob/7867e67e84f38f33016c1506764770982cfee503/pkg/asset/agent/joiner/clusterinfo.go#L428

zaneb: I'm reluctant to modify the code below when it won't be exercised in any tests.

For existing clusters that can't be upgraded, there is a workaround available: manually create the ConfigMap. It's not perfect but it's not nothing.

zaneb, line 112: Can we note in the comment what release we're able to delete this workaround in? (I think it's 4.20 right?)

pkg/cli/admin/nodeimage/basenodeimagecommand.go
```diff
@@ -87,7 +88,28 @@ func (c *BaseNodeImageCommand) getNodeJoinerPullSpec(ctx context.Context) error
...
+				installer, exists := imageMap["installer"]
+				if exists {
+					// Found pullSpec from configMap
+					c.log("installer pullspec obtained from installer-images configMap %s", installer)
+					c.nodeJoinerImage = installer
+					return nil
+				}
+			}
+		}
+	}
```
zaneb, line 110: If there's an error other than NotFound, we should at least log it.
And if it _is_ NotFound, we should probably log that we're trying the registry.

pkg/cli/admin/nodeimage/basenodeimagecommand.go
andfasano: > I'm reluctant to modify the code below when it won't be exercised in any tests.
This is a good point, and on the CI side it could be tested by performing a disconnected installation (plus add node) in a fully disconnected environment (there's a pending [PR](https://github.com/openshift-metal3/dev-scripts/pull/1742) for enabling that).

On the QE side I think the new disconnected scenario will be covered after the initial discussion (IIRC it was also reproduced)? cc @mhanss @bmanzari  @zniu1011 for confirmation

> For existing clusters that can't be upgraded, there is a workaround available: manually create the ConfigMap. It's not perfect but it's not nothing.

Without the fix all the combinations of clusters/oc between 4.17 - 4.19 will be affected (oc must provide a version skew support of -1/+1), and there will be a non-zeo cost in troubleshooting them / applying the workaround. Given that the fix itself it's pretty straightforward (get the mirror info if present and set them for the request), it seems to me a relevant simplification - with the implicit assumption to have full test coverage for that.

bfournie: Added comment

pkg/cli/admin/nodeimage/basenodeimagecommand.go
bfournie: Added logging.

pkg/cli/admin/nodeimage/basenodeimagecommand.go
bfournie: Added IDMS/ICSP support to handle backwards compatibility with installer versions that don't yet support the installer-images configMap.

zaneb: > on the CI side it could be tested by performing a disconnected installation (plus add node) in a fully disconnected environment

We'd have to do that _and_ test the latest `oc` against 4.17/4.18 clusters to exercise the fallback.
And even that would only give us coverage until we backport the installer change. After that we'd have to test against old .z releases.

pkg/cli/admin/nodeimage/basenodeimagecommand.go
```diff
@@ -87,9 +89,48 @@ func (c *BaseNodeImageCommand) getNodeJoinerPullSpec(ctx context.Context) error
...
+		c.log("configMap containing installer-images is not available, trying to get image from registry")
+	}
+
+	// If cannot obtain installer image from configMap, get it from the releaseImage.
+	// Note that after OCP 4.20, this should not be necessary since the configMap was added to installer in 4.19.
 	opts := ocrelease.NewInfoOptions(c.IOStreams)
 	opts.SecurityOptions = c.SecurityOptions
+	opts.IDMSFile = c.getIdmsFile(ctx)
+	if opts.IDMSFile == "" {
+		opts.ICSPFile = c.getIcspFile(ctx)
```
zaneb, line 123: Clusters are allowed to have both ICSP and IDMS objects. Ideally we would build a single IDMSFile from both sources rather than separate files from each and picking one.

pkg/cli/admin/nodeimage/basenodeimagecommand.go
```diff
@@ -87,9 +89,48 @@ func (c *BaseNodeImageCommand) getNodeJoinerPullSpec(ctx context.Context) error
...
+					// Found pullSpec from configMap
+					c.log("installer pullspec obtained from installer-images configMap %s", installer)
+					c.nodeJoinerImage = installer
+					return nil
+				}
+			}
+		}
+	} else if !kapierrors.IsNotFound(err) {
+		c.log("failure retrieving installer-images configMap %s", err.Error())
+	} else {
```
zaneb, line 113: On reflection, maybe just put this log outside of an `else` block, since there are other ways to get to the fallback code (e.g. ConfigMap has been incorrectly manually created on an old version, so it exists but does not have the expected keys).

pkg/cli/admin/nodeimage/basenodeimagecommand.go
bfournie: Are they both supported **at the same time**? I know there can be ICSP's after an upgrade but I thought it was [recommended](https://docs.redhat.com/en/documentation/openshift_container_platform/4.14/html/images/image-configuration#images-configuration-registry-mirror_image-configuration) to convert that to an IDMS.

I know with ABI only IDMS are created in the cluster, since 4.17 I believe - https://issues.redhat.com//browse/MGMT-14159. 

We could combine the mirrors but I assume that would just create duplicates, I don't think the intent is to split the mirroring config across the 2 objects.

pkg/cli/admin/nodeimage/basenodeimagecommand.go
bfournie: Sure, I'll move it so we log just one failure retrieving installer-images configMap.

pkg/cli/admin/nodeimage/basenodeimagecommand.go
zaneb: > Are they both supported **at the same time**?

Looks like yes unless something has changed: https://issues.redhat.com/browse/OCPNODE-1771

From your doc link:

>  For new clusters, you can use IDMS, ITMS, and ICSP CRs objects as desired. However, using IDMS and ITMS is recommended.
>
> If you upgraded a cluster, any existing ICSP objects remain stable, and both IDMS and ICSP objects are supported.

bfournie: OK, I was under the assumption that the ICSP was removed after it was converted to IDMS. If they're both there I'll concatenate all of the mirrors into one IDMS.

pkg/cli/admin/nodeimage/basenodeimagecommand.go
```diff
@@ -409,3 +441,37 @@ func (o *BaseNodeImageCommand) runNodeJoinerPod(ctx context.Context, tasks []fun
 	}
 	return nil
 }
+
+func (c *BaseNodeImageCommand) getIdmsFile(ctx context.Context) string {
+	imageDigestMirrorSets, idmsErr := c.ConfigClient.ConfigV1().ImageDigestMirrorSets().List(ctx, metav1.ListOptions{})
+	imageContentPolicies, icspErr := c.ConfigClient.ConfigV1().ImageContentPolicies().List(ctx, metav1.ListOptions{})
+	if idmsErr != nil && icspErr != nil {
+		return ""
```
andfasano, line 449: Only if `errors.IsNotFound()` for both the errors is true then it's ok to return an empty string, otherwise it may be a genuine error and should be reported as usual

pkg/cli/admin/nodeimage/basenodeimagecommand.go
```diff
@@ -409,3 +441,37 @@ func (o *BaseNodeImageCommand) runNodeJoinerPod(ctx context.Context, tasks []fun
...
+		contents, err = getIdmsContents(nil, imageContentPolicies)
+	}
+	if err != nil {
+		c.log("failure parsing imageDigestMirrorSet %s", err.Error())
+		return ""
+	}
+
+	// create temp file which will be removed by caller
+	idmsFile, err := os.CreateTemp("", "idms-file")
+	c.log("Building IDMS file with contents %s", contents)
```
andfasano, line 469: This `err` seems to be ignored

pkg/cli/admin/nodeimage/basenodeimagecommand.go
```diff
@@ -409,3 +441,37 @@ func (o *BaseNodeImageCommand) runNodeJoinerPod(ctx context.Context, tasks []fun
...
+	if _, err := idmsFile.Write(contents); err != nil {
+		idmsFile.Close()
+		os.Remove(idmsFile.Name())
+		return ""
```
andfasano, line 473: In this case I'd expect

```suggestion
		return "", err
```

pkg/cli/admin/nodeimage/mirrorbuilder.go
```diff
@@ -0,0 +1,64 @@
...
+	if idmsMirrorSetList != nil {
+		for _, idms := range idmsMirrorSetList.Items {
+			for _, digestMirror := range idms.Spec.ImageDigestMirrors {
+				imageDigestSet.Spec.ImageDigestMirrors = append(imageDigestSet.Spec.ImageDigestMirrors, digestMirror)
+			}
+		}
+	}
+
+	// Add mirrors from ICSP
+	if icspMirrorSetList != nil {
```
andfasano, line 31: I know this has been already previously discussed - and I keep finding a little bit awkward chaining both the sources. Since ICSP is deprecated,I think that at least logging its usage could make it more evident

pkg/cli/admin/nodeimage/basenodeimagecommand.go
bfournie: Fixed

zaneb:
I assume you have tested this manually against a disconnected 4.18 cluster?

/lgtm

bfournie:
> I assume you have tested this manually against a disconnected 4.18 cluster?
> 
> /lgtm

I tested it by removing the configMap settings so the IDMS was passed in to the command.

PR #2007, opened 2025-04-15
OCPBUGS-52164: Add the generation of initrd.addrsize and generic.ins file for ABI Day2 Artifacts

Neeraj8418:
This PR introduces a conditional filter to include .ins and .addrsize PXE artifacts only when the target CPU architecture is s390x during ABI-based Day-2 node image creation using oc adm node-image create.

Changes:

Updated the create.go logic to check for s390x CPU architecture before including:
*.ins
*.addrsize

Testing
Tested by:
Running the oc adm node-image create for s390x and verified only s390x-relevant PXE files are generated.
Ensured no regression on other platforms — .ins and .addrsize are excluded when architecture is not s390x.

### DIFF
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -381,7 +381,7 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
 	}
 	if o.GeneratePXEFiles {
 		rsyncOptions.Source.Path = "/assets/boot-artifacts/"
-		rsyncOptions.RsyncInclude = []string{"*.img", "*.*vmlinuz", "*.ipxe"}
+		rsyncOptions.RsyncInclude = []string{"*.img", "*.*vmlinuz", "*.ipxe", "*.ins", "*.addrsize"}
 		logMessage = "Saving PXE artifacts to %s"
 	}
 	rsyncOptions.Strategy = o.copyStrategy(rsyncOptions)


pkg/cli/admin/nodeimage/create.go
```diff
@@ -381,7 +381,7 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
 	}
 	if o.GeneratePXEFiles {
 		rsyncOptions.Source.Path = "/assets/boot-artifacts/"
-		rsyncOptions.RsyncInclude = []string{"*.img", "*.*vmlinuz", "*.ipxe"}
+		rsyncOptions.RsyncInclude = []string{"*.img", "*.*vmlinuz", "*.ipxe", "*.ins", "*.addrsize"}
```
zaneb, line 384: @andfasano would it be valid to make it just download all non-invisible files from boot-artifacts?
This looks fine, but ideally there shouldn't be any need to co-ordinate changes between the installer and oc whenever the installer changes, since oc has to work with the next version of the installer.

```suggestion
		rsyncOptions.RsyncInclude = nil
		rsyncOptions.RsyncExclude = []string{".*"}
```

andfasano: (my previous comment seems to be lost)
That's something we recently fixed in https://github.com/openshift/oc/pull/1931, so copying everything may bring back unwanted stuff.

pkg/cli/admin/nodeimage/create.go
```diff
@@ -381,7 +381,8 @@ func (o *CreateOptions) copyArtifactsFromNodeJoinerPod() error {
 	}
 	if o.GeneratePXEFiles {
 		rsyncOptions.Source.Path = "/assets/boot-artifacts/"
-		rsyncOptions.RsyncInclude = []string{"*.img", "*.*vmlinuz", "*.ipxe"}
+		rsyncOptions.RsyncInclude = nil
+		rsyncOptions.RsyncExclude = []string{".*"}
```
andfasano, line 385: Also commented in the other thread, but this seems to gather too much stuff (as previously fixed in https://github.com/openshift/oc/pull/1931), did you maybe verify it? Thanks

Neeraj8418: Hi @andfasano , I just verified for s390x and that only copied the required images. No additional stuff were copied.
```
[root@m42lp51 day2]# ll
total 1079148
-rw-r--r--. 1 root root        149 Apr 15 14:23 node.s390x-generic.ins
-rw-r--r--. 1 root root         16 Apr 15 14:23 node.s390x-initrd.addrsize
-rw-r--r--. 1 root root   64771460 Apr 15 14:23 node.s390x-initrd.img
-rw-r--r--. 1 root root    9253349 Apr 15 14:23 node.s390x-kernel.img
-rw-r--r--. 1 root root 1031003648 Apr 15 14:23 node.s390x-rootfs.img
-rw-r--r--. 1 root root        712 Apr 14 16:32 nodes-config.yaml
```
Does this looks fine or shall i revert it back to previous changes?

andfasano: I'd feel more comfortable into having an explicit list of files (extensions) to be included, rather than an exclusion list that may catch in future unwanted files.

Neeraj8418: Ok sure @andfasano . I have forced pushed the changed to first commit now. This branch has an explicit list of files to be included.

zaneb:
/retitle OCPBUGS-52164: Add the generation of initrd.addrsize and generic.ins file for ABI Day2 Artifacts
/lgtm

pkg/cli/admin/nodeimage/create.go
zaneb: Personally I'd feel more comfortable knowing that we can change the installer in future, but [OCPBUGS-45311](https://issues.redhat.com/browse/OCPBUGS-45311) doesn't specify what extra files were previously being added so it is hard to evaluate, and we have already messed Neeraj around enough, so I LGTMd.

PR #2009, opened 2025-04-15
Fix debug command auto-completion

sudomakeinstall2:
Currently `oc debug` auto-completes to files. This commit adds auto-completion for this command. Auto-completion suggests pods and resources that are valid for the debug command.

Implemenation is based on `completions.doPodResourceCompletion` expect that it includes `nodes` instead of `services`.

examples:
```
$ oc get pods
NAME                         READY   STATUS    RESTARTS   AGE
apiserver-5f58c584c4-6wvnr   2/2     Running   0          18d
apiserver-5f58c584c4-dlcbk   2/2     Running   0          18d
apiserver-5f58c584c4-ljlc9   2/2     Running   0          18d

$ oc debug [TAB]
apiserver-5f58c584c4-6wvnr  daemonsets/                 nodes/
replicationcontrollers/
apiserver-5f58c584c4-dlcbk  deployments/                pods/
statefulsets/
apiserver-5f58c584c4-ljlc9  jobs/                       replicasets/

$ oc debug nodes/vcl01-hv9-ctlplane-[TAB]
nodes/vcl01-hv9-ctlplane-0.redhat.com
nodes/vcl01-hv9-ctlplane-2.redhat.com
nodes/vcl01-hv9-ctlplane-1.redhat.com
```


pkg/cli/debug/debug.go
```diff
@@ -1349,3 +1350,54 @@ func (o *DebugOptions) resolveImageStreamTag(namespace, name, tag string) (strin
 	}
 	return image, nil
 }
+
+// debugCompletionFunc Returns a completion function that completes:
+// 1- pod names that match the toComplete prefix
+// 2- resource types containing pods and nodes which match the toComplete prefix
+func debugCompletionFunc(f kcmdutil.Factory) func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
```
ardaguclu, line 1357: There are many built-in completion functions in kubectl and it is always better to use them. For example, upstream debug uses https://github.com/kubernetes/kubectl/blob/72b3a7e9b0e99ba874ffe7dd85f9dfd756239dd1/pkg/cmd/cmd.go#L393 and some of the oc commands use https://github.com/openshift/oc/blob/beeaf0904b6e9a7c22a73af9e92327f50a4d6918/pkg/cli/rollout/rollout.go#L83. I think, in our case we can use ResourceTypeAndNameCompletionFunc like kubectl debug.

pkg/cli/debug/debug.go
```diff
@@ -1349,3 +1350,54 @@ func (o *DebugOptions) resolveImageStreamTag(namespace, name, tag string) (strin
...
+			return comps, directive
+		}
+
+		directive = cobra.ShellCompDirectiveNoFileComp
+		slashIdx := strings.Index(toComplete, "/")
+		if slashIdx == -1 {
+			// Standard case, complete pod names
+			comps = completion.CompGetResource(f, "pod", toComplete)
+
+			validResources := []string{
```
ardaguclu, line 1371: This list is not fully complete. oc debug supports resources https://github.com/openshift/oc/blob/beeaf0904b6e9a7c22a73af9e92327f50a4d6918/pkg/cli/debug/debug.go#L1021 more than this list
/hold

pkg/cli/debug/debug.go
sudomakeinstall2: I tried this first but it doesn't really fit here. It fits more with `oc ACTION resource name` rather than `oc ACTION resource/name`.

For example 
```
$ oc debug [tab]
nodes pods deployments ...

$ oc debug nod[tab] -> oc debug nodes [cursor is placed after space]
```
but you can't use `$ oc debug nodes my-node`, oc debug only accepts `$ oc debug resource/name` 
`kubectl` defines `doPodResourceCompletion` for this reason.

pkg/cli/debug/debug.go
sudomakeinstall2: Added imagestreamimages, imagestreamtags and cronjobs.
I didn't add deploymentConfig since it is deprecated since 4.14.

I can add it if you think it should be here.

PR #2011, opened 2025-04-21
CORENET-5412: adm, inspect, ns: Collect UserDefinedNetwork CRs

ormergi:
OVN-Kubernetes (OVN-K) UserDefinedNetwork (UDN) is namespace-scope CRD that allows non-admin users create OVN-K user-defined networks w/o a cluster-admin intervention.

Collect UDN CRs to allow troubleshooting them.

### DIFF
--- a/pkg/cli/admin/inspect/namespace.go
+++ b/pkg/cli/admin/inspect/namespace.go
@@ -31,6 +31,7 @@ func namespaceResourcesToCollect() []schema.GroupResource {
 		{Resource: "poddisruptionbudgets"},
 		{Resource: "secrets"},
 		{Resource: "servicemonitors"},
+		{Resource: "userdefinednetworks"},
 	}
 }
 


pkg/cli/admin/inspect/namespace.go
```diff
@@ -31,6 +31,7 @@ func namespaceResourcesToCollect() []schema.GroupResource {
 		{Resource: "poddisruptionbudgets"},
 		{Resource: "secrets"},
 		{Resource: "servicemonitors"},
+		{Resource: "userdefinednetworks"},
```
ardaguclu, line 34: As seen in the list, this list is very conservative and must be short. The recommended path is to add in must-gather (iff the collected data is small, because we have reached to a point where must-gather collection size and duration is not ideal currently)

/hold

arghosh93: We anticipate that userdefinednetworks is gonna be used widely and multiple namespace may have a userdefinednetworks defined in it. With the lifting of featuregate for BGP, we anticipate more use of userdefinednetworks as main use case of BGP feature is to advertise UDNs.
routeadvertisements(main CRD to implement BGP) are cluster scoped resources and a PR for inclusion was accepted in the must-gather repository.
https://github.com/openshift/must-gather/pull/485

arghosh93:
@ardaguclu any update on this? userdefinednetworks CR definition is definitely gonna be needed to troubleshoot any issue related to it.

ardaguclu:
> @ardaguclu any update on this? userdefinednetworks CR definition is definitely gonna be needed to troubleshoot any issue related to it.

Do you only want to get the yaml definition of UserDefinedNetworks CR in all namespaces?. Will there be any RelatedObjects defined in this resource that will trigger another nested resource inspection?. How much additional data will be more collected in must-gather and increase the output size?.

Basically, I'm trying to assess that did you evaluate what are the consequence of adding this CR in here?. I'm sure that this is needed, but we can't add any resources that collects significant amount of data.

arghosh93:
> Do you only want to get the yaml definition of UserDefinedNetworks CR in all namespaces?

Yes

>  Will there be any RelatedObjects defined in this resource that will trigger another nested resource inspection?

OVNKubernetes creates a NetworkAttachmentDefinition corresponding to the userdefinednetwork CR but we dont need NetworkAttachmentDefinition. Even if e store both, I dont anticipate much resource consumption as both are lightweight.
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/crd/userdefinednetwork/v1/udn.go

> How much additional data will be more collected in must-gather and increase the output size?.

A single UDN can take up to 4 KB.

PR #2022, opened 2025-05-15
OCPBUGS-55220: Replace colon to hypen in the filename while extracting tar on Windows

ardaguclu:
Extracting a tar file that contains colon in the name fails on Windows hosts, as this character is special for Windows. This PR replaces colon with hypen during extraction, if the host is Windows.

There are a couple of command in oc relies on this functionality like must-gather.

### DIFF
--- a/pkg/helpers/source-to-image/tar/tar.go
+++ b/pkg/helpers/source-to-image/tar/tar.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ardaguclu:
/hold
until pre-merge tests are completed.

pkg/helpers/source-to-image/tar/tar.go
```diff
@@ -456,7 +457,15 @@ func (t *stiTar) extractLink(dir string, header *tar.Header, tarReader io.Reader
 }
 
 func (t *stiTar) extractFile(dir string, header *tar.Header, tarReader io.Reader) error {
-	path := filepath.Join(dir, header.Name)
+	fileName := header.Name
+	if runtime.GOOS == "windows" {
+		// ":" is special character on Windows. If file name contains ":",
+		// extraction will fail. In order to overcome this problem, we are replacing ":"
+		// to "-".
+		fileName = strings.Replace(fileName, ":", "-", -1)
```
ingvagabund, line 465: Would the same change make sense for non Windows cases too?

ardaguclu: I'd expect this to be a safe change if we enable for all hosts. But isn't `:` only special for Windows?

ingvagabund: It's better if there's unconditional change. Less maintenance and testing.

ardaguclu: There is similar conditions in rsync side https://github.com/openshift/oc/blob/8a88a4de03faf9923f41bcc06e4d68dacb4a543c/pkg/cli/rsync/pathspec.go#L32. I agree replacement without condition means less maintenantce and testing. However, In my opinion, keeping the windows condition here emphasizes the motivation of operation better, because on other hosts we don't have any reasons to perform this replacement.

simonregn:
Hi. I still face this problem on Windows using Git Bash. My impression is, that in Git Bash 
`if runtime.GOOS == "windows"`
will not evaluate to true

PR #2035, opened 2025-06-09
pkg/cli/admin/upgrade/recommend/alerts: Ambiguate opinions

wking:
There's a chance we might be asked to drop the `OC_ENABLE_CMD_UPGRADE_RECOMMEND_PRECHECK` feature gate that's currently in front of this logic.  If we do, this wording will make it extra clear that we are not saying "don't update until this alert resolves". We are saying "this alert is firing.  You should investigate and form your own opinion".

As a concrete example, the [OCPBUGS-54658][1] series of bugs could cause issues pulling images, you would address the issue by launching an update to a release with the bugfix.  If you needed to take any manual steps to harden the core cluster components (e.g. [configuring a local mirror][2] that was more responsive), that could happen before or during the update.

[1]: https://issues.redhat.com/browse/OCPBUGS-54658
[2]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/disconnected_environments/mirroring-in-disconnected-environments

### DIFF
--- a/pkg/cli/admin/upgrade/recommend/alerts.go
+++ b/pkg/cli/admin/upgrade/recommend/alerts.go
@@ -108,7 +108,7 @@ func (o *options) alerts(ctx context.Context) ([]acceptableCondition, error) {
 					Type:    fmt.Sprintf("recommended/CriticalAlerts/%s/%d", alertName, i),
 					Status:  metav1.ConditionFalse,
 					Reason:  fmt.Sprintf("Alert:%s", alert.State),
-					Message: fmt.Sprintf("%s alert %s %s, suggesting significant cluster issues worth investigating. %s", alert.Labels.Severity, alert.Labels.AlertName, alert.State, details),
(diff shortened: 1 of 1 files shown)


petr-muller:
I am not a fan of this TBH. While technically correct, I do not think it is helpful to try weasel out of responsibility in the output itself. I think `may` is sufficiently vague to not be watered down further with "may or may not", and trying to accommodate corner cases by sentences that won't apply in 99% cases (updating a cluster will _not_ solve an alert in most cases) IMO just adds noise. Attention is precious and programs should try to be helpful.

I would not oppose having some language like this in `--help` / usage message, and in docs, but the actual output should try to avoid being ambiguous, IMO.

wking:
> While technically correct, I do not think it is helpful to try weasel out of responsibility in the output itself.

Ok, I'll just close this.  We can always revisit in the unlikely event that someone misinterprets the existing wording.

PR #2042, opened 2025-06-18
OCPBUGS-57635: Revert "OCPBUGS-56691: Rely on overall available disk space of the mounted volume"

ardaguclu:
Reverts openshift/oc#2028


ardaguclu:
/jira refresh
@ingvagabund it seems that some CI jobs have not sufficient disk space so that must-gather fails to continue. We need to update the must-gather tests with higher volume-percentage. But until fixing the tests, I think it is better to revert this.

PR #2058, opened 2025-07-11
CNTRLPLANE-1054: Update k8s  dependencies to 1.33 with bumping the other dependencies

ardaguclu:
Basically, the expectation is to update the k8s dependencies to 1.33. However, the other dependencies are usually updated only when it is required.

This PR ambitiously tries to update the other libraries, whose don't bring breaking changes, to the latest.

For example, `gopkg.in/yaml.v3` is archived we should use k8s fork. `github.com/pkg/errors` is redundant.


ardaguclu:
This is far from being ready for review. But it is better to cc'ing now.
/cc @ingvagabund @tchap

ardaguclu:
None of the failures are related and they all will be fixed by different PRs. This PR is ready to be reviewed.

ardaguclu:
Thanks for the tag. Due to the change in observe command, I'd also like to get review from Jan as well. If three of us are fine on the change, I'll be confident :).

ardaguclu:
It is unrelated
/test e2e-metal-ipi-ovn-ipv6

tchap:
LGTM

/unhold

PR #2072, opened 2025-08-08
WIP: Remove several tools that pull in too many dependencies

sdodson:
Since the tools image is a base for a few other images that get deployed as runtime components we should slim it down to both reduce its size and CVE exposure. Reduces package set by 133 and image size by ~ 200MiB.


ardaguclu:
> Since the tools image is a base for a few other images that get deployed as runtime components we should slim it down to both reduce its size and CVE exposure. Reduces package set by 133 and image size by ~ 200MiB.

I think, images should be based on `cli` instead of `tools`.

sdodson:
I need to come back to this once we finish other changes.

PR #2073, opened 2025-08-10
OCPBUGS-60281: Fix panic in oc describe templateinstance when Secret is nil

mhrivnak:
The TemplateInstanceDescriber.DescribeTemplateInstance function was attempting to access templateInstance.Spec.Secret.Name without checking if Secret is nil first. Since Secret is an optional pointer field, this caused a nil pointer dereference panic when describing template instances that don't have a secret specified.

Added a nil check before accessing Secret.Name and provide appropriate fallback behavior by displaying "No secret specified for parameters" when no secret is present.

### DIFF
--- a/pkg/helpers/describe/describer.go
+++ b/pkg/helpers/describe/describer.go
@@ -1348,7 +1348,12 @@ func (d *TemplateInstanceDescriber) DescribeTemplateInstance(templateInstance *t
 		d.DescribeObjects(templateInstance.Status.Objects, out)
 		out.Write([]byte("\n"))
 		out.Flush()
-		d.DescribeParameters(templateInstance.Spec.Template, namespace, templateInstance.Spec.Secret.Name, out)
+		if templateInstance.Spec.Secret != nil {
+			d.DescribeParameters(templateInstance.Spec.Template, namespace, templateInstance.Spec.Secret.Name, out)
+		} else {
+			formatString(out, "Parameters", " ")
+			fmt.Fprint(out, "    No secret specified for parameters.")
+		}
 		out.Write([]byte("\n"))
 		out.Flush()
 		return nil
--- a/pkg/helpers/describe/describer_test.go
+++ b/pkg/helpers/describe/describer_test.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/client-go/kubernetes/fake"
 	kubernetesscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"k8s.io/kubectl/pkg/describe"
 )
 
 type describeClient struct {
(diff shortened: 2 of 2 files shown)


pkg/helpers/describe/describer.go
```diff
@@ -1348,7 +1348,12 @@ func (d *TemplateInstanceDescriber) DescribeTemplateInstance(templateInstance *t
 		d.DescribeObjects(templateInstance.Status.Objects, out)
 		out.Write([]byte("\n"))
 		out.Flush()
-		d.DescribeParameters(templateInstance.Spec.Template, namespace, templateInstance.Spec.Secret.Name, out)
+		if templateInstance.Spec.Secret != nil {
+			d.DescribeParameters(templateInstance.Spec.Template, namespace, templateInstance.Spec.Secret.Name, out)
+		} else {
+			formatString(out, "Parameters", " ")
```
ardaguclu, line 1354: Would you please paste an example output to see how this fallback will be shown to users?.

tchap:
Thanks for the contribution. I think that it would also be nice to add a test case that covers this.

pkg/helpers/describe/describer.go
mhrivnak:
```
$ ./oc describe templateinstance my-configmap-instance
Name:		my-configmap-instance
Namespace:	foo
Created:	About an hour ago
Labels:		<none>
Annotations:	kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"template.openshift.io/v1","kind":"TemplateInstance","metadata":{"annotations":{},"name":"my-configmap-instance","namespace":"foo"},"spec":{"template":{"metadata":{"name":"my-configmap-template"}}}}
		

Conditions:		 
    Type:		Ready
    Status:		True
    LastTransitionTime:	2025-08-10 15:51:40 -0400 EDT
    Reason:		Created
    Message:		<none>


Objects:	 

Parameters:	 
    No secret specified for parameters.
```

mhrivnak:
> Thanks for the contribution. I think that it would also be nice to add a test case that covers this.

Tests added.

ardaguclu:
Before starting the code review, we need to see the CI jobs are green (apart from ci/prow/security)

ardaguclu:
@mhrivnak as far as I can see you just need to fix the `verify` failure.

mhrivnak:
> @mhrivnak as far as I can see you just need to fix the `verify` failure.

Any hints at what's going wrong, and what I could do to fix it? It kind of looks like the infrastructure of the test is timing out, but it's hard as a new contributor to make sense of the 300+ dense lines of log output from that test.

ardaguclu:
> > @mhrivnak as far as I can see you just need to fix the `verify` failure.
> 
> Any hints at what's going wrong, and what I could do to fix it? It kind of looks like the infrastructure of the test is timing out, but it's hard as a new contributor to make sense of the 300+ dense lines of log output from that test.

From what I understand from https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/2073/pull-ci-openshift-oc-main-verify/1955518728024100864, you just need to run `make update-gofmt`?

tchap:
@mhrivnak It seem you need to run `go fmt` or something: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/2073/pull-ci-openshift-oc-main-verify/1955518728024100864#1:build-log.txt%3A27

ardaguclu:
This looks good to me. Thank you
/lgtm

PR #2074, opened 2025-08-11
OCPNODE-3557: Fixes oc debug managed by label when keep labels option is not used

ngopalak-redhat:
This PR ensures the required `managed-by` label is correctly applied to debug pods created via `oc debug node/<nodename>` by default.

The previous pull request, [#2055](https://github.com/openshift/oc/pull/2055), introduced a label for debug pods. However, testing revealed that this label was not being set by default unless the keepLabels flag was used. This fix ensures the label is always present for all debug pods, which is the intended behavior.

Note to reviewer: The new unit test in debug_test.go was generated by Claude. The core logic for this fix is in debug.go does not use AI generated code.

Test:
From the invocation:
```
./oc debug node/ip-10-0-79-147.ec2.internal
Starting pod/ip-10-0-79-147ec2internal-debug-8l2bd ...
To use host binaries, run `chroot /host`. Instead, if you need to access host namespaces, run `nsenter -a -t 1`.
Pod IP: 10.0.79.147
sh-5.1#
```
From the describe output:
```
oc describe po ip-10-0-79-147ec2internal-debug-8l2bd                                         ✔  admin 󱃾  10:24:04
Name:                 ip-10-0-79-147ec2internal-debug-8l2bd
Namespace:            default
Priority:             1000000000
Priority Class Name:  openshift-user-critical
Service Account:      default
Node:                 ip-10-0-79-147.ec2.internal/10.0.79.147
Start Time:           Mon, 11 Aug 2025 10:23:54 +0530
Labels:               debug.openshift.io/managed-by=oc-debug
```


ngopalak-redhat:
cc: @BhargaviGudi @haircommander

pkg/cli/debug/debug.go
```diff
@@ -900,13 +901,8 @@ func (o *DebugOptions) transformPodForDebug(annotations map[string]string) (*cor
 	for k, v := range annotations {
 		pod.Annotations[k] = v
 	}
-	if o.KeepLabels {
-		if pod.Labels == nil {
-			pod.Labels = make(map[string]string)
-		}
-	} else {
-		pod.Labels = map[string]string{}
```
ardaguclu, line 908: Instead of traversing all the labels in a dedicated function, can't we simply add this code portion in here (of course with some refactoring such as defining the label name as const, etc.);

```go
		if _, ok := pod.Labels["debug.openshift.io/managed-by"]; ok {
			pod.Labels = map[string]string{}
			pod.Labels["debug.openshift.io/managed-b"] = "oc-debug"
		} else {
			pod.Labels = map[string]string{}
		}
```

pkg/cli/debug/debug_test.go
```diff
@@ -201,3 +201,68 @@ func TestGetNamespace(t *testing.T) {
 		}
 	}
 }
+
+// Generated by Claude
+func TestCreateLabelMap(t *testing.T) {
```
ardaguclu, line 206: What about adding a e2e test instead of unit test in here https://github.com/openshift/origin/blob/7e8f07d709c533890f1c2f29fddea27fe4acffae/test/extended/cli/debug.go#L25.

This test can initiate `oc debug node` without attaching to the terminal. After that we can check that the required label exists or not. What do you think?.

pkg/cli/debug/debug.go
ngopalak-redhat: We had discussed before to limit this label to only node debug and not for all the debug: https://github.com/openshift/oc/pull/2055#discussion_r2180486065. 
Hence its added while the node debug is handled and this code added in the PR just retains the label.

ardaguclu: My suggested code portion achieves the same

ngopalak-redhat: Thanks a lot. I have adopted the suggested code

pkg/cli/debug/debug_test.go
ngopalak-redhat: Thanks. I have added the test https://github.com/openshift/origin/pull/30093 . Please review that also

PR #2085, opened 2025-08-29
OCPBUGS-60989: adm release info: Print commit url if no associated PR number

ardaguclu:
`oc adm release info --changelog` discards the commits, if no associated pull request number is found. This is problematic because discarded commits makes it difficult to find the regression in the changelogs to investigate.

So that this PR prints commit message (with commit url) to at least show the change in the output. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Release generation now correctly handles merge commits without an associated pull request. Such commits are included in the output and no longer cause parsing errors or get skipped, resulting in more complete release notes and smoother runs when direct merges or backports occur. Commits with valid PR references continue to be processed as before.

### DIFF
--- a/pkg/cli/admin/release/git.go
+++ b/pkg/cli/admin/release/git.go
@@ -298,13 +298,13 @@ func mergeLogForRepo(g gitInterface, repo string, from, to string) ([]MergeCommi
 			m = squashRePR.FindStringSubmatch(mergeMsg)
 		}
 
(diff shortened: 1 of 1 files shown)


ardaguclu:
I'm closing this PR in favor of https://github.com/openshift/oc/pull/2086 (which is better approach)
/close

PR #2090, opened 2025-09-09
OCPBUGS-52936: oc adm policy: Only initialize UserClient if built-in OAuth is enabled

ardaguclu:
`UserClient` is strictly tied to `User` object in OpenShift clusters. However, `User` resource is not served, if the external OIDC is configured instead. 

That causes improper HTTP calls to the server against a resource that already does not exist. 

This PR uses discovery client to decide whether external OIDC is configured or not in the cluster by checking the existence of `user.openshift.io/v1` GroupVersion. And only initializes `UserClient`, if discovery client serves `user.openshift.io/v1`

### DIFF
--- a/pkg/cli/admin/policy/modify_roles.go
+++ b/pkg/cli/admin/policy/modify_roles.go
@@ -355,7 +355,7 @@ func (o *RoleModificationOptions) innerComplete(f kcmdutil.Factory, cmd *cobra.C
 		return err
 	}
 
-	found := false
+	userAPIAvailable := false
 	groupList, err := o.DiscoveryClient.ServerGroups()
 	if discovery.IsGroupDiscoveryFailedError(err) {
 		// proceed with partial results; treat missing groups as "not found"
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/policy/modify_roles.go
```diff
@@ -345,19 +346,40 @@ func (o *RoleModificationOptions) innerComplete(f kcmdutil.Factory, cmd *cobra.C
...
 		return err
 	}
-	o.UserClient, err = userv1client.NewForConfig(clientConfig)
+	o.RbacClient, err = rbacv1client.NewForConfig(clientConfig)
 	if err != nil {
 		return err
 	}
-	o.ServiceAccountClient, err = corev1client.NewForConfig(clientConfig)
+
+	found := false
```
liouk: nit -- suggested rename:
```suggestion
	userAPIAvailable := false
```

ardaguclu: Good idea. Updated.

ardaguclu:
> Could you also add a unit test proving that when the user API doesn't exist, the `o.UserClient` is nil? (or something along these lines)

@liouk thank you for the review. I've added unit tests to verify the behavior. Could you PTAL?.

PR #2092, opened 2025-09-09
NO-JIRA: Add some warnings into excluded list

ardaguclu:
Although we fixed these failures, snyk started raising warning to fix more. However, fixing these requires some changes that are prone to regression. For now, I'll enrich excluded list for these with appropriate comments. 

Thanks to that we'll have a clear signal of snyk failures to detect any vulnerabilities. Otherwise, it is perma-failing.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,6 +5,10 @@ exclude:
   global:
     - vendor/**
     - "**/*_test.go"
+    - pkg/cli/admin/internal/codesign/machoresign.go # the paths are fed by oc
+    - pkg/cli/admin/release/new.go # we sanitize the directory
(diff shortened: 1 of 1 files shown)


tchap:
Yeah, I checked these errors in the past and they are actually false positives. From what I found during my not exactly deep review, the paths could not actually be misused.

Let's just wait for the tests...

tchap:
Ah, I can see the test is passing,

/lgtm

ardaguclu:
Thank you

ardaguclu:
This can't cause any issues
/verified by @ardaguclu

PR #2095, opened 2025-09-11
OCPBUGS-61491: Update code to retrieve ICSP instead of ICP

mhanss:
This PR updates the code to retrieve ImageContentSourcePolicy (ICSP) instead of ImageContentPolicies (ICP) for mirror registries.


andfasano:
Related to https://github.com/openshift/installer/pull/9933

vendor/github.com/openshift/client-go/operator/applyconfigurations/utils.go
```diff
@@ -0,0 +1,462 @@
+// Code generated by applyconfiguration-gen. DO NOT EDIT.
+
+package applyconfigurations
```
ardaguclu, line 3: I'm not sure about why do we need changes in vendor directory without even updating the go.mod.

Could you please update the PR;
* First commit go.mod changes and why do we need it?. I think, overall we shouldn't update go.mod
* Second commit "go mod vendor". 
* Third commit required changes in cli/admin/nodeimage package.

/hold

ardaguclu:
Overall I don't think we need updates on vendor directory. I'd expect that the changes in this PR only modifies the nodeimage directory.

vendor/github.com/openshift/client-go/operator/applyconfigurations/utils.go
mhanss: Since github.com/openshift/client-go was already in go.mod, no update was needed. I’ve added two commits: one for the fake clientset and another for the changes needed for the node image package.

ardaguclu: Thank you. That makes sense now. Please ping me for another round of review when the CI passes
/hold cancel

ardaguclu:
/lgtm cancel
I think, it is better to get lgtm from the respective team

mhanss:
> /lgtm cancel I think, it is better to get lgtm from the respective team

@ardaguclu It seems the lgtm cancel has also removed the approved label. Could you please add it again?

PR #2109, opened 2025-09-30
OCPBUGS-62445: prevent crash when rootDeviceHints format is invalid

mhanss:
This PR adds validation for the --root-device-hint flag in the Create Node Image command.
Previously, providing an invalid format would cause the command to crash with a stack trace.

### DIFF
--- a/pkg/cli/admin/nodeimage/create.go
+++ b/pkg/cli/admin/nodeimage/create.go
@@ -635,6 +635,9 @@ func (o *CreateOptions) createConfigFileFromFlags() ([]byte, error) {
 	}
 	if o.SingleNodeOpts.RootDeviceHints != "" {
 		parts := strings.SplitN(o.SingleNodeOpts.RootDeviceHints, ":", 2)
+		if len(parts) != 2 {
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/nodeimage/create.go
```diff
@@ -635,6 +635,9 @@ func (o *CreateOptions) createConfigFileFromFlags() ([]byte, error) {
 	}
 	if o.SingleNodeOpts.RootDeviceHints != "" {
 		parts := strings.SplitN(o.SingleNodeOpts.RootDeviceHints, ":", 2)
+		if len(parts) != 2 {
```
andfasano, line 638: In general we don't want to duplicate the node-joiner logic here, since the oc command must be as much as possible a simple wrapper. So, if there is an issue it must be fixed on the node-joiner side, and reuse the already existing workflow for reporting back the error to the user

zaneb: The format of the CLI arg is effectively already defined here (on the previous line) isn't it? And node-joiner uses a different format (proper yaml).

andfasano: That's perfectly right, I misread the original issue, this is an `oc` interface and not related to node-joiner

/approve

PR #2113, opened 2025-10-02
NO-ISSUE: pkg/cli/admin/upgrade/recommend: Drop obsolete precheckEnabled knob

wking:
Since 8bdac163b5 (#2088), this knob has always been enabled. Save ourselves a no longer necessary level of indentation by removing it completely.

### DIFF
--- a/pkg/cli/admin/upgrade/recommend/recommend.go
+++ b/pkg/cli/admin/upgrade/recommend/recommend.go
@@ -76,7 +76,6 @@ type options struct {
 
 	mockData             mockData
 	showOutdatedReleases bool
-	precheckEnabled      bool
 
 	// quiet configures the verbosity of output.  When 'quiet' is true and 'version' is set, only print unaccepted issue names.
 	quiet bool
@@ -134,8 +133,6 @@ func (o *options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string
 		}
 	}
 
-	o.precheckEnabled = true
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/upgrade/recommend/recommend.go
```diff
@@ -188,47 +185,45 @@ func (o *options) Run(ctx context.Context) error {
 		issues.Insert(string(configv1.OperatorProgressing))
 	}
 
-	if o.precheckEnabled {
-		conditions, err := o.precheck(ctx)
```
wking, line 192: GitHub doesn't seem to have a knob like Git's `-w` / `--ignore-all-space`, but here's the diff rendered that way, to show that I'm just dropping that level of indentation:

```console
$ git --no-pager show --oneline -w
cb28d2778 (HEAD -> remove-obsolete-precheckEnabled, wking/remove-obsolete-precheckEnabled) pkg/cli/admin/upgrade/recommend: Drop obsolete precheckEnabled knob
... 
@@ -188,7 +185,6 @@ func (o *options) Run(ctx context.Context) error {
                issues.Insert(string(configv1.OperatorProgressing))
        }
 
-       if o.precheckEnabled {
        conditions, err := o.precheck(ctx)
        if err != nil {
                if !o.quiet {
@@ -232,7 +228,6 @@ func (o *options) Run(ctx context.Context) error {
                        }
                }
        }
-       }
 
        if c := findClusterOperatorStatusCondition(cv.Status.Conditions, configv1.RetrievedUpdates); c != nil && c.Status != configv1.ConditionTrue {
                if !o.quiet {
```

tchap:
/lgtm

Thanks!

pkg/cli/admin/upgrade/recommend/recommend.go
petr-muller: Check out the "Hide whitespace" here

Image

wking:
/verified by @wking based on no functional changes and green presubmits

PR #2114, opened 2025-10-03
NO-JIRA: Migrate to aws-sdk-go-v2 from archived aws-sdk-go

ardaguclu:
oc depends on [aws-sdk-go](https://github.com/openshift/oc/blob/a25e854ffddffcf2f8c1bc99b10afd1b34b8e0ca/go.mod#L13) for s3 based operations  (these are image related operations, such as mirroring to image hosted in a s3 bucket with `s3://` prefix url). 

However, https://github.com/aws/aws-sdk-go package has been archived. It is recommended to migrate to https://github.com/aws/aws-sdk-go-v2. But this change brings about breaking changes and requires to reimplement the s3.go in oc.

This PR migrates to aws-sdk-go-v2 by applying the required changes. We need to be careful to not introduce any regressions. This PR should keep the functionality as is.

### DIFF
--- a/pkg/cli/image/imagesource/s3.go
+++ b/pkg/cli/image/imagesource/s3.go
@@ -188,7 +188,7 @@ func (r *s3Repository) attemptCopy(id string, bucket, key string) bool {
 			sourceKey = path.Join(copyFrom, id)
 		}
 		_, err := r.s3.CopyObject(r.ctx, &s3.CopyObjectInput{
-			CopySource: aws.String(sourceKey),
+			CopySource: aws.String(url.QueryEscape(sourceKey)),
 			Bucket:     aws.String(bucket),
 			Key:        aws.String(key),
 		})
@@ -283,7 +283,7 @@ func (s *s3ManifestService) Put(ctx context.Context, manifest distribution.Manif
 		if _, err := s.r.s3.CopyObject(s.r.ctx, &s3.CopyObjectInput{
 			Bucket:      aws.String(s.r.bucket),
 			ContentType: aws.String(mediaType),
-			CopySource:  aws.String(path.Join(s.r.bucket, blob)),
+			CopySource:  aws.String(path.Join(s.r.bucket, url.QueryEscape(blob))),
 			Key:         aws.String(fmt.Sprintf("/v2/%s/manifests/%s", s.r.repoName, tag)),
 		}); err != nil {
 			return "", err


pkg/cli/image/imagesource/s3.go
```diff
@@ -34,76 +34,73 @@ type s3Driver struct {
...
 
 type s3CredentialStore struct {
-	store     auth.CredentialStore
-	url       *url.URL
-	retrieved bool
+	store auth.CredentialStore
+	url   *url.URL
 }
 
-func (s *s3CredentialStore) IsExpired() bool { return !s.retrieved }
```
ardaguclu, line 46: IsExpired was removed from the interface in v2.

ardaguclu:
/hold
until pre-merge tests are finished.

pkg/cli/image/imagesource/s3.go
```diff
@@ -34,76 +34,69 @@ type s3Driver struct {
...
 		return nil, err
 	}
-	s3obj = s3.New(s)
+
+	s3Client := s3.NewFromConfig(cfg, func(o *s3.Options) {
+		if insecure {
+			o.EndpointOptions.DisableHTTPS = true
+		}
+		if d.UserAgent != "" {
+			o.APIOptions = append(o.APIOptions, awsmiddleware.AddUserAgentKeyValue("oc", d.UserAgent))
```
ardaguclu: According to the CodeRabbit, we can inject custom user agent by using aws sdk middleware. So that we don't need to use custom HttpClient.

tchap: Shouldn't this use just `AddKey`? The original implementation uses `d.UserAgent` as the user agent string, there is no prefix added to that.

tchap: I mean `AddUserAgentKey`

ardaguclu: Thanks for review. Nice catch. I'll update with this

pkg/cli/image/imagesource/s3.go
```diff
@@ -34,76 +34,73 @@ type s3Driver struct {
...
 
-	awsConfig := aws.NewConfig()
+	ctx := context.Background()
 
-	var creds *credentials.Credentials
-	creds = credentials.NewChainCredentials([]credentials.Provider{
-		&s3CredentialStore{store: d.Creds, url: securityDomain},
-		&credentials.EnvProvider{},
-		&credentials.SharedCredentialsProvider{},
-	})
```
tchap, line 77: I see that both env and shared credentials are handled now in `config.LoadDefaultConfig`, so this should be ok. Just a note to myself since it may seem removed.

pkg/cli/image/imagesource/s3.go
```diff
@@ -34,76 +34,73 @@ type s3Driver struct {
...
 	switch {
 	case klog.V(10).Enabled():
-		awsConfig.WithLogLevel(aws.LogDebugWithHTTPBody | aws.LogDebugWithRequestErrors | aws.LogDebugWithSigning)
+		configOpts = append(configOpts, config.WithClientLogMode(aws.LogSigning|aws.LogRetries|aws.LogRequest|aws.LogResponse))
 	case klog.V(8).Enabled():
-		awsConfig.WithLogLevel(aws.LogDebugWithRequestErrors)
+		configOpts = append(configOpts, config.WithClientLogMode(aws.LogRetries|aws.LogRequest))
 	case klog.V(6).Enabled():
-		awsConfig.WithLogLevel(aws.LogDebug)
+		configOpts = append(configOpts, config.WithClientLogMode(aws.LogRetries))
```
tchap: Does this map to the former logging easily? It seem to me v6 is usually already super debug and you would just log everything there. So shouldn't we just do, log everything on v6 or higher?

pkg/cli/image/imagesource/s3.go
ardaguclu: Updated.

pkg/cli/image/imagesource/s3.go
ardaguclu: I was under the same impression that this fallback mechanism had been removed under `LoadDefaultConfig`.

pkg/cli/image/imagesource/s3.go
ardaguclu: Since log types in aws has changed, it is difficult to keep one to one mapping. But I updated the code to follow semantically same principal. It logs responseBody in verbosity 10. It logs response and request in verbosity 8. It logs request. 

Please let me know your thoughts about this?.

pkg/cli/image/imagesource/s3.go
```diff
@@ -208,19 +206,20 @@ func (r *s3Repository) attemptCopy(id string, bucket, key string) bool {
...
 }
 
-func (r *s3Repository) conditionalUpload(input *s3manager.UploadInput, id string) error {
+func (r *s3Repository) conditionalUpload(input *s3.PutObjectInput, id string) error {
 	if r.attemptCopy(id, *input.Bucket, *input.Key) {
 		return nil
 	}
-	_, err := s3manager.NewUploaderWithClient(r.s3).Upload(input)
+	uploader := manager.NewUploader(r.s3)
+	_, err := uploader.Upload(r.ctx, input)
```
tchap, line 214: NIT: I think that this reads better as `uploader := manager.NewUploader(r.s3).Upload(r.ctx, input)`
The same thing on line 412...

pkg/cli/image/imagesource/s3.go
tchap: I guess it's fine then. Although I don't know why it's split like that. But let's keep it if it was like that.

ardaguclu: I believe the reason is to follow k8s standards. In k8s, verbosity 10 logs everything but verbosity 6 usually does not contain any response body, etc.

pkg/cli/image/imagesource/s3.go
ardaguclu: That would be nicer. If you'll have blocker comments, I'll update these too. Otherwise, we can leave it as is.

pkg/cli/image/imagesource/s3.go
tchap: Ok, thanks for the explanation.

pkg/cli/image/imagesource/s3.go
tchap: I guess it's ok then.

ardaguclu:
Thanks again for review. I'm unsure we should do pre-merge tests or not. Probably CI won't test these bits.

pkg/cli/image/imagesource/s3.go
```diff
@@ -34,76 +34,69 @@ type s3Driver struct {
...
-	})
+	ctx := context.Background()
 
-	awsConfig.WithCredentials(creds)
-	awsConfig.WithRegion(region)
-	awsConfig.WithDisableSSL(insecure)
+	configOpts := []func(*config.LoadOptions) error{
+		config.WithRegion(region),
+		config.WithCredentialsProvider(&s3CredentialStore{store: d.Creds, url: securityDomain}),
+	}
```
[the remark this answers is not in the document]
tchap: @ardaguclu Have you checked the CodeRabbit input BTW? It's mentioning some potential issues...

tchap:
/lgtm cancel

until we research the review bot comment.

pkg/cli/image/imagesource/s3.go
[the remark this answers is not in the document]
ardaguclu: I checked once again. I think, `resolveCredentials` in `defaultAWSConfigResolvers` registers the default fallback options, if I'm not mistaken.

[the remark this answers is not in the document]
ardaguclu: As far as I understand, this line https://github.com/aws/aws-sdk-go-v2/blob/f382571d75aca804e358f4432515d2413799bca7/config/resolve_credentials.go#L70 builds a chain of credential providers including our custom, env provider and shared file provider.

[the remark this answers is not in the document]
tchap: Actually the resolution chain is assembled starting here:

https://github.com/openshift/oc/blob/a8cd402927b026cb23f8093dcd63caf3410c361d/vendor/github.com/aws/aws-sdk-go-v2/config/config.go#L208-L207

It means the order is, use `LoadOptions`, then shared config loader. Now if you use `config.WithCredentialsProvider`, that sets `LoadOptions.Credentials` and marks the whole chain as resolved as you never get to `resolveCredentialChain`, because `resolveCredentialProvider` returns found based on `LoadOptions`.

So actually the bot is correct.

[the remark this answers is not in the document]
ardaguclu: ah, you are right. I thought this https://github.com/aws/aws-sdk-go-v2/blob/f382571d75aca804e358f4432515d2413799bca7/config/resolve_credentials.go#L66 is `!found` incorrectly. So I'll check the viability of recommended by bot.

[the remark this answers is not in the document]
ardaguclu: @tchap I've updated the code. Could you PTAL?.

pkg/cli/image/imagesource/s3.go
```diff
@@ -66,16 +66,20 @@ func (d *s3Driver) newObject(server *url.URL, region string, insecure bool, secu
 
 	configOpts := []func(*config.LoadOptions) error{
 		config.WithRegion(region),
-		config.WithCredentialsProvider(&s3CredentialStore{store: d.Creds, url: securityDomain}),
+	}
+
+	credStore := &s3CredentialStore{store: d.Creds, url: securityDomain}
+	if _, err := credStore.Retrieve(ctx); err != nil {
```
tchap: Actually this should be `err == nil` as the bot noticed.

ardaguclu: Today, bot is definitely clever than me :)

tchap: Today is literally the first time I find the bot useful 😅

tchap:
I was too fast, one more time 😅 

/lgtm

pkg/cli/image/imagesource/s3.go
```diff
@@ -283,7 +283,7 @@ func (s *s3ManifestService) Put(ctx context.Context, manifest distribution.Manif
 		if _, err := s.r.s3.CopyObject(s.r.ctx, &s3.CopyObjectInput{
 			Bucket:      aws.String(s.r.bucket),
 			ContentType: aws.String(mediaType),
-			CopySource:  aws.String(path.Join(s.r.bucket, blob)),
+			CopySource:  aws.String(s.r.bucket + "/" + url.QueryEscape(blob)),
```
tchap: Was there any reason to remove `path.Join` ?

ardaguclu: Not actually. Critical change is using the QueryEscape

ardaguclu: Do you want me still using path.Join?

tchap: Yeah, it would be better to keep it.

ardaguclu: Updated.

PR #2116, opened 2025-10-06
NO-JIRA: Update shellescape and sspi to newer versions

ardaguclu:
* shellescape has been moved to new domain `al.essio.dev/pkg/shellescape`. This PR updates the version and domain name of this package.
* sspi has been released new version due to some new constraints in Azure (https://github.com/alexbrainman/sspi/pull/18). This PR updates it to newer version.
  * If we get bug reports about oc login does not work on Windows or against Azure, I'm linking this referenced PR https://github.com/go-ldap/ldap/pull/565 to be an example for us how we can adopt changes for channel binding.

Note: I tried to migrate from archived github.com/gonum/graph to gonum.org/v1/gonum. But it requires some major changes. So I'll defer this for now as a very nice to have in the future.


ardaguclu:
/hold
until reviewed
/cc @tchap 
could you PTAL?.

ardaguclu:
I think, we don't have to verify this manually. If CI is green, we are ok.
/verify bypass

Neustradamus:
@ardaguclu: Thanks for your update!

Linked to:
- Channel Binding: https://github.com/scram-sasl/info/issues/1

PR #2135, opened 2025-11-07
OCPBUGS-64743: fix(must-gather): do not set node affinity if nodename is set

ingvagabund:
Otherwise, conflicting scheduling constraints may get provided. Causing a pod to fail kubelet's admission.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -57,6 +57,9 @@ const (
 	unreachableTaintKey       = "node.kubernetes.io/unreachable"
 	notReadyTaintKey          = "node.kubernetes.io/not-ready"
 	controlPlaneNodeRoleLabel = "node-role.kubernetes.io/control-plane"
+	defaultMustGatherCommand  = "/usr/bin/gather"
+	defaultVolumePercentage   = 70
+	defaultSourceDir          = "/must-gather/"
 )
 
 var (
@@ -172,10 +175,10 @@ func NewMustGatherCommand(f kcmdutil.Factory, streams genericiooptions.IOStreams
 
 func NewMustGatherOptions(streams genericiooptions.IOStreams) *MustGatherOptions {
 	opts := &MustGatherOptions{
-		SourceDir:        "/must-gather/",
+		SourceDir:        defaultSourceDir,
 		IOStreams:        streams,
 		Timeout:          10 * time.Minute,
-		VolumePercentage: 70,
+		VolumePercentage: defaultVolumePercentage,
 	}
 	opts.LogOut = opts.newPrefixWriter(streams.Out, "[must-gather      ] OUT", false, true)
 	opts.RawOut = opts.newPrefixWriter(streams.Out, "", false, false)
@@ -636,7 +639,7 @@ func (o *MustGatherOptions) Run() error {
 	}
 	var hasMaster bool
 	for _, node := range nodes.Items {
-		if _, ok := node.Labels["node-role.kubernetes.io/control-plane"]; ok {
+		if _, ok := node.Labels[controlPlaneNodeRoleLabel]; ok {
 			hasMaster = true
 			break
 		}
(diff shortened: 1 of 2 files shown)


pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -478,3 +478,41 @@ sync && echo 'Caches written to disk'`,
...
+			name:        "node set, affinity provided, affinity not set",
+			node:        "node1",
+			affinity:    buildNodeAffinity([]string{"node2"}),
+			affinitySet: false,
+		},
+		{
+			name:        "node not set, affinity provided, affinity set",
+			node:        "",
+			affinity:    buildNodeAffinity([]string{"node2"}),
+			affinitySet: true,
```
tchap: I would find it nicer and more generic to use `expectedPod` and compare using `cmp.Equal/cmp.Diff`. Otherwise this test just tests the affinity. Which is also ok, but in that case I would rename the test.

pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -478,3 +478,41 @@ sync && echo 'Caches written to disk'`,
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			mgo := &MustGatherOptions{}
+			tPod := mgo.newPod(test.node, "image", false, test.affinity)
+			if test.affinitySet {
+				if tPod.Spec.Affinity == nil {
+					t.Errorf("Pod.Spec.Affinity is expected to be set, it's nil\n")
```
tchap: nit: I don't think we usually append a newline to test error messages.

ingvagabund:
/hold
for more unit tests

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -175,7 +178,7 @@ func NewMustGatherOptions(streams genericiooptions.IOStreams) *MustGatherOptions
 		SourceDir:        "/must-gather/",
 		IOStreams:        streams,
 		Timeout:          10 * time.Minute,
-		VolumePercentage: 70,
+		VolumePercentage: defaultVolumePercentage,
```
tchap, line 181: Right above there is `/must-gather/` used as `SourceDir`, I guess we should use `defaultSourceDir` there?

pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -478,3 +479,167 @@ sync && echo 'Caches written to disk'`,
...
+		}}
+		pod.Spec.Containers[1].VolumeMounts = []corev1.VolumeMount{{
+			Name:      "must-gather-output",
+			MountPath: cleanSourceDir,
+		}}
+	}
+
+	tests := []struct {
+		name        string
+		mgo         *MustGatherOptions
```
tchap: Can we just call this `options`, please?

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -1099,29 +1102,8 @@ func newClusterRoleBinding(ns *corev1.Namespace) *rbacv1.ClusterRoleBinding {
 	}
 }
 
-// newPod creates a pod with 2 containers with a shared volume mount:
-// - gather: init containers that run gather command
-// - copy: no-op container we can exec into
-func (o *MustGatherOptions) newPod(node, image string, hasMaster bool, affinity *corev1.Affinity) *corev1.Pod {
-
+func defaultMustGatherPod(image string, update func(pod *corev1.Pod)) *corev1.Pod {
```
tchap: I would be happier for this function not to contain `update` arg. You can just do the same with

```
pod := defaultMustGatherPod()
pod.XYZ = ...
```

I see it's handy for tests, though, so feel free to define a test helper that lets you pass a callback.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -1203,9 +1171,56 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool, affinity
...
+	return ret
+}
+
+// newPod creates a pod with 2 containers with a shared volume mount:
+// - gather: init containers that run gather command
+// - copy: no-op container we can exec into
+func (o *MustGatherOptions) newPod(node, image string, hasMaster bool, affinity *corev1.Affinity) *corev1.Pod {
+	nodeSelector := map[string]string{
+		corev1.LabelOSStable: "linux",
+	}
```
tchap: Shouldn't this be taken from the default pod, or at least merged somehow? Otherwise the values defined in the default pod are simply dropped.

ingvagabund:
@tchap thank you for the second review. PR updated.

ardaguclu:
~~I honestly don't think that we need such refactoring. Can't only make changes for affinity?~~

ardaguclu:
I added hold not due to any objection just to understand the reasons behind :)

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -1149,7 +1130,7 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool, affinity
 					Name:            gatherContainerName,
 					Image:           image,
 					ImagePullPolicy: corev1.PullIfNotPresent,
-					Command:         []string{"/bin/bash", "-c", podCmd},
+					Command:         []string{"/bin/bash", "-c", defaultMustGatherCommand},
```
ardaguclu: We will overwrite this with `podCmd`. I think this fake assignment should not be in here. It is better to fail instead of disabling volumechecker script, if there is any issue.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -1203,9 +1171,51 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool, affinity
...
+		ReadOnly:  false,
+	}}
+	ret.Spec.Containers[1].VolumeMounts = []corev1.VolumeMount{{
+		Name:      "must-gather-output",
+		MountPath: cleanedSourceDir,
+		ReadOnly:  false,
+	}}
+	ret.Spec.HostNetwork = o.HostNetwork
+	for key, value := range nodeSelector {
+		ret.Spec.NodeSelector[key] = value
```
ardaguclu: Why don't we do instead?

```go
if node == "" && hasMaster {
ret.Spec.NodeSelector[controlPlaneNodeRoleLabel] = ""
}
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -1203,9 +1171,51 @@ func (o *MustGatherOptions) newPod(node, image string, hasMaster bool, affinity
...
+	}
+
+	if node == "" {
```
ardaguclu: nit: Contextually this is more correct;

```
if node != nil {
ret.Spect.NodeName = node
} else {
// some documentation about why do we add affinity 
ret.Spec.Affinity = affinity
}
```

pkg/cli/admin/mustgather/mustgather.go
ingvagabund: The goal was to generate the minimally required must-gather pod spec. If the volumechecker script is a must (I still consider it as one way to limit the collection size) I can default the command to include the volumechecker script with the default size. Would that work?

ardaguclu: That would work. Because volumechecker script is indeed must.

ingvagabund:
@ardaguclu updated, thank you for the review

ardaguclu:
In main branch, it was difficult to track which fields in pod are static and which ones vary based on flag values. Now, it is easier to see this distinction. Thank you.
/hold cancel
/lgtm

PR #2137, opened 2025-11-07
fix(makefile): work OOTB on macOS

### LINKED ISSUE #2136: `make oc` fails on macOS 26.1
Platform: macOS
Version (output of `sws_vers -ProductVersion`): 26.1

## Issue

even after installing dependencies via `brew install gpgme heimdall`, I get the following when running `make oc`:

```text
go build -mod=vendor -tags 'include_gcs include_oss containers_image_openpgp gssapi' -ldflags "-X github.com/openshift/oc/pkg/version.versionFromGit="v4.2.0-alpha.0-2809-g1c5f490" -X github.com/openshift/oc/pkg/version.commitFromGit="1c5f490b7" -X github.com/openshift/oc/pkg/version.gitTreeState="clean" -X github.com/openshift/oc/pkg/version.buildDate="2025-11-07T19:18:37Z" -X k8s.io/component-base/version.gitMajor="1" -X k8s.io/component-base/version.gitMinor="34" -X k8s.io/component-base/version.gitVersion="v1.34.1" -X k8s.io/component-base/version.gitCommit="1c5f490b7" -X k8s.io/component-base/version.buildDate="2025-11-07T19:18:35Z" -X k8s.io/component-base/version.gitTreeState="clean" -X k8s.io/client-go/pkg/version.gitVersion="v4.2.0-alpha.0-2809-g1c5f490" -X k8s.io/client-go/pkg/version.gitCommit="1c5f490b7" -X k8s.io/client-go/pkg/version.buildDate="2025-11-07T19:18:35Z" -X k8s.io/client-go/pkg/version.gitTreeState="clean"" github.com/openshift/oc/cmd/oc
# github.com/apcera/gssapi
vendor/github.com/apcera/gssapi/name.go:213:9: could not determine what C.wrap_gss_canonicalize_name refers to
cgo: 
cc errors for preamble:
vendor/github.com/apcera/gssapi/name.go:90:2: error: unknown type name 'gss_const_name_t'
   90 |         gss_const_name_t input_name,
      |         ^
vendor/github.com/apcera/gssapi/name.go:95:16: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
   95 |                 OM_uint32 *, gss_const_name_t, const gss_OID, gss_name_t *)
      |                              ^
      |                              int
2 errors generated.

make: *** [build] Error 1
```

dronenb:
The README instructs to install `heimdall` via `brew` to build on macOS. However, this alone is insufficient. This change detects if you are building on macOS and have the `gssapi` tag, and if so, determines your `brew` prefix,, and includes the heimdall headers

Fixes #2136

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,16 +16,16 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 KUBE_GIT_MINOR_VERSION := "34"
 KUBE_GIT_VERSION := "v1.34.1"
 
-GO_LD_EXTRAFLAGS :=-X k8s.io/component-base/version.gitMajor="1" \
-                   -X k8s.io/component-base/version.gitMinor=$(KUBE_GIT_MINOR_VERSION) \
-                   -X k8s.io/component-base/version.gitVersion=$(KUBE_GIT_VERSION) \
-                   -X k8s.io/component-base/version.gitCommit="$(SOURCE_GIT_COMMIT)" \
-                   -X k8s.io/component-base/version.buildDate="$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')" \
-                   -X k8s.io/component-base/version.gitTreeState="clean" \
-                   -X k8s.io/client-go/pkg/version.gitVersion="$(SOURCE_GIT_TAG)" \
-                   -X k8s.io/client-go/pkg/version.gitCommit="$(SOURCE_GIT_COMMIT)" \
-                   -X k8s.io/client-go/pkg/version.buildDate="$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')" \
-                   -X k8s.io/client-go/pkg/version.gitTreeState="$(SOURCE_GIT_TREE_STATE)"
+GO_LD_EXTRAFLAGS := -X k8s.io/component-base/version.gitMajor="1" \
+					-X k8s.io/component-base/version.gitMinor=$(KUBE_GIT_MINOR_VERSION) \
+					-X k8s.io/component-base/version.gitVersion=$(KUBE_GIT_VERSION) \
+					-X k8s.io/component-base/version.gitCommit="$(SOURCE_GIT_COMMIT)" \
+					-X k8s.io/component-base/version.buildDate="$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')" \
+					-X k8s.io/component-base/version.gitTreeState="clean" \
+					-X k8s.io/client-go/pkg/version.gitVersion="$(SOURCE_GIT_TAG)" \
+					-X k8s.io/client-go/pkg/version.gitCommit="$(SOURCE_GIT_COMMIT)" \
+					-X k8s.io/client-go/pkg/version.buildDate="$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')" \
+					-X k8s.io/client-go/pkg/version.gitTreeState="$(SOURCE_GIT_TREE_STATE)"
 
 GO_BUILD_PACKAGES :=$(strip \
 	./cmd/... \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,6 +37,16 @@ GO_BUILD_FLAGS_DARWIN :=-tags 'include_gcs include_oss containers_image_openpgp'
...
+# Fix for macOS GSSAPI compatibility when gssapi tag is used
+# macOS system headers lack RFC 5587 support. Use Homebrew's heimdal for proper headers.
+# See vendor/github.com/apcera/gssapi/name.go for details.
+ifeq ($(shell uname),Darwin)
+ifneq (,$(findstring gssapi,$(GO_BUILD_FLAGS)))
+HOMEBREW_PREFIX := $(shell brew --prefix)
+export CGO_CFLAGS := -I$(HOMEBREW_PREFIX)/opt/heimdal/include
+endif
+endif
+
```
tchap, line 49: This is gonna fail on Windows, right?

I am wondering whether just documenting the issue in README wouldn't suffice TBH. But feel free to come up with a cross-platform way to improve this.

tchap: I am actually not sure how much we support Windows TBH 😅 Perhaps @ardaguclu knows better.

I would also like to understand first why my setup worked out of the box. I would prefer for the user to set up their env so that the lib is found just fine instead of us handling the env in the Makefile.

dronenb: I don't think this would affect Windows at all, since this only applies to Darwin, happy to test it though.

if we wanted users to setup their environment, I imagine that would mean including pkg-config in some capacity? I think this route would be the simplest...

Makefile
```diff
@@ -16,16 +16,16 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
...
-                   -X k8s.io/component-base/version.gitMinor=$(KUBE_GIT_MINOR_VERSION) \
-                   -X k8s.io/component-base/version.gitVersion=$(KUBE_GIT_VERSION) \
-                   -X k8s.io/component-base/version.gitCommit="$(SOURCE_GIT_COMMIT)" \
-                   -X k8s.io/component-base/version.buildDate="$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')" \
-                   -X k8s.io/component-base/version.gitTreeState="clean" \
-                   -X k8s.io/client-go/pkg/version.gitVersion="$(SOURCE_GIT_TAG)" \
-                   -X k8s.io/client-go/pkg/version.gitCommit="$(SOURCE_GIT_COMMIT)" \
-                   -X k8s.io/client-go/pkg/version.buildDate="$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')" \
-                   -X k8s.io/client-go/pkg/version.gitTreeState="$(SOURCE_GIT_TREE_STATE)"
+GO_LD_EXTRAFLAGS := -X k8s.io/component-base/version.gitMajor="1" \
```
ardaguclu, line 19: We don't need the changes in here.

Makefile
```diff
@@ -37,6 +37,16 @@ GO_BUILD_FLAGS_DARWIN :=-tags 'include_gcs include_oss containers_image_openpgp'
 GO_BUILD_FLAGS_WINDOWS :=-tags 'include_gcs include_oss containers_image_openpgp'
 GO_BUILD_FLAGS_LINUX_CROSS :=-tags 'include_gcs include_oss containers_image_openpgp'
 
+# Fix for macOS GSSAPI compatibility when gssapi tag is used
+# macOS system headers lack RFC 5587 support. Use Homebrew's heimdal for proper headers.
+# See vendor/github.com/apcera/gssapi/name.go for details.
+ifeq ($(shell uname),Darwin)
```
ardaguclu, line 43: `make oc` uses `GO_BUILD_FLAGS` (line:35) which includes `gssapi` dependecy. However, if you see darwin build flags (i.e. `GO_BUILD_FLAGS_DARWIN`, line:36), it does not have `gssapi`. 

Could you please test `make oc` by removing `gssapi` from GO_BUILD_FLAGS?. If the build succeeds, it would be great if you can try to use `oc login`. If `oc login` works too, I believe that we can modify `GO_BUILD_FLAGS` by removing `gssapi`, if the host is darwin.

Makefile
```diff
@@ -37,6 +37,16 @@ GO_BUILD_FLAGS_DARWIN :=-tags 'include_gcs include_oss containers_image_openpgp'
...
+ifneq (,$(findstring gssapi,$(GO_BUILD_FLAGS)))
+HOMEBREW_PREFIX := $(shell brew --prefix)
+export CGO_CFLAGS := -I$(HOMEBREW_PREFIX)/opt/heimdal/include
```
ardaguclu, line 46: I believe that we don't need this at all on macos.

Makefile
ardaguclu: Conditional will only work for Darwin. So this won't affect Windows.

ardaguclu (commented):
Thank you for filing and fixing this. Dropped a few comments.

Makefile
tchap: I am honestly bit of a greenhorn regarding Windows, so I was not sure whether `uname` is even available there. But I guess we require some core UNIX-like utils to be installed. Never mind.

PR #2143, opened 2025-11-13
OCPBUGS-65594: fix(must-gather): do not set node affinity if nodename is set

ingvagabund:
Backporting https://github.com/openshift/oc/pull/2142 by hand since there's no clean application of the patch in 4.19.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -56,6 +56,10 @@ const (
 	unreachableTaintKey       = "node.kubernetes.io/unreachable"
 	notReadyTaintKey          = "node.kubernetes.io/not-ready"
 	controlPlaneNodeRoleLabel = "node-role.kubernetes.io/control-plane"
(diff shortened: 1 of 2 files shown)


ingvagabund:
One could only backport the change of removing the setting of `.spec.nodeName` and
```go
	if node != "" {
		ret.Spec.NodeName = node
	} else {
		ret.Spec.Affinity = affinity
	}
```
to minimize the risk of introducing regressions. Yet:
- there's no unit test for testing the generated pod
- in the future we might need to backport/fix more so running an already existing unit test is a huge plus
- the changes in the PR are mostly about refactoring the code (besides the fix there's no other functional change)

To increase the confidence while reviewing this PR I can break the changes into more commits so every step is clear, simple and obvious in what the intention is and how it is done. If needed.

ardaguclu:
Changes looks good to me. I agree with your comments above ^^
/lgtm
/hold
but we definitely need pre-merge testing to not break 4.19 z-streams (albeit of a low probability)

ingvagabund:
pre-merge testing done
/verified by @zhouying7780

PR #2149, opened 2025-11-20
OCPBUGS-58038: Clear defaultNodeSelector to run on any node

ardaguclu:
`oc adm copy-to-node` and `oc adm restart-kubelet` commands can run on any node type (either control-plane or worker) by creating a pod in its dedicated namespace. If defaultNodeSelector is configured within the cluster, its dedicated namespace is not allowed to run pods in excluded node types. 

Therefore, we intentionally clear `openshift.io/node-selector` annotation to disable `defaultNodeSelector` for this namespace.


ardaguclu:
/hold
until per-merge tests are performed.

PR #2153, opened 2025-11-25
OCPBUGS-65925: Fall back to simpler behavior, if setsid,ps,pkill are not installed

ardaguclu:
`setsid` brings us robust session killing, when volumesizechecker detects any exceeded volume usage. This works well in default must-gather image as well as others that `setsid` is installed. 

However, there are some images that `setsid` is **not** installed. This method does not work and more importantly must-gather does not work. So we must introduce a fall back approach for the images that lack such binaries. 

This PR checks the presence of `setsid`. If it is present, it will continue using the current approach. If it is not, it will run the scripts in simpler way. If volumesize checker detects the exceeded volume usage, it will run `kill 0` to kill every process. We can't provide elegant options because these images do not include any binaries that we can use such as `ps`, `setsid`, etc.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -99,10 +99,14 @@ usage_percentage=$(df -P "$target_dir" | awk 'NR==2 {print $5}' | sed 's/%%//')
 echo "[disk usage checker] Volume usage percentage: current = ${usage_percentage} ; allowed = ${usage_percentage_limit}"
 if [ "$usage_percentage" -gt "$usage_percentage_limit" ]; then
 	echo "[disk usage checker] Disk usage exceeds the volume percentage of ${usage_percentage_limit} for mounted directory, terminating..."
-	ps -o sess --no-headers | sort -u | while read sid; do
-		[[ "$sid" -eq "${$}" ]] && continue
-		pkill --signal SIGKILL --session "$sid"
-	done
+	if [ "$HAVE_SESSION_TOOLS" = "true" ]; then
+		ps -o sess --no-headers | sort -u | while read sid; do
+			[[ "$sid" -eq "${$}" ]] && continue
+			pkill --signal SIGKILL --session "$sid"
+		done
+	else
+		kill 0
+	fi
 	exit 1
 fi
 sleep 5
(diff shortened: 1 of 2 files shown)


ardaguclu:
I got this PR https://github.com/openshift/oc/pull/2152 into this PR.

ardaguclu:
/cc @tchap could you PTAL?.
/hold
until reviews, pre-merge tests are completed.

tchap:
Read it as thoroughly as I managed, looks principally ok.

If we wanted to be super pretty, we could do `SETSID="$(command -v setsid)"` and then just check `SETSID` everywhere, but just using `command` directly is no problem, it's just 2 places really.

/lgtm

ardaguclu:
> Read it as thoroughly as I managed, looks principally ok.
> 
> If we wanted to be super pretty, we could do `SETSID="$(command -v setsid)"` and then just check `SETSID` everywhere, but just using `command` directly is no problem, it's just 2 places really.
> 
> /lgtm

Thank you. So next step would have a pre-merge tests.

tchap:
Note that #2152 is unfinished and the change does not fix anything yet. But no problem with merging it, I guess.

ardaguclu:
> Note that #2152 is unfinished and the change does not fix anything yet. But no problem with merging it, I guess.

Why do you think that?. The ask is to printing this message to stderr and this achieves this. What am I missing?

tchap:
> Why do you think that?. The ask is to printing this message to stderr and this achieves this. What am I missing?

We are streaming pod logs, right? So there are two separate steps involved:

1. Actually print the message to stderr in the pod.
2. Stream pod stderr to local stderr.

While the first is trivial, the latter is not and it seems the relevant functionality is behind an alpha feature gate `PodLogsQuerySplitStreams`, so it's disabled by default: https://kubernetes.io/docs/concepts/cluster-administration/logging/#container-log-streams

We cannot even implement this currently, I guess.

ardaguclu:
> > Why do you think that?. The ask is to printing this message to stderr and this achieves this. What am I missing?
> 
> We are streaming pod logs, right? So there are two separate steps involved:
> 
> 1. Actually print the message to stderr in the pod.
> 2. Stream pod stderr to local stderr.
> 
> While the first is trivial, the latter is not and it seems the relevant functionality is behind an alpha feature gate `PodLogsQuerySplitStreams`, so it's disabled by default: https://kubernetes.io/docs/concepts/cluster-administration/logging/#container-log-streams
> 
> We cannot even implement this currently, I guess.

Thanks. That makes sense. So I think, we can close the bug. Because it sounds like a feature request. I'd prefer reverting that feature from this PR as well.

tchap:
It might be more of a feature request.

/lgtm

ardaguclu:
> It might be more of a feature request.

That is correct. However, if we rely on a feature on upstream, I think this can be considered as a feature.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -99,10 +99,14 @@ usage_percentage=$(df -P "$target_dir" | awk 'NR==2 {print $5}' | sed 's/%%//')
 echo "[disk usage checker] Volume usage percentage: current = ${usage_percentage} ; allowed = ${usage_percentage_limit}"
 if [ "$usage_percentage" -gt "$usage_percentage_limit" ]; then
 	echo "[disk usage checker] Disk usage exceeds the volume percentage of ${usage_percentage_limit} for mounted directory, terminating..."
-	ps -o sess --no-headers | sort -u | while read sid; do
-		[[ "$sid" -eq "${$}" ]] && continue
-		pkill --signal SIGKILL --session "$sid"
-	done
+	if command -v setsid >/dev/null 2>&1 && command -v ps >/dev/null 2>&1 && command -v pkill >/dev/null 2>&1; then
```
tchap: I would honestly just set a helper variable initially and then use it everywhere. This is getting ugly. Like, set a variable to non-empty value if we can do sessions, and do that at the beginning of `buildPodCommand`.

ardaguclu: I'm open to any suggestions as long as we can inject this helper variable from buildPodCommand to volumeCheckerScript. Would you please elaborate your exact suggestion?

tchap: I think that everything is just concatenated in `buildPodCommand`, so isn't any variable visible later on? Perhaps you need to export it for the subshells to make it work...

ardaguclu: yeah, when I tested this, subshells did not see the variables in the main session. Claude suggested me exporting this variable and I rejected the suggestion. But I'll retry exporting this environment variable.

ardaguclu: I made a mistake about ordering. So that's why variable didn't work previously. This works now. Could you PTAL?

tchap: Looks better, thanks!

danielerez:
Hey @ardaguclu, can you please backport this to 4.20 once merged?

ardaguclu:
> Hey @ardaguclu, can you please backport this to 4.20 once merged?

Yes, I am planning to backport this to 4.20.

PR #2155, opened 2025-11-26
OCPBUGS-63189: Unify ErrAlreadyExists to correctly match

ardaguclu:
When the uploaded blob is already existed, [S3 raises](https://github.com/openshift/oc/blob/9e4292b2df204c3d29a4441aa653cdc048e6b797/pkg/cli/image/imagesource/s3.go#L357) ErrAlreadyExists error correctly. However, `mirror.go` compares the error against duplicate of `ErrAlreadyExists` error. So it mismatches.

This PR removes duplicate and unused `ErrAlreadyExists` in `mappings.go` and correctly compare the error in S3.

### DIFF
--- a/pkg/cli/image/mirror/mappings.go
+++ b/pkg/cli/image/mirror/mappings.go
@@ -14,9 +14,6 @@ import (
 	"github.com/openshift/oc/pkg/cli/image/imagesource"
 )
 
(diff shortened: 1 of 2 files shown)


tchap:
/lgtm

Would be lovely to have tests for this, though.

ardaguclu:
/verified by pass
It is a minor and relatively safe change. We can trust CI results.

PR #2157, opened 2025-12-01
MCO-1961: Allow multiple machine-os versions

sdodson:


pkg/cli/admin/release/image_mapper.go
```diff
@@ -198,7 +198,8 @@ func loadImageStreamTransforms(input, local *imageapi.ImageStream, allowMissingI
 			}
 			existing, ok := versions[k]
 			if ok {
-				if existing.Version != v.Version {
+				// we allow multiple machine-os versions due to dual stream efforts
+				if existing.Version != v.Version && k != "machine-os" {
```
ardaguclu, line 202: This disables component version skew check entirely for machine-os component. Would it be better if we only allow 2 different machine-os versions and fail if there are 3 (to prevent any divergence) as we did https://github.com/openshift/oc/pull/1662 (and https://github.com/openshift/oc/pull/1656).

Additionally, same changes need to go Hypershift https://github.com/openshift/hypershift/blob/02f528cd5364b03168500ada33c9a0d9d0593025/support/releaseinfo/releaseinfo.go#L185

tchap:
We should also add some tests BTW.

pkg/cli/admin/release/image_mapper.go
sdodson: Thanks for pointing out the hypershift use case. For the time being we're simply dropping the `machine-os` version from the rhel-coreos-10 images but I think we need to figure out how to effectively re-introduce this. Ideally for 4.22 but perhaps deferred to the next release after 4.22. I'm going to close this for the time being but open up a ticket for the RHCOS team to dive into this across the problem space.

sdodson:
/close
for now, RHCOS team to take a ticket to chase this across the problem space to ensure that we can have multiple unique versions for RHCOS in the release payload.

jbtrystram:
Picking that up. 
So i did build a payload with two machine-os images, and with this patch, oc does not complain. 
```
oc adm release info quay.io:443/jbtrystramtestimages/rhel-coreos:4.22.0-ec.3-full-payload                                
Name:           4.22.0-ec.3
Digest:         sha256:d6514d0ec05b7e09e8f0dedcc68fe23d3afdd3a899dba391258210cc5399efa2
Created:        2026-03-11T20:08:31Z
OS/Arch:        linux/amd64
Manifests:      859
Metadata files: 2

Pull From: quay.io:443/jbtrystramtestimages/rhel-coreos@sha256:d6514d0ec05b7e09e8f0dedcc68fe23d3afdd3a899dba391258210cc5399efa2

Release Metadata:
  Version:  4.22.0-ec.3
  Upgrades: 4.21.0, 4.21.1, 4.21.2, 4.21.3, 4.21.4, 4.22.0-ec.0, 4.22.0-ec.1, 4.22.0-ec.2
  Metadata:

Component Versions:
  kubectl          1.34.1        
  kubernetes       1.34.2        
  kubernetes-tests 1.34.1        
  machine-os       10.2.20260311 Red Hat Enterprise Linux CoreOS

Images:
  NAME                                           DIGEST
  agent-installer-api-server               
<...> 
  rhel-coreos                                    sha256:42e2e4d11674c603398469770700ef5a30432294ffa245ed1c67852e93fc380e
  rhel-coreos-10                                 quay.io:443/jbtrystramtestimages/rhel-coreos:4.22-10.2-tagged
  rhel-coreos-10-extensions                      sha256:ee90288a9dfc801da76f890438957d38702c09c94ecaaaeef0f56b5cd2eb19a9
  rhel-coreos-extensions                         sha256:b7d9bb1dee4383e0e64baaabb915c194deebf9cbe19726aaea7e3498eaa2c59e
<...>

Warnings:
* multiple versions or display names reported for the following component(s): machine-os
```

pkg/cli/admin/release/image_mapper.go
jbtrystram: Limiting to two values here will probably backfire later because we may have more streams coming up : the nvidia kernel, confidential clusters.. Now that this can of worms is open, more may even be added

sdodson:
/reopen
@ardaguclu We're now looking at potentially having multiple streams which means we'll have N number of images labeled with `machine-os` version. Can you think of any issues that will arise in product if we allow that?

Thanks @jbtrystram for testing this out, I was never able to get to that point.

ardaguclu:
> /reopen @ardaguclu We're now looking at potentially having multiple streams which means we'll have N number of images labeled with `machine-os` version. Can you think of any issues that will arise in product if we allow that?
> 
> Thanks @jbtrystram for testing this out, I was never able to get to that point.

It is difficult to see the implications of this change (maybe there isn't any). But `machine-os` is critical parts of this releasing process https://github.com/openshift/oc/blob/05fa7bb1b8758ab42834aeb1045f681ff26f2ef4/pkg/cli/admin/release/new.go#L55-L58, https://github.com/openshift/oc/blob/05fa7bb1b8758ab42834aeb1045f681ff26f2ef4/pkg/cli/admin/release/info.go#L2113

In my opinion, by allowing any arbitrary stream we would permanently lose the control. Would it be too difficult to embed all possible streams in oc and do not allow the others?. That would bring about some maintenance burden, but at least we intentionally update the list.

ardaguclu:
Also cc'ing @wking @bradmwilliams (please ignore if you are not related to this topic)

sdodson:
> In my opinion, by allowing any arbitrary stream we would permanently lose the control. Would it be too difficult to embed all possible streams in oc and do not allow the others?. That would bring about some maintenance burden, but at least we intentionally update the list.

In the past few months its become clear that we're likely to see a proliferation of streams. It's probably right on the tipping point as to whether or not it will be manageable to do this with a lot of cross repo coordination, right now the MCO, Installer, and RHCOS are all oriented toward dynamic discovery so that we don't have to coordinate.

jbtrystram:
@sdodson here is a patch to remove the warning when creating the payload : 
```
From 81fa5a71b3bbfb1eda6c001df6634f90d4743afc Mon Sep 17 00:00:00 2001
From: jbtrystram <jbtrystram@redhat.com>
Date: Tue, 24 Mar 2026 11:26:34 +0100
Subject: [PATCH] release/info: allow multiple machine-os versions

Rhcos will now come in two flavors : rhel-9 base and rhel-10. We remove
the warning when more than one version is reported.
Let's not put a limit here as  we're likely to see a proliferation of
streams with nvidia and confidential clusters comming up.

Assisted-by: OpenCode.ai <Opus 4.6>
---
 pkg/cli/admin/release/info.go | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

diff --git a/pkg/cli/admin/release/info.go b/pkg/cli/admin/release/info.go
index 2367751d1..ff9b60718 100644
--- a/pkg/cli/admin/release/info.go
+++ b/pkg/cli/admin/release/info.go
@@ -1058,7 +1058,8 @@ func readComponentVersions(is *imageapi.ImageStream, errOut io.Writer) (Componen
        sort.Strings(keys)
        for _, k := range keys {
                v := combined[k]
-               if v.Len() > 1 {
+               // we allow multiple machine-os versions due to dual stream efforts
+               if v.Len() > 1 && k != "machine-os" {
                        multiples = multiples.Insert(k)
                }
                if _, ok := out[k]; ok {
@@ -1075,7 +1076,8 @@ func readComponentVersions(is *imageapi.ImageStream, errOut io.Writer) (Componen
                if !ok {
                        continue
                }
-               if v.Len() > 1 {
+               // we allow multiple machine-os display names due to dual stream efforts
+               if v.Len() > 1 && k != "machine-os" {
                        multiples = multiples.Insert(k)
                }
                version, ok := out[k]
-- 
2.53.0
```

ardaguclu:
I'll defer the decision to @sdodson about relaxing the validation

sdodson:
> @sdodson here is a patch to remove the warning when creating the payload :

Thanks, I wasn't able to `git apply` the patch, not sure why. But I've applied the changes. PTAL

sdodson:
I'd like to move forward with this so we can make sure that this works properly in time for branch cut.

PR #2160, opened 2025-12-04
CNTRLPLANE-2167:Integrate OpenShift Test Extension (OTE) into oc

gangwgr:
As per discussion https://redhat-internal.slack.com/archives/CC3CZCQHM/p1762421042028719
[CNTRLPLANE-2167](https://issues.redhat.com//browse/CNTRLPLANE-2167):Integrate OpenShift Test Extension (OTE) into oc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,9 +71,8 @@ oc: GO_BUILD_PACKAGES :=./cmd/oc
 oc: build
 .PHONY: oc
 
-oc-tests-ext: GO_BUILD_PACKAGES :=./cmd/oc-tests-ext
-oc-tests-ext: GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp'
-oc-tests-ext: build
+oc-tests-ext:
+	go build -o oc-tests-ext ./cmd/oc-tests-ext
 .PHONY: oc-tests-ext
 
 update: update-generated-completions
--- a/images/cli/Dockerfile.rhel
+++ b/images/cli/Dockerfile.rhel
@@ -1,13 +1,10 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.24-openshift-4.22 AS builder
 WORKDIR /go/src/github.com/openshift/oc
 COPY . .
-RUN make build --warn-undefined-variables \
-      && make oc-tests-ext \
-      && gzip oc-tests-ext
+RUN make build --warn-undefined-variables
 
 FROM registry.ci.openshift.org/ocp/4.22:base-rhel9
 COPY --from=builder /go/src/github.com/openshift/oc/oc /usr/bin/
-COPY --from=builder /go/src/github.com/openshift/oc/oc-tests-ext.gz /usr/bin/
 RUN for i in kubectl openshift-deploy openshift-docker-build openshift-sti-build openshift-git-clone openshift-manage-dockerfile openshift-extract-image-content openshift-recycle; do ln -sf /usr/bin/oc /usr/bin/$i; done
 LABEL io.k8s.display-name="OpenShift Client" \
       io.k8s.description="OpenShift is a platform for developing, building, and deploying containerized applications." \


images/cli/Dockerfile.rhel
```diff
@@ -1,10 +1,13 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.24-openshift-4.21 AS builder
 WORKDIR /go/src/github.com/openshift/oc
 COPY . .
-RUN make build --warn-undefined-variables
+RUN make build --warn-undefined-variables \
+      && make oc-tests-ext \
```
ardaguclu: Many people, CI systems rely on this `cli` image. This change increases the storage size which may have unexpected impacts.

ardaguclu (commented):
Even for reviewing this PR requires dedicated time and plan which I don't have. So we need to defer this until we plan this change in oc accordingly. Because I don't know the motivations and reasoning behind this change yet.
So I have to add hold for this, until there is someone willing to review this in-detail
/hold

images/cli/Dockerfile.rhel
gangwgr: @ardaguclu We have same discussed with @p0lyn0mial  here https://redhat-internal.slack.com/archives/CC3CZCQHM/p1762421042028719 regarding kas-o

p0lyn0mial: This PR integrates with https://docs.google.com/document/d/1cFZj9QdzW8hbHc3H0Nce-2xrJMtpDJrwAse9H7hLiWk/edit?tab=t.0#heading=h.66y4kqbj468a

TLDR: it’s about integrating tests from this repository with openshift-tests to improve coverage and protect against regressions. 

BTW: do we actually have any e2e tests defined for this repository? If not, maybe this PR isn’t needed at the moment.

gangwgr: @p0lyn0mial In qe repo we have, need to migrate those here 
@zhouying7780 Please confirm

gangwgr: @p0lyn0mial as checked with @zhouying7780 We have cases here

ardaguclu: @gangwgr I built this binary on my local. Its size is 56 mb (zipped size is 24 mb). That means this change statically increases every image that relies on `cli` image with almost 30%.  Additionally, this binary will be only used by our CI system so there is no point of building and exposing this to customers. 

I think, we can add a separate Makefile target (e.g. `make oc-tests-ext`) in oc and build this binary separately only in our CI system (https://github.com/openshift/release/blob/104543c7c4e0bc5ad2b60e6d452b266233313efc/ci-operator/config/openshift/oc/openshift-oc-main.yaml#L30). WDYT?.

cmd/oc-tests-ext/main.go
```diff
@@ -0,0 +1,53 @@
+package main
```
ardaguclu, line 1: Changes in this file make sense to me.

Makefile
```diff
@@ -71,6 +71,11 @@ oc: GO_BUILD_PACKAGES :=./cmd/oc
 oc: build
 .PHONY: oc
 
+oc-tests-ext: GO_BUILD_PACKAGES :=./cmd/oc-tests-ext
+oc-tests-ext: GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp'
+oc-tests-ext: build
```
ardaguclu: This should be a separate build target independent from main `build` and should be executed separately when it is required.

Makefile
```diff
@@ -71,6 +71,11 @@ oc: GO_BUILD_PACKAGES :=./cmd/oc
 oc: build
 .PHONY: oc
 
+oc-tests-ext: GO_BUILD_PACKAGES :=./cmd/oc-tests-ext
+oc-tests-ext: GO_BUILD_FLAGS :=-tags 'include_gcs include_oss containers_image_openpgp'
```
ardaguclu: There is no need to use these flags for `oc-tests-ext` binary.

ardaguclu (commented):
There are merge conflicts. So this PR needs to be updated. I dropped some comments.

gangwgr:
@ardaguclu Regarding this https://github.com/openshift/oc/pull/2160#discussion_r2661089503 we need Dockerfile changes and ote is independent from oc binary.  In ci ote binary extracted and run test cases. We can't build ote binary in ci job as ote team asked us not to build ote binary in jobs.
We have only 2 options
Use Dockerfile
or migrate cases to origin repo
CC [@Ke Wang(kewang)](https://redhat.enterprise.slack.com/team/URQL37FLM)
slack discussion with [@polynomial](https://redhat.enterprise.slack.com/team/UK6NS15HS)

PR #2161, opened 2025-12-16
Bump containerd version from 1.7.11  to 1.7.29

atheo89:
This PR updates the containerd dependency to `v1.7.29` to address [CVE-2024-25621](https://www.cve.org/CVERecord?id=CVE-2024-25621)

Reason for Change The previous version (v1.7.11) is affected by `CVE-2024-25621` (High Severity). 
Updating to v1.7.29 resolves this vulnerability and brings in the latest patch fixes.

**Steps Taken:**
The update was applied using standard Go module commands to ensure go.mod, go.sum, and vendor/ are synchronized:
```
$ go get github.com/containerd/containerd@v1.7.29
$ go mod tidy
$ go mod vendor
```


go.mod
```diff
@@ -92,7 +92,7 @@ require (
 	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/chai2010/gettext-go v1.0.2 // indirect
-	github.com/containerd/containerd v1.7.11 // indirect
+	github.com/containerd/containerd v1.7.29 // indirect
```
ardaguclu, line 95: As far as I can see from the CVE https://www.cve.org/CVERecord?id=CVE-2024-25621, this is the commit https://github.com/containerd/containerd/commit/7c59e8e9e970d38061a77b586b23655c352bfec5 fixing the issue. 

However, as you can see from the `vendor` directory changes, oc is not affected by this CVE. So basically we don't fix anything, other than silencing the security scanners.

atheo89: Thanks for checking the code path. However, because the oc binary explicitly reports linking against containerd v1.7.11, our internal security compliance gates (ProdSec) automatically [flag this as a blocker](https://issues.redhat.com/browse/RHOAIENG-41457). We cannot release or deploy artifacts that trigger High/Critical CVE alerts.

From an operational standpoint, upgrading the dependency is a more sustainable solution than maintaining manual waivers or suppression rules for every scan in perpetuity. Bumping the version resolves the compliance debt and ensures we are on a clean, supported release.

ardaguclu: > our internal security compliance gates (ProdSec) automatically [flag this as a blocker](https://issues.redhat.com/browse/RHOAIENG-41457). We cannot release or deploy artifacts that trigger High/Critical CVE alerts.

I'm not a blocker of merging this bump. It looks safe. But I'd like to point out that oc does not have to follow dependent components' policies. True positives should be bumped. On the other hand, we usually bump our packages to silence false positives, unless the bump is risky for any regressions.

atheo89: Thanks for the context. We are definitely in the 'silencing false positives' territory here to satisfy our ProdSec gates. I appreciate you reviewing the safety of the bump so we can get this cleared.

ardaguclu: I realized that this PR updates indirect import. We need to be cautious about consumer of this library won't regress. So I took back my `safe` change statement.

ardaguclu: I checked the dependency graph and the changes in this PR is safe.

atheo89: Since this is a patch-level update (v1.7.11 → v1.7.29), it should technically be backward compatible. Aside from the standard unit tests and verification scripts in this repo, are there specific downstream integration tests or consumer projects you recommend I check to verify this?

ardaguclu: We can rely on our CI results. I've opened this https://github.com/openshift/oc/pull/2162. So we may entirely drop this dependency. If https://github.com/openshift/oc/pull/2162 fails in CI, we can merge yours.

ardaguclu:
/hold
until we decide which one we are going to merge.

atheo89:
Closing this PR in favor https://github.com/openshift/oc/pull/2162

PR #2183, opened 2026-01-15
OCPBUGS-73844: Let import mode be decided by the cluster when not specified

Prashanth684:
Today, we default to "Legacy" importmode when no option is specified. With the introduction of
https://issues.redhat.com/browse/MULTIARCH-4552, the importmode is determined by the Cluster version's `desired.architecture` - which if set to "Multi", the import mode is PreserveOriginal else Legacy. Hence we don't need to default to Legacy all the time.

### DIFF
--- a/pkg/cli/importimage/importimage.go
+++ b/pkg/cli/importimage/importimage.go
@@ -100,7 +100,6 @@ func NewImportImageOptions(streams genericiooptions.IOStreams) *ImportImageOptio
 		PrintFlags:      genericclioptions.NewPrintFlags("imported"),
 		IOStreams:       streams,
 		ReferencePolicy: tag.SourceReferencePolicy,
-		ImportMode:      string(imagev1.ImportModeLegacy),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/cli/importimage/importimage.go
```diff
@@ -129,7 +128,7 @@ func NewCmdImportImage(f kcmdutil.Factory, streams genericiooptions.IOStreams) *
 	cmd.Flags().BoolVar(&o.Confirm, "confirm", o.Confirm, "If true, allow the image stream import location to be set or changed")
 	cmd.Flags().BoolVar(&o.All, "all", o.All, "If true, import all tags from the provided source on creation or if --from is specified")
 	cmd.Flags().StringVar(&o.ReferencePolicy, "reference-policy", o.ReferencePolicy, "Allow to request pullthrough for external image when set to 'local'. Defaults to 'source'.")
-	cmd.Flags().StringVar(&o.ImportMode, "import-mode", o.ImportMode, "Imports the full manifest list of a tag when set to 'PreserveOriginal'. Defaults to 'Legacy'.")
+	cmd.Flags().StringVar(&o.ImportMode, "import-mode", o.ImportMode, "Imports the full manifest list of a tag when set to 'PreserveOriginal'. When set to 'Legacy', imports a single sub-manifest. When unspecified, the cluster determines the import mode.")
```
ardaguclu, line 131: Wouldn't this be a breaking change?. Because we implicitly change a default behavior. So people who rely on Legacy behavior by default will start getting failures?.

Prashanth684: The default behavior will only change for clusters installed with the multi-arch payload, where the expectation is that images are manifestlisted. That being said, we should document that on multi-arch clusters, the default behavior is importing manifestlisted images and to get the legacy behavior, `--import-mode=Legacy` needs to be added.

xiuwang:
/verified by @xiuwang 

Test `oc import-image`, `oc tag`, `oc new-app`, `oc new-build` in multi arch clusters, without `--import-mode` specified, the import mode is default to `PreserveOriginal`, and no regression in single arch cluster, default to `Legacy` when not specify `--import-mode`

pkg/cli/importimage/importimage.go
ardaguclu: That makes sense. Let's update `When unspecified, the cluster determines the import mode.` to more explicit about the behavior, such as "When unspecified, if the cluster is single-arch, it will be Legacy; if the cluster is multi-arch, it will PreserveOriginal", etc.

ardaguclu:
Just dropped a comment about documentation update. Other than that this looks good to me. Additionally, you need to fix invalid-jira label.

PR #2249, opened 2026-04-07
OCPBUGS-77845: Fix extraction of gated manifests using --install-config

JoelSpeed:
We were filtering out the release-metadata too early in the cycle which meant that we didn't have it available within the feature gate extraction mechanism. This meant we couldn't identify the release version, and therefore bailed out without any feature gate aware filtering.

This removes the early check for credentials request (it's filtered during the write to disk) so that we still keep these files in memory to be used by later stages

Also turns out we don't need/care about the OC version so dropping that part too


wking (approved):
/lgtm

wking:
Nothing in [e2e-aws-ovn][1] excercises this code; I suggest we `/override ci/prow/e2e-aws-ovn`.  In the absence of an `oc` root approver who can do that, maybe we'll get lucky on a retest:

/test e2e-aws-ovn

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/2249/pull-ci-openshift-oc-main-e2e-aws-ovn/2041507124462227456

PR #2252, opened 2026-04-13
RFE-5434: Add cluster ID and timestamp to must-gather directory name

asadawar:
## Summary

- Include the last 12 characters of the cluster ID and a UTC timestamp in the auto-generated must-gather destination directory name
- Helps support engineers distinguish must-gather archives from different clusters and collection times when customers upload multiple archives to the same support case
- Falls back gracefully if cluster ID is unavailable (e.g. cluster is unreachable)

**New format:** `must-gather.local.<cluster-id-suffix>.<timestamp>.<rand>`
**Example:** `must-gather.local.76708af6b91c.20260413T141030Z.5119224030749742202`

**Fallback (cluster unreachable):** `must-gather.local.<timestamp>.<rand>`

## Details

The change is in `Complete()` within `pkg/cli/admin/mustgather/mustgather.go`. A new `generateDestDir()` method:
1. Queries `ClusterVersion` "version" via the existing `ConfigClient` to retrieve `Spec.ClusterID`
2. Extracts the last 12 characters of the UUID (low collision probability per [UUID substring analysis](https://stackoverflow.com/questions/70789567/uniqueness-of-uuid-substring))
3. Appends a UTC timestamp in compact ISO 8601 format (`20060102T150405Z`)
4. Appends a random numeric suffix for uniqueness
5. Falls back to `must-gather.local.<timestamp>.<rand>` if `ClusterVersion` is unreachable

The ordering (cluster ID, then timestamp, then rand) ensures `ls` groups archives by cluster and sorts chronologically within each cluster.

No new CLI flags are introduced -- the consensus in the RFE discussion was to change the default naming rather than adding an opt-in option.

## Test plan

- [x] Unit tests added for `generateDestDir()` covering: full cluster ID, short cluster ID, empty cluster ID, nil ConfigClient
- [x] All existing mustgather package tests pass (24/24)
- [ ] Manual validation on a live cluster: `oc adm must-gather` produces directory with cluster ID suffix and timestamp
- [ ] Manual validation with `--dest-dir` flag: custom directory name is still respected (no change in behavior)
- [ ] Manual validation with unreachable cluster: directory falls back to `must-gather.local.<timestamp>.<rand>`

See: https://issues.redhat.com/browse/RFE-5434


tchap:
Wouldn't it be nicer to use `must-gather.local.<cluster-id-suffix>.<timestamp>.<rand>` ? Like this it would be sorted when listing by name according to cluster, timestamp.

asadawar:
Good point, reordered to `must-gather.local.<cluster-id-suffix>.<timestamp>.<rand>` so that `ls` groups by cluster and sorts chronologically. Updated.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +266,35 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+// The format includes a random ID, a partial cluster ID (last 12 characters), and a UTC
+// timestamp to help distinguish must-gather archives from different clusters and collection
+// times. If the cluster ID cannot be retrieved (e.g. cluster is unreachable), it falls back
+// to the random ID and timestamp only.
+// See: https://issues.redhat.com/browse/RFE-5434
+func (o *MustGatherOptions) generateDestDir() string {
+	clusterIDSuffix := ""
+	if o.ConfigClient != nil {
+		cv, err := o.ConfigClient.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
+		if err == nil {
```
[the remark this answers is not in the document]
tchap: @asadawar I don't see the commit in the PR. Please make fix this issue. It would be better, though, to pass a context into this function, inherit the context in WithTimeout and possibly pass context.TODO() in the caller so that there is a proper context interface. Thanks.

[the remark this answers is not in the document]
asadawar: Sorry about the confusion — I had initially added a 3s timeout commit but then reverted it via force-push after reconsidering (the rest of Complete() also uses context.TODO() without timeouts, so a standalone timeout felt inconsistent). The force-push removed it from the PR history which is why you didn't see it.

Updated now: generateDestDir() accepts a context.Context parameter, wraps it with a 10s WithTimeout internally, and the caller passes context.TODO(). Thanks for the suggestion.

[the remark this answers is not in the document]
tchap: It's not super consistent, but this is an optional thing, so basically we use the old name when timeout or error, which makes sense. I think that a 5s or 3s timeout could be enough, what do you think? The user may start wondering what is happening when there is an issue. Or we could add a log statement saying we are generating the target directory name.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +266,37 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 	return nil
 }
 
+// generateDestDir builds the default destination directory name for must-gather output.
+// The format includes a partial cluster ID (last 12 characters), a UTC timestamp, and a
+// random ID to help distinguish must-gather archives from different clusters and collection
+// times. If the cluster ID cannot be retrieved (e.g. cluster is unreachable), it falls back
+// to the timestamp and random ID only.
+// See: https://issues.redhat.com/browse/RFE-5434
+func (o *MustGatherOptions) generateDestDir(ctx context.Context) string {
```
tchap, line 279: Honestly would be nicer to pass a `clock.Clock` interface so that the tests are deterministic. And then align the tests to check for exact names instead of just `contains`.

pkg/cli/admin/mustgather/mustgather.go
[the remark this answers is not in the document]
asadawar: Thanks for the feedback. Updated with the following changes:

1. **Timeout reduced to 5s** — since this is just a single GET for a small ClusterVersion resource, 5 seconds should be more than enough. If it doesn't respond in that time, the cluster is likely having issues and we should just fall back to the name without cluster ID rather than keep the user waiting.

2. **Added a `klog.V(4)` debug log on failure** — this way if someone is troubleshooting why the directory name doesn't include a cluster ID, they can run with `-v=4` and see the reason. I kept it at V(4) to stay consistent with how other non-critical failures are logged in this file (e.g. pod deletion errors on line 809, rsync retries on line 926).

3. **Injected `clock.PassiveClock`** — addressed in the other comment thread.

pkg/cli/admin/mustgather/mustgather.go
asadawar: Good suggestion, done. Added a `clock.PassiveClock` field to `MustGatherOptions` — that's a smaller interface from `k8s.io/utils/clock` that only requires `Now()` and `Since()`, which is all we need here (the full `Clock` interface includes timers that we don't use).

In production, it's initialized with `clock.RealClock{}` in `NewMustGatherOptions()`, so it calls the real `time.Now()`. In tests, I created a simple `fakeClock` struct that returns a fixed time (`2026-04-14T12:00:00Z`), which lets us assert exact prefixes like `must-gather.local.76708af6b91c.20260414T120000Z.` instead of regex patterns.

The random suffix from `rand.Int63()` is still non-deterministic, but that's fine — the important parts (cluster ID and timestamp) are now fully testable.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -73,7 +73,7 @@ var (
 	`)
 
 	mustGatherExample = templates.Examples(`
-		# Gather information using the default plug-in image and command, writing into ./must-gather.local.<rand>
+		# Gather information using the default plug-in image and command, writing into ./must-gather.local.<cluster-id-suffix>.<timestamp>.<rand>
 		  oc adm must-gather
```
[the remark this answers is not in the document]
asadawar: Good catch — updated the example text to document both formats: with cluster ID suffix and the fallback without it.

pkg/cli/admin/mustgather/mustgather.go
asadawar: @tchap just checking in on this. All your feedback has been addressed (reordering, clock.PassiveClock, klog.V(4) debug logging, fallback format in example text). Could you take a quick look when you get a chance? Also, would you be able to run /ok-to-test so CI can pick it up? Thanks!

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -185,6 +188,7 @@ func NewMustGatherOptions(streams genericiooptions.IOStreams) *MustGatherOptions
 		IOStreams:        streams,
 		Timeout:          10 * time.Minute,
 		VolumePercentage: defaultVolumePercentage,
+		Clock:            clock.RealClock{},
```
tchap: Better to export as few fields as necessary. I would rename this to `clock` and only set it in tests. And in `generateDestDir()` you can do

```golang
clock := o.clock
if clock == nil {
    clock = clock.RealClock{}
}
```

Or feel free to add a helper method on `MustGatherOptions` like `getClock()` and that one returns `o.clock` or `RealClock{}` when that field is unset.

asadawar: Good point, updated. Renamed to unexported `clock` field, removed the default from the constructor, and added a `getClock()` helper that returns `RealClock{}` when nil. Went with the helper approach since there's already a similar `getNamespace()` pattern in the same file. Tests set the field directly since they're in the same package.

tchap:
/lgtm cancel

Please fix the import ordering as mentioned by CodeRabbit.

asadawar:
Fixed the import ordering. Ready for another look.

tchap:
Let me wait for CI, I will tag this when it seems ok.

asadawar:
Hello @atiratree, this PR has received /lgtm from @tchap and all CI tests are passing. Would you be able to take a look and /approve when you get a chance?  Happy to address any feedback. Thanks!

tchap:
I think that @ardaguclu is more likely to review/approve.

asadawar:
Hello @ardaguclu this PR has received /lgtm from @tchap and all CI tests are passing. Would you be able to take a look and /approve when you get a chance? Happy to address any feedback. Thanks!

ardaguclu:
As stated in the Jira ticket, this needs to be reviewed and approved by the maintainers of must-gather (openshift/must-gather), since this change has an impact on it. RFE exists but as far as I can see it is not accepted yet.

asadawar:
@shivprakashmuley Hi, Nick mentioned in the RFE ticket that he's passed this along to you for review. This PR adds cluster ID and timestamp to must-gather directory names to make it easier to identify must-gathers from different clusters in the same support case. It has /lgtm from tchap and all CI is green. Would appreciate your thoughts or approval when you get a chance. Thanks!

asadawar:
@ardaguclu Hello, Nick has approved the RFE Jira ticket https://redhat.atlassian.net/browse/RFE-5434. The PR is ready with all CI checks passing and /lgtm from @tchap. Could you please review and provide your feedback/approval? Thanks.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -73,7 +74,9 @@ var (
 	`)
 
 	mustGatherExample = templates.Examples(`
-		# Gather information using the default plug-in image and command, writing into ./must-gather.local.<rand>
+		# Gather information using the default plug-in image and command, writing into
+		# ./must-gather.local.<cluster-id-suffix>.<timestamp>.<rand>
```
ardaguclu: It would be better to mention that timestamp is UTC.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +269,46 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 	return nil
 }
 
+// generateDestDir builds the default destination directory name for must-gather output.
```
ardaguclu: This document should belong to generateDestDir not getClock.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +269,46 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+// times. If the cluster ID cannot be retrieved (e.g. cluster is unreachable), it falls back
+// to the timestamp and random ID only.
+func (o *MustGatherOptions) getClock() clock.PassiveClock {
+	if o.clock != nil {
+		return o.clock
+	}
+	return clock.RealClock{}
+}
+
+// See: https://issues.redhat.com/browse/RFE-5434
```
ardaguclu: No need to link Jira

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +269,46 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+func (o *MustGatherOptions) generateDestDir(ctx context.Context) string {
```
ardaguclu: Could you please update with this;

```go
func (o *MustGatherOptions) generateDestDir(ctx context.Context) string {
	parts := []string{"must-gather.local"}

	if clusterID := o.getClusterIDSuffix(ctx); clusterID != "" {
		parts = append(parts, clusterID)
	}

	timestamp := o.getClock().Now().UTC().Format("20060102T150405Z")
	parts = append(parts, timestamp)

	parts = append(parts, fmt.Sprintf("%06d", rand.Int63()))

	return strings.Join(parts, ".")
}

func (o *MustGatherOptions) getClusterIDSuffix(ctx context.Context) string {
	if o.ConfigClient == nil {
		return ""
	}

	lookupCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
	defer cancel()

	cv, err := o.ConfigClient.ConfigV1().ClusterVersions().Get(lookupCtx, "version", metav1.GetOptions{})
	if err != nil {
		klog.V(4).Infof("unable to retrieve cluster ID for directory name: %v", err)
		return ""
	}

	id := string(cv.Spec.ClusterID)
	if length := len(id); length > 12 {
		return id[length-12:]
	}
	return id
}
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +269,46 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+	clusterIDSuffix := ""
+	if o.ConfigClient != nil {
+		lookupCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		defer cancel()
+		cv, err := o.ConfigClient.ConfigV1().ClusterVersions().Get(lookupCtx, "version", metav1.GetOptions{})
+		if err == nil {
+			id := string(cv.Spec.ClusterID)
```
ardaguclu: Did you check that cv.Spec can be nil?. If yes, it will panic.

asadawar:
Hi @ardaguclu, thank you for the review. I've addressed all the feedback:

1. **Timestamp UTC** — Updated example text to clarify the timestamp is UTC.
2. **Godoc placement** — Moved the comment to `generateDestDir` and added a one-liner for `getClock`.
3. **Jira link** — Removed.
4. **Refactor** — Split into `generateDestDir()` and `getClusterIDSuffix()` using `strings.Join` as you suggested.
5. **cv.Spec nil check** — `ClusterVersionSpec` is a value type (not a pointer), so it cannot be nil. This is consistent with the code you suggested as well.

All tests passing. Please take a look when you get a chance, thanks!

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -73,7 +74,9 @@ var (
 	`)
 
 	mustGatherExample = templates.Examples(`
-		# Gather information using the default plug-in image and command, writing into ./must-gather.local.<rand>
+		# Gather information using the default plug-in image and command, writing into
+		# ./must-gather.local.<cluster-id-suffix>.<timestamp-utc>.<rand>
```
ardaguclu:
```suggestion
		# ./must-gather.local.<cluster-id-suffix>.<timestamp(UTC)>.<rand>
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -73,7 +74,9 @@ var (
...
+		# or ./must-gather.local.<timestamp-utc>.<rand> if the cluster ID is unavailable
```
ardaguclu:
```suggestion
		# or ./must-gather.local.<timestamp(UTC)>.<rand> if the cluster ID is unavailable
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -230,7 +233,7 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
 		}
 	}
 	if len(o.DestDir) == 0 {
-		o.DestDir = fmt.Sprintf("must-gather.local.%06d", rand.Int63())
+		o.DestDir = o.generateDestDir(context.TODO())
```
ardaguclu:
```suggestion
		o.DestDir = o.generateDestDir()
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +269,55 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+	}
+	return clock.RealClock{}
+}
+
+// generateDestDir builds the default destination directory name for must-gather output.
+// The format includes a partial cluster ID (last 12 characters), a UTC timestamp, and a
+// random ID to help distinguish must-gather archives from different clusters and collection
+// times. If the cluster ID cannot be retrieved (e.g. cluster is unreachable), it falls back
+// to the timestamp and random ID only.
+func (o *MustGatherOptions) generateDestDir(ctx context.Context) string {
```
ardaguclu:
```suggestion
func (o *MustGatherOptions) generateDestDir() string {
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +269,55 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+
+	timestamp := o.getClock().Now().UTC().Format("20060102T150405Z")
+	parts = append(parts, timestamp)
+
+	parts = append(parts, fmt.Sprintf("%06d", rand.Int63()))
+
+	return strings.Join(parts, ".")
+}
+
+func (o *MustGatherOptions) getClusterIDSuffix(ctx context.Context) string {
```
ardaguclu:
```suggestion
func (o *MustGatherOptions) getClusterIDSuffix() string {
```

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -266,6 +269,55 @@ func (o *MustGatherOptions) Complete(f kcmdutil.Factory, cmd *cobra.Command, arg
...
+	if o.ConfigClient == nil {
```
ardaguclu: Please remove this. ConfigClient can not be empty.

pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -28,6 +29,80 @@ import (
...
+	}
+}
+
+func TestGenerateDestDirNoConfigClient(t *testing.T) {
+	fixedTime := time.Date(2026, 4, 14, 12, 0, 0, 0, time.UTC)
+	fc := fakeClock{t: fixedTime}
+
+	options := MustGatherOptions{
+		IOStreams:    genericiooptions.NewTestIOStreamsDiscard(),
+		ConfigClient: nil,
```
ardaguclu: You can use fakeclient which returns err

pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -28,6 +29,80 @@ import (
...
+			}
+			destDir := options.generateDestDir(context.TODO())
+			if !strings.HasPrefix(destDir, tc.expectedPrefix) {
+				t.Errorf("expected prefix %q, got %q", tc.expectedPrefix, destDir)
+			}
+		})
+	}
+}
+
+func TestGenerateDestDirNoConfigClient(t *testing.T) {
```
ardaguclu: This function can be merged with TestGenerateDestDir

pkg/cli/admin/mustgather/mustgather.go
ardaguclu: ??

ardaguclu: You answered already;

ClusterVersionSpec is a value type (not a pointer), so it cannot be nil. This is consistent with the code you suggested as well.

pkg/cli/admin/mustgather/mustgather.go
asadawar: Done, updated to `<timestamp(UTC)>`.

pkg/cli/admin/mustgather/mustgather.go
asadawar: Done.

pkg/cli/admin/mustgather/mustgather.go
asadawar: Done, removed ctx parameter.

pkg/cli/admin/mustgather/mustgather.go
asadawar: Done, using context.Background() internally now.

pkg/cli/admin/mustgather/mustgather.go
asadawar: Makes sense, removed. Thanks for pointing this out.

pkg/cli/admin/mustgather/mustgather_test.go
asadawar: I merged the test into TestGenerateDestDir -- the "without cluster ID" case uses an empty fake clientset which covers the fallback path. Let me know if you'd prefer a fakeclient error case instead.

pkg/cli/admin/mustgather/mustgather_test.go
asadawar: Done, merged into TestGenerateDestDir.

ardaguclu:
@asadawar thank you for spending time on this. Did you verify/test this on live cluster?. So that everything works properly?

asadawar:
Hi @ardaguclu, yes I tested this on a live OCP 4.21 cluster. Here are the results:

```
$ oc adm must-gather
...
$ ls
must-gather.local.dc6154fc5d69.20260514T111751Z.970975954159320464

$ oc get clusterversion version -o jsonpath='{.spec.clusterID}'
4d2559f6-9712-47ab-9001-dc6154fc5d69
```

The last 12 characters of the cluster ID (`dc6154fc5d69`) are correctly included in the directory name, along with the UTC timestamp. Everything works as expected.

Thank you for the thorough review and the suggestions — I learned a lot from your feedback on this PR. Really appreciate you taking the time!

ardaguclu:
/approve
I'll defer the last tag to @tchap to have a look at one last time

pkg/cli/admin/mustgather/mustgather_test.go
```diff
@@ -28,6 +29,65 @@ import (
 	imageclient "github.com/openshift/client-go/image/clientset/versioned/fake"
 )
 
+// fakeClock implements clock.PassiveClock for deterministic tests.
+type fakeClock struct {
```
tchap: There is k8s.io/utils/clock/testing

asadawar: Good catch, thanks! Replaced with `clocktesting.NewFakePassiveClock` from `k8s.io/utils/clock/testing`.

tchap:
Please squash this change into a single commit, thanks.

asadawar:
Squashed into a single commit. Also replaced the custom fakeClock with `clocktesting.NewFakePassiveClock` from `k8s.io/utils/clock/testing` as you suggested. Thanks!

tchap:
Thanks.

/lgtm

asadawar:
Hi @tchap, two of the e2e tests (`e2e-aws-ovn-serial-1of2` and `e2e-aws-ovn`) are still failing after `/retest-required`. These appear to be infrastructure-related — our change only touches the must-gather directory naming and all unit/verify/build tests pass consistently.

Could you advise on the next step here? Should we continue retesting, or is there a way to get these unblocked? The other two e2e tests (`e2e-aws-oc-ote` and `e2e-aws-ovn-serial-2of2`) passed on the retest.

Also noting that tide shows "Not mergeable. Needs verified label" — I believe that gets applied automatically once CI is green.

ardaguclu:
These tests are required. So we need to retest.
/retest

tchap:
/verified by @asadawar

He tested this manually.

PR #2258, opened 2026-04-20
DO-NOT-MERGE: Prove OCPBUGS-78768

liouk:
This PR introduces a unit test which will be used as proof of both the existence of OCPBUGS-78768 and the validity of the fix. The unit test is expected to fail without the fix, and succeed with the fix in place.

Do not merge.

/hold

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for LDAP synchronization configuration validation.

PR #2259, opened 2026-04-20
OCPBUGS-78768: Skip filter validation for RFC2307 group query when groupUIDAttribute is "dn"

liouk:
When `groupUIDAttribute` is set to `"dn"`, the group query uses the base DN directly rather than performing a search, so requiring a filter is unnecessary. Pass the `isDNOnly` flag to `validateLDAPQuery` to allow filterless group queries in this case.

## Summary by CodeRabbit

* **Bug Fixes**
  * LDAP group sync validation no longer incorrectly rejects group queries that include a filter when the group UID attribute resolves to "dn"; validation accepts both DN-only and filtered group queries as appropriate.

* **Tests**
  * Added unit tests covering validation for DN-based group UID attributes, with and without group query filters.

### DIFF
--- a/pkg/helpers/groupsync/ldap/ldap.go
+++ b/pkg/helpers/groupsync/ldap/ldap.go
@@ -165,7 +165,8 @@ func ValidateLDAPClientConfig(url, bindDN, bindPassword, CA string, insecure boo
 func ValidateRFC2307Config(config *legacyconfigv1.RFC2307Config) validation.ValidationResults {
 	validationResults := validation.ValidationResults{}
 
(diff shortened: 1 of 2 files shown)


liouk:
Will hold until fix proven with proof PR #2258.

/hold

liouk:
Proof PR unit tests passing with this PR: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-oc-2258-openshift-oc-2259-unit/2046617175610888192

/hold cancel

liouk:
Apart from CI, also manually tested with tests similar to the unit tests in this PR.

/verified by CI

PR #2263, opened 2026-05-05
OCPBUGS-85052: feat: add etcdconfig to operator inspect command

eggfoobar:
in 4.22 and 5.0 we added a new pacemaker cr under the etcd config for two node fencing clusters, adding the ability to pull down that information during inspect

added a remove step if the resources do not exist, so if the command is used against a non fencing cluster or something that does not have any etcd config we remove the empty directory to avoid noise

## Summary by CodeRabbit

* **New Features**
  * Expanded cluster inspection to collect etcd OpenShift API resources and ensure no empty diagnostic folders are left behind.

* **Tests**
  * Added test coverage for API group discovery and retrieval, including etcd API resources.

### DIFF
--- a/pkg/cli/admin/inspect/inspect.go
+++ b/pkg/cli/admin/inspect/inspect.go
@@ -366,6 +366,56 @@ func (o *InspectOptions) gatherOperatorResourceData(ctx context.Context, destDir
 	return nil
 }
 
+// gatherEtcdConfigResourceData gathers all etcd.openshift.io resources.
+func (o *InspectOptions) gatherEtcdConfigResourceData(ctx context.Context, destDir string, resourceCtx *resourceContext) error {
+	// determine if we've already collected etcdConfigResourceData
+	if resourceCtx.visited.Has(etcdConfigResourceDataKey) {
+		klog.V(1).Infof("Skipping previously-collected etcd.openshift.io resource data")
+		return nil
+	}
+	resourceCtx.visited.Insert(etcdConfigResourceDataKey)
+
+	// ensure destination path exists
+	if err := os.MkdirAll(destDir, os.ModePerm); err != nil {
+		return err
+	}
+
+	resources, err := retrieveAPIGroupVersionResourceNames(o.discoveryClient, "etcd.openshift.io")
+	if err != nil {
+		return err
+	}
+
+	errs := []error{}
+	for _, resource := range resources {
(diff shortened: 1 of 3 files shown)


eggfoobar:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-upgrade

pkg/cli/admin/inspect/inspect.go
```diff
@@ -366,6 +366,56 @@ func (o *InspectOptions) gatherOperatorResourceData(ctx context.Context, destDir
 	return nil
 }
 
+// gatherEtcdConfigResourceData gathers all etcd.openshift.io resources.
+func (o *InspectOptions) gatherEtcdConfigResourceData(ctx context.Context, destDir string, resourceCtx *resourceContext) error {
```
ardaguclu, line 370: That means every cluster starts inspecting a new resource which will have implications in terms of duration and size. So that this is strongly undesirable. 

The recommended way is to update the inspection of this resource in https://github.com/openshift/must-gather/blob/main/collection-scripts/gather_etcd_more

/hold

eggfoobar: Hey @ardaguclu, so my thinking was because we added an API level configuration at the etcd operator config level it felt appropriate for it to be pulled down when an admin ran `oc adm inspect clusteroperators` since this is a valuable cluster scope resource that gives them an idea of what pacemaker under the etcd operator is  doing, I added a remove step as well since this will only be relevant information for clusters that are running pacemaker.

Forgive me, I don't have all the background, but do you think that type of information should be put in a must-gather, rather than a targeted inspect for cluster operators?

ardaguclu: No worries. Yes this data needs to be collected in must-gather (via `oc adm inspect custom-resource`) first. 
After that if it is decided that this resource should be collected in `oc adm inspect clusteroperators` (e.g. decided by Architects, etc.) in the future, we can reconsider it. Because collected resource set in `oc adm inspect clusteroperators` is not frozen but it requires good justification.

eggfoobar: Perfect, thank you so much for the info, I'll close this out and follow it up with must gather, much appreciated it!

eggfoobar:
Closing, it's determined that this is not the appropriate place for this addition at this time, we'll be adding this fetching call to the must gather tooling.

PR #2276, opened 2026-05-25
inspectalerts: improve error message when no bearer token is available

danielmellado:
When using certificate-based authentication (e.g. system:admin from installer kubeconfig), the monitoring API on Thanos Querier rejects the request because kube-rbac-proxy only supports TokenReview-based authentication on the web port (9091). The previous error message ("no token is currently in use for this session") did not explain why it failed or what the user should do.

Update the message to clarify that token-based auth is required and point users to 'oc login' or creating a service account with the 'cluster-monitoring-view' role.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messaging for token-based authentication failures. When no token is detected, the message now clearly states token-based authentication is required for the monitoring API and directs users to authenticate (for example, by running `oc login` or using an appropriately scoped service account).

### DIFF
--- a/pkg/cli/admin/inspectalerts/inspectalerts.go
+++ b/pkg/cli/admin/inspectalerts/inspectalerts.go
@@ -99,7 +99,7 @@ func (o *options) Run(ctx context.Context) error {
 // other forms of authentication.
 func ValidateRESTConfig(restConfig *rest.Config) error {
 	if restConfig.BearerToken == "" && restConfig.BearerTokenFile == "" {
-		return fmt.Errorf("no token is currently in use for this session")
+		return fmt.Errorf("no bearer token available for this session; the monitoring API requires token-based authentication, use 'oc login'")
 	}
 	return nil
 }


pkg/cli/admin/inspectalerts/inspectalerts.go
```diff
@@ -99,7 +99,7 @@ func (o *options) Run(ctx context.Context) error {
 // other forms of authentication.
 func ValidateRESTConfig(restConfig *rest.Config) error {
 	if restConfig.BearerToken == "" && restConfig.BearerTokenFile == "" {
-		return fmt.Errorf("no token is currently in use for this session")
+		return fmt.Errorf("no bearer token available for this session; the monitoring API requires token-based authentication, use 'oc login' or create a service account bound to the 'cluster-monitoring-view' role")
```
simonpasquier: "oc login" is straightforward but I'm unclear about what it means to create a service account.

danielmellado: I was thinking about the service account path is documented in the [monitoring docs](https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html/monitoring/accessing-monitoring-apis-by-using-the-cli) as the way to access monitoring APIs programmatically (e.g. from scripts or CI), but maybe I'm mixing up thiongs

I can either:

Drop the service account mention entirely and keep it simple: "no bearer token available: the monitoring API requires token-based authentication; run 'oc login' to authenticate with a token"

Keep it but link to the docs: "no bearer token available: the monitoring API requires token-based authentication; run 'oc login' or see https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html/monitoring/accessing-monitoring-apis-by-using-the-cli for service account setup"

I think that maybe we can just drop that and keep just the login.
WDYT?

simonpasquier: Yeah, for now keeping only the mention to "oc login" is probably fine (better than what we have anyway).

danielmellado: done!

simonpasquier (commented):
/lgtm

PR #2278, opened 2026-06-02
Fix adm upgrade status timeouts: Optimize MachineConfig fetches and enforce client deadlines

YamunadeviShanmugam:
The oc adm upgrade status command frequently triggers context deadline exceeded in CI/CD pipelines. This PR add some fixes that improves the performance and fail faster rather than hanging indefinitely.

Fix includes modifying the command querying MachineConfigs sequentially node-by-node via multiple GET requests to single query and adding timeouts to avoid indefinite hang

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented status operations from hanging by introducing stricter operation timeouts and more reliable timeout handling.
  * Fixed how machine configurations are sourced so node upgrade status is reported correctly.

* **Improvements**
  * More robust status assessment that yields more accurate and consistent upgrade status results.

### DIFF
--- a/pkg/cli/admin/upgrade/status/status.go
+++ b/pkg/cli/admin/upgrade/status/status.go
@@ -55,7 +55,9 @@ func New(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Command 
 		Short: "Display the status of the current cluster version update or multi-arch migration",
 		Run: func(cmd *cobra.Command, args []string) {
 			kcmdutil.CheckErr(o.Complete(f, cmd, args))
-			kcmdutil.CheckErr(o.Run(cmd.Context()))
+			ctx, cancel := context.WithTimeout(cmd.Context(), 5*time.Minute)
+			defer cancel()
+			kcmdutil.CheckErr(o.Run(ctx))
 		},
 	}
 
@@ -109,6 +111,7 @@ func (o *options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string
 			return err
 		}
 		cfg.UserAgent = rest.DefaultKubernetesUserAgent() + "(upgrade-status)"
+		cfg.Timeout = 60 * time.Second
 
 		roundTripper, err := rest.TransportFor(cfg)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


ardaguclu:
This command is maintained by Update team. It is better to ask review from https://github.com/openshift/oc/blob/9557cf3d482ecbc4e271eb4eefeefff5eaf4bdac/OWNERS_ALIASES#L21-L24

hongkailiu:
> The oc adm upgrade status command frequently triggers context deadline exceeded in CI/CD pipelines. This PR add some fixes that improves the performance and fail faster rather than hanging indefinitely.

I will start to see if it is an issue from CI (how we run the test) or from the core code of the cmd, or neither.

Thanks to @YamunadeviShanmugam for the link to [a failing job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/79812/rehearse-79812-pull-ci-openshift-origin-main-e2e-aws-tls-observed-config-fips/2061304434863902720).

```
{  oc adm upgrade status failed 1 times (of 213)
- 2026-06-01T07:59:00Z: failed to run oc adm upgrade status after 2 attempts: context deadline exceeded
recorded attempts and outputs:
Attempt #1 failed: Error running oc --kubeconfig=/tmp/kubeconfig-819642240 adm upgrade status --details=all:
StdOut>
Error from server (Timeout): the server was unable to return a response in the time allotted, but may still be processing the request (get clusterversions.config.openshift.io version)
StdErr>
Error from server (Timeout): the server was unable to return a response in the time allotted, but may still be processing the request (get clusterversions.config.openshift.io version)
exit status 1
...
```

In the monitor test, `oc adm upgrade status` has been executed 213 times and failed only once* (with retries. See details below). It is not that bad.
I do not have stat about non-100% success rate leading to a job failure. But I would imagine it is not annoying enough because the test has been there almost 1 year (and I am aware of no complaint until today). I assume `/retest` would fix the flakiness in most cases.

The error in the run is from https://github.com/openshift/origin/blob/2d60dffe5a572f3078f9ee8018a9d31e61eadf17/pkg/monitortests/cli/adm_upgrade/status/monitortest.go#L128-L140

The timeout `2*time.Minute` (the command may have no timeout while the test does) and thus it is NOT "hanging indefinitely". ^_^

If the API server has a glitch and does not respond (see the logs ^^^ from the job) to our oc cmd, the change in the cmd (like tuning timeout in the client side) would not help much.

PR #2287, opened 2026-06-11
OTA-1956: oc adm release new: Include base image's image-references in pruning

jhadvig:
## Summary
The CVO image is the release base image (`ToImageBaseTag: "cluster-version-operator"`) but does not carry the `io.openshift.release.operator=true` label. This means `extractManifests` skips it entirely, and any images declared in the CVO's `install/image-references` are invisible to `pruneUnreferencedImageStreams` — they get dropped from the release payload.

This blocks [openshift/cluster-version-operator#1398](https://github.com/openshift/cluster-version-operator/pull/1398),  which adds a `cluster-update-console-plugin` deployment to the CVO's own manifests. The plugin image exists in the CI ImageStream but is pruned from the release because no operator's `image-references` declares it.

## Root cause

1. `extractManifests` iterates all tags and checks for `io.openshift.release.operator=true` ([line 1037](https://github.com/openshift/oc/blob/9557cf3d482ecbc4e271eb4eefeefff5eaf4bdac/pkg/cli/admin/release/new.go#L1037-L1039)). CVO has no such label → extraction is skipped → no files in the temp directory.
2. `pruneUnreferencedImageStreams` ([line 1591](https://github.com/openshift/oc/blob/9557cf3d482ecbc4e271eb4eefeefff5eaf4bdac/pkg/cli/admin/release/new.go#L1591-L1608)) scans each extracted operator's directory for `image-references`. Since CVO was never extracted, its `image-references` is never read.
3. `cluster-update-console-plugin` is not referenced by any operator and not in `AlwaysInclude` → it is pruned from the release.

## Implementation

Two small changes in `pkg/cli/admin/release/new.go`:

1. **`extractManifests` ConditionFn**: When an image has no operator label but IS the `ToImageBaseTag`, still create the destination directory and proceed with extraction. This makes the CVO's `image-references` file available on disk for the pruning step.

2. **Exclude base image from `ordered`**: After the metadata-based filter, remove the base image tag from `ordered`. Its manifests already live in `/manifests/` from the base layer and must not be duplicated into `release-manifests/`.

## Tests

Added `TestPruneUnreferencedImageStreams` (3 cases):
  - Operator `image-references` keeps referenced images (existing behavior)
  - Base image `image-references` prevents pruning (new behavior)
  - Without base image `image-references`, the image is pruned (the bug)

Added `TestBaseImageExcludedFromOrdered` (2 cases):
  - Base image tag is removed from `ordered`
  - No-op when base image is not in the list

## Related

- [openshift/cluster-version-operator#1398](https://github.com/openshift/cluster-version-operator/pull/1398) — CVO-side fixes (YAML split, service name, serving-cert, image template mechanism)

/assign @wking @DavidHurta 

## Summary by CodeRabbit

* **Bug Fixes**
  * Base image tags are now excluded from component ordering during release construction.
  * Manifest extraction now explicitly handles base image tags so their manifests are extracted when present.

* **Tests**
  * Added tests covering pruning of unreferenced image tags based on operator and base image references.
  * Added tests for base image tag list handling.

### DIFF
--- a/pkg/cli/admin/release/new_test.go
+++ b/pkg/cli/admin/release/new_test.go
@@ -1,7 +1,12 @@
 package release
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


jhadvig:
## Local validation

Tested the fix locally using `--from-dir` to simulate the release assembly pipeline without needing registry auth.

### Setup

Created a minimal release directory with a `cluster-version-operator` subdirectory containing an `image-references` file that declares `cluster-update-console-plugin`:

```
/tmp/release-test/
├── image-references              # release image-references (192 tags including cluster-update-console-plugin)
├── release-metadata              # Cincinnati metadata
└── cluster-version-operator/
    └── image-references          # CVO's image-references declaring cluster-update-console-plugin
```

### With the fix — CVO's image-references prevents pruning

```console
$ /tmp/oc-custom adm release new --from-dir=/tmp/release-test --to-file=/tmp/test-release.tar
info: Using /tmp/release-test as the input to the release
info: Found 1 operator manifest directories on disk
info: Included 2 images from 1 input operators into the release
```

Output release contains both `cluster-version-operator` and `cluster-update-console-plugin`.

### Without CVO's image-references — image is pruned

```console
$ rm /tmp/release-test/cluster-version-operator/image-references
$ /tmp/oc-custom adm release new --from-dir=/tmp/release-test --to-file=/tmp/test-release-no-ref.tar
info: Using /tmp/release-test as the input to the release
info: Found 1 operator manifest directories on disk
info: Included 1 images from 1 input operators into the release
```

Output release only contains `cluster-version-operator` — `cluster-update-console-plugin` was pruned because nothing references it.

pkg/cli/admin/release/new_test.go
```diff
@@ -12,6 +17,167 @@ import (
...
+	t.Run("base image tag is removed from ordered", func(t *testing.T) {
+		ordered := []string{"my-operator", "cluster-version-operator", "another-operator"}
+		ordered = slices.DeleteFunc(ordered, func(s string) bool {
+			return s == baseTag
+		})
+		if slices.Contains(ordered, baseTag) {
+			t.Errorf("expected %s to be removed from ordered", baseTag)
+		}
+		if len(ordered) != 2 {
+			t.Errorf("expected 2 entries, got %d", len(ordered))
```
wking: is this just excercising the `slices` library?  I'm ok trusting the stdlib to do what it says it will do, and not building our own custom tests that just excercise the stdlib.  Maybe we can drop this?  Or maybe you want to bump up a level and excercise some of the `new.go` logic you're adding that uses those `slices` functions?

pkg/cli/admin/release/new_test.go
```diff
@@ -12,6 +17,167 @@ import (
...
+	return names
+}
+
+func TestPruneUnreferencedImageStreams(t *testing.T) {
+	makeIS := func(names ...string) *imageapi.ImageStream {
+		is := &imageapi.ImageStream{}
+		for _, name := range names {
+			is.Spec.Tags = append(is.Spec.Tags, imageapi.TagReference{
+				Name: name,
+				From: &corev1.ObjectReference{Kind: "DockerImage", Name: "example.com/" + name + ":latest"},
```
wking: Overlaps a bit with `writeImageReferences`.  Maybe adjust `writeImageReferences` to return the ImageStream structure it wrote to disk, if we need an in-memory copy?  Or refactor to pull the ImageStream structure creation out into a `createImageStream` helper that both the tests and `writeImageReferences` can consume?

wking (approved):
I left two nits with the test suite, but I don't think either needs to block the merge.  I'm approving and holding, in case you want to address either.  But feel free to lift the hold whenever you like, whether or not you make any changes.

/llgtm
/hold in case you want to address test-case feedback

pkg/cli/admin/release/new_test.go
jhadvig: Good point, dropped it 🎤

pkg/cli/admin/release/new_test.go
jhadvig: Refactored... extracted a shared `createImageStream` helper that both `writeImageReferences` and the test cases use.

wking (approved):
/lgtm

PR #2289, opened 2026-06-11
OCPBUGS-88342: openshift-clients RPM has a hard dependency on bash-completion

sdodson:
Fixes https://issues.redhat.com/browse/OCPBUGS-88342

## Summary

- Downgrade `bash-completion` from a hard `Requires` to a soft `Recommends` in `oc.spec`

## Motivation

The `openshift-clients` RPM has an unnecessary hard dependency on `bash-completion`. Changing this to `Recommends` allows `oc` to be installed in minimal client environments that do not need shell completion support. Packages listed as `Recommends` are still installed by default when available but will not block installation if absent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/oc.spec
+++ b/oc.spec
@@ -52,7 +52,7 @@ BuildRequires:  rsync
 
 Provides:       atomic-openshift-clients = %{version}
 Obsoletes:      atomic-openshift-clients <= %{version}
-Requires:       bash-completion
+Recommends:     bash-completion
 
 %description
 %{summary}


sdodson:
I've created a bug because we may want to backport, I'm unsure right now whether or not we just do this for 5.0 and move on.

ardaguclu:
That makes sense to me
/lgtm
/approve
/retest

PR #2291, opened 2026-06-19
NO-JIRA: Fix tar slip path traversal in codesign archive extraction

rhamitarora:
Jira: [ARO-26358](https://redhat.atlassian.net/browse/ARO-26358)
Jira: [ARO-26359](https://redhat.atlassian.net/browse/ARO-26359)
Jira: [ARO-26373](https://redhat.atlassian.net/browse/ARO-26373)

## Summary

* **Bug Fixes**
  * Strengthened tar archive extraction safeguards to prevent malicious entry names from writing outside the intended extraction directory during code-signing and image layer unpacking.

* **Tests**
  * Added unit tests that verify path-traversal attempts are rejected safely (no files written outside the extraction area) and that normal regular-file entries extract correctly.

### DIFF
--- a/pkg/cli/admin/internal/codesign/machoresign.go
+++ b/pkg/cli/admin/internal/codesign/machoresign.go
@@ -160,8 +160,10 @@ func extractTarToTmpAndSign(path string) (string, error) {
 
 		// Guard against tar slip: ensure the resolved path stays within tempDir.
 		// filepath.Clean/Join normalize but do not constrain the result, so an
-		// entry name containing ".." could escape the extraction directory.
-		if tempExtractionPath != cleanTempDir && !strings.HasPrefix(tempExtractionPath, cleanTempDir+string(os.PathSeparator)) {
+		// entry name containing ".." could escape the extraction directory. The
+		// safe branch is determined solely by strings.HasPrefix so that static
+		// analyzers recognize this as a sanitizer.
+		if !strings.HasPrefix(tempExtractionPath, cleanTempDir+string(os.PathSeparator)) {
 			return "", fmt.Errorf("invalid archive entry %q: escapes extraction directory", header.Name)
 		}
 
--- a/pkg/cli/admin/internal/codesign/machoresign.go
+++ b/pkg/cli/admin/internal/codesign/machoresign.go
@@ -4,10 +4,12 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"debug/macho"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 func ResignMacho(path string, asArchive bool, command string, links []string) error {
@@ -125,6 +127,7 @@ func extractTarToTmpAndSign(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	cleanTempDir := filepath.Clean(tempDir)
 
 	gzipFile, err := os.Open(path)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/cli/admin/internal/codesign/machoresign.go
```diff
@@ -155,6 +158,13 @@ func extractTarToTmpAndSign(path string) (string, error) {
 
 		tempExtractionPath := filepath.Clean(filepath.Join(tempDir, header.Name))
 
+		// Guard against tar slip: ensure the resolved path stays within tempDir.
+		// filepath.Clean/Join normalize but do not constrain the result, so an
+		// entry name containing ".." could escape the extraction directory.
+		if tempExtractionPath != cleanTempDir && !strings.HasPrefix(tempExtractionPath, cleanTempDir+string(os.PathSeparator)) {
```
ardaguclu: How we do expect that `tempExtractionPath` can equal to `cleanTempDir`?

pkg/cli/admin/internal/codesign/machoresign_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	writeTarGz(t, archivePath, map[string][]byte{
+		"../escaped.txt": []byte("owned"),
+	})
+
+	tempDir, err := extractTarToTmpAndSign(archivePath)
+	if tempDir != "" {
+		defer os.RemoveAll(tempDir)
+	}
+	if err == nil {
```
ardaguclu, line 62: We should also assert the expected `tempDir`

pkg/cli/admin/internal/codesign/machoresign_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestExtractTarToTmpAndSign_ExtractsRegularFile(t *testing.T) {
+	dir := t.TempDir()
+	archivePath := filepath.Join(dir, "benign.tar.gz")
+
+	writeTarGz(t, archivePath, map[string][]byte{
+		"tool": []byte("not a macho binary"),
+	})
+
+	tempDir, err := extractTarToTmpAndSign(archivePath)
```
ardaguclu, line 82: Can we assert the expected `tempDir`?

pkg/cli/image/archive/archive.go
```diff
@@ -141,6 +141,19 @@ func unpackLayer(dest string, layer io.Reader, options *TarOptions) (size int64,
 			}
 		}
 
+		// Ensure the resolved destination stays within dest before performing any
+		// filesystem operation (including parent-directory creation below), so a
+		// malicious entry containing '..' cannot escape the extraction directory.
+		path := filepath.Join(dest, hdr.Name)
```
ardaguclu, line 147: What about moving this just above `// Skip AUFS metadata dirs`?

ardaguclu (commented):
Thank you for spending time on this. Dropped a few comments
/ok-to-test

pkg/cli/admin/internal/codesign/machoresign.go
rhamitarora: Good point. A tar entry always includes a name, which means the cleaned extraction path can never be the temporary directory itself. That made the additional check unnecessary. I've removed it in commit 9710f19, and the guard now uses only:

`strings.HasPrefix(tempExtractionPath, cleanTempDir+string(os.PathSeparator))
`
Resolving this comment.

pkg/cli/image/archive/archive.go
rhamitarora: I'd prefer to keep the containment check where it is today, immediately after `AlterHeaders` and before any filesystem operations.

If we move it down just above `// Skip AUFS metadata dirs`, it would end up after the `system.MkdirAll(parentPath, 0600)` call. Since `MkdirAll` is itself a filesystem sink, a crafted path such as `../foo/bar` could result in directory creation outside the destination before the validation check is reached.

The goal of this change is to ensure the containment check happens before any filesystem sink is invoked. That also aligns with the CodeQL finding, which treats the parent-directory creation path as part of the Zip Slip sink set.

For that reason, I intentionally placed the guard before `MkdirAll`. I'm happy to move it if you feel strongly about the code structure, but I wanted to highlight the potential traversal issue that could be introduced by doing so.

pkg/cli/admin/internal/codesign/machoresign_test.go
rhamitarora: added

pkg/cli/admin/internal/codesign/machoresign_test.go
rhamitarora: added

ardaguclu:
/retitle NO-JIRA: Fix tar slip path traversal in codesign archive extraction
/retest

PR #2295, opened 2026-06-29
OCPEDGE-2416: Add oc adm transition topology command for SNO to HA transitions

jaypoulz:
- Implement preflight validation with tiered checks (Error/Warning severity)
- Add user prompts with bypass flag support (--allow-transition-with-warnings)
- Gate command behind OC_ENABLE_CMD_TRANSITION_TOPOLOGY and MutableTopology feature gates
- Include discovery, initiate, and status subcommand modes
- Update vendored openshift/api to include Infrastructure.spec.controlPlaneTopology field

## Summary by CodeRabbit

* **New Features**
  * Added a feature-gated `oc adm transition topology` workflow to discover, dry-run validate, and initiate control-plane/infrastructure topology transitions, with confirmation and an option to proceed despite warnings.
  * Added `oc adm transition topology status` to display desired vs current topology and transition controller readiness, including etcd quorum/voting-member details.
* **Bug Fixes**
  * Improved preflight/validation behavior with clearer blocking vs non-blocking results and early short-circuiting on errors.
* **Tests**
  * Expanded command and preflight validation coverage, including multi-warning and multi-API failure scenarios.
* **Chores**
  * Updated Go module and related dependency versions; normalized OWNERS alias formatting and added governance files.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -44,7 +44,7 @@ aliases:
     - mpatlasov
     - rhrmo
     - dfajmon
-  storage-approvers: 
+  storage-approvers:
     - jsafrane
     - tsmetana
     - gnufied
@@ -53,3 +53,45 @@ aliases:
     - mpatlasov
     - rhrmo
     - dfajmon
+  openshift-edge-reviewers:
+    - Neilhamza
+    - agullon
+    - copejon
+    - dhensel-rh
+    - eggfoobar
+    - eslutsky
+    - fonta-rh
+    - fracappa
+    - ggiguash
+    - jaypoulz
+    - jeff-roche
+    - jerpeter1
+    - jogeo
+    - kasturinarra
+    - lucaconsalvi
+    - mmakwana30
+    - pacevedom
+    - pmtk
+    - qJkee
+    - vimauro
+  openshift-edge-approvers:
+    - Neilhamza
+    - agullon
+    - brandisher
+    - copejon
+    - dhensel-rh
+    - eggfoobar
+    - eslutsky
+    - fonta-rh
+    - fracappa
+    - ggiguash
+    - jaypoulz
+    - jeff-roche
+    - jerpeter1
+    - jogeo
+    - kasturinarra
+    - lucaconsalvi
+    - mmakwana30
+    - pacevedom
+    - pmtk
+    - qJkee
--- a/pkg/cli/admin/transition/preflight/client_validator.go
+++ b/pkg/cli/admin/transition/preflight/client_validator.go
@@ -1,0 +1,535 @@
+package preflight
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+	operatorv1client "github.com/openshift/client-go/operator/clientset/versioned"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// ClientSideValidator implements Validator interface using client-side validation.
+// It duplicates CCO's preflight validation logic by directly querying the API server.
+type ClientSideValidator struct {
+	kubeClient     kubernetes.Interface
+	configClient   configv1client.Interface
+	operatorClient operatorv1client.Interface
+}
+
+// NewClientSideValidator creates a new client-side validator
+func NewClientSideValidator(
+	kubeClient kubernetes.Interface,
+	configClient configv1client.Interface,
+	operatorClient operatorv1client.Interface,
+) *ClientSideValidator {
+	return &ClientSideValidator{
+		kubeClient:     kubeClient,
+		configClient:   configClient,
+		operatorClient: operatorClient,
+	}
+}
+
+// Validate checks if transition from current to target topology is possible.
+// Uses tiered validation approach:
+// - Phase 1: Run ALL Error-severity checks (blocking issues: transition support, feature gates, version)
+// - Phase 2: Run ALL Warning-severity checks (cluster readiness: nodes, operators, etcd)
+// Short-circuits after Phase 1 if any Error checks fail to avoid wasting time on cluster state checks.
+func (v *ClientSideValidator) Validate(ctx context.Context, current, target TopologyState) (*ValidationResult, error) {
+	result := NewValidationResult(current, target)
+
+	// ========================================================================
+	// PHASE 1: Error-severity checks (blocking - cannot be bypassed)
+	// ========================================================================
+	// These checks determine if the transition is fundamentally supported.
+	// Show ALL blocking issues to the user before short-circuiting.
+
+	// Check 1: Is the control plane transition supported? (CEL-based validation)
+	// This validates what the CCO will accept in spec.controlPlaneTopology
+	result.AddCheck(v.validateSupportedTransition(ctx, current, target))
+
+	// Check 2: Is MutableTopology feature gate enabled?
+	result.AddCheck(v.validateFeatureGateEnabled(ctx))
+
+	// Short-circuit if ANY Error-severity checks failed
+	// Don't waste time checking cluster state if transition isn't even valid
+	if result.HasErrorCheckFailures() {
+		return result, nil
+	}
+
+	// ========================================================================
+	// PHASE 2: Warning-severity checks (cluster readiness - bypassable)
+	// ========================================================================
+	// These checks validate cluster state is ready for the transition.
+	// Can be bypassed with --allow-transition-with-warnings.
+
+	result.AddCheck(v.validateClusterOperatorsStable(ctx))
+	result.AddCheck(v.validateControlPlaneNodeCount(ctx, RequiredControlPlaneNodeCount))
+	result.AddCheck(v.validateExactInfrastructureNodeCount(ctx, RequiredInfrastructureNodeCount))
+	result.AddCheck(v.validateControlPlaneNodesSchedulable(ctx, RequiredControlPlaneNodeCount))
+	result.AddCheck(v.validateControlPlaneNodesReady(ctx, RequiredControlPlaneNodeCount))
+	result.AddCheck(v.validateEtcdQuorum(ctx))
+	result.AddCheck(v.validateEtcdNotProgressing(ctx))
+	result.AddCheck(v.validateEtcdVotingMembers(ctx, RequiredEtcdVotingMembers))
+
+	return result, nil
+}
+
+// Helper functions for creating CheckResults with common patterns
+
+// checkUnknown creates a CheckResult with Unknown status (API error)
+func checkUnknown(name string, severity CheckSeverity, message string, err error) CheckResult {
+	return CheckResult{
+		Name:     name,
+		Severity: severity,
+		Status:   CheckStatusUnknown,
+		Message:  fmt.Sprintf("%s: %v", message, err),
+	}
+}
+
+// checkFailed creates a CheckResult with Failed status
+func checkFailed(name string, severity CheckSeverity, message string) CheckResult {
+	return CheckResult{
+		Name:     name,
+		Severity: severity,
+		Status:   CheckStatusFailed,
+		Message:  message,
+	}
+}
+
+// checkPassed creates a CheckResult with Passed status
+func checkPassed(name string, severity CheckSeverity) CheckResult {
+	return CheckResult{
+		Name:     name,
+		Severity: severity,
+		Status:   CheckStatusPassed,
+	}
+}
+
+// listNodes fetches all nodes or returns an Unknown CheckResult on error
+func (v *ClientSideValidator) listNodes(ctx context.Context, checkName string, severity CheckSeverity) (*corev1.NodeList, *CheckResult) {
+	nodes, err := v.kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		result := checkUnknown(checkName, severity, "failed to list nodes", err)
+		return nil, &result
+	}
+	return nodes, nil
+}
+
+// getEtcd fetches the Etcd operator CR or returns an Unknown CheckResult on error
+func (v *ClientSideValidator) getEtcd(ctx context.Context, checkName string, severity CheckSeverity) (*operatorv1.Etcd, *CheckResult) {
+	etcd, err := v.operatorClient.OperatorV1().Etcds().Get(ctx, EtcdOperatorResourceName, metav1.GetOptions{})
+	if err != nil {
+		result := checkUnknown(checkName, severity, "failed to get etcd operator CR", err)
+		return nil, &result
+	}
+	return etcd, nil
+}
+
+// validateSupportedTransition checks if the control plane and infrastructure transition is supported.
+// This validates what will be set in spec.controlPlaneTopology and the expected status changes.
+//
+// Constraint priority (most fundamental first):
+// 1. Already at target (no-op check)
+// 2. Current state must be SNO (only SNO → HA transitions are supported)
+// 3. Control plane must be transitioning to HighlyAvailable (not SingleReplica, infrastructure cannot transition alone)
+// 4. Infrastructure must also transition to HighlyAvailable (both topologies transition together)
+//
+// This ordering ensures that attempts to transition from HighlyAvailable state show the fundamental
+// constraint ("only SNO → HA supported") rather than derivative constraints.
+func (v *ClientSideValidator) validateSupportedTransition(ctx context.Context, current, target TopologyState) CheckResult {
+	// Check if control plane is transitioning
+	cpTransitioning := current.ControlPlane != target.ControlPlane
+	infraTransitioning := current.Infrastructure != target.Infrastructure
+
+	// Constraint 1: If nothing is transitioning, cluster is already at target
+	if !cpTransitioning && !infraTransitioning {
+		return checkFailed(CheckNameSupportedTransition, CheckSeverityError,
+			fmt.Sprintf("cluster is already at target topology (control plane: %s, infrastructure: %s)",
+				current.ControlPlane, current.Infrastructure))
+	}
+
+	// Constraint 2: Current state must be SingleReplica for both topologies
+	// This is the fundamental requirement - we only support SNO → HA transitions
+	if current.ControlPlane != configv1.SingleReplicaTopologyMode || current.Infrastructure != configv1.SingleReplicaTopologyMode {
+		return checkFailed(CheckNameSupportedTransition, CheckSeverityError,
+			fmt.Sprintf("only SingleReplica -> HighlyAvailable transitions are supported (current state: control plane %s, infrastructure %s)",
+				current.ControlPlane, current.Infrastructure))
+	}
+
+	// At this point: current state is SNO, something is transitioning
+
+	// Constraint 3: Control plane must be transitioning to HighlyAvailable
+	if !cpTransitioning {
+		// Infrastructure is transitioning but control plane is not
+		return checkFailed(CheckNameSupportedTransition, CheckSeverityError,
+			fmt.Sprintf("control plane must transition to HighlyAvailable (infrastructure cannot transition alone)"))
+	}
+
+	if target.ControlPlane != configv1.HighlyAvailableTopologyMode {
+		// Control plane is transitioning but not to HighlyAvailable
+		return checkFailed(CheckNameSupportedTransition, CheckSeverityError,
+			fmt.Sprintf("control plane target must be HighlyAvailable, got %s (only SingleReplica -> HighlyAvailable is allowed)",
+				target.ControlPlane))
+	}
+
+	// At this point: current is SNO, control plane is transitioning SNO → HA
+
+	// Constraint 4: Infrastructure must also transition to HighlyAvailable
+	if target.Infrastructure != configv1.HighlyAvailableTopologyMode {
+		return checkFailed(CheckNameSupportedTransition, CheckSeverityError,
+			fmt.Sprintf("infrastructure must transition to HighlyAvailable with control plane (infrastructure: %s -> %s, control plane: %s -> %s)",
+				current.Infrastructure, target.Infrastructure, current.ControlPlane, target.ControlPlane))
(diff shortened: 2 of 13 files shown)


jaypoulz:
@coderabbitai review

jaypoulz:
@coderabbitai review

pkg/cli/admin/transition/preflight/client_validator.go
```diff
@@ -0,0 +1,491 @@
...
+		}
+	}
+
+	if count != required {
+		return CheckResult{
+			Name:     CheckNameInfrastructureNodeCount,
+			Severity: CheckSeverityWarning,
+			Status:   CheckStatusFailed,
+			Message:  fmt.Sprintf("need exactly %d, have %d (HighlyAvailable compact topology requires no dedicated worker nodes)", required, count),
+		}
```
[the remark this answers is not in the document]
jaypoulz: Phase one is just single node to HA compact. 0 compute nodes is a requirement. That's the "compact" part of HA compact.

jaypoulz:
@coderabbitai review

jaypoulz:
@coderabbitai review

jaypoulz:
@coderabbitai review

pkg/cli/admin/transition/preflight/client_validator.go
```diff
@@ -0,0 +1,561 @@
...
+	activeVersion := featureGate.Status.FeatureGates[0]
+	for _, enabledFeature := range activeVersion.Enabled {
+		if string(enabledFeature.Name) == MutableTopologyFeatureGate {
+			return CheckResult{
+				Name:     CheckNameFeatureGateEnabled,
+				Severity: CheckSeverityError,
+				Status:   CheckStatusPassed,
+			}
+		}
+	}
```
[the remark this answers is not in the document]
jaypoulz: no, it can't. We added a comment that explains why this isn't true. [0] is guaranteed to match the active cluster version. see read lines 203 to 204

jaypoulz:
@coderabbitai review

jaypoulz:
@coderabbitai review

jaypoulz:
@coderabbitai review

pkg/cli/admin/transition/preflight/OWNERS
```diff
@@ -0,0 +1,7 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
```
ardaguclu: We probably do not need this file. Upper one should also cover here.

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/templates"
+
+	"github.com/openshift/oc/pkg/cli/admin/transition/preflight"
+)
+
+var (
+	transitionLong = templates.LongDesc(`
```
ardaguclu: This is too verbose. Can you update it more concise?

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+
+	// Confirm actually applies the transition (--confirm flag)
+	// Without this flag, the command runs in dry-run mode
+	Confirm bool
+
+	// AllowTransitionWithWarnings bypasses Warning-severity check failures (--allow-transition-with-warnings flag)
+	AllowTransitionWithWarnings bool
+
+	// Context for API calls
+	ctx context.Context
```
ardaguclu: This is not best practice. We can create ctx whenever it is needed.

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+
+// NewTransitionOptions creates a new TransitionOptions with default values
+func NewTransitionOptions(streams genericclioptions.IOStreams) *TransitionOptions {
+	return &TransitionOptions{
+		IOStreams: streams,
+	}
+}
+
+// errorWriter accumulates write errors to reduce repetitive error handling
+type errorWriter struct {
```
ardaguclu: I think we definitely do not need this struct. Why can't we use fmt?

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+		Example: transitionExample,
+		Args:    cobra.NoArgs,
+		Run: func(cmd *cobra.Command, args []string) {
+			kcmdutil.CheckErr(o.Complete(f, cmd, args))
+			kcmdutil.CheckErr(o.Validate())
+			kcmdutil.CheckErr(o.Run())
+		},
+	}
+
+	cmd.Flags().StringVar(&o.ControlPlane, "control-plane", "", "Target control plane topology (HighlyAvailable or SingleReplica)")
```
ardaguclu:
```suggestion
cmd.Flags().StringVar(&o.ControlPlane, "control-plane", o.ControlPlane, "Target control plane topology (HighlyAvailable or SingleReplica)")

```

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+
+		# Monitor transition progress
+		oc adm transition topology status
+	`)
+)
+
+// TransitionOptions holds all options for the topology transition command
+type TransitionOptions struct {
+	// Target control plane topology (--control-plane flag)
+	ControlPlane string
```
ardaguclu, line 112: Why do you export all these?

pkg/cli/admin/transition/transition.go
ardaguclu: same for the other flags.

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+		},
+	}
+
+	cmd.Flags().StringVar(&o.ControlPlane, "control-plane", "", "Target control plane topology (HighlyAvailable or SingleReplica)")
+	cmd.Flags().StringVar(&o.Infrastructure, "infrastructure", "", "Target infrastructure topology (HighlyAvailable or SingleReplica)")
+	cmd.Flags().BoolVar(&o.Confirm, "confirm", false, "Apply the transition (default is dry-run)")
+	cmd.Flags().BoolVar(&o.AllowTransitionWithWarnings, "allow-transition-with-warnings", false, "Bypass warning-severity preflight check failures (requires --confirm)")
+
+	// Add status subcommand
+	cmd.AddCommand(NewCmdStatus(f, o))
```
ardaguclu: Can we define this command under `transition`?. This is too nested.

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+
+	return nil
+}
+
+// Run executes the topology transition command
+func (o *TransitionOptions) Run() error {
+	// Create a timeout-scoped context for initial API calls (2 minutes)
+	// This prevents unbounded waits on cluster API calls during validation
+	apiCtx, apiCancel := context.WithTimeout(o.ctx, 2*time.Minute)
```
ardaguclu: Create context here.

ardaguclu: move timeout value to constant

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,474 @@
...
+		return o.runDiscoveryMode(currentCP, currentInfra)
+	}
+
+	// Mode 2: Initiate mode (at least one topology flag provided)
+	return o.runInitiateMode(apiCtx, currentCP, currentInfra)
+}
+
+// runDiscoveryMode displays current topology and available transitions
+func (o *TransitionOptions) runDiscoveryMode(currentCP, currentInfra configv1.TopologyMode) error {
+	w := &errorWriter{w: o.Out}
```
ardaguclu: Simple fmt.Fprintf should be sufficient

pkg/cli/admin/transition/constants.go
```diff
@@ -0,0 +1,56 @@
+package transition
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// Kubernetes and OpenShift resource names
+const (
```
ardaguclu: We don't need to store them in a separate file. This can be moved under transition.go

pkg/cli/admin/transition/transition.go
jaypoulz: I thought about this. I'm fine with pulling it up to a higher level, but I'm worried it won't age as we look at other kinds of Infrastructure transitions (e.g. "platform" transitions).

My thinking was it would be helpful to be able to monitor separate transitions seperately. The advantage of your suggestion is that we be encourage a maximum of one active transition of any kind at a time, which is probably best practice.

ardaguclu: This is not a blocker for me and it is up to your team. But in my opinion, it is too nested.

pkg/cli/admin/transition/transition.go
jeff-roche: Updated in `dfa0716ba`. Output now uses direct `fmt.Fprintf` calls, with errors returned to the caller.

pkg/cli/admin/transition/transition.go
jeff-roche: Updated in `dfa0716ba`. Context is no longer stored on the options struct. Each command passes `cmd.Context()` into `run`.

pkg/cli/admin/transition/transition.go
jeff-roche: Updated in `dfa0716ba`, including the other flags. Each flag now uses its corresponding option value as the default.

pkg/cli/admin/transition/transition.go
jeff-roche: Updated in `dfa0716ba`. The API timeout is now the package-level `apiRequestTimeout` constant.

jeff-roche: Updated in `dfa0716ba`. The timeout context is now created locally in `run`, derived from `cmd.Context()`, and canceled with `defer`.

pkg/cli/admin/transition/transition.go
jeff-roche: Updated in `dfa0716ba`. All `transitionOptions` fields are now unexported.

pkg/cli/admin/transition/transition.go
jeff-roche: Updated in `dfa0716ba`. The custom writer struct is gone; output uses `fmt.Fprintf` directly and propagates write errors.

pkg/cli/admin/transition/preflight/client_validator.go
```diff
@@ -0,0 +1,535 @@
...
+			Message:  fmt.Sprintf("need %d, have %d", required, schedulable),
+		}
+	}
+
+	return CheckResult{
+		Name:     CheckNameControlPlaneNodesSchedulable,
+		Severity: CheckSeverityWarning,
+		Status:   CheckStatusPassed,
+	}
+}
```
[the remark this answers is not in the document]
jeff-roche: Fixed in `566375869`. Both `NoSchedule` and `NoExecute` now prevent a node from being counted as schedulable, with regression coverage for all taint effects.

pkg/cli/admin/transition/transition.go
jeff-roche: Kept `status` directly under `transition`, as discussed, so the command is `oc adm transition status` rather than nesting it under `topology`.

pkg/cli/admin/transition/transition.go
jeff-roche: Updated in `dfa0716ba`. Removed the unused resource and condition constants and condensed the remaining declarations.

pkg/cli/admin/transition/preflight/types_test.go
```diff
@@ -0,0 +1,201 @@
...
+	if check.String() != expected {
+		t.Errorf("expected %q, got %q", expected, check.String())
+	}
+}
+
+// TestClientSideValidator_ImplementsInterface tests interface compliance
+func TestClientSideValidator_ImplementsInterface(t *testing.T) {
+	// Compile-time check that ClientSideValidator implements Validator
+	var _ Validator = (*ClientSideValidator)(nil)
+}
```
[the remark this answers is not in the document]
jeff-roche: Added in `566375869`. The tests cover passed, failed, and unknown results across both severities, aggregate error output, and `Unknown` precedence.

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,615 @@
...
+			showing := degradedOps
+			if len(showing) > maxShow {
+				showing = showing[:maxShow]
+			}
+			w.fprintf("         Degraded: %s", strings.Join(showing, ", "))
+			if len(degradedOps) > maxShow {
+				w.fprintf(" (+%d more)", len(degradedOps)-maxShow)
+			}
+			w.fprintf("\n")
+		}
```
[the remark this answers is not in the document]
jeff-roche: This no longer applies because the client-side operator monitor loop was removed. Transition progress is reported through `oc adm transition status`.

pkg/cli/admin/transition/status.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return cmd
+}
+
+// Run executes the status subcommand
+func (o *StatusOptions) Run() error {
+	ctx := context.TODO()
+
+	// Get Infrastructure resource
+	infra, err := o.configClient.ConfigV1().Infrastructures().Get(ctx, InfrastructureResourceName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
jeff-roche: Fixed in `dfa0716ba`. Status now derives a bounded context from `cmd.Context()` and returns API and output failures. Tests cover Infrastructure reads, operator Config reads, and output errors.

pkg/cli/admin/transition/preflight/OWNERS
jeff-roche: Removed the redundant nested `preflight/OWNERS`; ownership is inherited from the parent directory.

jeff-roche:
Just a heads up for reviewers: I'm taking over this PR from @jaypoulz due to priority rebalancing within our team

ardaguclu:
I don't think, I will be able to find anytime review thousands of line of changes. Isn't the purpose of this PR just adding a new single command (with 2 subcommands)?. I think, we definitely need to prompt the AI with prioritizing simplicity rather than pile of code. That would reduce the number of changes.

jeff-roche:
> I don't think, I will be able to find anytime review thousands of line of changes. Isn't the purpose of this PR just adding a new single command (with 2 subcommands)?. I think, we definitely need to prompt the AI with prioritizing simplicity rather than pile of code. That would reduce the number of changes.

@ardaguclu while on the surface these are simple commands, they are not in actuality simple commands. We are currently tackling work in the 5.1 cycle that will greatly simplify these commands but in the meantime I think this is the minimum viable set of code changes needed to verify the commands in the state they are in. In order to progress on those simplification efforts (and ultimately reduce the code changes in the long term), we need to land this PR to use as our base.

jeff-roche:
@ardaguclu  I dropped the following:
    
- All client-side preflight validator tests (these will become obsolete in the coming weeks):
    - Transition support and feature gate checks
    - Cluster operator health
    - Control-plane and infrastructure node counts
    - Node schedulability and readiness
    - etcd quorum, progress, and member count
    - API failure behavior
    - Validation result status and error formatting
    - Taint handling, including NoExecute
- Command tests for preflight blocking and bypassing:
    - Blocking error checks
    - Warning checks
    - --allow-transition-with-warnings
- Admin-level env feature-gate registration test (didn't know if this required adding a new test file for it):
    - OC_ENABLE_CMD_TRANSITION_TOPOLOGY unset, false, and true.

jaypoulz:
> I don't think, I will be able to find anytime review thousands of line of changes. Isn't the purpose of this PR just adding a new single command (with 2 subcommands)?. I think, we definitely need to prompt the AI with prioritizing simplicity rather than pile of code. That would reduce the number of changes.

@ardaguclu @jeff-roche The complexity of the command is due to the fact that we're duplicating the business logic of the server side-validations. If we provide explicitly "available transitions" via the status object, this client patch has one validation - does the request match what the server allows?

The length of this PR maps to complexity of the business logic being maintained. Simplify (or in this case, remove) that logic, and the PR is simplified.

jeff-roche:
@ardaguclu  After discussion with @jaypoulz we dropped the preflight validations since we would be removing them as our next step anyway. The controller PR (already merged) does all the preflight checks as well so we are still covered from that perspective. The CLI now just does basic topology validation (only valid values allowed), current != requested check, and permits the change. I left the validation framework in place because we will be utilizing it in the next part where we query the server for validity. 

I think where it currently stands creates a solid base for us to build on going forward (in future PRs)

ardaguclu:
@jeff-roche Thank you. That makes sense. I will try to review this PR on Monday.

pkg/cli/admin/transition/status.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get Infrastructure resource: %w", err)
+	}
+
+	// Display Control Plane Topology
+	if _, err := fmt.Fprintf(o.Out, "Control Plane Topology:\n"); err != nil {
+		return err
+	}
+	if infra.Spec.ControlPlaneTopology != "" {
+		if _, err := fmt.Fprintf(o.Out, "  Spec (desired):   %s\n", infra.Spec.ControlPlaneTopology); err != nil {
```
ardaguclu: nit: We can just ignore the linter for the error check of Fprintf. That will improve the readability of the PR.

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// runDiscoveryMode displays current topology and available transitions
+func (o *transitionOptions) runDiscoveryMode(currentCP, currentInfra configv1.TopologyMode) error {
+	if _, err := fmt.Fprintf(o.Out, "Current Topology:\n  Control Plane:    %s\n  Infrastructure:   %s\n\n", currentCP, currentInfra); err != nil {
+		return err
+	}
+
+	if currentCP == configv1.SingleReplicaTopologyMode && currentInfra == configv1.SingleReplicaTopologyMode {
+		_, err := fmt.Fprintf(o.Out, "Available Transition:\n  Control Plane:    SingleReplica -> HighlyAvailable\n  (Infrastructure also transitions to HighlyAvailable)\n\nTo validate transition readiness:\n  oc adm transition topology --control-plane=HighlyAvailable\n\nTo initiate transition:\n  oc adm transition topology --control-plane=HighlyAvailable --confirm\n")
```
ardaguclu: Raw string would be better

pkg/cli/admin/transition/transition.go
```diff
@@ -0,0 +1,306 @@
...
+
+// applyTransition patches the Infrastructure resource to initiate the transition
+func (o *transitionOptions) applyTransition(ctx context.Context, targetCP, targetInfra configv1.TopologyMode) error {
+	if _, err := fmt.Fprintf(o.Out, "Initiating topology transition...\n"); err != nil {
+		return err
+	}
+
+	// Update Infrastructure resource with retry on conflict
+	// Use RetryOnConflict to handle resourceVersion conflicts with other controllers
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ardaguclu, line 338: How can conflict error occur?. Is there any other component that may update this?. I assume yes.

OWNERS_ALIASES
```diff
@@ -44,12 +44,54 @@ aliases:
...
+    - jaypoulz
+    - jeff-roche
+    - jerpeter1
+    - jogeo
+    - kasturinarra
+    - lucaconsalvi
+    - mmakwana30
+    - pacevedom
+    - pmtk
+    - qJkee
```
ardaguclu: nit: New line

ardaguclu (commented):
Overall structure looks nice to me. Thank you. I've dropped a few comments.

pkg/helpers/describe/describer_test.go
```diff
@@ -137,6 +137,7 @@ var MissingDescriberGroupCoverageExceptions = []schema.GroupVersion{
 
 // MissingDescriberGVKExceptions lists specific GroupVersionKind that don't need describers
 var MissingDescriberGVKExceptions = []schema.GroupVersionKind{
+	{Group: "etcd.openshift.io", Version: "v1", Kind: "PacemakerCluster"},
```
ardaguclu: Is this change redundant?

ardaguclu:
My comments are mostly nitpicks and it is preferable if you can fix them in this PR. However, this new command is well written.
/approve

Once you make the necessary changes based on my comments, you can get lgtm from your team.

pkg/cli/admin/transition/transition.go
jeff-roche: CCO is now doing updates to this so conflicts can occur

pkg/helpers/describe/describer_test.go
jeff-roche: I'm not sure where this came from, maybe dirty starting branch? I'll drop it

jeff-roche:
@ardaguclu I made a bunch of updates to status.go, transition.go, validator.go and the tests. I tried to increase reuse, reduce the amount of hard coded pieces, make current and target topology state explicit, as well as some string building cleanup and formatting cleanup.

pkg/cli/admin/transition/status.go
jeff-roche: This no longer points to the right place in the file but I believe I cleaned this up

pkg/cli/admin/transition/transition.go
jeff-roche: again no longer pointing to the correct place but I switched up the formatting of the strings to make it more clear

OWNERS_ALIASES
jeff-roche: I pushed up a newline in the most recent fix but it is not showing in github and I'm not sure why

jeff-roche:
> @ardaguclu I made a bunch of updates to status.go, transition.go, validator.go and the tests. I tried to increase reuse, reduce the amount of hard coded pieces, make current and target topology state explicit, as well as some string building cleanup and formatting cleanup.

@jaypoulz can you review?

jaypoulz (commented):
/lgtm

ardaguclu:
I think, it would be better to add a few e2e tests in https://github.com/openshift/oc/tree/main/test/e2e in the followup PR. That would prevent the future changes causing regression.

PR #2297, opened 2026-07-02
Fix must-gather hang: apply --timeout to log streaming phase

afcollins:
getGatherContainerLogs had no timeout, blocking forever if gather scripts ran long. Wrap all gather phases in a shared deadline context.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gather reliability by applying a single timeout across the gather lifecycle, including container startup, log streaming, and completion checks.
  * Reduced noisy errors when gather log streaming stops due to cancellation or timeout.
  * Ensured the gathered output download/copy step continues independently so it can finish after the gather process completes.

### DIFF
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -833,9 +833,10 @@ func (o *MustGatherOptions) processNextWorkItem(ctx context.Context, ns string, 
 		log("gather did not start: %s", err)
 		return fmt.Errorf("gather did not start for pod %s: %w", pod.Name, err)
 	}
+
 	// stream gather container logs
 	if err := o.getGatherContainerLogs(ctx, pod); err != nil {
-		if !errors.Is(err, context.Canceled) {
+		if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
 			log("gather logs unavailable: %v", err)
 		}
 	}
@@ -947,6 +948,12 @@ func (o *MustGatherOptions) copyFilesFromPod(ctx context.Context, pod *corev1.Po
 }
 
 func (o *MustGatherOptions) getGatherContainerLogs(ctx context.Context, pod *corev1.Pod) error {
+	// Create a timeout context covering the entire gather lifecycle (waiting for
+	// container start, streaming logs, waiting for completion). Copying is
+	// intentionally excluded — per flag docs, it continues until finished.
+	gatherCtx, gatherCancel := context.WithTimeout(ctx, o.Timeout)
+	defer gatherCancel()
+
 	since2s := int64(2)
 	opts := &logs.LogsOptions{
 		Namespace:   pod.Namespace,
@@ -967,13 +974,13 @@ func (o *MustGatherOptions) getGatherContainerLogs(ctx context.Context, pod *cor
 		// gather script might take longer than the default API server time,
 		// so we should check if the gather script still runs and re-run logs
 		// thus we run this in a loop
-		if err := opts.RunLogsContext(ctx); err != nil {
+		if err := opts.RunLogsContext(gatherCtx); err != nil {
 			return err
 		}
 
 		// to ensure we don't print all of history set since to past 2 seconds
 		opts.Options.(*corev1.PodLogOptions).SinceSeconds = &since2s
-		if done, _ := o.isGatherDone(ctx, pod); done {
+		if done, _ := o.isGatherDone(gatherCtx, pod); done {
 			return nil
 		}
 		klog.V(4).Infof("lost logs, re-trying...")
--- a/pkg/cli/admin/mustgather/mustgather.go
+++ b/pkg/cli/admin/mustgather/mustgather.go
@@ -948,9 +948,7 @@ func (o *MustGatherOptions) copyFilesFromPod(ctx context.Context, pod *corev1.Po
 }
 
 func (o *MustGatherOptions) getGatherContainerLogs(ctx context.Context, pod *corev1.Pod) error {
-	// Create a timeout context covering the entire gather lifecycle (waiting for
-	// container start, streaming logs, waiting for completion). Copying is
-	// intentionally excluded — per flag docs, it continues until finished.
+	// Use a timeout context to prevent RunLogsContext or isGatherDone from waiting indefinitely
 	gatherCtx, gatherCancel := context.WithTimeout(ctx, o.Timeout)
 	defer gatherCancel()
 


pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -831,21 +831,27 @@ func (o *MustGatherOptions) processNextWorkItem(ctx context.Context, ns string,
...
 
+	// Create a timeout context covering the entire gather lifecycle (waiting for
+	// container start, streaming logs, waiting for completion). Copying is
+	// intentionally excluded — per flag docs, it continues until finished.
+	gatherCtx, gatherCancel := context.WithTimeout(ctx, o.Timeout)
+	defer gatherCancel()
+
 	// wait for gather container to be running (gather is running)
-	if err := o.waitForGatherContainerRunning(ctx, pod); err != nil {
+	if err := o.waitForGatherContainerRunning(gatherCtx, pod); err != nil {
```
tchap: I would use `ctx` here. No need to include this kind of waiting in the timeout IMO.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -831,21 +831,27 @@ func (o *MustGatherOptions) processNextWorkItem(ctx context.Context, ns string,
...
+	if err := o.getGatherContainerLogs(gatherCtx, pod); err != nil {
+		if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
 			log("gather logs unavailable: %v", err)
 		}
 	}
 
 	// wait for pod to be running (gather has completed)
 	log("waiting for gather to complete")
-	if err := o.waitForGatherToComplete(ctx, pod); err != nil {
+	if err := o.waitForGatherToComplete(gatherCtx, pod); err != nil {
```
tchap: Since you are passing a timeout context here, we don't need to use `PollUntilContextTimeout` and `o.Timeout` within this function, so it can be removed, I think. Not that it really matters, but better not to repeat the same thing on multiple places.

tchap:
Also I am not sure about our policy, but I tend to remove `Co-Authored-By: Claude Opus 4.6` from commits. In Kubernetes it's actually forbidden.

afcollins:
> Also I am not sure about our policy, but I tend to remove `Co-Authored-By: Claude Opus 4.6` from commits. In Kubernetes it's actually forbidden.

I thought Red Hat policy was to identify commits so we know which ones are AI-assisted.

Anyway, removed. Please check again.

pkg/cli/admin/mustgather/mustgather.go
afcollins: I see. I think we could revert to `ctx` here as well.

afcollins: Although, the difference between passing gatherCtx and the PollUntil loop within each function is each one of those loops will timeout at 35 minutes. Instead of the overall gather timing out at the shared timeout.

Is that the behavior we want?

I think the user experience of setting `--timeout` is that I expect the command to terminate after that timeout is reached, not that the timeout is applied for sub steps of the overall command.

tchap:
> I thought Red Hat policy was to identify commits so we know which ones are AI-assisted.

Right. I think the right tag is `Assisted-by: Claude Code`

pkg/cli/admin/mustgather/mustgather.go
tchap: Discussing some issues with Claude, namely that now only one of the phases is covered by `gatherCtx`:

The timeout in `getGatherContainerLogs` doesn't cover `waitForGatherToComplete`, so the gather phase can hang for up to **2x the timeout**:

1. `getGatherContainerLogs` creates `gatherCtx` with `o.Timeout` (e.g. 10m).
2. Log streaming runs for 9m, then `gatherCtx` expires → `DeadlineExceeded`.
3. The error is silently swallowed (line 842).
4. `waitForGatherToComplete` starts a **fresh** `o.Timeout` poll — another 10m.

If the gather pod is truly stuck (never terminates), `waitForGatherToComplete` blocks for the full additional timeout — which is the Prow hang scenario this PR is trying to fix.

**Suggested fix:** lift `gatherCtx` into `processNextWorkItem` so it covers `waitForGatherContainerRunning`, `getGatherContainerLogs`, and `waitForGatherToComplete` under a single timeout. The copy phase stays on the parent `ctx` as intended.

```go
// in processNextWorkItem, before the gather calls:
gatherCtx, gatherCancel := context.WithTimeout(ctx, o.Timeout)
defer gatherCancel()

if err := o.waitForGatherContainerRunning(gatherCtx, pod); err != nil { ... }
if err := o.getGatherContainerLogs(gatherCtx, pod); err != nil { ... }
if err := o.waitForGatherToComplete(gatherCtx, pod); err != nil { ... }

// copy phase uses ctx, not gatherCtx
if err := o.copyFilesFromPod(ctx, pod); err != nil { ... }
```

---

I think the proposed solution is good in spirit. Wondering still whether to include waiting for the container to come up. Strictly speaking no. The point of the timeout is not to terminate `oc` after the given timeout, so the user can't expect `oc` to always unblock, I think. And it's not the case even now. What do you think?

afcollins: So Claude suggested the same fix for both of us. :)

I am OK with treating each step with its own timeout. That behavior wasn't clear to me before, but I understand it now.

I believe the issue when one or more nodes cannot be reached is purely from the lack of timeout in `getGatherContainerLogs`, so we can keep that as the focused change of this PR.

I am not sure if the PR tests will cover this scenario, so I will need to try it out to make sure it works as expected.

afcollins:
I tried this out on a cluster where I stopped kubelet on one of the nodes.

Without these changes, the gather retries until 'Waiting for performance profile collector pods to become ready' reaches 300 attempts, looping on these two lines, and then it proceeds and completes. So this is different behavior than the environment where we saw it loop indefinitely.
```
[must-gather-9mcfn] POD 2026-07-13T18:03:31.064571996Z [disk usage checker] Volume usage percentage: current = 19 ; allowed = 70
[must-gather-9mcfn] POD 2026-07-13T18:03:31.354493414Z Waiting for performance profile collector pods to become ready: 140
```

With the changes, when the timeout is reached, gather falls back to inspect:
```
[must-gather-cwwln] POD 2026-07-13T17:32:38.024131739Z Gathering data for ns/openshift-apiserver-operator...
[must-gather-cwwln] OUT 2026-07-13T17:32:38.318949Z waiting for gather to complete
[must-gather-cwwln] OUT 2026-07-13T17:33:38.322175Z gather never finished: context deadline exceeded

Error running must-gather collection:
    gather never finished for pod must-gather-cwwln: context deadline exceeded

Falling back to `oc adm inspect clusterversion.v1.config.openshift.io,clusteroperators.v1.config.openshift.io` to collect basic cluster types.
```

I would prefer it not to fall back to inspect, but in lieu of making more changes than necessary, I am OK with this behavior.

tchap:
@afcollins Could you perhaps write down the actual intended behavior you want to implement? What should happen with all the phases when the timeout is reached? I feel like we just keep guessing here and I am getting lost in the discussion a bit, which is not great since I don't have too much bandwidth for this right now.

afcollins:
Thank you for taking a look. Sorry for making this confusing.

The problem we are trying to solve is that must-gather sometimes does not obey timeout when the cluster is in a broken state. I tried to reproduce the broken state but was unable to produce the exact scenario where gather does not obey the timeout. However, it does happen occasionally in our scheduled runs. One example [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-eng-ocp-qe-perfscale-ci-main-aws-5.0-nightly-x86-cudn-incremental-700-24nodes/2072621600754110464/artifacts/cudn-incremental-700-24nodes/gather-must-gather/build-log.txt), where `--timeout` was used, but the step ran until an external 2 hour timeout was reached.

The goal here is to enforce the context timeout for gatherContainerLogs step, which is where we suspect the problem is. These the changes currently in the branch. The context used for `RunLogsContext` does not use any timeout.

pkg/cli/admin/mustgather/mustgather.go
```diff
@@ -950,6 +951,12 @@ func (o *MustGatherOptions) copyFilesFromPod(ctx context.Context, pod *corev1.Po
 }
 
 func (o *MustGatherOptions) getGatherContainerLogs(ctx context.Context, pod *corev1.Pod) error {
+	// Create a timeout context covering the entire gather lifecycle (waiting for
+	// container start, streaming logs, waiting for completion). Copying is
+	// intentionally excluded — per flag docs, it continues until finished.
+	gatherCtx, gatherCancel := context.WithTimeout(ctx, o.Timeout)
```
tchap, line 952: This comment is not precise, this is only for log streaming. Actually the other branches may use their own timeout, but this is handled sequentially, so the user can in the end wait `N * Timeout`. We need to implement this so that there is only a single timeout handling really.

afcollins: I updated the comment.

Moving this PR to draft as I haven't seen this case for a while now. Without a solid reproducer, I will just be guessing at the behavior.

Thanks for all your help reviewing.

PR #2303, opened 2026-07-08
NO-JIRA: Bump gitignore dependency

rh-roman:
`joelanford/ignore` pulls in a vulnerable version of go-git. Vulnerable
code isn't in the execution path, but gets flagged in security scans.
See https://redhat.atlassian.net/browse/OCPBUGS-97660

Updating ignore dependency to latest, which pulls in a non-vulnerable
go-git and doesn't trip security scans.

## Summary by CodeRabbit

* **Chores**
  * Updated several third-party dependencies to newer versions for improved stability and compatibility.
  * Included newer versions of networking and Git-related libraries, along with a refreshed experimental package.


chrisahl:
Will this be backported to 4.22?

PR #2308, opened 2026-07-09
Export pod inspect functions for external consumption

kaovilai:
Export the following unexported functions in pkg/cli/admin/inspect/pod.go so they can be imported by external Go packages without vendoring the full oc binary (which pulls in CVE-affected transitive dependencies):

- gatherPodData → GatherPodData
- gatherContainerInfo → GatherContainerInfo
- gatherContainerAllLogs → GatherContainerAllLogs
- gatherContainerLogs → GatherContainerLogs
- gatherContainerRotatedLogFiles → GatherContainerRotatedLogFiles
- filterContainerLogsErrors → FilterContainerLogsErrors
- rotatedLogFilename → RotatedLogFilename

All internal callers in pod.go and namespace.go are updated to use the new exported names. Godoc comments are added to each exported function per project conventions. No behavioral changes.

## Summary by CodeRabbit

* **New Features**
  * Expanded inspection coverage to collect data from both regular and init containers.
  * Added support for gathering current, previous, and rotated container logs more consistently.
* **Bug Fixes**
  * Improved log collection error handling so expected missing-previous-log messages are filtered out, reducing noisy failures.

### DIFF
--- a/pkg/cli/admin/inspect/namespace.go
+++ b/pkg/cli/admin/inspect/namespace.go
@@ -82,7 +82,7 @@ func (o *InspectOptions) gatherNamespaceData(ctx context.Context, baseDir, names
 			klog.V(1).Infof("        Gathering data for pod %q\n", pod.GetName())
 			structuredPod := &corev1.Pod{}
 			runtime.DefaultUnstructuredConverter.FromUnstructured(pod.Object, structuredPod)
-			if err := o.gatherPodData(ctx, path.Join(destDir, "/pods/"+pod.GetName()), namespace, structuredPod); err != nil {
+			if err := o.GatherPodData(ctx, path.Join(destDir, "/pods/"+pod.GetName()), namespace, structuredPod); err != nil {
 				errs = append(errs, err)
 				continue
 			}
--- a/pkg/cli/admin/inspect/pod.go
+++ b/pkg/cli/admin/inspect/pod.go
@@ -17,7 +17,9 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func (o *InspectOptions) gatherPodData(ctx context.Context, destDir, namespace string, pod *corev1.Pod) error {
+// GatherPodData writes the pod resource YAML to destDir and gathers log
+// data for every container (init and regular) defined in the pod spec.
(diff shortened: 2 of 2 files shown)


ardaguclu:
Exporting those functions means that we should provide backwards compatibility guarantees for anyone using them. We wouldn't not want to expose those functions, since they can likely change. 

However, I understand the inconvenience here. If you only depend on these functions, why don't you directly copy to your repository and fully drop the dependency from oc?. That would be easier.
/hold

kaovilai:
> Exporting those functions means that we should provide backwards compatibility guarantees for anyone using them. We wouldn't not want to expose those functions, since they can likely change.

We don't require backwards compatibility guarantees. If we do we'd use older go.mod

kaovilai:
>  why don't you directly copy to your repository and fully drop the dependency from oc?. That would be easier.

We want future compatibility updates to flow in without recopying in the future. If things break for older version then we pin for release branch for older releases.

ardaguclu:
> > Exporting those functions means that we should provide backwards compatibility guarantees for anyone using them. We wouldn't not want to expose those functions, since they can likely change.
> 
> We don't require backwards compatibility guarantees. If we do we'd use older go.mod

I was referring to the backwards compatibility guarantees of oc.

kaovilai:
I understand the compatibility concern, but I don’t think this PR materially changes `oc`’s existing compatibility burden.

These functions already exist and are already part of the inspect implementation. The PR does not introduce new behavior, new flags, new output, or a new CLI contract; it only allows downstream Go consumers to call the same implementation instead of copying it. If these helpers need to change in the future, downstreams can pin to an older `github.com/openshift/oc` revision for older release branches and update intentionally when moving forward.

Copying the code would solve the immediate compile problem, but it creates a worse long-term maintenance model: downstream copies drift from upstream, miss fixes, and need manual re-syncing. Exporting the helpers keeps the implementation centralized and lets future inspect fixes flow naturally.

If the exported names are the issue, I’m happy to add wording to the GoDoc making the intended support level clear. But as-is, this PR has essentially no downside for current `oc` users and gives downstream consumers a cleaner, less divergent integration path.

tchap:
There is a difference between locking behavior and locking package API, but I guess the additional burden is not big.

PR #2311, opened 2026-07-13
NO-JIRA: Handle wrapped ENOTSUP error

rh-roman:
system.Lsetxattr wraps syscall errors in XattrError, so comparing the
result with == against syscall.ENOTSUP and ErrNotSupportedPlatform never
matched, causing layer extraction to fail on filesystems that do not
support xattrs instead of silently skipping them.

Replace both == comparisons with errors.Is.

Co-Authored-By: Claude Sonnet 4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when restoring archived files across platforms.
  * Unsupported extended-attribute operations are now handled reliably without interrupting extraction.

### DIFF
--- a/pkg/cli/image/archive/archive.go
+++ b/pkg/cli/image/archive/archive.go
@@ -2,6 +2,7 @@ package archive
 
 import (
 	"archive/tar"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/cli/image/archive/archive.go
```diff
@@ -402,12 +403,12 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 			continue
 		}
 		if err := system.Lsetxattr(path, key, []byte(value), 0); err != nil {
-			if err == system.ErrNotSupportedPlatform {
+			if errors.Is(err, system.ErrNotSupportedPlatform) {
```
ardaguclu, line 406: But this was working, does new system.Lsetxattr wrap errors?. Is there any code link for wrapping?

rh-roman: This is our vendored `Lsetxattr` https://github.com/openshift/oc/blob/main/vendor/github.com/docker/docker/pkg/system/xattrs_linux.go#L43-L49. Around 2 years ago we brought in a version that started wrapping errors. We must not have hit this case (or didn't notice the errors).

ardaguclu:
Thank you
/lgtm
/approve

ardaguclu:
Failure is unrelated and supposed to be fixed by https://github.com/openshift/oc/pull/2315
/override ci/prow/okd-scos-images

PR #2338, opened 2026-08-03
OCPBUGS-61061: add blank line before error in upgrade recommend --version output

savio87:
## Summary

When running `oc adm upgrade recommend --version`, the conditional-update risk message runs directly into the error line with no visual separation. This adds a blank line between the output and the error to improve readability.

Bug: https://redhat.atlassian.net/browse/OCPBUGS-61061

## Changes

- Added `fmt.Fprintln(o.Out)` before the error return in the `--version` path of `recommend.go` (only when not in quiet mode)
- Updated 5 golden test fixtures to include the new blank line

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved non-quiet upgrade recommendation output by adding clearer spacing before error messages.
  * Standardized error formatting when unaccepted issues remain or no updates are available.
* **Tests**
  * Updated command output examples to reflect the revised formatting.

### DIFF
--- a/pkg/cli/admin/upgrade/recommend/examples/4.14.1-all-recommended.version-4.12.51-output
+++ b/pkg/cli/admin/upgrade/recommend/examples/4.14.1-all-recommended.version-4.12.51-output
@@ -10,5 +10,4 @@ recommended/Alert=Unknown:
 
 Upstream update service: https://api.integration.openshift.com/api/upgrades_info/graph
 Channel: candidate-4.14 (available channels: candidate-4.14, candidate-4.15, eus-4.14, fast-4.14, stable-4.14)
-
 error: no updates to 4.12 available, so cannot display context for the requested release 4.12.51
--- a/pkg/cli/admin/upgrade/recommend/recommend.go
+++ b/pkg/cli/admin/upgrade/recommend/recommend.go
@@ -366,6 +366,9 @@ func (o *options) Run(ctx context.Context) error {
 					}
 					unaccepted := issues.Difference(accept)
 					if unaccepted.Len() > 0 {
+						if !o.quiet {
+							fmt.Fprintln(o.Out)
+						}
 						if cvoChecking {
 							return fmt.Errorf("There are issues that apply to this cluster and have not been accepted. `oc adm upgrade accept` can be used to accept them: %s\n", strings.Join(sets.List(unaccepted), ","))
 						}
--- a/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.version-4.12.51-output
+++ b/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-not-recommended.version-4.12.51-output
@@ -16,4 +16,5 @@ Message: An unintended reversion to the default kubelet nodeStatusReportFrequenc
   
   After rebooting into kernel-4.18.0-372.88.1.el8_6 or later, kernel nodes experience high load average and io_wait times. The nodes might fail to start or stop pods and probes may fail. Workload and host processes may become unresponsive and workload may be disrupted. https://issues.redhat.com/browse/COS-2705
 
+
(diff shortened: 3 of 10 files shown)


savio87:
Hi, this is a small formatting fix that adds a blank line between the conditional-update risk message and the error line in oc adm upgrade recommend --version output.

The Jira issue (OCPBUGS-61061) currently targets 4.20.z, could a maintainer please update the target version to 5.0.0 and run /jira refresh?

Also requesting /ok-to-test when ready. Thanks!

tchap:
According to the examples changes this is adding an extra newline where there already is one. Or were the examples wrong?

savio87:
Good question! The blank line before `error:` in the existing fixtures comes from the test harness, not from the actual CLI output. In `examples_test.go` (line 104), the test joins stdout and error with `\nerror:` — that `\n` creates the visual separation you see in the fixtures. But in real CLI usage, Cobra prints the error immediately after stdout with no blank line, so the output looks cramped. This change adds the blank line to the actual program output (via `fmt.Fprintln(o.Out)`), which means the fixtures now need the extra line to account for both the program's newline and the test harness's newline.

pkg/cli/admin/upgrade/recommend/recommend.go
```diff
@@ -366,6 +366,9 @@ func (o *options) Run(ctx context.Context) error {
 					}
 					unaccepted := issues.Difference(accept)
 					if unaccepted.Len() > 0 {
+						if !o.quiet {
+							fmt.Fprintln(o.Out)
+						}
```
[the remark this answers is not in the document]
savio87: The existing code throughout this file consistently ignores write errors from `fmt.Fprintf`/`fmt.Fprintln` (e.g., lines 349, 353, 361, 374). Handling it only here would be inconsistent. Happy to change if a maintainer prefers it.

tchap:
Could we align the test harness to be consistent with the actual CLI output?

savio87:
Good point — I've updated the test harness in `examples_test.go` to remove the extra `\n` before `error:`, so the fixtures now match the actual CLI output. The blank line separation now comes entirely from the program's `fmt.Fprintln(o.Out)`. Pushed in 3217d7f.

pkg/cli/admin/upgrade/recommend/examples/4.14.1-all-recommended.version-4.12.51-output
```diff
@@ -10,5 +10,4 @@ recommended/Alert=Unknown:
 
 Upstream update service: https://api.integration.openshift.com/api/upgrades_info/graph
 Channel: candidate-4.14 (available channels: candidate-4.14, candidate-4.15, eus-4.14, fast-4.14, stable-4.14)
-
```
tchap, line 13: Shouldn't the singular empty line be here actually?

savio87:
You're right @tchap — this test case goes through the error path where the program now adds the blank line via fmt.Fprintln(o.Out), so the fixture should have it. Restored the trailing blank line. Also updated the Jira target version to 5.0.0 only.

pkg/cli/admin/upgrade/recommend/examples/4.14.1-all-recommended.version-4.12.51-output
savio87: Good catch @tchap , this error path (no updates to 4.12 available) doesn't go through the fmt.Fprintln(o.Out) we added, so the blank line here was an artifact of the old test harness's \n prefix. Removed it. The blank line only appears on the "unaccepted issues" error paths (lines 370-375) where we explicitly add it.

savio87:
Hey @tchap , gentle ping to check this. All tests have passed.

savio87:
Hey @tchap , gentle ping to check this. All tests have passed.

savio87:
Hi @hongkailiu — this PR has lgtm from @tchap and all tests have passed. Could you please /approve when you have a moment? Thanks!

PR #2340, opened 2026-08-04
OCPBUGS-8247: return error when --to-multi-arch is used on a multi-arch cluster

savio87:
## Summary

Running `oc adm upgrade --to-multi-arch` on a cluster that is already multi-architecture did not return an error. The existing check only caught the case where a migration was *requested* (via `spec.desiredUpdate.architecture`) but not yet completed. After completion, CVO clears that field, so the command would attempt the patch again.

This adds a check on `status.desired.architecture` to detect clusters that are already running multi-architecture and return a clear error message.

Bug: https://redhat.atlassian.net/browse/OCPBUGS-8247

## Changes

- Added check: if `cv.Status.Desired.Architecture == ClusterVersionArchitectureMulti`, return `"Cluster is already multi-architecture"`
- This runs before the existing `spec.desiredUpdate` check, which catches in-progress migrations

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and error messaging for multi-architecture upgrades.
  * Added clear messages when a cluster is already configured for multi-architecture or has an upgrade request in progress.

### DIFF
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -246,10 +246,10 @@ func (o *Options) Run() error {
 
 	case o.ToMultiArch:
 		if cv.Status.Desired.Architecture == configv1.ClusterVersionArchitectureMulti {
-			return fmt.Errorf("Cluster is already multi-architecture")
+			return errors.New("the cluster already targets a multi-architecture release")
 		}
 		if cv.Spec.DesiredUpdate != nil && cv.Spec.DesiredUpdate.Architecture == configv1.ClusterVersionArchitectureMulti {
-			return fmt.Errorf("Cluster has already been requested to update to multi-architecture")
+			return errors.New("a multi-architecture release has already been requested")
 		}
 
 		if err := checkForUpgrade(cv); err != nil {
--- a/pkg/cli/admin/upgrade/upgrade.go
+++ b/pkg/cli/admin/upgrade/upgrade.go
@@ -245,8 +245,11 @@ func (o *Options) Run() error {
 		return nil
 
 	case o.ToMultiArch:
+		if cv.Status.Desired.Architecture == configv1.ClusterVersionArchitectureMulti {
+			return fmt.Errorf("Cluster is already multi-architecture")
+		}
 		if cv.Spec.DesiredUpdate != nil && cv.Spec.DesiredUpdate.Architecture == configv1.ClusterVersionArchitectureMulti {
-			return fmt.Errorf("info: Update to multi cluster architecture has already been requested")
+			return fmt.Errorf("Cluster has already been requested to update to multi-architecture")
 		}
 
 		if err := checkForUpgrade(cv); err != nil {


savio87:
Adds a check on status.desired.architecture to detect clusters already running multi-architecture, returning a clear error instead of attempting a redundant patch. Could a maintainer update the target version on [OCPBUGS-8247](https://redhat.atlassian.net/browse/OCPBUGS-8247) and run /ok-to-test? Thanks!

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -245,8 +245,11 @@ func (o *Options) Run() error {
 		return nil
 
 	case o.ToMultiArch:
+		if cv.Status.Desired.Architecture == configv1.ClusterVersionArchitectureMulti {
+			return fmt.Errorf("Cluster is already multi-architecture")
```
tchap: I am not sure the message is precise, the docs say:

```golang
	// desired is the version that the cluster is reconciling towards.
	// If the cluster is not yet fully initialized desired will be set
	// with the information available, which may be an image or a tag.
	// +required
	Desired Release `json:"desired"`
```

So it may be still progressing, I guess.

tchap: I have to say I don't know exactly how this works, it's just based on the docs.

savio87:
Good point @tchap — `Status.Desired` reflects what the cluster is reconciling towards, not necessarily its current state. Updated the message to "Cluster is already updating to multi-architecture" to account for the case where the transition is still in progress.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -245,8 +245,11 @@ func (o *Options) Run() error {
 		return nil
 
 	case o.ToMultiArch:
+		if cv.Status.Desired.Architecture == configv1.ClusterVersionArchitectureMulti {
+			return fmt.Errorf("Cluster is already updating to multi-architecture")
+		}
 		if cv.Spec.DesiredUpdate != nil && cv.Spec.DesiredUpdate.Architecture == configv1.ClusterVersionArchitectureMulti {
-			return fmt.Errorf("info: Update to multi cluster architecture has already been requested")
+			return fmt.Errorf("Cluster has already been requested to update to multi-architecture")
```
tchap: Since there is no formatting needed, can we just use `errors.New` to create the error?

savio87:
Good point @tchap , updated to errors.New since there are no format parameters.

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -245,8 +245,11 @@ func (o *Options) Run() error {
 		return nil
 
 	case o.ToMultiArch:
+		if cv.Status.Desired.Architecture == configv1.ClusterVersionArchitectureMulti {
+			return errors.New("cluster is already updating to multi-architecture")
+		}
 		if cv.Spec.DesiredUpdate != nil && cv.Spec.DesiredUpdate.Architecture == configv1.ClusterVersionArchitectureMulti {
-			return fmt.Errorf("info: Update to multi cluster architecture has already been requested")
+			return fmt.Errorf("cluster has already been requested to update to multi-architecture")
```
tchap: This line also needs updating to `errors.New`

pkg/cli/admin/upgrade/upgrade.go
```diff
@@ -245,8 +245,11 @@ func (o *Options) Run() error {
 		return nil
 
 	case o.ToMultiArch:
+		if cv.Status.Desired.Architecture == configv1.ClusterVersionArchitectureMulti {
+			return errors.New("cluster is already updating to multi-architecture")
```
wking: instead of "updating to", can we say "the cluster already targets a multi-architecture release" or something?  We might be updating from single-arch to multi-arch, or from multi-arch to multi-arch, or happily settled on multi-arch.  "targets" doesn't care about where we were in the past, or whether we're mid-update or steady-state now.  It's only talking about the release that the CVO is already trying to reconcile, and that's enough to error out here, telling the admin that there's no need to request a multi-arch transition if we're already aiming at a multi-arch release.

savio87:
Thanks @wking and @tchap for the feedback. I've updated both error messages:

Line 249 (Status.Desired.Architecture == Multi): "the cluster already targets a multi-architecture release" — adopting @wking's suggested wording

Line 252 (Spec.DesiredUpdate.Architecture == Multi): "a multi-architecture release has already been requested" — switched from fmt.Errorf to errors.New as @tchap suggested

savio87:
@tchap @wking — feedback addressed and CI is green. Could you /lgtm when you have a moment? Thanks!

tchap:
I will piggy-back on what Trevor suggested.

/lgtm

savio87:
Hi @wking — feedback addressed (errors.New + your “targets a multi-architecture release” wording), and @tchap has given /lgtm. The failed e2e checks look like infra (no build client found for build10/build02); I’ve triggered /retest. Could you please /approve when you have a moment? Thanks!

PR #2351, opened 2026-08-06
OCPBUGS-100066: skip wildcard sources in IDMS/ICSP image reference parsing

savio87:
## Bug

https://redhat.atlassian.net/browse/OCPBUGS-100066

`oc adm node-image create` fails with `invalid source "*.redhat.com": invalid reference format` when the cluster has an ImageDigestMirrorSet (IDMS) containing wildcard domain sources like `*.redhat.com` or `*.redhat.io`.

## Root Cause

In `pkg/cli/image/strategy/onerror.go`, the `alternativeImageSourcesIDMS()` and `alternativeImageSourcesICSP()` functions call `reference.Parse(rdm.Source)` on every IDMS/ICSP source entry. Wildcard sources (e.g. `*.redhat.com`) are valid in the IDMS CRD schema and handled correctly by CRI-O at runtime, but `reference.Parse()` rejects them because `*` is not a valid character in a Docker image reference.

## Fix

Skip IDMS/ICSP source entries that start with `*.` before attempting to parse them as Docker image references. These wildcard entries cannot be meaningfully matched to a specific image reference anyway, so skipping them is the correct behavior. A debug-level log message is emitted when a wildcard source is skipped.

## Bug Fixes

* **Bug Fixes**
  * Wildcard mirror sources beginning with `*.` are now skipped during image mirror resolution for ICSP and IDMS configurations.
  * Valid mirror entries continue to be used when mixed with wildcard entries.
  * Configurations containing only wildcard sources now correctly fall back to the original image source.
  * Wildcard entries are logged and excluded before further processing.

### DIFF
--- a/pkg/cli/image/strategy/onerror.go
+++ b/pkg/cli/image/strategy/onerror.go
@@ -96,7 +96,7 @@ func alternativeImageSourcesICSP(imageRef reference.DockerImageReference, icspLi
 		repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
 		for _, rdm := range repoDigestMirrors {
 			if strings.HasPrefix(rdm.Source, "*.") {
-				klog.V(5).Infof("Skipping wildcard ICSP source %q", rdm.Source)
+				klog.V(5).Infof("Skipping wildcard ICSP source entry")
 				continue
 			}
 			var suffix string
@@ -234,7 +234,7 @@ func alternativeImageSourcesIDMS(imageRef reference.DockerImageReference, idmsLi
 		repoDigestMirrors := idms.Spec.ImageDigestMirrors
 		for _, rdm := range repoDigestMirrors {
 			if strings.HasPrefix(rdm.Source, "*.") {
-				klog.V(5).Infof("Skipping wildcard IDMS source %q", rdm.Source)
+				klog.V(5).Infof("Skipping wildcard IDMS source entry")
 				continue
 			}
 			var suffix string
--- a/pkg/cli/image/strategy/onerror.go
+++ b/pkg/cli/image/strategy/onerror.go
@@ -95,6 +95,10 @@ func alternativeImageSourcesICSP(imageRef reference.DockerImageReference, icspLi
 	for _, icsp := range icspList {
 		repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
 		for _, rdm := range repoDigestMirrors {
+			if strings.HasPrefix(rdm.Source, "*.") {
+				klog.V(5).Infof("Skipping wildcard ICSP source %q", rdm.Source)
+				continue
+			}
 			var suffix string
 			var err error
 			rdmSourceRef, err := reference.Parse(rdm.Source)
@@ -229,6 +233,10 @@ func alternativeImageSourcesIDMS(imageRef reference.DockerImageReference, idmsLi
 	for _, idms := range idmsList {
 		repoDigestMirrors := idms.Spec.ImageDigestMirrors
 		for _, rdm := range repoDigestMirrors {
+			if strings.HasPrefix(rdm.Source, "*.") {
+				klog.V(5).Infof("Skipping wildcard IDMS source %q", rdm.Source)
+				continue
+			}
 			var suffix string
 			var err error
 			rdmSourceRef, err := reference.Parse(rdm.Source)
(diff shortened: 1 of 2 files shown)


savio87:
This fixes a crash in oc adm node-image create when IDMS contains wildcard source entries like *.redhat.com. Could a maintainer please set the target version on [OCPBUGS-100066](https://redhat.atlassian.net/browse/OCPBUGS-100066) and run /ok-to-test? Thank you!

tchap:
Shouldn't we rather implement logic that handles wildcards properly? If we want to mirror what CRI-O does, shouldn't `oc` also support these wildcards? Just a question to start with...

savio87:
Good question @tchap, the current fix prevents the crash by skipping wildcards that cause `reference.Parse` to fail with "invalid reference format". Implementing proper wildcard matching (like CRI-O does) would be a larger effort and could be tracked as a follow-up enhancement. This PR at minimum ensures `oc adm node-image create` doesn't crash when wildcards are present in IDMS/ICSP entries. Would you prefer to handle full wildcard support in this PR or as a separate issue?

ardaguclu:
> Good question @tchap, the current fix prevents the crash by skipping wildcards that cause `reference.Parse` to fail with "invalid reference format". Implementing proper wildcard matching (like CRI-O does) would be a larger effort and could be tracked as a follow-up enhancement. This PR at minimum ensures `oc adm node-image create` doesn't crash when wildcards are present in IDMS/ICSP entries. Would you prefer to handle full wildcard support in this PR or as a separate issue?

I think `oc adm node-image` can catch this in validation step before passing the yaml to the internals. This validation can error out. This would be the better user experience. 
/hold

savio87:
Thanks @ardaguclu , makes sense to catch this earlier. I can move the wildcard check to getIdmsContents in mirrorbuilder.go so it's caught before reaching the image strategy internals. One question: should it error out entirely, or skip wildcard entries with a visible warning? Erroring out would block oc adm node-image create on any cluster that has wildcard IDMS/ICSP entries (which are valid CRI-O configs the user may not have control over). Skipping with a warning feels like a better user experience, the user sees the skip and the command continues with valid entries. What do you think?

savio87:
Hello @ardaguclu , gentle ping, please check my previous comment.

ardaguclu:
> Hello @ardaguclu , gentle ping, please check my previous comment.

I'm not sure about the UX. I think, it would be better to ask that command's maintainers.

savio87:
@rwsu @andfasano — @ardaguclu placed /hold on this PR and suggested checking with oc adm node-image maintainers on UX for wildcard IDMS/ICSP sources (e.g. *.redhat.com).

oc adm node-image create currently crashes on these entries even though they're valid in CRI-O. Which behavior do you prefer?

Skip + warn — filter wildcards, continue with valid mirrors
Hard error — fail early if wildcards are present
Full wildcard matching — CRI-O-like behavior (likely a follow-up)
Happy to move the check to getIdmsContents in mirrorbuilder.go once we agree. Thanks!

/cc @ardaguclu

andfasano:
> @rwsu @andfasano — @ardaguclu placed /hold on this PR and suggested checking with oc adm node-image maintainers on UX for wildcard IDMS/ICSP sources (e.g. *.redhat.com).
> 
> oc adm node-image create currently crashes on these entries even though they're valid in CRI-O. Which behavior do you prefer?
> 
> Skip + warn — filter wildcards, continue with valid mirrors Hard error — fail early if wildcards are present Full wildcard matching — CRI-O-like behavior (likely a follow-up) Happy to move the check to getIdmsContents in mirrorbuilder.go once we agree. Thanks!
> 
> /cc @ardaguclu

@savio87 it's not very clear to me what kind of crash generated this change, do you have specific job link to share for us to have a look on it? Thanks

savio87:
@andfasano Thanks — it's not a CI job crash, it's a local/command failure when oc adm node-image create reads cluster IDMS/ICSP entries.

Repro / error (from OCPBUGS-100066):

invalid source "*.redhat.com": invalid reference format
when the cluster has an ImageDigestMirrorSet with wildcard sources like *.redhat.com / *.redhat.io.

Where it fails: pkg/cli/image/strategy/onerror.go calls reference.Parse(rdm.Source) on every IDMS/ICSP source. Wildcards are valid for CRI-O, but reference.Parse rejects *.

Bug: https://redhat.atlassian.net/browse/OCPBUGS-100066

This PR currently skips *. sources in the image strategy. @ardaguclu prefers catching this earlier in oc adm node-image validation (getIdmsContents / mirrorbuilder.go) and put /hold pending UX.

Which UX do you prefer?

1. Skip + warn — filter wildcards, continue with valid mirrors
2. Hard error — fail early if wildcards are present
3. Full wildcard matching — CRI-O-like (likely a follow-up)

Happy to move the check once we agree. /cc @rwsu @ardaguclu

PR #2360, opened 2026-08-11
NO-JIRA: Add --proxy-url support to oc login

### LINKED ISSUE #2314: oc login should support proxy-url option
The `oc login` command does not currently support the `proxy-url` option supported in the kubeconfig:

https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/

Ideally it would be possible to pass something like `--proxy-url=http://squid.redhat.com:3128` and `oc login` would use the proxy setting the same as if HTTPS_PROXY were set and configure the `proxy-url` setting for the cluster in the kubeconfig.

Current behavior is unfortunate because if `proxy-url` is manually set in the kubeconfig it works for normal commands, but when `oc login` is run again it not only doesn't work, but when HTTPS_PROXY environment variable is set it removes the `proxy-url` setting, breaking functionality for other oc commands.

This functionality is particularly important for cases where multiple clusters are managed through `oc` and some should use a proxy, some not, or different proxy servers should be used for different clusters.

kchawlani19:
## Summary
- Add `--proxy-url` to `oc login` so login uses a per-cluster proxy and persists it to kubeconfig `cluster.proxy-url`.
- Reuse an existing cluster `proxy-url` when the flag is omitted, so re-login no longer drops it (including when `HTTPS_PROXY` is set).
- Leave `rest.Config.Proxy` unset when neither flag nor kubeconfig proxy is present, so env proxies still apply without being written into kubeconfig.

Fixes https://github.com/openshift/oc/issues/2314

## Test plan
- [x] Unit: `--proxy-url` flag accepted and validated (`http`/`https`/`socks5`; reject invalid schemes)
- [x] Unit: login client uses `--proxy-url` during dial
- [x] Unit: existing cluster `proxy-url` is reused when flag is omitted
- [x] Unit: `--proxy-url` overrides an existing cluster proxy
- [x] Unit: `SaveConfig` / `CreateConfig` persist `proxy-url`; unrelated clusters keep distinct proxies
- [x] Unit: `HTTPS_PROXY` alone is not written into kubeconfig
- [x] CI: `unit` / `verify` / `verify-deps` green
- [ ] Manual (optional): `oc login --proxy-url=http://<proxy>:3128 ...` and confirm kubeconfig `clusters.*.proxy-url`

## Summary by CodeRabbit

* **New Features**
  * Added `--proxy-url` support to the login command.
  * Supports HTTP, HTTPS, and SOCKS5 proxy URLs.
  * Reuses existing cluster proxy settings when no new proxy is specified.
  * Saves explicitly configured proxy settings to login configuration.

* **Bug Fixes**
  * Invalid or unsupported proxy URLs now produce validation errors.
  * Proxy settings are preserved correctly when configurations are merged.
  * Cluster selection now prioritizes the canonical server nickname.

### DIFF
--- a/pkg/cli/login/helpers.go
+++ b/pkg/cli/login/helpers.go
@@ -17,6 +17,34 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
+// parseProxyURL validates a kubeconfig-style proxy URL.
+// Schemes match client-go's kubeconfig proxy-url validation.
+func parseProxyURL(proxyURL string) (*url.URL, error) {
+	u, err := url.Parse(proxyURL)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse: %v", proxyURL)
+	}
+
+	switch u.Scheme {
+	case "http", "https", "socks5":
+	default:
+		return nil, fmt.Errorf("unsupported scheme %q, must be http, https, or socks5", u.Scheme)
+	}
+	return u, nil
+}
+
+// setClientConfigProxy configures rest.Config.Proxy from a proxy URL string.
+// When set, client-go uses this instead of HTTPS_PROXY/HTTP_PROXY, and
+// CreateConfig persists it to the kubeconfig cluster's proxy-url field.
+func setClientConfigProxy(clientConfig *restclient.Config, proxyURL string) error {
+	u, err := parseProxyURL(proxyURL)
+	if err != nil {
+		return err
+	}
+	clientConfig.Proxy = http.ProxyURL(u)
+	return nil
+}
+
 // getMatchingClusters examines the kubeconfig for all clusters that point to the same server
 func getMatchingClusters(clientConfig restclient.Config, kubeconfig clientcmdapi.Config) sets.String {
 	ret := sets.String{}
--- a/pkg/cli/login/helpers.go
+++ b/pkg/cli/login/helpers.go
@@ -3,6 +3,7 @@ package login
 import (
 	"bytes"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -19,20 +20,21 @@ import (
 )
 
 // parseProxyURL validates a kubeconfig-style proxy URL.
-// Schemes match client-go's kubeconfig proxy-url validation.
+// Supported schemes match client-go's kubeconfig proxy-url validation.
+// Errors never include the raw URL, which may contain credentials.
 func parseProxyURL(proxyURL string) (*url.URL, error) {
 	u, err := url.Parse(proxyURL)
 	if err != nil {
-		return nil, fmt.Errorf("could not parse: %v", proxyURL)
+		return nil, errors.New("invalid --proxy-url: could not parse")
 	}
 
 	switch u.Scheme {
 	case "http", "https", "socks5":
 	default:
-		return nil, fmt.Errorf("unsupported scheme %q, must be http, https, or socks5", u.Scheme)
+		return nil, fmt.Errorf("invalid --proxy-url: unsupported scheme %q, must be http, https, or socks5", u.Scheme)
 	}
-	if len(u.Host) == 0 {
-		return nil, fmt.Errorf("host must be specified")
+	if u.Hostname() == "" {
+		return nil, errors.New("invalid --proxy-url: host must be specified")
 	}
 	return u, nil
 }
--- a/pkg/cli/login/helpers.go
+++ b/pkg/cli/login/helpers.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strings"
 
+	cliconfig "github.com/openshift/oc/pkg/helpers/kubeconfig"
 	"github.com/openshift/oc/pkg/helpers/term"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/cli-runtime/pkg/printers"
@@ -29,6 +30,9 @@ func parseProxyURL(proxyURL string) (*url.URL, error) {
 	case "http", "https", "socks5":
 	default:
 		return nil, fmt.Errorf("unsupported scheme %q, must be http, https, or socks5", u.Scheme)
+	}
+	if len(u.Host) == 0 {
+		return nil, fmt.Errorf("host must be specified")
 	}
 	return u, nil
 }
@@ -58,8 +62,15 @@ func getMatchingClusters(clientConfig restclient.Config, kubeconfig clientcmdapi
 	return ret
 }
 
-// findClusters returns the first cluster matching the host.
+// findCluster returns a cluster matching the host. Prefer the canonical
+// nickname CreateConfig would write, so login reuses the same cluster entry
+// (and its proxy-url) when multiple stanzas share a server URL.
 func findCluster(host string, kubeconfig clientcmdapi.Config) *clientcmdapi.Cluster {
+	if nick, err := cliconfig.GetClusterNicknameFromURL(host); err == nil {
+		if cluster, ok := kubeconfig.Clusters[nick]; ok && cluster.Server == host {
+			return cluster
+		}
+	}
 	for _, cluster := range kubeconfig.Clusters {
 		if cluster.Server == host {
 			return cluster
--- a/pkg/cli/login/login.go
+++ b/pkg/cli/login/login.go
@@ -52,6 +52,9 @@ var (
 
 		# Log in to the external OIDC issuer through Auth Code + PKCE by starting a local server listening on port 8080
 		oc login localhost:8443 --exec-plugin=oc-oidc --client-id=client-id --extra-scopes=email,profile --callback-port=8080
+
+		# Log in using a proxy and persist proxy-url in the kubeconfig cluster entry
+		oc login localhost:8443 --username=myuser --password=mypass --proxy-url=http://squid.example.com:3128
 	`)
 )
 
@@ -102,6 +105,8 @@ func NewCmdLogin(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.
 	cmds.Flags().StringVar(&o.OIDCIssuerURL, "issuer-url", o.OIDCIssuerURL, "Experimental: Issuer url for external issuer. Required.")
 	cmds.Flags().StringVar(&o.OIDCCAFile, "oidc-certificate-authority", o.OIDCCAFile, "Experimental: The path to a certificate authority bundle to use when communicating with external OIDC issuer.")
 	cmds.Flags().BoolVar(&o.OIDCAutoOpenBrowser, "auto-open-browser", o.OIDCAutoOpenBrowser, "Experimental: Automatically open browser for login. When used with --web, defaults to true. When used with --exec-plugin for external OIDC, defaults to false.")
+
+	cmds.Flags().StringVar(&o.ProxyURL, "proxy-url", o.ProxyURL, "Proxy to use for all requests sent by oc login, and stored as proxy-url in the kubeconfig cluster entry")
 	return cmds
 }
 
@@ -229,6 +234,12 @@ func (o LoginOptions) Validate(cmd *cobra.Command, serverFlag string, args []str
 
 	if cmd.Flags().Changed("auto-open-browser") && !o.WebLogin && o.OIDCExecPluginType == "" {
 		return errors.New("--auto-open-browser can only be specified along with --web or --exec-plugin")
+	}
+
+	if len(o.ProxyURL) > 0 {
+		if _, err := parseProxyURL(o.ProxyURL); err != nil {
+			return fmt.Errorf("invalid --proxy-url %q: %w", o.ProxyURL, err)
+		}
 	}
 
 	return nil
--- a/pkg/cli/login/login.go
+++ b/pkg/cli/login/login.go
@@ -238,7 +238,7 @@ func (o LoginOptions) Validate(cmd *cobra.Command, serverFlag string, args []str
 
 	if len(o.ProxyURL) > 0 {
 		if _, err := parseProxyURL(o.ProxyURL); err != nil {
-			return fmt.Errorf("invalid --proxy-url %q: %w", o.ProxyURL, err)
+			return err
 		}
 	}
 
--- a/pkg/cli/login/loginoptions.go
+++ b/pkg/cli/login/loginoptions.go
@@ -180,7 +180,7 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
 	}
 	if len(proxyURL) > 0 {
 		if err := setClientConfigProxy(clientConfig, proxyURL); err != nil {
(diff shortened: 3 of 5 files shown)


kchawlani19:
Updated the PR title to `NO-JIRA:` so Tide can pick up `jira/valid-reference` (this change tracks GitHub issue #2314 rather than an OCPBUGS ticket). The description now links the issue and includes a test plan.

pkg/cli/login/helpers.go
```diff
@@ -17,6 +17,34 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("could not parse: %v", proxyURL)
+	}
+
+	switch u.Scheme {
+	case "http", "https", "socks5":
+	default:
+		return nil, fmt.Errorf("unsupported scheme %q, must be http, https, or socks5", u.Scheme)
+	}
+	return u, nil
```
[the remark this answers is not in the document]
kchawlani19: Addressed in b807546: `parseProxyURL` now rejects empty hosts (`http://`, `http:/proxy`) and tests cover both cases.

pkg/cli/login/loginoptions.go
```diff
@@ -162,15 +168,31 @@ func (o *LoginOptions) getClientConfig() (*restclient.Config, error) {
...
+	existingCluster := findCluster(clientConfig.Host, *o.StartingKubeConfig)
+
+	// Prefer an explicitly provided --proxy-url. Otherwise reuse an existing
+	// cluster proxy-url so login uses it and CreateConfig does not drop it.
+	// Leave Proxy unset when neither is present so HTTPS_PROXY/HTTP_PROXY still
+	// apply via the default transport without being written into kubeconfig.
+	proxyURL := o.ProxyURL
+	if len(proxyURL) == 0 && existingCluster != nil {
+		proxyURL = existingCluster.ProxyURL
+	}
```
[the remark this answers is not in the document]
kchawlani19: Addressed in b807546: `findCluster` now prefers the canonical CreateConfig nickname when multiple clusters share a server URL, with a unit test for distinct proxy-urls.

kchawlani19:
Hi @ardaguclu  @atiratree  — could you take a look when you have a chance?

This adds `oc login --proxy-url` (fixes https://github.com/openshift/oc/issues/2314):
- uses the proxy for the login client
- persists it to kubeconfig `cluster.proxy-url`
- preserves an existing proxy-url when the flag is omitted

First-stage CI is green. Happy to address any feedback.
/cc @ardaguclu @atiratree

ardaguclu:
Thanks for spending time on this. But I think, we don't have any capacity to review these changes for now.

kchawlani19:
Thanks @ardaguclu for the update — understood. Happy to keep the PR open and follow up later when you have bandwidth. Please let me know if there's a better time or another reviewer I should ping.

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +658,437 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
+				t.Fatalf("expected error containing %q, got: %v", tc.expectErrSubstr, err)
+			}
+		})
+	}
+}
+
+func TestGetClientConfigUsesProxyURL(t *testing.T) {
+	apiHit := make(chan struct{}, 1)
+	apiServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		apiHit <- struct{}{}
```
tchap: I don't have time to review all of this right now, so I may continue later, but you could use an `atomic` type here so that you don't have to do `select` at the end, but just load a value and compare.

kchawlani19: Thanks @tchap — updated in the latest commit to use `atomic.Bool` / `atomic.Value` instead of channel+select.

tchap: The cast to `string` will panic, though, if the value is not there. Not too great. Use `atomic.Pointer[string]` or check the cast worked and fail if not. Thanks.

kchawlani19: Good catch — switched to `atomic.Pointer[string]` and nil-check on Load(). Thanks!

pkg/cli/login/helpers.go
```diff
@@ -30,8 +62,15 @@ func getMatchingClusters(clientConfig restclient.Config, kubeconfig clientcmdapi
...
-// findClusters returns the first cluster matching the host.
+// findCluster returns a cluster matching the host. Prefer the canonical
+// nickname CreateConfig would write, so login reuses the same cluster entry
+// (and its proxy-url) when multiple stanzas share a server URL.
 func findCluster(host string, kubeconfig clientcmdapi.Config) *clientcmdapi.Cluster {
+	if nick, err := cliconfig.GetClusterNicknameFromURL(host); err == nil {
+		if cluster, ok := kubeconfig.Clusters[nick]; ok && cluster.Server == host {
+			return cluster
+		}
+	}
```
tchap, line 75: I am not sure this is needed, strictly speaking, but it also doesn't hurt...

pkg/cli/login/login.go
```diff
@@ -231,6 +236,12 @@ func (o LoginOptions) Validate(cmd *cobra.Command, serverFlag string, args []str
 		return errors.New("--auto-open-browser can only be specified along with --web or --exec-plugin")
 	}
 
+	if len(o.ProxyURL) > 0 {
+		if _, err := parseProxyURL(o.ProxyURL); err != nil {
+			return fmt.Errorf("invalid --proxy-url %q: %w", o.ProxyURL, err)
```
tchap: I would improve the error messages in `parseProxyURL` so that we don't have to wrap here. The same goes for `setClientConfigProxy` call below. Wrapping it down there with `invalid proxy URL` leaks internals of `setClientConfigProxy`. How can you know the error is related to proxy URL? Better to generate meaningful and complete errors in `parseProxyURL` and just return it without wrapping...

pkg/cli/login/loginoptions.go
```diff
@@ -71,6 +71,12 @@ type LoginOptions struct {
 	WebLogin     bool
 	CallbackPort int32
 
+	// ProxyURL, when set, is used for the login HTTP client and persisted to the
+	// kubeconfig cluster's proxy-url field. When empty, an existing cluster
+	// proxy-url is preserved; otherwise HTTPS_PROXY/HTTP_PROXY may still apply
+	// via the default transport when rest.Config.Proxy is nil.
```
tchap: We should rather be sure the usual `HTTPS_PROXY/HTTP_PROXY` still works. Does it still work? We should make the docstring more exact.

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +659,432 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
 
+func TestLoginProxyURLFlagAccepted(t *testing.T) {
+	cmd := NewCmdLogin(nil, genericiooptions.NewTestIOStreamsDiscard())
+	flag := cmd.Flags().Lookup("proxy-url")
+	if flag == nil {
+		t.Fatal("expected oc login to define --proxy-url")
+	}
+	if err := cmd.Flags().Set("proxy-url", "http://squid.example.com:3128"); err != nil {
+		t.Fatalf("expected --proxy-url to accept a value: %v", err)
+	}
```
tchap: Is this really testing anything useful? I mean the `Set` part...

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +659,432 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
+}
+
+func TestValidateProxyURL(t *testing.T) {
+	testCases := []struct {
+		name            string
+		proxyURL        string
+		expectErrSubstr string
```
tchap: I think that we know the exact full error message, so let's match against the full error message exactly and rename this to `expectedErrMsg` or something like that.

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +659,432 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
+	}
+	got, err := clientConfig.Proxy(req)
+	if err != nil {
+		t.Fatalf("Proxy func failed: %v", err)
+	}
+	if got == nil || got.String() != proxyServer.URL {
+		t.Fatalf("expected proxy %q, got %v", proxyServer.URL, got)
+	}
+
+	// dialToServer requests the API root; absolute form may include a trailing slash.
```
tchap: Pls extend this to mention that `getClientConfig` is actually making the real request.

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +659,432 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	got, err := clientConfig.Proxy(req)
+	if err != nil {
+		t.Fatalf("Proxy func failed: %v", err)
+	}
+	if got == nil || got.String() != proxyServer.URL {
+		t.Fatalf("expected proxy %q, got %v", proxyServer.URL, got)
+	}
```
tchap: IMO this whole section can be dropped, it's not really testing anything relevant that is not tested by the following check.

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +659,432 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
+	if !strings.HasPrefix(*u, apiServer.URL) {
+		t.Fatalf("proxy saw unexpected URL %q", *u)
+	}
+	if apiHit.Load() {
+		t.Fatal("api server should not have been reached directly when proxy is configured")
+	}
+}
+
+func TestGetClientConfigPrefersCanonicalClusterProxyURL(t *testing.T) {
```
tchap: This is not a particularly robust test, because even if we just iterate as we did before, we can accidentally find the right cluster since map iteration is non-deterministic.

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +659,432 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	got, err := clientConfig.Proxy(req)
+	if err != nil {
+		t.Fatalf("Proxy func failed: %v", err)
+	}
+	if got == nil || got.String() != proxyServer.URL {
+		t.Fatalf("expected preserved proxy %q, got %v", proxyServer.URL, got)
+	}
```
tchap: I don't really like that we are sending the real request in `getClientConfig` and then also calling `Proxy` directly and checking the result. We can store what server was called in the request handlers and then just check that, we don't need to create another request and call `Proxy`. We should simplify this in all tests where possible.

pkg/cli/login/loginoptions_test.go
```diff
@@ -654,6 +659,432 @@ func TestValidateAutoOpenBrowser(t *testing.T) {
...
+	clientConfig, err := options.getClientConfig()
+	if err != nil {
+		t.Fatalf("getClientConfig failed: %v", err)
+	}
+	if clientConfig.Proxy != nil {
+		t.Fatal("expected rest.Config.Proxy to remain nil so HTTPS_PROXY/HTTP_PROXY can still apply")
+	}
+}
+
+func TestSaveConfigPersistsAndPreservesProxyURL(t *testing.T) {
```
tchap: I reviewed up until here for now. Can we pls restructure the tests so that we have a top-level wrapper `TestProxyURL` or `TestProxyURLFlag` and then we can use `t.Run()` for each test that is now a top-level thing? So that we can see these are all related.

Also pls add a doc comment to each test quickly explaining what that test is testing. This really helps with review.

pkg/cli/login/loginoptions_test.go
kchawlani19: Follow-up: dial tests later switched from atomics to a simple \`contacted\` string recorded by the httptest handlers (same process, no races). Behavior coverage is unchanged.

pkg/cli/login/helpers.go
kchawlani19: Kept intentionally: the canonical-nickname lookup still checks \`Server == host\` so we do not reuse a nickname that points at a different API URL.

pkg/cli/login/login.go
kchawlani19: Done — \`parseProxyURL\` now returns complete \`invalid --proxy-url: ...\` errors (never interpolating the raw URL/credentials), and both \`Validate\` and \`getClientConfig\` return those errors directly with no wrapping.

pkg/cli/login/loginoptions.go
kchawlani19: Clarified the \`ProxyURL\` docstring and the \`getClientConfig\` comment: when neither \`--proxy-url\` nor an existing cluster proxy-url is set, \`rest.Config.Proxy\` stays nil so client-go still honors \`HTTPS_PROXY\`/\`HTTP_PROXY\`, and env proxies are not written into kubeconfig. Covered by \`leaves Proxy nil without flag or cluster proxy\` plus the SaveConfig env-proxy test.

pkg/cli/login/loginoptions_test.go
kchawlani19: Agreed — dropped the \`Flags().Set\` assertion. The test now only checks that \`--proxy-url\` is registered.

pkg/cli/login/loginoptions_test.go
kchawlani19: Done — validation cases now use \`expectedErrMsg\` and compare the full error string exactly (plus an extra credential-leak guard).

pkg/cli/login/loginoptions_test.go
kchawlani19: Done — dial tests now note that \`getClientConfig\` performs the real request via \`dialToServer\`, and they assert which httptest endpoint was contacted.

pkg/cli/login/loginoptions_test.go
kchawlani19: Agreed — removed the redundant \`Proxy()\`/implementation-detail checks. The dial tests keep a single behavioral observation: which server was contacted.

pkg/cli/login/loginoptions_test.go
kchawlani19: Addressed in production + test: \`findCluster\` prefers CreateConfig's canonical nickname, and the test asserts the selected cluster's \`ProxyURL\` is the canonical one (so a map-iteration alias hit fails).

pkg/cli/login/loginoptions_test.go
kchawlani19: Done — dial tests record the contacted endpoint in the handlers and assert that; no extra request / direct \`Proxy()\` checks.

pkg/cli/login/loginoptions_test.go
kchawlani19: Done — proxy coverage is grouped under \`TestProxyURL\` with \`t.Run\` subtests, plus short comments describing each important case.

PR #2376, opened 2026-08-19
RFE-8872: warn and confirm before deleting PV/PVC

yupputho:
## Summary
- Prompt in a terminal before `oc delete` removes a persistent volume or claim, matching [RFE-8872](https://redhat.atlassian.net/browse/RFE-8872).
- Skip the prompt for non-interactive sessions (CI/scripts) and when `--interactive=false` is set, so existing automation is unchanged.
- Detect PV/PVC deletes from resource args (`pvc`, `pv`, `pvc/name`, mixed types) and from local `-f` manifests.

## Test plan
- [ ] `go test -mod=vendor ./pkg/cli/kubectlwrappers/`
- [ ] In a terminal, `oc delete pvc <name>` shows the warning and `(y/N)`; `n` or empty input prints `deletion is cancelled`
- [ ] `echo y | oc delete pvc <name>` (non-TTY) deletes without hanging
- [ ] `oc delete pod <name>` is unchanged
- [ ] `oc delete --interactive=false pvc <name>` skips the new prompt
- [ ] `oc delete --dry-run=client pvc <name>` does not prompt

## Summary by CodeRabbit

* **New Features**
  * Added a safety confirmation prompt when deleting persistent volumes or claims through `oc delete` in an interactive terminal.
  * Supports detecting storage resources specified directly or through local YAML/JSON files.
  * Deletion is canceled when confirmation is declined or invalid.

* **Bug Fixes**
  * Prompts are appropriately skipped for dry runs, raw output, explicit interactive settings, non-storage resources, and non-interactive sessions.

* **Documentation**
  * Updated guidance describing the storage deletion safety check.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -98,7 +98,7 @@ func (o *ExampleOptions) Run() error { ... }
 ## Contributing Rules
 
 - **Do not modify `pkg/cli/cli.go`** unless it is part of a kubectl rebase process (to reflect changes from `kubectl/cmd.go`).
-- **Do not diverge from wrapped kubectl commands.** If a command is a pure kubectl wrapper, behavioral changes belong upstream in `k8s.io/kubectl`.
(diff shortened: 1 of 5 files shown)


ardaguclu:
In oc (and kubectl) officially recommended path is to use Kuberc for that configuration. We won't change the default behaviors of commands.
/close

PR #2378, opened 2026-08-24
RFE-9789: oc whoami --show-token should return exec plugin tokens

### LINKED ISSUE #2377: `oc whoami --show-token` fails with ExecCredential (exec) auth plugins, e.g. `oc get-token`
## Summary

`oc whoami --show-token` fails with:

```
error: no token is currently in use for this session
```

when the active kubeconfig context authenticates via an `ExecCredential`
(exec) auth provider — for example `oc get-token`, which is invoked through
the client-go exec credential plugin mechanism
(`client.authentication.k8s.io`).

## Expected behavior

`oc whoami --show-token` should return a valid, current bearer token when the
active context is configured with an exec-based (`ExecCredential`) auth
provider, the same way it does for statically configured tokens.

## Actual behavior

`oc whoami --show-token` errors out with `no token is currently in use for
this session` instead of returning a token, even though the same kubeconfig
successfully authenticates every other `oc`/`kubectl` command against the
cluster (i.e. the exec plugin itself works fine for normal API requests).

## Root cause (as I understand it)

In an exec-credential setup there is no static, long-lived bearer token
stored directly in the kubeconfig or in the in-memory client-go config, in
the form `oc whoami --show-token` currently expects. Instead, credentials are
obtained dynamically by invoking the configured exec plugin (e.g. `oc
get-token`) — the same mechanism `oc`/kubectl already use to authenticate
every other API request. `oc whoami --show-token` does not invoke this exec
credential flow at all, so it has nothing to print and errors out.

## Impact

Any workflow that shells out to `oc whoami --show-token` to obtain a bearer
token for use elsewhere (CI/CD pipelines, scripts, tools that wrap the
OpenShift/Kubernetes API, docs/automation that assume `oc whoami
--show-token` is a reliable way to get a usable token) breaks for
clusters/users configured with `ExecCredential`-based authentication such as
`oc get-token`, even though the underlying auth mechanism is fully supported
by `oc` for normal API calls.

## Suggested fix

Enhance `oc whoami --show-token` (and/or the underlying client-go/`oc`
credential resolution path it uses) so that when the active context is
configured with an exec-based (`ExecCredential`) auth provider such as `oc
get-token`, `oc` invokes that plugin (respecting its caching/TTL behavior like
it does for normal requests) and returns the resulting token, rather than
only supporting statically configured tokens.

This would let `oc whoami --show-token` behave consistently across all
supported authentication methods, including `oc get-token` and other
`ExecCredential` providers.

## Environment

- `oc` / OpenShift CLI (client-go exec credential plugin auth)
- Any cluster/user configured with an `ExecCredential` (exec-based) auth
  provider in kubeconfig (`client.authentication.k8s.io`), such as via `oc
  get-token`

## Steps to reproduce

1. Configure a kubeconfig context to authenticate via an exec-based auth
   provider, e.g. `oc get-token` as the `command` in a
   `user.exec.command`/`ExecConfig` entry (`client.authentication.k8s.io/v1`).
2. Confirm normal `oc`/`kubectl` commands work against the cluster with this
   context (e.g. `oc get projects`), proving the exec plugin itself is
   functioning correctly.
3. Run `oc whoami --show-token`.
4. Observe:
   ```
   error: no token is currently in use for this session
   ```

## Related

Filed as a Red Hat support RFE case: 04524345.

kchawlani19:
## Summary
- `oc whoami --show-token` currently only reads `rest.Config.BearerToken`, which is empty for ExecCredential plugins such as `oc get-token`.
- Resolve the session token through the same client-go transport stack used for API requests so exec plugins (and token files/auth providers) are invoked with their normal caching/TTL.
- Static bearer tokens and non-token auth (cert/basic) keep the existing behavior, including `no token is currently in use for this session` when no bearer token is available.

Fixes https://github.com/openshift/oc/issues/2377
Jira: https://issues.redhat.com/browse/OCPBUGS-113632
Support case: 04524345

## Test plan
- [ ] `go test ./pkg/cli/whoami/` (covers static token, token file, exec plugin success/failure, and no-token)
- [ ] With a kubeconfig user configured for `oc get-token` exec auth, `oc get projects` succeeds and `oc whoami --show-token` prints the current bearer token
- [ ] With a static token kubeconfig, `oc whoami --show-token` still prints that token
- [ ] With cert-only or basic auth, `oc whoami --show-token` still errors with `no token is currently in use for this session`

## Summary by CodeRabbit

- **New Features**
  - `whoami --show-token` now supports tokens from static configuration, token files, exec-based credential providers, and authentication providers.
  - Added clear errors when token resolution fails or no token is available.
  - Updated the flag description and usage example.

- **Bug Fixes**
  - Removed the requirement for a statically configured bearer token when using `--show-token`.

### DIFF
--- a/pkg/cli/whoami/whoami.go
+++ b/pkg/cli/whoami/whoami.go
@@ -3,6 +3,8 @@ package whoami
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -40,6 +42,9 @@ var whoamiLong = templates.LongDesc(`
 var whoamiExample = templates.Examples(`
 	# Display the currently authenticated user
 	oc whoami
+
+	# Display the token for this session
+	oc whoami --show-token
 `)
 
 type WhoAmIOptions struct {
@@ -83,7 +88,7 @@ func NewCmdWhoAmI(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra
 		},
 	}
 
-	cmd.Flags().BoolVarP(&o.ShowToken, "show-token", "t", o.ShowToken, "Print the token the current session is using. This will return an error if you are using a different form of authentication.")
+	cmd.Flags().BoolVarP(&o.ShowToken, "show-token", "t", o.ShowToken, "Print the token the current session is using, including tokens from exec credential plugins. This will return an error if you are using a different form of authentication.")
 	cmd.Flags().BoolVarP(&o.ShowContext, "show-context", "c", o.ShowContext, "Print the current user context name")
 	cmd.Flags().BoolVar(&o.ShowServer, "show-server", o.ShowServer, "If true, print the current server's REST API URL")
 	cmd.Flags().BoolVar(&o.ShowConsoleUrl, "show-console", o.ShowConsoleUrl, "If true, print the current server's web console URL")
@@ -152,9 +157,6 @@ func (o *WhoAmIOptions) Validate() error {
 	if o.PrintFlags.OutputFlagSpecified() && (o.ShowToken || o.ShowContext || o.ShowServer || o.ShowConsoleUrl) {
 		return fmt.Errorf("--output cannot be used with --show-token, --show-context, --show-server, or --show-console")
 	}
-	if o.ShowToken && len(o.ClientConfig.BearerToken) == 0 {
-		return fmt.Errorf("no token is currently in use for this session")
-	}
 	if o.ShowContext && len(o.RawConfig.CurrentContext) == 0 {
 		return fmt.Errorf("no context has been set")
 	}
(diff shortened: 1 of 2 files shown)


ardaguclu:
This has never supported. So, this is a feature request rather than a bug. I would recommend filing a RFE request to be triaged.
/hold

kchawlani19:
Thanks @ardaguclu — agreed this expands `--show-token` beyond static kubeconfig tokens rather than fixing a regression.

Converted OCPBUGS-113632 to [RFE-9789](https://issues.redhat.com/browse/RFE-9789) and retitled this PR accordingly.

/jira refresh

kchawlani19:
Gentle ping @ardaguclu — [RFE-9789](https://issues.redhat.com/browse/RFE-9789) has been in Backlog since 26 Aug with a working implementation in this PR (CI green).

Could you take another look when you have a chance? If the approach looks good, `/unhold` (and `/lgtm` if you are happy with the code) would unblock review. The RFE still needs PM triage, but the hold is what is stopping any further PR progress.

Thanks.

ardaguclu:
@kchawlani19 I think, it would be better to wait for the approval of RFE. Because I will need to allocate time to review this PR accordingly.

ardaguclu:
I'm seeing bot-looking accounts try to add this functionality and I think we need to wait for a while to see whether this change exposes a vulnerability or not. So we should hold off this feature.

PR #2380, opened 2026-08-25
OCPBUGS-106186: Fix registry blob upload request headers

liouk:
`oc image mirror` intermittently fails when uploading blobs over HTTP/2 to a local Docker registry on bare-metal CI. The `httpBlobUpload.ReadFrom` method in the vendored `distribution/distribution` library sends PATCH requests without `Content-Length`, `Content-Range`, or `Content-Type` headers, violating the OCI Distribution Spec. Over HTTP/2, this causes a flow control race: the client fills the server's flow control window and pauses; if the server sends response headers before granting more window, Go's HTTP/2 transport aborts the stream with `RST_STREAM CANCEL`. On retry, the server retains the partial upload but the client restarts from offset 0, resulting in an unrecoverable `"upload resumed at wrong offset"` error.

This PR:
- Carries a vendor patch on `distribution/distribution` `ReadFrom` to set `Content-Length`, `Content-Range`, and `Content-Type` headers when the reader provides a `ContentLength()` method
- Wraps the blob reader in `copyBlob` with a `readerWithContentLength` struct that exposes the known blob size

The upstream library has deprecated its client package (moved to `internal/` in v3.0.0), so an upstream fix is not viable. See [OCPBUGS-106186](https://issues.redhat.com/browse/OCPBUGS-106186) for details.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image mirroring uploads by providing known content-length information during transfers.
  * Prevented incorrect or unnecessary upload headers from being set automatically.
  * Improved compatibility and reliability for blob uploads.

### DIFF
--- a/deps.diff
+++ b/deps.diff
@@ -1,0 +1,13 @@
+diff --no-dereference -N -r current/vendor/github.com/distribution/distribution/v3/registry/client/blob_writer.go updated/vendor/github.com/distribution/distribution/v3/registry/client/blob_writer.go
+28,31d27
+< type ContentLengthResolver interface {
+< 	ContentLength() int64
+< }
+< 
+49,54d44
+< 
+< 	req.Header.Set("Content-Type", "application/octet-stream")
+< 	if clr, ok := r.(ContentLengthResolver); ok {
+< 		req.ContentLength = clr.ContentLength()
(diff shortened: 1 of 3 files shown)


liouk:
Awaiting feedback from workloads folks on the fix and carry patch before proceeding.

/hold

liouk:
As discussed internally, the overhead of a carry patch isn't justified for this issue -- this is an issue that will rarely occur and can easily be resolved with a retry.

/close

PR #2382, opened 2026-08-27
OCPBUGS-95885,OCPBUGS-95884: Bump golang.org/x/net to v0.55.0 via go patch version v035.0-sec.4

tonyxrmdavidson:
Bump golang.org/x/net to v0.55.0 via go patch version v035.0-sec.4


tonyxrmdavidson:
The bump inadvertently bump the go version.

PR #2397, opened 2026-09-08
OCPBUGS-122005: Add IDMS support to `oc image info`

zaneb:
Deprecate the `--icsp-file` option to `oc image info` and add a `--idms-file` option for consistency with other commands.

Use the Explicit mirror strategy instead of OnError, because users are making an explicit request to respect the mirroring settings. Experience in the field has shown that contacting quay first does not always fail cleanly in some proxy setups, which can break the command altogether.

## Summary by CodeRabbit

- **New Features**
  - Added the `--idms-file` option for specifying an IDMS alternate source when retrieving image information.
  - Image retrieval now supports explicit ICSP or IDMS alternate-source strategies.
  - Configured mirrors are tried first when locating images through IDMS.

- **Bug Fixes**
  - Prevented simultaneous use of `--idms-file` and `--icsp-file`.
  - Standardized tag warnings across both alternate-source options so they appear only once.

- **Deprecations**
  - Deprecated the `--icsp-file` option.

### DIFF
--- a/pkg/cli/image/info/info.go
+++ b/pkg/cli/image/info/info.go
@@ -83,7 +83,7 @@ func NewInfo(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Comm
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be read from.")
 	flags.StringVar(&o.ICSPFile, "icsp-file", o.ICSPFile, "Path to an ImageContentSourcePolicy file.  If set, data from this file will be used to find alternative locations for images.")
 	flags.MarkDeprecated("icsp-file", "support for it will be removed in a future release. Use --idms-file instead.")
-	flags.StringVar(&o.IDMSFile, "idms-file", o.IDMSFile, "Path to an ImageDigestMirrorSet file. If set, data from this file will be used to find alternative locations for images.")
+	flags.StringVar(&o.IDMSFile, "idms-file", o.IDMSFile, "Path to an ImageDigestMirrorSet file. If set, data from this file will be used to find alternative locations for images. Mirrors will be tried first.")
 	flags.BoolVar(&o.ShowMultiArch, "show-multiarch", o.ShowMultiArch, "Show information even if the image is multiarch image. If not set, error is thrown for multiarch images.")
 
 	return cmd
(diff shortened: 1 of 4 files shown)


pkg/cli/image/info/info.go
```diff
@@ -82,6 +82,8 @@ func NewInfo(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Comm
 	flags.StringVarP(&o.Output, "output", "o", o.Output, "Print the image in an alternative format: json")
 	flags.StringVar(&o.FileDir, "dir", o.FileDir, "The directory on disk that file:// images will be read from.")
 	flags.StringVar(&o.ICSPFile, "icsp-file", o.ICSPFile, "Path to an ImageContentSourcePolicy file.  If set, data from this file will be used to find alternative locations for images.")
+	flags.MarkDeprecated("icsp-file", "support for it will be removed in a future release. Use --idms-file instead.")
+	flags.StringVar(&o.IDMSFile, "idms-file", o.IDMSFile, "Path to an ImageDigestMirrorSet file. If set, data from this file will be used to find alternative locations for images.")
```
ardaguclu: We can document the behavior of the strategy we adopted in this new flag.

pkg/cli/image/info/info.go
```diff
@@ -123,7 +129,10 @@ func (o *InfoOptions) Run() error {
 		return err
 	}
 	if len(o.ICSPFile) > 0 {
-		registryContext = registryContext.WithAlternateBlobSourceStrategy(strategy.NewICSPOnErrorStrategy(o.ICSPFile))
+		registryContext = registryContext.WithAlternateBlobSourceStrategy(strategy.NewICSPExplicitStrategy(o.ICSPFile))
```
ardaguclu: In my opinion, we should leave this as is to not break anyone after the upgrades.

pkg/cli/image/info/info.go
```diff
@@ -123,7 +129,10 @@ func (o *InfoOptions) Run() error {
...
+	}
+	if len(o.IDMSFile) > 0 {
+		registryContext = registryContext.WithAlternateBlobSourceStrategy(strategy.NewIDMSExplicitStrategy(o.IDMSFile))
```
ardaguclu, line 137: It makes sense to me setting new strategy for this new flag.

pkg/cli/image/info/info.go
zaneb: Done (for all 4 commands).

pkg/cli/image/info/info.go
zaneb: Done.

PR #2403, opened 2026-09-14
OCPBUGS-122354: Fix inconsistent terminology in oc set probe help example

abhoje:
In `oc set probe` help text, one example comment uses inconsistent terminology.

Actual result:

Set only the initial-delay-seconds field on all deployments
oc set probe dc --all --readiness --initial-delay-seconds=30

Expected result:

Set only the initial-delay-seconds field on all deployment configs
oc set probe dc --all --readiness --initial-delay-seconds=30

The command uses dc, which refers to DeploymentConfig, but the comment says "deployments". In OpenShift, deployment/deploy and dc are different resource types, so this wording is misleading.

**Why this matters**

This can be especially confusing in OCP-related learning and certification scenarios, where candidates may assume the example applies to Kubernetes Deployment resources. Since OpenShift 4.14+, Deployment is more commonly used than DeploymentConfig, so incorrect help text increases the chance of misunderstanding which resource type the command targets.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the `probe` command help example to refer specifically to all deployment configurations.


abhoje:
Bug for reference : https://redhat.atlassian.net/browse/OCPBUGS-122354

abhoje:
/retitle OCPBUGS-122354: oc set probe --help example mention "all deployments" but uses dc --all (DeploymentConfig)

tchap:
/retitle OCPBUGS-122354: Fix inconsistent terminology in oc set probe help example

